PR #3, opened 2023-06-28
Setup general application skeleton

wkozyra95:
Minimal application skeleton

- Add HTTP server that listens for API calls
- Add TCP socket that can be used to connect input and output connections
  - one thread to listen for new connections
  - one thread for each input to listen for incoming data from tcp connection
  - one thread for each output to listen on a channel and write data to tcp connection
- Add Pipeline abstraction that is responsible for processing video (from rtp on input to rtp on output, excluding transport layer)
  - Passing data to the pipeline is done just via a function call (all the processing before the queue will be done synchronously)
  - Passing output data from the pipeline is done via the channel (but when we implement everything I suspect we can remove that)

Not tested in any way,

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -1,0 +1,13 @@
+use anyhow::Result;
+
+pub struct Decoder {}
+
+impl Decoder {
+    pub fn new() -> Self {
+        Decoder {}
+    }
+
+    pub fn decode(&self, _buffer: bytes::Bytes) -> Result<Vec<bytes::Bytes>> {
+        todo!()
+    }
+}
--- a/src/http.rs
+++ b/src/http.rs
@@ -1,0 +1,86 @@
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+use std::{
+    io::{Cursor, Read},
+    sync::Arc,
+    thread,
+};
+use tiny_http::{Response, Server, StatusCode};
+
+use super::state::State;
+
+#[derive(Serialize, Deserialize)]
+struct RegisterInputRequest {
+    pub port: u16,
+    pub input_id: u32,
+}
+
+#[derive(Serialize, Deserialize)]
+struct RegisterOutputRequest {
+    pub port: u16,
+    pub output_id: u32,
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+enum Request {
+    RegisterInput(RegisterInputRequest),
+    RegisterOutput(RegisterOutputRequest),
+}
+
+fn handle_event(state: Arc<State>, event: Request) -> Result<()> {
+    match event {
+        Request::RegisterInput(RegisterInputRequest { port, input_id }) => {
+            state.add_input(port, input_id)?;
+            Ok(())
+        }
+        Request::RegisterOutput(RegisterOutputRequest { port, output_id }) => {
+            state.add_output(port, output_id)?;
+            Ok(())
+        }
+    }
+}
+
+fn handle_request(state: Arc<State>, request: &mut tiny_http::Request) -> Response<impl Read> {
+    let event = match serde_json::from_reader::<_, Request>(request.as_reader()) {
+        Ok(event) => event,
+        Err(err) => {
+            return Response::new(
+                StatusCode(500),
+                vec![],
+                Cursor::new(format!("{}", err).into_bytes()),
+                None,
+                None,
+            );
+        }
+    };
+    if let Err(err) = handle_event(state, event) {
+        return Response::new(
+            StatusCode(500),
+            vec![],
+            Cursor::new(format!("{}", err).into_bytes()),
+            None,
+            None,
+        );
+    }
+    Response::new(
+        StatusCode(200),
+        vec![],
+        Cursor::new("{}".as_bytes().to_vec()),
+        None,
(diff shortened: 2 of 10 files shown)


src/http.rs
```diff
@@ -0,0 +1,86 @@
...
+        }
+        Request::RegisterOutput(RegisterOutputRequest { port, output_id }) => {
+            state.add_output(port, output_id)?;
+            Ok(())
+        }
+    }
+}
+
+fn handle_request(state: Arc<State>, request: &mut tiny_http::Request) -> Response<impl Read> {
+    let event = match serde_json::from_reader::<_, Request>(request.as_reader()) {
```
wkozyra95, line 45: obviously better error handling is need here

src/pipeline.rs
```diff
@@ -0,0 +1,136 @@
...
+                eprintln!("render loop");
+                thread::sleep(Duration::from_millis(1000));
+            }
+        });
+    }
+}
+
+impl PipelineInput {
+    fn process_input(&self, buffer: bytes::Bytes) -> Result<Vec<Frame>> {
+        self.rtp_parser
```
wkozyra95, line 107: no idea if parsing rtp and decoding will be 2 separate elements or maybe we will go from rtp to raw frames directly. This POC does not address how the timestamp would be passed there.

src/state.rs
```diff
@@ -0,0 +1,106 @@
...
+
+#[allow(dead_code)]
+pub struct PendingConnection {
+    pub port: u16,
+    pub tcp_stream: TcpStream,
+}
+
+pub struct SyncHashMap<K, V>(std::sync::Mutex<HashMap<K, V>>);
+
+impl<K, V> SyncHashMap<K, V>
```
wkozyra95, line 36: We could use some lib, but here we are dealing with very small maps, so this approach is good enough, and this way we can customize it more for our needs.

src/state.rs
```diff
@@ -0,0 +1,106 @@
...
+            pending_connections: SyncHashMap::new(),
+            pipeline,
+        }
+    }
+
+    pub fn add_output(&self, port: u16, output_id: u32) -> Result<()> {
+        let pending_output = self
+            .pending_connections
+            .remove(&port)
+            .ok_or_else(|| anyhow!("no pending connection for port {}", port))?;
```
WojciechBarczynski, line 87: If we want to return some useful information in the HTTP response, maybe we should use [thiserror](https://docs.rs/thiserror/latest/thiserror/) here.

src/pipeline.rs
```diff
@@ -0,0 +1,136 @@
...
+                //
+                // let input_frames = pipeline.queue.next();
+                // let pipeline.render.render(output_frames);
+                // for let (output_id, frames) in input_frames {
+                //     self.on_output_data_received(output_id, frames)
+                // }
+                eprintln!("render loop");
+                thread::sleep(Duration::from_millis(1000));
+            }
+        });
```
WojciechBarczynski, line 101: In the future, we might want to implement different queuing strategies, that better fit certain use cases. With this render loop, queues have to produce frames in stable intervals, which might not always be the case. I think that this should be a queue responsibility to "push" frames further in the pipeline.

WojciechBarczynski (changes requested):

src/decoder.rs
```diff
@@ -0,0 +1,13 @@
...
+pub struct Decoder {}
+
+impl Decoder {
+    pub fn new() -> Self {
+        Decoder {}
+    }
+
+    pub fn decode(&self, _buffer: bytes::Bytes) -> Result<Vec<bytes::Bytes>> {
+        todo!()
+    }
```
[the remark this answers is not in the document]
wkozyra95: I would leave it as minimal as it is, I expect stuff like this to change when implementing it, so adding stuff like that now just adds work to future refactor.

src/state.rs
```diff
@@ -0,0 +1,106 @@
...
+    net::TcpStream,
+    sync::{Arc, Mutex},
+};
+
+use crate::{pipeline::Pipeline, tcp_connections};
+
+#[allow(dead_code)]
+pub struct Frame {
+    pub data: bytes::Bytes,
+}
```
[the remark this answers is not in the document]
WojciechBarczynski: It's useful to add here `resolution` as well. The `renderer` needs the information about frame resolution to upload them correctly to GPU.

src/pipeline.rs
wkozyra95: it is in this implementation, I expect that next call to be blocking.

WojciechBarczynski: I got it wrong. Now I understand what you meant here :) The `.next()` will block the `queue`

src/state.rs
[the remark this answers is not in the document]
wkozyra95: This is just a placeholder, I expect this type will be defined in the renderer or queu code

src/decoder.rs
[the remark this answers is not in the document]
WojciechBarczynski: Ok

WojciechBarczynski (approved):
LGTM

jerzywilczek (approved):
:rocket:

wkozyra95:
Closing in favor of https://github.com/membraneframework/video_compositor/pull/17

PR #10, opened 2023-07-03
Add common crate

jerzywilczek:
For now I opted for every scene node to have a manually specified
resolution. I think we can add other ways of specifying resolution later
on.

The transformation hold their parameters as `Box<dyn Any>` for now. This will possibly change in the future, when we'll work on deserializing the scene from JSON.

### DIFF
--- a/compositor_common/src/scene.rs
+++ b/compositor_common/src/scene.rs
@@ -1,10 +1,10 @@
-use std::{any::Any, collections::HashMap, rc::Rc};
+use std::{any::Any, collections::HashMap, sync::Arc};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct VideoId(usize);
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub struct TransformationRegistryKey(String);
+pub struct TransformationRegistryKey(pub String);
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct Resolution {
@@ -26,12 +26,12 @@ pub enum Node {
 
     Transformation {
         registry_key: TransformationRegistryKey,
-        inputs: HashMap<String, Rc<Node>>,
+        inputs: HashMap<String, Arc<Node>>,
         resolution: Resolution,
         params: Box<dyn Any>,
     },
 }
 
 pub struct Scene {
-    pub final_nodes: Vec<Rc<Node>>,
+    pub final_nodes: Vec<Arc<Node>>,
 }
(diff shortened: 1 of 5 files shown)


compositor_common/src/scene.rs
```diff
@@ -0,0 +1,37 @@
...
+    },
+
+    Image {
+        data: Vec<u8>,
+        resolution: Resolution,
+    },
+
+    Transformation {
+        registry_key: TransformationRegistryKey,
+        inputs: HashMap<String, Rc<Node>>,
```
wkozyra95, line 29: This object will be constructed in response to some API call, but values will be read inside the renderer (on a different thread), so unless you are planning to construct and keep this object only in render thread then everything here needs to an Arc

wkozyra95 (approved):

compositor_common/src/scene.rs
```diff
@@ -0,0 +1,37 @@
...
+pub struct VideoId(usize);
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct TransformationRegistryKey(String);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Resolution {
+    width: usize,
+    height: usize,
+}
```
WojciechBarczynski, line 13: Isn't `u32` more appropriate here? `Resolution` isn't just for array indexing.  

```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub struct Resolution {
    width: u32,
    height: u32,
}
```

WojciechBarczynski (approved):

compositor_common/src/scene.rs
jerzywilczek: yeah, but resolution is limited by your address space size, not by what fits in 32 bits

compositor_common/src/scene.rs
jerzywilczek: true, I looked at the `Arc`s I had here yesterday, thought they're not needed and replaced them with `Rc`s... 😅

compositor_common/src/scene.rs
wkozyra95: I'm pretty sure resolution is limited by sth a lot smaller than u32. In the purely semantic sense u32 or different fixed size is more correct because public api and value itself does not represent anything that should change with an architecture, on the other hand when using those values we will almost everywhere need usize, so it will be easier to use.

jerzywilczek: I meant that in our application it's limited by what we can download into ram, which is limited by ptr length. There probably is a much stricter limit on what wgpu can render, but on the CPU side this is going to be used mostly for buffer allocation.

PR #11, opened 2023-07-05
Add an initial implementation of the transformation registry

### LINKED ISSUE #8: Design "Scene" (processing graph) nodes API
TODO: Create Rust API for processing nodes. Specify their inputs, outputs and parameters.

jerzywilczek:
This patch adds an initial implementation of the transformation registry. It also defines a `Transformation` trait. implementing it allows adding custom transformations.

Since we only accept manually defined resolutions, the API is quite simple.

Closes #8 .

### DIFF
--- a/compositor_render/src/registry.rs
+++ b/compositor_render/src/registry.rs
@@ -1,0 +1,52 @@
+use std::collections::HashMap;
+
+use compositor_common::scene::TransformationRegistryKey;
+
+use crate::renderer::transformation::Transformation;
+
+#[derive(Debug, thiserror::Error)]
+pub enum GetError {
+    #[error("a transformation with a key {0} could not be found")]
+    KeyNotFound(String),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum RegisterError {
+    #[error("a transformation with a key {0} is already registered")]
+    KeyTaken(String),
+}
+
+pub struct TransformationRegistry {
+    registry: HashMap<TransformationRegistryKey, Box<dyn Transformation>>,
+}
+
+impl TransformationRegistry {
+    pub fn new() -> Self {
+        Self {
+            registry: HashMap::new(),
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn get(&self, key: &TransformationRegistryKey) -> Result<&dyn Transformation, GetError> {
+        match self.registry.get(key) {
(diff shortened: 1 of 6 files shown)


jerzywilczek:
I realized I should define the `Texture` struct before marking this 'ready to review'

compositor_render/src/renderer.rs
```diff
@@ -9,15 +18,32 @@ pub enum RendererNewError {
...
+            .register(Box::new(provider(self.wgpu_ctx.clone())))?;
+
+        Ok(())
+    }
 }
 
-struct WgpuCtx {
+pub struct WgpuCtx {
     #[allow(dead_code)]
     device: wgpu::Device,
```
noituri, line 48: I think the fields have to be public as well

noituri (changes requested):

compositor_render/src/renderer.rs
jerzywilczek: true

WojciechBarczynski (approved):
Aside from naming confusions addressed in the comment LGTM

compositor_render/src/renderer/transformation.rs
```diff
@@ -0,0 +1,22 @@
...
+pub enum TransformationParams {
+    String(String),
+    Binary(Vec<u8>),
+}
+
+pub trait Transformation: 'static {
+    fn apply(
+        &self,
+        params: &TransformationParams,
+        sources: &HashMap<String, Texture>,
```
WojciechBarczynski, line 17: I think this is more self-explanatory and gives us compile time checks.
```suggestion
        sources: &HashMap<NodeID, Texture>,
```

Obviously, we need to define `pub struct NodeID(String)` as well.

noituri (approved):

compositor_render/src/registry.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum RegisterError {
+    #[error("a transformation with a key {0} is already registered")]
+    KeyTaken(String),
+}
+
+pub struct TransformationRegistry {
+    registry: HashMap<TransformationRegistryKey, Box<dyn Transformation>>,
+}
```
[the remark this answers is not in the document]
jerzywilczek: I think `Transformation` and `Transformer` are fine

PR #17, opened 2023-07-10
Setup general application skeleton

wkozyra95:


compositor_common/src/frame.rs
```diff
@@ -0,0 +1,14 @@
+use crate::scene::Resolution;
+
+#[allow(dead_code)]
+pub struct Frame {
+    pub data: YuvData,
+    pub resolution: Resolution,
+    pub pts: i64,
```
WojciechBarczynski, line 7: Negative presentation timestamps are really weird here. I don't know exactly why FFmpeg starts producing negative frames, but I think we should fix it somehow. I found a similar issue with the patch solution described [here](https://stackoverflow.com/questions/60264279/ffmpeg-returns-negative-pts-and-dts-for-first-non-key-frame). 

Overall, we probably want to receive streams starting with PTS 0 and give users an option to add some sort of `pts_offset` (to synchronize newly added videos with ones added earlier). I think we should use `u64` as `pts` to make invalid states unrepresentable.

src/http.rs
```diff
@@ -0,0 +1,112 @@
...
+            }
+            Request::RegisterOutput(RegisterOutputRequest { port, resolution }) => {
+                self.state.register_output(port, resolution)
+            }
+            Request::Start => todo!(),
+        }
+    }
+    fn handle_request_before_init(&self, raw_request: &mut tiny_http::Request) -> Result<()> {
+        let request = Server::parse_request(raw_request)?;
+        match request {
```
WojciechBarczynski, line 76: [NIT] For code formatting coherency
```suggestion
            Request::Start => todo!(),
        }
    }
    
    fn handle_request_before_init(&self, raw_request: &mut tiny_http::Request) -> Result<()> {
        let request = Server::parse_request(raw_request)?;
        match request {
```

src/source.rs
```diff
@@ -0,0 +1,118 @@
...
+        let mut output_ctx =
+            format::output_as(&PathBuf::from(format!("rtp://127.0.0.1:{}", port)), "rtp")?;
+        let h264_codec = codec::encoder::find(Id::H264).unwrap();
+        let mut stream = output_ctx.add_stream(h264_codec)?;
+        unsafe {
+            (*(*stream.as_mut_ptr()).codecpar).codec_id = codec::Id::H264.into();
+        }
+
+        let mut encoder = Context::new().encoder().video()?;
+        encoder.set_time_base(Rational::new(1, 90000));
```
WojciechBarczynski, line 37: I have no idea why we use `Rational::new(1, 90000)` here. 
I think it should look like this:
```rust
        let some_self_explaing_name = Rational::new(1, 90000);
        encoder.set_time_base(some_self_explaing_name);
```

src/sink.rs
```diff
@@ -0,0 +1,101 @@
...
+                        continue;
+                    }
+                };
+                pipeline.push_input_data(port.into(), frame);
+            }
+        }
+
+        Ok(())
+    }
+}
```
WojciechBarczynski, line 83: Why do we define `decoder` in `RtpSink` and `encoder` in `RtpSource`? Should it be the other way around?
The `source` node is the first node in this pipeline (producer node), while the `sink` node is the last one.

src/source.rs
```diff
@@ -0,0 +1,118 @@
...
+    frame, Dictionary, Rational,
+};
+
+pub struct RtpSource {
+    #[allow(dead_code)]
+    port: u16,
+    sender: Sender<Frame>,
+}
+
+impl RtpSource {
```
WojciechBarczynski, line 19: Up

WojciechBarczynski (changes requested):

src/sink.rs
```diff
@@ -0,0 +1,101 @@
...
+fn frame_from_av(decoded: &mut Video) -> Result<Frame> {
+    if decoded.format() != format::pixel::Pixel::YUV420P {
+        return Err(anyhow!("only YUV420P is supported"));
+    }
+    Ok(Frame {
+        data: YuvData {
+            y_plane: bytes::Bytes::copy_from_slice(decoded.data(0)),
+            u_plane: bytes::Bytes::copy_from_slice(decoded.data(1)),
+            v_plane: bytes::Bytes::copy_from_slice(decoded.data(2)),
+        },
```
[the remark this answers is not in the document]
wkozyra95: data returns &[u8]. there is no way to take ownership of that data. To avoid this copy I would have to pass ffpmeg frame all the way to the renderer code, which would complicate api.

Note that I'm only copying the data, but memory for that frame is allocated only once, because ffmpeg frame is reused. Addi tonally this approach would allow us to change that in the future to have pool of allocated buffers. I didn't do any benchmarks, but I believe allocation is more expensive than copy.

src/source.rs
wkozyra95: Yes, you are right I got confused by gstreamer naming convention where:
- in case of pipeline it's first element is a source and result is a sink
- in case of single element it's input pads are sink pads and output pads are source past

RtpSender/RtpReceiver is probably a lot cleaner.

compositor_common/src/frame.rs
wkozyra95: I agree with the point that we should avoid negative values, but I think pts should still be represented by a signed integer.
- ffmpeg frame/packet is using a signed value for pts.
- If I understand correctly negative pts is valid (even though it's not desirable)
- It does not complicate offset logic

> Overall, we probably want to receive streams starting with PTS 0 

Should this be handled here or in queue? On the one hand probably it would be easier to do it here, but if want to use pipeline create as standalone composer in the future, it might make sense to move offset synchronizations to the queue, so we don't reimplement that.

src/sink.rs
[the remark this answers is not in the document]
WojciechBarczynski: Ok. It's not crucial at this point. We can leave it here as it is. 

Alternatively, in future, we can use a buffers pool or `renderer` that could have function like:
```rust
pub struct Planes {
    y_plane: &[u8],
    u_plane: &[u8],
    v_plane: &[u8]
}

pub fn wgpu_texture_from_slices(&self, planes: &Planes, resolution: Resolution) -> wgpu::Texture {...}
``` 
and we can keep `wgpu::Texture` later on (but we will be using a lot of VRAM with this solution).
If we create GPU decoding/encoding, then it'll replace it anyway.

WojciechBarczynski (approved):
GJ

compositor_common/src/frame.rs
WojciechBarczynski: In the queue, the code handling new streams will be the same. I don't understand what negative PTS should even mean in our case, so I would use `u64` here. Even though the queue will handle negative timestamps as well, I think that the output of the `RtpReceiver` should always return increasing timestamps starting from 0.

src/rtp_sender.rs
```diff
@@ -0,0 +1,119 @@
...
+                ("tune", "zerolatency"),
+                ("qcomp", "0.6"),
+                ("i_qfactor", "0.71"),
+                ("g", "250"),
+                ("me_range", "16"),
+                ("partitions", "+parti8x8+parti4x4+partp8x8+partb8x8"),
+            ]),
+        )?;
+        output_ctx.write_header()?;
+        let mut packet = codec::packet::Packet::empty();
```
jerzywilczek, line 59: don't we also need to do something like `packet.rescale_ts()`?

jerzywilczek (changes requested):
That's the stuff I found so far. We should talk about the decoder and encoder later.

compositor_pipeline/src/pipeline.rs
```diff
@@ -0,0 +1,69 @@
...
+
+pub trait PipelineOutput {
+    fn send_frame(&self, frame: Frame);
+}
+
+pub struct Pipeline<Source: PipelineOutput> {
+    outputs: SyncHashMap<u32, Arc<Source>>,
+    //queue: LiveQueue,
+    //renderer: Renderer,
+}
```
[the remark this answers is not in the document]
wkozyra95: Ahh sorry, I addressed it here https://github.com/membraneframework/video_compositor/pull/17#discussion_r1259996655, but forgoto to change the names in all the places

src/rtp_sender.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    fn run(port: u16, resolution: Resolution, receiver: Receiver<Frame>) -> Result<()> {
+        let mut output_ctx =
+            format::output_as(&PathBuf::from(format!("rtp://127.0.0.1:{}", port)), "rtp")?;
+        let h264_codec = codec::encoder::find(Id::H264).unwrap();
+        let mut stream = output_ctx.add_stream(h264_codec)?;
+        unsafe {
+            (*(*stream.as_mut_ptr()).codecpar).codec_id = codec::Id::H264.into();
+        }
```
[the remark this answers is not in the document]
wkozyra95: this required a lot of try an error :D We can talk about it more offline

src/rtp_sender.rs
wkozyra95: no, if the input timestamp has the same time_base as the output, and for video it should always be 90000.

src/rtp_sender.rs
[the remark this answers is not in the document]
jerzywilczek: maybe we should make an issue in ffmpeg-next

jerzywilczek (approved):
seems good

PR #21, opened 2023-07-11
Add web rendering

### LINKED ISSUE #9: Research HTML/CSS transformation node
TODO: Research possible technologies, to implement transformation node receiving frames and rendering them with custom HTML/CSS specifications.

noituri:
This pull request adds initial implementation of web rendering.
Rendering textures onto a canvas is not implemented yet.

Closes #9


compositor_render/src/transformations/web_renderer/packet.rs
```diff
@@ -0,0 +1,56 @@
...
+impl<'a> Packet<'a> {
+    pub fn send(&self, stream: &mut TcpStream) -> Result<(), PacketError> {
+        let header = (self.0.len() as u32).to_be_bytes();
+        let buffer = [&header, self.0].concat();
+        stream.write(&buffer).map_err(PacketError::SendFailure)?;
+
+        Ok(())
+    }
+
+    pub fn read(stream: &mut TcpStream) -> Result<Vec<u8>, PacketError> {
```
wkozyra95, line 17: This function could use some rewrite:
- better naming e.g. `len` of what?  msg?
- `4 - header.len() + buf_offset` avoid arithmetic like that, approach like that removes a lot of safety of rust.
- Use a range like `[buf_offset..n]` only when it's clear the the values are correct, if some more complicated logic is involved and it's not avoidable use function that will return result and won't panic
- consider using bytes package instead of Vec

In terms of more general code structure:
- impl on Packet might be responsible for serialization, but sending/reading from socket should not be part of the implementation. If you want to keep current signature of read/send function it should be on some wrapper around TcpStream struct
- When the last read is executed it can read the end of the current packet and the start of the new one. In this implementation, you are losing that information.

let me know if you are not sure how to approach it

web_renderer/src/common.ts
```diff
@@ -0,0 +1 @@
+export type Url = string;
```
wkozyra95, line 1: add newline (also I recommend to fix that in your editor settings)

web_renderer/src/packet.ts
```diff
@@ -0,0 +1,26 @@
...
+            (this.length & 0x000000ff)
+        ]);
+        
+        sock.write(Buffer.concat([header, this.message]))
+    }
+
+    public toString(): string {
+        return this.message.toString();
+    }
+}
```
wkozyra95, line 26: newline

web_renderer/src/session.ts
```diff
@@ -0,0 +1,49 @@
...
+        this.height = height;
+        
+        const factor = screen.getPrimaryDisplay().scaleFactor;
+        this.window.setSize(this.width / factor, this.height / factor);
+    }
+
+    public get frame(): Buffer {
+        return this.last_frame;
+    }
+}
```
wkozyra95, line 49: newline

compositor_render/src/transformations/web_renderer/packet.rs
```diff
@@ -0,0 +1,56 @@
...
+    io::{Read, Write},
+    net::TcpStream,
+};
+
+pub struct Packet<'a>(pub &'a [u8]);
+
+impl<'a> Packet<'a> {
+    pub fn send(&self, stream: &mut TcpStream) -> Result<(), PacketError> {
+        let header = (self.0.len() as u32).to_be_bytes();
+        let buffer = [&header, self.0].concat();
```
wkozyra95, line 11: unnecessary copy, you can just run write twice

compositor_render/src/transformations/web_renderer/packet.rs
```diff
@@ -0,0 +1,56 @@
...
+        stream.write(&buffer).map_err(PacketError::SendFailure)?;
+
+        Ok(())
+    }
+
+    pub fn read(stream: &mut TcpStream) -> Result<Vec<u8>, PacketError> {
+        let mut len = -1;
+        let mut header = Vec::with_capacity(4);
+        let mut msg = Vec::new();
+        let mut buf = [0u8; 65535];
```
wkozyra95, line 21: Correct me if I'm wrong, but I think this will allocate this buffer on the stack. It should be on the heap, it's too large for the stack, use vector instead.

noituri: As far as I know it should fit on the stack but I changed the method's implementation anyway

compositor_render/src/transformations/web_renderer/transformation.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    fn registry_key(&self) -> TransformationRegistryKey {
+        TransformationRegistryKey("web_renderer".to_owned())
+    }
+}
+
+impl WebRenderer {
+    pub fn new(wgpu_ctx: Rc<WgpuCtx>, port: u16) -> Result<Self, WebRendererNewError> {
+        let stream = PacketStream::connect(port)?;
```
jerzywilczek, line 49: Launch the electron app here using `Command`

jerzywilczek: or throw an error if node/npm/whatever is not installed.

jerzywilczek (changes requested):
ask for a rereview when ready, take this one comment for now

compositor_render/src/transformations/web_renderer/http_client.rs
```diff
@@ -0,0 +1,40 @@
...
+use compositor_common::scene::Resolution;
+use reqwest::blocking::Client;
+use serde::Serialize;
+
+pub struct HttpClient {
+    port: u16,
+    client: Client,
+}
+
+impl HttpClient {
```
wkozyra95, line 11: http client is sth that would have a method like get/post. This is a wrapper that is doing sth specific and it should be named appropriately .e.g  ElectronApi, WebRenderererClient, WebRenderererHttpClient

compositor_render/src/transformations/web_renderer/transformation.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    fn registry_key(&self) -> TransformationRegistryKey {
+        TransformationRegistryKey("web_renderer".to_owned())
+    }
+}
+
+impl Drop for WebRenderer {
+    fn drop(&mut self) {
+        self.renderer_process.kill().unwrap();
```
wkozyra95, line 62: log in case of error, if electron already crashed we don't want to panic

compositor_render/src/transformations/web_renderer/transformation.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+impl WebRenderer {
+    pub fn new(wgpu_ctx: Rc<WgpuCtx>) -> Result<Self, WebRendererNewError> {
+        let port: u16 = env::var("WEB_RENDERER_PORT")?.parse()?;
```
wkozyra95, line 68: why are we defining that port via env? 
- Easy approach would be to just hardcode the port.
- A bit more involved would be to pick free port and pass it to the electron when starting it.
- If we want to make it configurable, it should be passed inside our api, not via env.

web_renderer/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "productName": "web_renderer",
+  "version": "1.0.0",
+  "description": "OSR Web Renderer",
+  "main": ".webpack/main",
+  "scripts": {
+    "start": "electron-forge start",
+    "package": "electron-forge package",
+    "make": "electron-forge make",
+    "publish": "electron-forge publish",
+    "lint": "eslint --ext .ts,.tsx ."
```
wkozyra95, line 12: `npm run lint` returns some errors
Fix them and it would be good to add those checks to the CI (also check if tsc compiles without warnings)

web_renderer/src/server.ts
```diff
@@ -0,0 +1,53 @@
...
+        if (this.sessions.has(data.url)) {
+            session = this.sessions.get(data.url);
+        } else {
+            console.log(`Starting rendering for ${data.url}`)
+            session = new Session(data.url, data.resolution);
+            session.run();
+            this.sessions.set(data.url, session);
+        }
+
+        if (session.resolution.width != data.resolution.width ||
```
wkozyra95, line 41: we will not change output resolution in the middle of the sessions

web_renderer/src/server.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    public listen(port: number): void {
+        this.server.listen(port, () => {
+            console.log(`Listening on ${port}`);
+        });
+    }
+
+    private initRoutes(): void {
+        // TODO: better endpoint name
+        this.server.post("/get_frame", this.get_frame.bind(this));
```
wkozyra95, line 25: I would prefer it if api rendering and creating session was a separate call. The logic when you create session if it does not exists seems a bit hacky

web_renderer/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+    const port = parseInt(process.argv[2]);
+    startServer(port);
+}
+
+app.whenReady().then(main);
+app.on('window-all-closed', () => {
+    if (process.platform !== 'darwin') {
+        app.quit();
+    }
+});
```
WojciechBarczynski:
```suggestion
app.on('window-all-closed', () => {
    app.quit();
});
```

We want to close electron on mac as well.

web_renderer/src/server.ts
```diff
@@ -0,0 +1,71 @@
...
+
+app.use(express.json());
+
+app.post("/new_session", (req: Request, res: Response<NewSessionResponse>) => {
+    try {
+        const data = NewSessionRequest.parse(req.body)
+        console.log(`Starting rendering for ${data.url}`)
+        const session_id = randomUUID();
+        const session = new Session(data.url, data.resolution);
+        sessions.set(session_id, session);
```
WojciechBarczynski, line 19: Shouldn't we check if `session_id` is already used?

web_renderer/src/server.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const app = express();
+
+app.use(express.json());
+
+app.post("/new_session", (req: Request, res: Response<NewSessionResponse>) => {
+    try {
+        const data = NewSessionRequest.parse(req.body)
+        console.log(`Starting rendering for ${data.url}`)
+        const session_id = randomUUID();
```
WojciechBarczynski, line 17: [NIT]
For consistency use semicolons everywhere or nowhere.

WojciechBarczynski (approved):
GJ 🚀

web_renderer/src/index.ts
WojciechBarczynski: Fixed

web_renderer/src/server.ts
WojciechBarczynski: Improbable

compositor_render/src/transformations/web_renderer/mod.rs
```diff
@@ -0,0 +1,8 @@
+mod electron_api;
+pub mod transformation;
+
+pub use transformation::*;
+
+type Url = String;
+
+type SessionId = String;
```
wkozyra95, line 8: check https://github.com/membraneframework/video_compositor/pull/22#discussion_r1271925086 - similiar reasoning applies in this case.

I would suggest:
- SessionId either keep as just a String or create new type if you want to have a better type safety
- Url either keep String or maybe check for some already existing Url representations in rust.

compositor_render/src/transformations/web_renderer/transformation.rs
```diff
@@ -0,0 +1,176 @@
...
+        _sources: &HashMap<NodeId, Texture>,
+        target: &Texture,
+    ) -> Result<(), Box<dyn Error>> {
+        let TransformationParams::String(url) = params else {
+            return Err(Box::new(WebRendererApplyError::InvalidParam));
+        };
+
+        let mut session_ids = self.session_ids.borrow_mut();
+        let session_id = match session_ids.get(url) {
+            Some(id) => id.to_owned(),
```
wkozyra95, line 47: Do you need to clone it? Wouldn't a borrowing session id be enough?

Also if you have a case where you need to clone string it might be more efficient to use Arc<str>/Rc<str> instead. I did not benchmark this, but I suspect creating ref counted pointer is cheaper than allocating memorry

compositor_render/src/transformations/web_renderer/transformation.rs
```diff
@@ -0,0 +1,176 @@
...
+            None => {
+                let size = target.size();
+                let id = self.api.new_session(
+                    url.to_owned(),
```
wkozyra95, line 51: the same as above, borrow should be enough

web_renderer/src/server.ts
```diff
@@ -0,0 +1,71 @@
...
+import { sessions } from './state';
+
+const app = express();
+
+app.use(express.json());
+
+app.post("/new_session", (req: Request, res: Response<NewSessionResponse>) => {
+    try {
+        const data = NewSessionRequest.parse(req.body);
+        console.log(`Starting rendering for ${data.url}`);
```
wkozyra95, line 16: it's fine for now, but keep in mind that if we stay with this approach we will need real logger

web_renderer/src/server.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+});
+
+export function startServer(port: number): void {
+    app.listen(port, () => {
+        console.log(`Listening on ${port}`);
+    });
+}
+
+type NewSessionResponse =
```
wkozyra95, line 61: usually types are placed at the top of the file, or just before endpoint implementation they relate to

web_renderer/src/common.ts
```diff
@@ -0,0 +1,3 @@
...
```
wkozyra95, line 1: the same issue as in rust side

web_renderer/src/server.ts
```diff
@@ -0,0 +1,72 @@
...
+interface ErrorResponse {
+    error: string
+}
+
+type NewSessionResponse =
+    { session_id: SessionId }
+    | ErrorResponse;
+
+type GetFrameResponse =
+    Buffer
```
wkozyra95, line 18:
```suggestion
    | Buffer
```

web_renderer/src/server.ts
```diff
@@ -0,0 +1,72 @@
...
+import { randomUUID } from 'crypto';
+import { GetFrameRequest, NewSessionRequest } from './schemas';
+import { sessions } from './state';
+
+interface ErrorResponse {
+    error: string
+}
+
+type NewSessionResponse =
+    { session_id: SessionId }
```
wkozyra95, line 14:
```suggestion
    | { session_id: SessionId }
```

web_renderer/src/server.ts
```diff
@@ -0,0 +1,72 @@
...
+        session.run();
+
+        res
+            .status(HttpConstants.HTTP_STATUS_CREATED)
+            .send({
+                session_id: session_id
+            });
+    } catch (err) {
+        res
+            .status(HttpConstants.HTTP_STATUS_BAD_REQUEST)
```
wkozyra95, line 42: let's leave for now, but usually you are just throwing an error and add middleware that will convert it to the response

wkozyra95 (approved):

compositor_render/src/transformations/web_renderer/transformation.rs
```diff
@@ -0,0 +1,177 @@
...
+        let renderer_process = Command::new("npm")
+            .args(["run", "start", "--", "--", &port.to_string()])
+            .current_dir(web_renderer_path)
+            .spawn()
+            .map_err(WebRendererNewError::WebRendererInitError)?;
+
+        Ok(renderer_process)
+    }
+
+    fn upload(&self, data: &[u8], target: &Texture) -> Result<(), WebRendererApplyError> {
```
jerzywilczek, line 124: It's not clear 'which way' this method uploads. It could be uploading to the electron instance or to the GPU. The name of the method should state which way the upload goes.

compositor_render/src/transformations/web_renderer/transformation.rs
```diff
@@ -0,0 +1,177 @@
...
+        let size = target.size();
+        let img = image::load_from_memory(data)?;
+
+        self.wgpu_ctx.queue.write_texture(
```
jerzywilczek, line 128: You could do a safety check whether the image has the correct resolution before this.

jerzywilczek (approved):
Seems ok, but shouldn't this wait for #30 to avoid changing the API few days after merging?

wkozyra95:
> https://github.com/membraneframework/video_compositor/pull/30 to avoid changing the API few days after merging?

Given that I'm introducing the most breaking changes, I prefer for this PR and queue one to land first and address everything in my PR.

PR #22, opened 2023-07-13
Live queue

WojciechBarczynski:
Adds live frames queueing


compositor_pipeline/src/queue.rs
```diff
@@ -0,0 +1,94 @@
...
+            check_queue_receiver.recv().unwrap();
+            Self::start_ticker(tick_duration, check_queue_sender);
+
+            let start = Instant::now();
+            loop {
+                let mut internal_queue = internal_queue.lock().unwrap();
+                let buffer_pts = internal_queue.get_next_output_buffer_pts();
+
+                if start.elapsed() > buffer_pts || internal_queue.check_all_inputs_ready(buffer_pts)
+                {
```
wkozyra95, line 63:
```suggestion
                if start.elapsed() > buffer_pts || internal_queue.check_all_inputs_ready(buffer_pts) {
```

compositor_common/src/frame.rs
```diff
@@ -1,15 +1,48 @@
...
+pub struct Framerate(pub u32);
+
+impl Framerate {
+    pub fn get_interval_duration(self) -> Duration {
+        Duration::from_nanos((1_000_000_000 / self.0).into())
+    }
+}
+
+#[derive(Debug)]
+pub struct FramesBatch {
```
wkozyra95, line 27: I think a common crate should describe types that are used everywhere, but this value describes queue output and renderer input, so I think it should be defined in the renderer crate.

Also, the name is technically correct, but I feel it's a bit misleading, when I hear batch I understand there would be a list of consecutive frames for single input "batched" together. Maybe consider InputFrames, FrameSet or maybe using rust naming InstantFrameSet?

compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -0,0 +1,136 @@
...
+        match self.inputs_queues.get_mut(&input_id) {
+            Some(input_queue) => {
+                if let Vacant(e) = self.timestamp_offsets.entry(input_id) {
+                    e.insert(frame.pts);
+                }
+
+                frame.pts += *self
+                    .timestamp_offsets
+                    .get(&input_id)
+                    .ok_or_else(|| anyhow!("Timestamp offset unregistered for pad {input_id}"))
```
wkozyra95, line 52: no strong opinion, but I think we wanted to use anyhow only in top-level crates, and stick to precise errors in lib packages

compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -0,0 +1,136 @@
...
+        self.inputs_queues.remove(&input_id);
+        self.timestamp_offsets.remove(&input_id);
+    }
+
+    pub fn enqueue_frame(&mut self, input_id: InputID, mut frame: Frame) -> Result<(), QueueError> {
+        match self.inputs_queues.get_mut(&input_id) {
+            Some(input_queue) => {
+                if let Vacant(e) = self.timestamp_offsets.entry(input_id) {
+                    e.insert(frame.pts);
+                }
```
wkozyra95, line 47:
```suggestion
                let offset = self.timestamp_offsets.entry(input_id).or_insert(frame.pts);
```
to avoid that unwrap bellow

compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub fn remove_input(&mut self, input_id: InputID) {
+        self.inputs_queues.remove(&input_id);
+        self.timestamp_offsets.remove(&input_id);
+    }
+
+    pub fn enqueue_frame(&mut self, input_id: InputID, mut frame: Frame) -> Result<(), QueueError> {
+        match self.inputs_queues.get_mut(&input_id) {
+            Some(input_queue) => {
+                if let Vacant(e) = self.timestamp_offsets.entry(input_id) {
+                    e.insert(frame.pts);
```
wkozyra95, line 46: this does not seem correct, offset should be calculated based on a difference between get_next_output_buffer_pts and frame.pts. In current implementation first frame will always have pts 0, so offset will always be zero. Sorry if I'm missing sth.

compositor_pipeline/src/queue.rs
```diff
@@ -0,0 +1,94 @@
...
+            // Wait for first frame
+            check_queue_receiver.recv().unwrap();
+            Self::start_ticker(tick_duration, check_queue_sender);
+
+            let start = Instant::now();
+            loop {
+                let mut internal_queue = internal_queue.lock().unwrap();
+                let buffer_pts = internal_queue.get_next_output_buffer_pts();
+
+                if start.elapsed() > buffer_pts || internal_queue.check_all_inputs_ready(buffer_pts)
```
wkozyra95, line 62: I don't see anything related to the additional buffer we want to keep. shouldn't this be 
start.elapsed() + fixed_buffer_duration > buffer_pts

or I guess this could be taken into account when enqueueing

wkozyra95 (commented):
I don't see anything related to the

compositor_pipeline/src/queue/internal_queue.rs
WojciechBarczynski: Obviously, my bad. Fixed it

compositor_common/src/frame.rs
```diff
@@ -1,15 +1,48 @@
...
     pub resolution: Resolution,
     pub pts: Duration,
 }
 
+pub type Pts = Duration;
+pub type InputID = u32;
+
+/// TODO: This should be a rational.
+#[derive(Debug, Clone, Copy)]
+pub struct Framerate(pub u32);
```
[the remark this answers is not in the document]
WojciechBarczynski: I suppose, that you wanted to keep the duration between two frames as framerate (e.g. `1/30s` with `30fps`).

Theoretically, this causes rounding error accumulation in the `pts` calculation of the next `"input frame batches"` (With some framerates, the exact interval between two frames has no precise representation with natural numbers, e.g. `30fps`, `60fps`). Using framerate directly, we can avoid this error accumulation with:
```rust
pub struct InternalQueue {
    ...
    output_framerate: Framerate,
    pub send_batches_counter: u32,
}

impl InternalQueue {
...

    pub fn get_next_output_buffer_pts(&self) -> Duration {
        Duration::from_secs_f64(self.send_batches_counter as f64 / self.output_framerate.0 as f64)
    }
}
```

In reality, it shouldn't be a problem, since with each frame we would accumulate at most one nanosecond of error, which would be hardly noticeable even after a few days of constant streaming.

Aside from that, I think it's simpler to keep framerate as `u32` or some `rational`, than as `Duration`. With `Duration` it isn't obvious to me what it should represent. I only guessed that you meant the interval between the next two frames. 
I wouldn't use `Duration` here mostly for clarity.

compositor_pipeline/src/queue.rs
WojciechBarczynski: Fixed

compositor_pipeline/src/queue.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub fn start(&self, sender: Sender<InputFrames>) {
+        let (check_queue_sender, check_queue_receiver) = self.check_queue_channel.clone();
+        let internal_queue = self.internal_queue.clone();
+        let tick_duration = self.output_framerate.get_interval_duration();
+        let buffer_duration = self.buffer_duration;
+
+        thread::spawn(move || {
+            // Wait for first frame
+            check_queue_receiver.recv().unwrap();
+            sleep(buffer_duration);
```
wkozyra95, line 56: why we need sleep?

compositor_pipeline/src/queue.rs
```diff
@@ -0,0 +1,97 @@
...
+            Self::start_ticker(tick_duration, check_queue_sender);
+
+            let start = Instant::now();
+            loop {
```
wkozyra95, line 60: this is busy loop, you missed reading from channel. This would explain that 100% cpu usage you posted on slack :D

compositor_pipeline/src/queue.rs
```diff
@@ -0,0 +1,97 @@
...
+    #[allow(dead_code)]
+    pub fn remove_input(&self, input_id: InputId) {
+        let mut internal_queue = self.internal_queue.lock().unwrap();
+        // TODO: gracefully remove input - wait until last enqueued frame PTS is smaller than output PTS
+        internal_queue.remove_input(input_id);
+    }
+
+    #[allow(dead_code)]
+    pub fn start(&self, sender: Sender<InputFrames>) {
+        let (check_queue_sender, check_queue_receiver) = self.check_queue_channel.clone();
```
wkozyra95, line 48: you only use that channel for ticker, but it should also send event after enqueue

WojciechBarczynski: Are you sure? I think that cloned `crossbeam channel` will work as indented here, based on [this example](https://docs.rs/crossbeam-channel/latest/crossbeam_channel/struct.Sender.html#examples). 

I modified the code a little bit to check that and according to logs it should work.
```rust
#[derive(Debug)]
enum Event {
    Tick,
    Enqueue
}

pub struct Queue {
    internal_queue: Arc<Mutex<InternalQueue>>,
    check_queue_channel: (Sender<Event>, Receiver<Event>),
    output_framerate: Framerate,
    buffer_duration: Duration,
}

impl Queue {
    ...

    pub fn start(&self, sender: Sender<InputFrames>) {
        let (check_queue_sender, check_queue_receiver) = self.check_queue_channel.clone();
        let internal_queue = self.internal_queue.clone();
        let tick_duration = self.output_framerate.get_interval_duration();
        let buffer_duration = self.buffer_duration;

        thread::spawn(move || {
            // Wait for first frame
            check_queue_receiver.recv().unwrap();
            Self::start_ticker(tick_duration, check_queue_sender);

            let start = Instant::now();
            loop {
                let event = check_queue_receiver.recv().unwrap();
                info!("Render loop {:#?}", event);

                let mut internal_queue = internal_queue.lock().unwrap();
                let buffer_pts = internal_queue.get_next_output_buffer_pts();

                if start.elapsed() + buffer_duration > buffer_pts
                    || internal_queue.check_all_inputs_ready(buffer_pts)
                {
                    let frames_batch = internal_queue.get_frames_batch(buffer_pts);

                    sender.send(frames_batch).unwrap();

                    internal_queue.drop_useless_frames();
                }
            }
        });
    }

    pub fn enqueue_frame(&self, input_id: InputId, frame: Frame) -> Result<(), QueueError> {
        let mut internal_queue = self.internal_queue.lock().unwrap();

        internal_queue.enqueue_frame(input_id, frame)?;
        internal_queue.drop_pad_useless_frames(input_id)?;

        self.check_queue_channel.0.send(Event::Enqueue).unwrap();

        Ok(())
    }

    fn start_ticker(tick_duration: Duration, check_queue_sender: Sender<Event>) {
        thread::spawn(move || {
            let ticker = tick(tick_duration);
            loop {
                ticker.recv().unwrap();
                check_queue_sender.send(Event::Tick).unwrap();
            }
        });
    }
}
```

logs:
```rust
[2023-07-24T19:13:59Z INFO  compositor_pipeline::queue] Render loop Tick
[2023-07-24T19:13:59Z INFO  compositor_pipeline::queue] Render loop Enqueue
[2023-07-24T19:13:59Z INFO  compositor_pipeline::queue] Render loop Tick
[2023-07-24T19:13:59Z INFO  compositor_pipeline::queue] Render loop Enqueue
[2023-07-24T19:13:59Z INFO  compositor_pipeline::queue] Render loop Tick
[2023-07-24T19:13:59Z INFO  compositor_pipeline::queue] Render loop Enqueue
[2023-07-24T19:13:59Z INFO  compositor_pipeline::queue] Render loop Tick
[2023-07-24T19:13:59Z INFO  compositor_pipeline::queue] Render loop Tick
[2023-07-24T19:13:59Z INFO  compositor_pipeline::queue] Render loop Enqueue
[2023-07-24T19:14:00Z INFO  compositor_pipeline::queue] Render loop Tick
[2023-07-24T19:14:00Z INFO  compositor_pipeline::queue] Render loop Enqueue
[2023-07-24T19:14:00Z INFO  compositor_pipeline::queue] Render loop Tick

```

compositor_pipeline/src/queue.rs
WojciechBarczynski: You are right. I totally forgot about that. It's great that you catch it 😃 I will run those benchmarks once again. Now that 100% CPU usage makes sense 😛

compositor_pipeline/src/queue.rs
WojciechBarczynski: I put it there for the ticker to match enforced buffer pushes (when `start.elapsed() + buffer_duration > buffer_pts` is true, because some inputs haven't pushed frames yet), but putting it here probably is a bad idea. 

</br>

Do you think that it should be placed in `"ticker thread"` like this:
```rust
    fn start_ticker(
        tick_duration: Duration,
        check_queue_sender: Sender<()>,
        buffer_duration: Duration,
    ) {
        thread::spawn(move || {
            sleep(buffer_duration);
            let ticker = tick(tick_duration);
            loop {
                ticker.recv().unwrap();
                check_queue_sender.send(()).unwrap();
            }
        });
    }
```

or maybe we should ignore it, since those first ticks won't `"push"` frames out of the queue, because of `start.elapsed() + buffer_duration > buffer_pts` condition?
I think leaving it in the `"ticker thread"` is the more appropriate solution.

compositor_pipeline/src/queue.rs
WojciechBarczynski: Well, even without this loop constantly running, encoding Big Buck Bunny in 4k 30fps was "a little bit" too much for M1 pro 😄
Screenshot 2023-07-24 at 21 51 33

compositor_pipeline/src/queue.rs
wkozyra95: ahh sorry, I was checking where check_queue_sender is passed and did not realize that you are cloning it there not creating

compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -0,0 +1,142 @@
...
+                    best_diff_frame_index = index;
+                    best_diff_nanos = diff;
+                } else {
+                    break;
+                }
+            }
+
+            if best_diff_frame_index > 0 {
+                input_queue.drain(0..best_diff_frame_index);
+            }
```
wkozyra95, line 118:
```suggestion
            input_queue.drain(0..best_diff_frame_index);
```

compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -0,0 +1,142 @@
...
+                .as_nanos()
+                .abs_diff(next_output_buffer_nanos);
+
+            for (index, frame) in input_queue.iter().enumerate() {
+                let diff = frame.pts.as_nanos().abs_diff(next_output_buffer_nanos);
+                if diff <= best_diff_nanos {
+                    best_diff_frame_index = index;
+                    best_diff_nanos = diff;
+                } else {
+                    break;
```
wkozyra95, line 112: I assume you can break here because you know that there is no point to check further?

compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub fn drop_pad_useless_frames(&mut self, input_id: InputId) -> Result<(), QueueError> {
+        let next_output_buffer_nanos = self.get_next_output_buffer_pts().as_nanos();
+
+        let input_queue = self
+            .inputs_queues
+            .get_mut(&input_id)
+            .ok_or(QueueError::UnknownInputId(input_id))?;
+
+        if let Some(first_frame) = input_queue.first() {
+            let mut best_diff_frame_index = 0;
```
wkozyra95, line 100: In general I have mixed feelings into pushing functional approach, but consider sth like this

```
let (best_diff_frame_index, best_diff_nanos) = input_queue
  .iter()
  .enumerate()
  .min_by_key(|(index, frame)|{
      frame.pts.as_nanos().abs_diff(next_output_buffer_nanos)
   })
   .unwrap_or((0, first_frame));
```

compositor_pipeline/src/queue.rs
wkozyra95: I don't really like the way the queue is stored and copied, but I don't have any better idea on how to handle it.

wkozyra95 (approved):

compositor_render/src/renderer.rs
```diff
@@ -57,12 +60,20 @@ impl Renderer {
...
+            Some(frame) => {
+                let frame = Frame {
+                    data: frame.data.clone(),
+                    resolution: frame.resolution,
+                    pts: inputs.pts,
+                };
+                Ok(frame)
+            }
+            None => Err(RendererRenderError::NoInput(0)),
+        }
```
jerzywilczek, line 76: Doesn't this do exactly the same thing?

compositor_render/src/renderer.rs
```diff
@@ -1,16 +1,19 @@
...
+};
 
 use self::transformation::Transformation;
 
 pub mod texture;
 pub mod transformation;
 
 pub struct Renderer {
-    wgpu_ctx: Rc<WgpuCtx>,
+    wgpu_ctx: Arc<WgpuCtx>,
```
jerzywilczek, line 16: I don't see why this is necessary

compositor_render/src/renderer.rs
jerzywilczek: But now with two clones instead of one?

jerzywilczek (changes requested):
Most of these are minor changes, but I do want some questions answered :cowboy_hat_face:

compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        frames_batch
+    }
+
+    /// Checks if all inputs have frames closest to buffer_pts.
+    /// Every input queue should have frame with larger or equal pts than buffer pts.
+    /// We assume, that queue receives frames with monotonically increasing timestamps,
+    /// so when all inputs queues have frame with pts larger or equal than buffer timestamp,
+    /// queue won't receive frame with pts "closer" to buffer pts.
+    /// In other cases, queue might receive frame "closer" to buffer pts in future.
```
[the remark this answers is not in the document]
WojciechBarczynski: It's not about this assumption. The phrase `In other cases` is rather about how we can't be certain that we won't receive any frame "closer" to buffer pts on any input, without having received a `frame with pts larger or equal than buffer timestamp` on every input.

compositor_render/src/renderer.rs
WojciechBarczynski: > Doesn't this do exactly the same thing?

Now the frame has assigned pts of `InputFrames` (the batch). Without it, we would send to the encoder a frame with pts determined by the encoder. If the queue will send the same frame twice, this will pass it to the encoder, which will. crash on `non monotonically increasing timestamps`.

I haven't optimized this part, since it will be completely changed anyway.

compositor_render/src/renderer.rs
WojciechBarczynski: It was useful when I wanted to add `Renderer` directly to the pipeline, but since it's not done here in that way, we can keep it as `Rc`.

compositor_pipeline/src/queue/internal_queue.rs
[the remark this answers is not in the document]
WojciechBarczynski: I modified the doc, with more clear and straight forward explanation.

compositor_render/src/renderer.rs
jerzywilczek: oh, I see. makes sense.

PR #27, opened 2023-07-17
Add long running example

wkozyra95:
Added a long-running example to see how the compositor behaves when running for longer periods.

I  noticed one issue for 4K streams encoder is keeping up only on the ultrafast preset. In all the other cases the number of messages in a channel grows steadily increasing memory usage.

I decided to start dropping the oldest frames when a number of messages in the queue is over 100 as a stop-gap solution, but when queue and renderer will be implemented this needs to be handled better.

### Potential solutions I considered

1. Current approach (dropping frames in encoder): Easiest to implement and in most cases it gives the intended results, but it does not solve that issue for other places (e.g. what if renderer will not keep up). Also if new streams is pushing more data at the beginning we can stop dropping frames too early even if the encoder could handle the average traffic.

2. Using bound channel in encoder: This would just push the problem to the queue. It's hard to say now if this would be a good thing before queue is ready. 

3. Using bound channels(or other blocking mechanisms) before encoder and queue: It would block decoding if the rest of the pipeline can't handle new data, so it's most optimal because we would buffer data before decoding. The main problem is that buffering on udp channel is handled internally in libav, so we have no control over the buffer that stores received udp data and how it handles dropping them.

With approaches 2 and 3 we technically risking that we will stay more and more behind, but the queue will already handle frames that are too old, and drop them.

### DIFF
--- a/examples/common/common.rs
+++ b/examples/common/common.rs
@@ -22,6 +22,7 @@ pub fn write_example_sdp_file(port: u16) -> Result<String> {
                     m=video {} RTP/AVP 96\n\
                     a=rtpmap:96 H264/90000\n\
                     a=fmtp:96 packetization-mode=1\n\
+                    a=rtcp-mux\n\
                 ",
             port
         )
--- a/examples/common/common.rs
+++ b/examples/common/common.rs
@@ -43,6 +43,7 @@ pub fn post<T: Serialize + ?Sized>(json: &T) -> Result<Response> {
     Ok(response)
 }
 
+#[allow(dead_code)]
 pub fn download(url: &str, destination: &Path) -> Result<()> {
     let mut resp = reqwest::blocking::get(url)?;
     let mut out = File::create(destination)?;
@@ -50,6 +51,7 @@ pub fn download(url: &str, destination: &Path) -> Result<()> {
     Ok(())
 }
 
+#[allow(dead_code)]
 pub fn ensure_downloaded(url: &str, destination: &Path) -> Result<()> {
     if destination.exists() {
         return Ok(());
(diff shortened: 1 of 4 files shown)


examples/common/common.rs
```diff
@@ -43,13 +43,15 @@ pub fn post<T: Serialize + ?Sized>(json: &T) -> Result<Response> {
     Ok(response)
 }
 
+#[allow(dead_code)]
```
wkozyra95, line 46: this is needed because file is imported in 2 examples, and one of them is not using some functions.

This was the only way I saw to share code between examples without creating a new crate.

examples/long_ffmpeg.rs
```diff
@@ -0,0 +1,82 @@
+use anyhow::Result;
```
wkozyra95, line 1: mostly copy of existing example, but instead of mp4 it sends infinite test source stream

src/rtp_sender.rs
```diff
@@ -102,7 +102,10 @@ impl RtpSender {
 
     fn run(opts: Options, receiver: Receiver<Frame>) -> Result<()> {
         let mut output_ctx = format::output_as(
-            &PathBuf::from(format!("rtp://127.0.0.1:{}", opts.port)),
+            &PathBuf::from(format!(
+                "rtp://127.0.0.1:{}?rtcpport={}",
```
wkozyra95, line 106: rtcp was not muxed on the same port before

examples/common/common.rs
jerzywilczek: sadly this is also the only way I know of.

jerzywilczek (approved):
Makes sense, I think dropping in the decoder is OK for now

WojciechBarczynski (approved):

PR #29, opened 2023-07-18
[ci] Trigger checks only when needed

wkozyra95:
Trigger checks on PR to any branch or push to master. (push to branch without PR will not trigger CI now)
When some checks are in progress and new changes are pushed the old workflow will be cancelled (it does not apply to commits on master)


jerzywilczek (approved):
v cool

PR #31, opened 2023-07-20
Design proposal for handling dynamic shader parameter amounts.

jerzywilczek:


design_docs/2023_07_20_dynamic_shader_parameter_amount.md
```diff
@@ -0,0 +1,55 @@
+# Passing variables to the shader
+We need to be able to pass an arbitrary number of textures and parameters to the custom shaders. For parameters, we need to be able to pass some "header parameters" -- constants used in all videos, as well as some per-video data.
```
WojciechBarczynski, line 2: I don't like the phrase `"header parameters"`. First what comes to my mind is the request header, which has nothing to do with `constants used in all videos`.

Maybe we should use sth like:
- `transformation params`
- `video params`

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
```diff
@@ -0,0 +1,55 @@
...
+Binding const-sized uniform/storage buffers works everywhere too. This would work approximately like this in the shaders (ofc the len can be defined in a different way, e.g. as the length of a runtime-sized texture array):
+```wgsl
+struct MyList {
+	// you declare an array with a constant, large size
+	per_video_data: array<MyStruct, 64>
+	// and then in some way specify how much of this buffer is filled
+	per_video_data_len: u32
+}
+
+@group(0) @binding(0) var<uniform> per_video_data: MyList
```
WojciechBarczynski, line 19: Change names `MyList`, `per_video_data`, and `per_video_data_len` to more descriptive ones.

WojciechBarczynski, line 19: Same bind group and binding as before. I think we should specify in this doc what bind groups and bindings will be used (in which one texture, video parameters, output frame pts, etc. will be passed)

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
```diff
@@ -0,0 +1,55 @@
+# Passing variables to the shader
+We need to be able to pass an arbitrary number of textures and parameters to the custom shaders. For parameters, we need to be able to pass some "header parameters" -- constants used in all videos, as well as some per-video data.
+
+### Possible solutions
```
WojciechBarczynski, line 4: [NIT]
```suggestion
## Possible solutions
```

https://github.com/DavidAnson/markdownlint/blob/v0.27.0/doc/md001.md

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
WojciechBarczynski, line 1: [NIT]
IMO we should use some markdown lint (like https://github.com/DavidAnson/vscode-markdownlint), to keep consistency between those design docs and use common linting.

WojciechBarczynski (changes requested):

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
jerzywilczek: I'm against doing that, since this is supposed to represent the users capabilities of defining their own structs and using them here, which I feel these names demonstrate pretty well

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
jerzywilczek: good idea

jerzywilczek: or we could have a template markdown document with the headings ready which you would copy and just fill in

jerzywilczek: or both

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
wkozyra95: I think it's not clear here how passing user-defined parameters and textures from other nodes fit together. Based on this description I understand that:
- solution 1 - Textures are passed totally separate from params, example code is only addressing passing textures. 
- soultion 2 
  - If this is user-defined I don't see where to texture is passed. Is it MyStruct? If yes is texture + params or just texture?
  - If It's user-defined how can distinguish which values in the struct are params that the user should specify and which are buffers/textures with image data? How would we construct those structs on the rust side? (this applies to all solutions after this one too)

jerzywilczek: Yeah, this is a little bit unclear. I'll try to add a section explaining this better.

Textures are indeed passed separately from params, in a different bind group. The full set of bindings the user declares would look approximately like this:
```wgsl
@group(0) @binding(0) var textures: texture_2d_array<f32>;
@group(1) @binding(0) var<uniform> users_custom_parameters: UsersCustomStruct;
// this can include, for example, the timestamp of this frame 
@group(1) @binding(1) var<uniform> parameters_received_from_the_compositor: OurStruct;
@group(2) @binding(0) var sampler_: sampler;
```

jerzywilczek: what do you think of this:
```wgsl
@group(0) @binding(0) var textures: texture_2d_array<f32>;
@group(1) @binding(0) var<uniform> users_custom_parameters: UsersCustomStruct;
// this can include, for example, the timestamp of this frame 
@group(1) @binding(1) var<uniform> parameters_received_from_the_compositor: OurStruct;
@group(2) @binding(0) var sampler_: sampler;
```

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
```diff
@@ -0,0 +1,55 @@
...
+- We need to be able to allocate an appropriately sized buffer for the parameters. This means that every node has to have it's own buffer. This is not a big problem, since the buffers are small, especially compared to textures.
+- Careful error handling is necessary, since if the user gives us a buffer that is too short or misaligned the `wgpu`'s default reaction is to panic the process (there are mechanisms that change this behavior -- error scopes, we just need to remember to use them).
+- It's becoming more and more apparent that we will need to expose some pipeline configuration options to the shader defining API, so that shaders can define the background color, or the formula used for alpha blending etc.
+- We need to decide whether we want to use storage or uniform buffers for the parameters. Uniform buffers are faster, but limited in size to 64 KiB. Storage buffers are a bit slower, but support a much more generous 2 GiB of max size. Changing this later would be a breaking change, since the custom shaders would need to replace `var<uniform>` with `var<storage>` in their shader code. Do we want to allow the users to choose which kind of buffer they want to use?
```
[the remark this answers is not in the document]
jerzywilczek: It needs to be defined on both sides

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
WojciechBarczynski: LGTM

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+### Proposed solution
+
+Use the constant-sized texture arrays. Use const-sized uniforms, since not being compatible with Metal would harm development.
+
+This document proposes the following bind groups structure:
+
+```wgsl
+// The textures rendered by previous nodes
```
wkozyra95, line 78: So if I understand correctly end result of what user provides would have:
- This entire code excatlly like it is here, but potentially different variable names(group and binding are defining what is what)
- User would need to copy CompositorBuffers definition from somewhere, or will that be injected by us in the source?
- UserDefinedStructs will be only in code, and we will parse the structure from shader source

Is that correct?

wkozyra95 (approved):

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
jerzywilczek: c) is definitely true. For a) and b) I was thinking of some solutions, but figured those were out of scope of this design doc. In particular, the rust-based game engine, [bevy](https://bevyengine.org/) created a [shader preprocessor](https://lib.rs/crates/naga_oil) for wgpu shaders. The preprocessor allows cool stuff like `#import` directives, which we could use for declaring bind groups and the `CompositorBuffers` structure. I only found out about this preprocessor late yesterday, I will add some mention of it to the doc today.

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
```diff
@@ -0,0 +1,102 @@
...
+
+This is also (sort of) possible using the first solution with a const limit on the video amount:
+
+```wgsl
+struct MyStruct {
+    some_param: u32,
+    some_other_param: f32,
+    per_video_data: array<MyVideoData, 64>
+}
+```
```
WojciechBarczynski, line 69: What about memory alignment here? Wouldn't it require adding some extra "padding"? https://www.w3.org/TR/WGSL/#memory-layouts

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
```diff
@@ -0,0 +1,102 @@
...
+### Shader parameters
+
+Binding const-sized uniform/storage buffers works on every backend. This would work approximately like this in the shaders (ofc the len can be defined in a different way, e.g. as the length of the texture array):
+
+```wgsl
+struct MyList {
+    // you declare an array with a constant, large size
+    per_video_data: array<MyStruct, 64>
+    // and then in some way specify how much of this buffer is filled
+    per_video_data_len: u32
```
WojciechBarczynski, line 37: 1. From the comment, I get the impression, that this represents how many structs are in the array, while this variable name suggests something like a single `MyStruct` size/len. Maybe something like `videos_count` would be better?
2. Wouldn't wgpu memory layout force us to fill up this array with exactly `sizeof(MyStruct) * 64` bytes? Have you tried passing it as uniform without filling this list up entirely?

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
jerzywilczek: depends on the alignment of the `MyVideoData`, but potentially yes. The users will have to remember about this, it's not something that we can do for them.

design_docs/2023_07_20_dynamic_shader_parameter_amount.md
jerzywilczek: I did try and it works, but the array should be the last element in the struct. Then we can only send, for example, the first half of the buffer.

jerzywilczek: Also this is an example struct, not the API proposal for all users to use, each shader will declare it's own structs.

WojciechBarczynski (approved):

jerzywilczek:
> I'm not sure if it is outside of this design's scope but it's unclear to me in what order the textures would be passed to the shader. The order would be important for things such us transparency.
 
The order has to be entirely user-specified somewhere

PR #37, opened 2023-07-31
Format conversion YUV <-> RGBA

wkozyra95:
This PR adds:
- conversion from YUV and RGBA
- conversion from RGBA to YUV
- some additional abstractions to wrap Textures
- fix simple_ffmpeg example

Code for interacting with wgpu is almost 1:1 from membrane plugin, only repackaged with some additional abstraction layers.

Tested on both examples:
- simple_ffmpeg (pass input into output)
- long_ffmpeg (use web renderer output)

Next steps:
- rework uploading and downloading between cpu<->gpu
- rescale frame on resolution change
- texture.rs is growing, so it's also time to brake it out into multiple modules

### DIFF
--- a/compositor_render/src/renderer.rs
+++ b/compositor_render/src/renderer.rs
@@ -1,13 +1,16 @@
-use std::{collections::HashMap, sync::Arc};
+use std::sync::Arc;
 
 use compositor_common::{
     scene::{InputId, OutputId, SceneSpec},
     transformation::{TransformationRegistryKey, TransformationSpec},
-    Frame,
 };
 use log::error;
 
-use crate::{frame_set::FrameSet, registry::TransformationRegistry, render_loop::populate_inputs};
+use crate::{
+    frame_set::FrameSet,
+    registry::TransformationRegistry,
+    render_loop::{populate_inputs, read_outputs},
+};
 use crate::{
     registry::{self, RegistryType},
     render_loop::run_transforms,
@@ -19,8 +22,14 @@ use crate::{
     },
 };
 
-use self::scene::{Scene, SceneUpdateError};
+use self::{
+    color_converter_pipeline::{RGBAToYUVConverter, YUVToRGBAConverter},
+    scene::{Scene, SceneUpdateError},
+    texture::{RGBATexture, YUVTextures},
+};
 
+mod color_converter_pipeline;
+mod common_pipeline;
 pub mod scene;
 pub mod texture;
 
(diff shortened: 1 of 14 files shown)


compositor_render/src/renderer.rs
```diff
@@ -187,7 +186,19 @@ impl WgpuCtx {
...
         ))?;
 
-        Ok(Self { device, queue })
+        let yuv_bind_group_layout = YUVTextures::new_bind_group_layout(&device);
+        let rgba_bind_group_layout = RGBATexture::new_bind_group_layout(&device);
+        let yuv_to_rgba_converter = YUVToRGBAConverter::new(&device, &yuv_bind_group_layout);
+        let rgba_to_yuv_converter = RGBAToYUVConverter::new(&device, &rgba_bind_group_layout);
+
+        Ok(Self {
+            device,
```
wkozyra95, line 195: I'm not a fan of dumping everything in context like that, but for now I don't have any better idea

compositor_render/src/renderer/common_pipeline.rs
```diff
@@ -0,0 +1,164 @@
...
+    };
+}
+
+pub struct RectangleRenderBuffers {
+    pub vertex: Buffer,
+    pub index: Buffer,
+}
+
+/// Vertex and index buffer that describe render area as an rectangle mapped to texture.
+impl RectangleRenderBuffers {
```
wkozyra95, line 24: Couldn't find a better name for the vertex and index buffer together, and as far as I can see wgpu does not have any concept named render buffer, so hopefully it won't be confusing

compositor_render/src/renderer/color_converter_pipeline.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub fn new(
+        device: &wgpu::Device,
+        yuv_textures_bind_group_layout: &wgpu::BindGroupLayout,
+    ) -> Self {
+        let shader_module = device.create_shader_module(wgpu::include_wgsl!("yuv_to_rgba.wgsl"));
+        let sampler = Sampler::new(device);
+        let buffers = RectangleRenderBuffers::new(device);
+
+        let pipeline_layout = device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
+            label: Some("YUV to RGBA colour converter render pipeline layout"),
```
jerzywilczek, line 33:
```suggestion
            label: Some("YUV to RGBA color converter render pipeline layout"),
```
Let's get rid of the British spelling that elixir used to enforce on us

compositor_render/src/renderer/color_converter_pipeline.rs
```diff
@@ -0,0 +1,236 @@
...
+                render_pass.set_bind_group(2, &self.plane_selector.bind_group, &[]);
+                render_pass.set_vertex_buffer(0, self.buffers.vertex.slice(..));
+                render_pass.set_index_buffer(
+                    self.buffers.index.slice(..),
+                    RectangleRenderBuffers::INDEX_FORMAT,
+                );
+                render_pass.draw_indexed(0..RectangleRenderBuffers::INDICES.len() as u32, 0, 0..1);
+            }
+
+            ctx.queue.submit(Some(encoder.finish()));
```
jerzywilczek, line 233: This is suboptimal, can you add a TODO mentioning this should be optimized?

compositor_render/src/renderer/common_pipeline.rs
```diff
@@ -0,0 +1,164 @@
...
+            bind_group_layout,
+        }
+    }
+}
+
+pub struct U32Uniform {
+    pub buffer: Buffer,
+    pub bind_group: wgpu::BindGroup,
+    pub bind_group_layout: wgpu::BindGroupLayout,
+}
```
jerzywilczek, line 122:
```suggestion
// TODO: This should be done with push-constants, not with a buffer
pub struct U32Uniform {
    pub buffer: Buffer,
    pub bind_group: wgpu::BindGroup,
    pub bind_group_layout: wgpu::BindGroupLayout,
}
```

compositor_render/src/renderer/texture.rs
```diff
@@ -126,8 +72,157 @@ impl Texture {
...
+            visibility: wgpu::ShaderStages::FRAGMENT,
+            count: None,
+        };
+        device.create_bind_group_layout(&wgpu::BindGroupLayoutDescriptor {
+            label: Some("yuv all textures bind group layout"),
+            entries: &[create_entry(0), create_entry(1), create_entry(2)],
+        })
+    }
+
+    pub(self) fn new_bind_group(
```
jerzywilczek, line 117:
```suggestion
    fn new_bind_group(
```
`pub(self)` is equivalent to no `pub` at all, I believe

jerzywilczek (approved):
Looks good for now.

compositor_render/src/renderer/texture.rs
wkozyra95: it makes sense if you have more than one module in the same file, for some reason I assumed that you need it to access private fields in the same file, but this does not seem to be necessary

WojciechBarczynski (approved):
GJ 🚀

PR #39, opened 2023-08-01
Generic transformation

jerzywilczek:
This PR adds working shader nodes. They don't accept parameters for now, but, as demonstrated in the provided example, work!


compositor_render/src/transformations/shader/pipeline.rs
```diff
@@ -0,0 +1,181 @@
...
+            }],
+        });
+
+        let uniforms_bgl = device.create_bind_group_layout(&wgpu::BindGroupLayoutDescriptor {
+            label: Some("uniforms bgl"),
+            entries: &[
+                wgpu::BindGroupLayoutEntry {
+                    binding: 0,
+                    count: None,
+                    visibility: wgpu::ShaderStages::FRAGMENT,
```
noituri, line 156: I think we may want to make it visible in vertex shader as well

compositor_render/examples/silly.wgsl
```diff
@@ -0,0 +1,49 @@
...
+    output.tex_coords = input.tex_coords;
+
+    return output;
+}
+
+struct CustomStruct {
+    a: u32,
+}
+
+struct SomeStruct {
```
noituri, line 25: Maybe consider using more descriptive name like `CompositorData`

compositor_render/src/transformations/shader/pipeline.rs
```diff
@@ -0,0 +1,181 @@
...
+                    ty: wgpu::BindingType::Buffer {
+                        has_dynamic_offset: false,
+                        min_binding_size: None,
+                        ty: wgpu::BufferBindingType::Uniform,
+                    },
+                },
+                wgpu::BindGroupLayoutEntry {
+                    binding: 1,
+                    count: None,
+                    visibility: wgpu::ShaderStages::FRAGMENT,
```
noituri, line 166: Correct me if I'm wrong, but this too might be useful in vertex shader

noituri (commented):
Overall lgtm

compositor_render/examples/silly_shader.rs
```diff
@@ -0,0 +1,115 @@
...
+    std::process::Command::new("ffmpeg")
+        .arg("-i")
+        .arg(path.as_ref().as_os_str())
+        .arg("input.yuv")
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .spawn()
+        .unwrap()
+        .wait()
+        .unwrap();
```
WojciechBarczynski, line 21: It's better to use some rust lib for this, but if this is only a temporary example, you don't have to do that.

WojciechBarczynski (approved):

compositor_render/examples/crab.jpg
wkozyra95, line 1: if the image is hosted somewhere consider fetching it from public url instead, it's small enough that it does not really matter, but there is already helper function in examples that you could use for that

compositor_render/examples/silly_shader.rs
jerzywilczek: I figured that since we're already using ffmpeg in other examples it's simpler to do this than to include a dev dependency

compositor_render/examples/crab.jpg
jerzywilczek: yeah, but it's in examples of the main crate and I feel this example should be in the renderer crate. This may be lazy, but I think we can leave the picture in the repo, especially that it's in the public domain and free of copyright

wkozyra95: ok I missed that it's not in the main crate, make sense

PR #41, opened 2023-08-02
Add Dockerfile

wkozyra95:
- Add dockerfile for composer
- Add `ip` field to `register_output` request.
- Add CI job to run linter on dockerfile and try to build the image.
- Add example that builds docker image, starts a container and streams using ffmpeg from outside
  - I don't expect this example to work on macOS
  - If example is killed with Ctr-C then it's leaving terminal in bad state, so I often had to kill terminal window or run `tput reset`
- Few fixes:
  - pipeline start was only called in examples, so cargo run wouldn't work
  - allow overriding web renderer path using env variable  

There are some problems with electron rendering, but given that we are going to replace it with chrome, so I left if for now (images sometimes show, but full page is never loaded plus there is a bunch libva and egl initiazation errors in the terminal).
  
20230803_12h20m43s_grim

But shaders work fine

20230802_13h28m17s_grim


WojciechBarczynski (approved):

jerzywilczek (approved):
Looks ok, but I'm no expert in docker

noituri (approved):

PR #43, opened 2023-08-02
Text renderer

wojtechb:
Adds basic text renderer.

Text renderer always renders text starting from the right-left corner. Currently, it doesn't support text wrapping (longer texts are simply clipped).

Specifying text [width](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass) and [weight](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass) is not supported due to rendering artifacts.

Examples:
`cargo run -r --example text_renderer`

Questions for reviewers:
1. Are there any other features that the text renderer should support?
2. How to handle text alignment in the API? Since now we want to render text starting from the top right corner, how the user should be able to e.g. align text to the middle of the screen? Doing it with other transformations seems painful.

### DIFF
--- a/compositor_common/src/scene.rs
+++ b/compositor_common/src/scene.rs
@@ -96,7 +96,7 @@ pub enum Style {
 #[derive(Serialize, Deserialize, Clone)]
 #[serde(tag = "type")]
 pub struct TextParams {
-    pub content: String,
+    pub content: Arc<str>,
     /// in pixels
     pub font_size: f32,
     /// default: white (255, 255, 255, 255)
--- a/compositor_render/src/transformations/text_renderer.rs
+++ b/compositor_render/src/transformations/text_renderer.rs
@@ -1,3 +1,5 @@
+use std::sync::{Arc, Mutex};
+
 use compositor_common::scene::TextParams;
 use glyphon::{
     AttrsOwned, Buffer, Color, FontSystem, Metrics, Shaping, SwashCache, TextArea, TextAtlas,
@@ -13,7 +15,7 @@ use crate::renderer::{texture::NodeTexture, RenderCtx};
 
 #[allow(dead_code)]
 pub struct TextSpec {
-    content: String,
+    content: Arc<str>,
     attributes: AttrsOwned,
     font_size: f32,
     line_height: f32,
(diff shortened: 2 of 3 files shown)


compositor_common/src/scene.rs
```diff
@@ -81,3 +84,28 @@ pub enum ShaderParams {
...
+pub enum Style {
+    Normal,
+    Italic,
+    Oblique,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(tag = "type")]
+pub struct TextParams {
+    pub content: String,
```
wkozyra95, line 99:
```suggestion
    pub content: Arc<str>,
```

examples/text_renderer.rs
```diff
@@ -0,0 +1,133 @@
...
+                "input_id": "input 1",
+                "resolution": { "width": VIDEO_RESOLUTION.width, "height": VIDEO_RESOLUTION.height },
+            }
+        ],
+        "transforms": [
+           {
+                "node_id": "text_renderer",
+                "type": "text_renderer",
+                "text_params": {
+                    "content": "VideoCompositor🚀\nSecond Line",
```
wkozyra95, line 97: not a strong opinion, but if you don't need to nest that, maybe flatten text_params into parent object

compositor_render/src/transformations/text_renderer.rs
```diff
@@ -0,0 +1,197 @@
...
+
+impl Default for TextRendererCtx {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[allow(dead_code)]
+pub struct TextRenderer {
+    was_rendered: bool,
```
wkozyra95, line 88:
```suggestion
    was_rendered: Mutex<bool>,
```
maybe? It would make for a cleaner api

wkozyra95 (approved):

examples/text_renderer.rs
WojciechBarczynski: I would leave it in this form because it clearly separates type-specific values from values used in all nodes.

In `long_ffmpeg` example we used `shader_params`  as:
```
"transforms": [
           {
               "node_id": "side-by-side",
               "type": "shader",
               "shader_id": "example shader",
               "shader_params": {
                   "example": {"type": "string", "value": "param"},
               },
               "input_pads": [
                   "input 1",
               ],
               "resolution": { "width": VIDEO_RESOLUTION.width, "height": VIDEO_RESOLUTION.height },
           },
...
]
```
so flattening it here would be inconsistent in my opinion.

wkozyra95:
```
           {
                "node_id": "text_renderer",
                "type": "text_renderer",
                "text": "VideoCompositor🚀\nSecond Line",
                "font_size": 100.0,
                "font_family": "Comic Sans MS",

                // mode 1
                "mode": "wrap",
                "width": 1080,
                "max-height": 200, // optional
                "justify-content": "center " | "left" | "right",

                // mode 2
                "mode": "single-line",
                "max-width": 1080, //optional
                "overflow": "hidden" | "font-rescale" | "suffix", // actual suffix could be
                                                                  // additonal parm
                // mode 3 (would adjust font size and wrapping to fit text best)
                "mode": "best-fit",
                "resolution": { "width": VIDEO_RESOLUTION.width, "height": VIDEO_RESOLUTION.height },
                
                // mode 4 (single line, adds padding to fit resolution - currently this is implemented)
                "mode": "padded",
                "resolution": { "width": VIDEO_RESOLUTION.width, "height": VIDEO_RESOLUTION.height },

                "input_pads": [],
           }
```

Just a few random ideas for an API, I'm not saying everything there makes sense, but it might be a good starting point

PR #45, opened 2023-08-02
Shader parameter json format design doc

jerzywilczek:
I need some feedback before I start implementing this

### DIFF
--- a/design_docs/2023_08_02_shader_parameter_format.md
+++ b/design_docs/2023_08_02_shader_parameter_format.md
@@ -1,6 +1,36 @@
 # Shader parameter format
 
-What kind of format should we use for transferring shader parameters in json? I've been thinking about two designs.
+We need to transfer shader parameters in JSON. This document defines the transfer format.
+
+## Format definition
+
+The full format can be generated with the following rust type definitions and serde macros:
+
+```rust
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(tag = "type", rename_all = "snake_case", content = "value")]
+pub enum ShaderParams {
+    F32(f32),
+    U32(u32),
+    I32(i32),
+    List(Vec<ShaderParams>),
+    Struct(Vec<ShaderParamStructField>),
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct ShaderParamStructField {
+    pub field_name: String,
+    #[serde(flatten)]
+    pub value: ShaderParams,
+}
+```
+
+Additionally:
+
+- all `wgsl` `vecN` types are represented as lists
+- all `wgsl` `matNxM` types are represented as lists of lists
+
+## Format demonstration
 
 Let's look at a single type that we would like to represent. This is what it looks like in wgsl:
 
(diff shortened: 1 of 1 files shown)


design_docs/2023_08_02_shader_parameter_format.md
```diff
@@ -0,0 +1,100 @@
...
+            "type": "u32",
+            "val": 42
+        }
+    ]
+}
+```
+
+There are varying degrees of the strictness of this approach, in this example I chose to treat `vec4` as a `"list"`. `vec2`, `vec3` and `array` would probably all be considered a `"list"`, but in a more strict version of this approach all of them could be separate types. The more annoying thing about this approach is the necessary validation. We should validate that all values in a list type are of the same type. Using more strict typing we would have to enforce that the `vec`s and matrices all have correct amounts of values and don't include non-primitive types.
+
+What format should we choose?
```
WojciechBarczynski, line 100: IMO second version looks cleaner and adding run-time type validations to it will be easier

design_docs/2023_08_02_shader_parameter_format.md
```diff
@@ -0,0 +1,100 @@
...
+## Second idea
+
+We include information about the structure of the data:
+
+```json
+{
+    "type": "struct",
+    "val": [
+        {
+            "type": "list",
```
wkozyra95, line 66:
```suggestion
            "type": "list",
            "name": "field_name"
```
We don't need field names, but it would be useful for validation and error messages. Relaying on order of the fields only would be hard for users

design_docs/2023_08_02_shader_parameter_format.md
```diff
@@ -0,0 +1,100 @@
...
+
+        {
+            "type": "u32",
+            "val": 42
+        }
+    ]
+}
+```
+
+There are varying degrees of the strictness of this approach, in this example I chose to treat `vec4` as a `"list"`. `vec2`, `vec3` and `array` would probably all be considered a `"list"`, but in a more strict version of this approach all of them could be separate types. The more annoying thing about this approach is the necessary validation. We should validate that all values in a list type are of the same type. Using more strict typing we would have to enforce that the `vec`s and matrices all have correct amounts of values and don't include non-primitive types.
```
wkozyra95, line 98: agree, list is better, but we still need to do the same validation based on shader sources, so it does not help us there

design_docs/2023_08_02_shader_parameter_format.md
```diff
@@ -0,0 +1,100 @@
...
+            "type": "list",
+            "val": [
+                {
+                    "type": "f32",
+                    "val": 0.0
+                },
+
+                {
+                    "type": "f32",
+                    "val": 0.0
```
wkozyra95, line 75: I know it's just an example, but I would avoid abbreviations like val and just put `value`/`values`

wkozyra95 (approved):

noituri (approved):

wkozyra95 (approved):

noituri (approved):

WojciechBarczynski (approved):

PR #49, opened 2023-08-03
Add a possibility to pass parameters to shaders.

jerzywilczek:
This includes custom shader parameters and parameters provided by the compositor.


noituri (approved):

compositor_render/src/render_loop.rs
```diff
@@ -82,6 +82,18 @@ pub(super) fn run_transforms(ctx: &RenderCtx, scene: &Scene) {
     }
 }
 
+pub(super) fn update_global_buffers(ctx: &RenderCtx, pts: Duration) {
```
wkozyra95, line 85: You would prefer to keep render_loop.rs without any logic for specific type of transformation. Can you move this method to shader.rs?

Also, it's not a strong opinion, but maybe name sth more generic? e.g. Shader::prepare_new_render_loop or sth like that? We could use consistent naming for transforms that also need to do sth at the beginning of the loop.

compositor_render/src/renderer.rs
```diff
@@ -184,13 +189,45 @@ impl WgpuCtx {
         let yuv_to_rgba_converter = YUVToRGBAConverter::new(&device, &yuv_bind_group_layout);
         let rgba_to_yuv_converter = RGBAToYUVConverter::new(&device, &rgba_bind_group_layout);
 
+        let shader_parameters_bind_group_layout = Shader::new_parameters_bind_group_layout(&device);
+
+        let global_shader_parameters_buffer = device.create_buffer(&wgpu::BufferDescriptor {
```
wkozyra95, line 194: global is a bit misleading, values here are global to a single render loop or scene, so maybe
- render_ctx_shader_paramaters_buffer
- render_ctx_parameters_buffer (shader in this context is a bit redundant)
- render_loop_paramaters_buffer

examples/long_ffmpeg.rs
```diff
@@ -117,7 +117,8 @@ fn start_example_client_code() -> Result<()> {
                "type": "shader",
                "shader_id": "example shader",
                "shader_params": {
-                   "example": {"type": "string", "value": "param"},
+                   "type": "u32",
```
wkozyra95, line 120: remember to rebase and add that fix to docker example too

wkozyra95 (approved):

compositor_render/src/transformations/shader/pipeline.rs
```diff
@@ -10,24 +10,36 @@ use super::INPUT_TEXTURES_AMOUNT;
...
+        });
+
         let pipeline_layout = device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
             label: Some("shader transformation pipeline layout"),
-            bind_group_layouts: &[
-                &bgls.textures_bgl,
-                &bgls.uniforms_bgl,
-                &sampler.bind_group_layout,
-            ],
+            bind_group_layouts: &[&textures_bgl, uniforms_bgl, &sampler.bind_group_layout],
```
WojciechBarczynski, line 42: What if the shader doesn't require any parameters? We should allow to register shaders without any params. What do you think?

jerzywilczek: I want to address that in a subsequent PR

WojciechBarczynski (approved):

compositor_render/src/render_loop.rs
jerzywilczek: So you're saying I should remove this function from `render_loop` and place it in `Shader`, but still call `Shader::prepare_new_render_loop` in the renderer? Do I understand correctly, or did you mean something else?

wkozyra95: yes

PR #53, opened 2023-08-04
Decouple transform registration from render thread

wkozyra95:
Currently, transform registrations are blocking the render thread, this PR separates it out, so registering long transforms does not affect rendering piepline


noituri (approved):
lgtm

WojciechBarczynski (approved):

PR #54, opened 2023-08-04
Add scene validation checks: cycles detection, repeated node ids and unused nodes checks

WojciechBarczynski:


compositor_common/src/validators.rs
```diff
@@ -98,4 +115,269 @@ impl SceneSpec {
...
+        if !unused_transforms.is_empty() {
+            return Err(SpecValidationError::UnusedNodes(unused_transforms));
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod test {
```
wkozyra95, line 243: this will grow to much, move it to separate file

compositor_common/src/validators.rs
```diff
@@ -98,4 +115,269 @@ impl SceneSpec {
...
+                &mut visited,
+            )?;
+        }
+
+        Ok(())
+    }
+
+    fn validate_node_ids_uniqueness(
+        &self,
+        defined_node_ids: &Vec<NodeId>,
```
wkozyra95, line 172: this could be an iterator instead

compositor_common/src/validators.rs
```diff
@@ -98,4 +115,269 @@ impl SceneSpec {
...
+    ) -> Result<(), SpecValidationError> {
+        let mut nodes_ids: HashSet<NodeId> = HashSet::new();
```
wkozyra95, line 174:
```suggestion
        let mut nodes_ids: HashSet<&NodeId> = HashSet::new();
```
clone only when returning error, but there is no need for that to just keep track of which nodes were visited

compositor_common/src/validators.rs
```diff
@@ -98,4 +115,269 @@ impl SceneSpec {
...
+        for node_id in defined_node_ids {
+            match nodes_ids.get(node_id) {
+                Some(_) => {
+                    return Err(SpecValidationError::DuplicateNames(node_id.0.clone()));
+                }
+                None => {
+                    nodes_ids.insert(node_id.clone());
+                }
+            }
+        }
```
wkozyra95, line 185:
```suggestion
        for node_id in defined_node_ids {
             if nodes_ids.contains(node_id) {
                    return Err(SpecValidationError::DuplicateNames(node_id.0.clone()));
             }
             nodes_ids.insert(node_id);
        }
```

wkozyra95 (approved):

compositor_common/src/validators.rs
WojciechBarczynski: I modified those functions' lifetimes and removed unnecessary cloning

compositor_common/src/validators.rs
```diff
@@ -98,4 +115,125 @@ impl SceneSpec {
...
+            }
+            nodes_ids.insert(node_id);
+        }
+
+        Ok(())
+    }
+
+    fn validate_nodes_are_used(
+        &self,
+        input_nodes: &HashSet<NodeId>,
```
wkozyra95, line 188: the name here is input_nodes, but you are passing ids of registered inputs (they are not the same). Make sure to use consistent naming in other places too

But it's also not just naming, I think what we are validating here is also inconsistent. C
- You can have registered input that is not used
- You shouldn't have an input node that is not used

but this code ignores if input nodes are unused

WojciechBarczynski: Good point. Fixed that

wkozyra95 (approved):

compositor_common/src/validators.rs
```diff
@@ -98,4 +122,134 @@ impl SceneSpec {
...
+
+        let mut visited: HashMap<&NodeId, NodeState> = HashMap::new();
+
+        fn visit<'a>(
+            node: &'a NodeId,
+            transform_nodes: &'a HashMap<&NodeId, &TransformNodeSpec>,
+            inputs: &HashSet<NodeId>,
+            visited: &mut HashMap<&'a NodeId, NodeState>,
+        ) -> Result<(), SpecValidationError> {
+            if let Some(_input) = inputs.get(node) {
```
jerzywilczek, line 144:
```suggestion
            if inputs.get(node).is_some() {
```

compositor_common/src/validators.rs
```diff
@@ -98,4 +122,134 @@ impl SceneSpec {
...
+        &self,
+        defined_node_ids: I,
+    ) -> Result<(), SpecValidationError> {
+        let mut nodes_ids: HashSet<&NodeId> = HashSet::new();
+
+        for node_id in defined_node_ids {
+            if nodes_ids.contains(&node_id) {
+                return Err(SpecValidationError::DuplicateNames(node_id.0.clone()));
+            }
+            nodes_ids.insert(node_id);
```
jerzywilczek, line 187:
```suggestion
            if !nodes_ids.insert(node_id) {
                return Err(SpecValidationError::DuplicateNames(node_id.0.clone()));
            }
```

`insert` returns whether the insertion actually happened or the value was already in the set

compositor_common/src/validators.rs
```diff
@@ -98,4 +122,134 @@ impl SceneSpec {
...
+
+        let mut unused_transforms: HashSet<Arc<str>> = HashSet::new();
+
+        let nodes_iter = input_nodes.keys().chain(transform_nodes.keys());
+
+        for node in nodes_iter {
+            if !visited.contains(node) {
+                unused_transforms.insert(node.0.clone());
+            }
+        }
```
jerzywilczek, line 244: You could do this using `HashSet::difference` instead

jerzywilczek (approved):
only cosmetic stuff

PR #58, opened 2023-08-07
Port electron web renderer to CEF

### LINKED ISSUE #46: Implement web renderer using CEF

noituri:
In this PR I reimplemented the web rendering using CEF.

Closes #46


compositor_render/src/transformations/web_renderer/browser.rs
```diff
@@ -0,0 +1,73 @@
...
+    resolution: Resolution,
+}
+
+impl cef::RenderHandler for RenderHandler {
+    fn resolution(&self, _browser: &cef::Browser) -> Resolution {
+        self.resolution
+    }
+
+    fn on_paint(&self, _browser: &cef::Browser, buffer: &[u8], _resolution: Resolution) {
+        if !self.frame_tx.is_full() {
```
wkozyra95, line 60: - This might not be bug if chrome has some guarantees about how on_paint is called, but regardless of that we should not be relying on logic like that. The problem I see here is that is_full can return false, but in the meantime other thread could write sth to that channel and this write would block.
- The other issue is that you are not sending the latest possible frame. This would probably give at most 1 frame, but it's not necessary if chrome is able to keep up, so we should not do it.

You could create an unbounded channel and when reading frames in the render drain it entirely and use last frame

compositor_pipeline/src/pipeline.rs
```diff
@@ -37,7 +39,7 @@ pub struct Pipeline<Input: PipelineInput, Output: PipelineOutput> {
 #[derive(Serialize, Deserialize)]
 pub struct Options {
     pub framerate: Framerate,
-    pub init_web_renderer: Option<bool>,
+    pub web_renderer: Option<WebRendererOptions>,
```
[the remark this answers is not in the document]
noituri: I had to use additional `#[serde(default)]` to make it work without `Option`

compositor_render/src/event_loop.rs
```diff
@@ -0,0 +1,46 @@
...
+            return Err(EventLoopRunError::WrongThread);
+        }
+
+        ctx.run_message_loop();
+        Ok(())
+    }
+
+    /// Fallback event loop used when web renderer is disabled
+    fn fallback_event_loop(&self) {
+        let mut signals = Signals::new([consts::SIGINT]).unwrap();
```
wkozyra95, line 37: I missed that before, but neither the renderer nor the pipeline crate should depend on signal_hook. To be clear this is the correct place for the Event loop struct, but not for the code that handles interrupts.

compositor_render/src/renderer.rs
```diff
@@ -197,8 +209,10 @@ impl WgpuCtx {
 
         let yuv_bind_group_layout = YUVTextures::new_bind_group_layout(&device);
         let rgba_bind_group_layout = RGBATexture::new_bind_group_layout(&device);
+        let bgra_bind_group_layout = BGRATexture::new_bind_group_layout(&device);
```
wkozyra95, line 212: let's leave it for now, but those specific layouts could have been part of the chrome code

compositor_render/src/transformations/web_renderer.rs
```diff
@@ -1,86 +1,94 @@
...
-                expected: Resolution {
-                    width: size.width as usize,
-                    height: size.height as usize,
-                },
-                received: Resolution {
-                    width: img.width() as usize,
-                    height: img.height() as usize,
-                },
-            });
+        if !frame.is_empty() {
```
wkozyra95, line 84: I know that it was like that before, but
- Why it would be empty?
- Can we refactor the code so it can't be empty?
- What would be rendered if this happens?
- If we can't do anything about it, at least use Option to signal when the value is missing instead of empty byte array

compositor_render/src/transformations/web_renderer/browser.rs
```diff
@@ -0,0 +1,76 @@
...
+            frame_data: Vec::new(),
+        }
+    }
+
+    pub fn retrieve_frame(&mut self) -> &[u8] {
+        if let Some(frame) = self.painted_frames_receiver.try_iter().last() {
+            self.frame_data = frame;
+        }
+
+        &self.frame_data
```
wkozyra95, line 23:
```suggestion
        self.frame_data = self.painted_frames_receiver.try_iter().last().or(self.frame_data);
        &self.frame_data
```

compositor_render/src/transformations/web_renderer.rs
noituri: On second thought, there is a chance that chromium would take longer time to startup a website so it's possible that compositor requests a render when the frame is not yet ready.

compositor_render/src/transformations/web_renderer/browser.rs
noituri: I don't think it will work. `self.frame_data` is behind mutable reference and `.or()`  takes ownership

Cargo.toml
```diff
@@ -33,12 +34,13 @@ serde = { workspace = true }
...
 ffmpeg-next = "6.0.0"
 crossbeam-channel = { workspace = true }
 env_logger = { workspace = true }
-log = { workspace = true } 
+log = { workspace = true }
+signal-hook = { workspace = true }
 
 [dev-dependencies]
 reqwest = { workspace = true }
+signal-hook = { workspace = true }
```
wkozyra95, line 46: if sth is in dependencies already, does it need to be in dev dependencies too?

wkozyra95 (approved):

Cargo.toml
noituri: It's not needed. I forgot to remove it 😅

compositor_render/src/renderer/bgra_to_rgba.wgsl
```diff
@@ -0,0 +1,27 @@
...
+
+    return output;
+}
+
+@group(0) @binding(0) var texture: texture_2d<f32>;
+@group(1) @binding(0) var sampler_: sampler;
+
+@fragment
+fn fs_main(input: VertexOutput) -> @location(0) vec4<f32> {
+    var color = textureSample(texture, sampler_, input.tex_coords);
```
WojciechBarczynski, line 25: [NIT] It can be immutable.
```suggestion
    let color = textureSample(texture, sampler_, input.tex_coords);
```

WojciechBarczynski (approved):

compositor_render/src/event_loop.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+impl EventLoop {
+    pub fn new(cef_ctx: Option<Arc<cef::Context>>) -> Self {
+        Self { cef_ctx }
+    }
+
+    /// Runs the event loop. It must run on the main thread.
+    /// `fallback` is used when web rendering is disabled.
+    /// Blocks the thread indefinitely.
```
jerzywilczek, line 16: What happens when you run it on a non-main thread?

compositor_render/src/renderer.rs
```diff
@@ -47,24 +54,24 @@ pub struct Renderer {
...
 
 #[derive(Debug, thiserror::Error)]
 pub enum RendererNewError {
     #[error("failed to initialize a wgpu context")]
     FailedToInitWgpuCtx(#[from] WgpuCtxNewError),
 
-    #[error("failed to start an electron instance")]
-    FailedToStartElectron(#[from] ElectronNewError),
+    #[error("failed to init chromium context")]
+    FailedToStartElectron(#[from] ChromiumContextError),
```
jerzywilczek, line 74: `FailedToStartElectron`?

jerzywilczek (changes requested):
:nerd_face:

compositor_render/src/event_loop.rs
jerzywilczek: agreed

compositor_render/src/renderer.rs
noituri: Oh I missed that one 😅

compositor_render/src/event_loop.rs
jerzywilczek: ok cool

compositor_chromium/src/browser.rs
```diff
@@ -10,6 +10,10 @@ pub struct Browser {
     inner: Validated<chromium_sys::cef_browser_t>,
 }
 
+// TODO: Is it safe?
+unsafe impl Send for Browser {}
+unsafe impl Sync for Browser {}
+
```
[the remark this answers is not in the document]
noituri: I could make it `Sync` by using `Mutex` but I'm not sure how to make it `Send` without implementing it manually (unsafe).

jerzywilczek (approved):
:1st_place_medal:

PR #59, opened 2023-08-07
Add `align` and `wrap` options to text renderer. Add rendered text texture size calculation

WojciechBarczynski:


compositor_render/src/transformations/text_renderer.rs
```diff
@@ -19,6 +22,8 @@ pub struct TextSpec {
     attributes: AttrsOwned,
     font_size: f32,
     line_height: f32,
+    align: glyphon::cosmic_text::Align,
+    wrap: glyphon::cosmic_text::Wrap,
 }
 
 impl From<TextParams> for TextSpec {
```
wkozyra95, line 29: I missed that before, but in the current convention, Spec was used to describe values that are part of the scene that are used to create nodes. 

I don't see a problem with Spec being part of the node if we use the same value in API and inside Node, but here we use a different type in API, so I think that name is not appropriate.

wkozyra95 (approved):

compositor_render/src/transformations/text_renderer.rs
WojciechBarczynski: I changed the naming and moved implementations to `From`. Even though it technically makes `compositor_common` dependent on `glyphon`, changing `glyphon` for anything else, if needed, shouldn't be a problem. At first I wanted to avoid this, but this implementation looks a lot cleaner. 

Let me know what do you think about it (whether you find placing those `From`-s inside `compositor_common` ok).

wkozyra95: changing `glyphon` is not the issue, the problem is that we will likely use the common crate in rust client SDK whenever we get to writing on, so this would pull glyphon and it's dependencies to the project including wgpu

btw I'm not objecting to the changes, I just wanted to raise that concern. We can address if and when this becomes a problem

wkozyra95 (approved):

compositor_common/src/scene/text_spec.rs
```diff
@@ -0,0 +1,116 @@
...
+        match wrap {
+            Wrap::None => glyphon::cosmic_text::Wrap::None,
+            Wrap::Glyph => glyphon::cosmic_text::Wrap::Glyph,
+            Wrap::Word => glyphon::cosmic_text::Wrap::Word,
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(rename_all = "snake_case")]
```
wkozyra95, line 53:
```suggestion
```

compositor_common/src/scene/text_spec.rs
```diff
@@ -0,0 +1,116 @@
...
+pub enum Style {
+    Normal,
+    Italic,
+    Oblique,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(rename_all = "snake_case")]
```
wkozyra95, line 35:
```suggestion
```

compositor_common/src/scene/text_spec.rs
```diff
@@ -0,0 +1,116 @@
...
+fn default_align() -> Align {
+    Align::Left
+}
+
+fn default_wrap() -> Wrap {
+    Wrap::None
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(rename_all = "snake_case")]
```
wkozyra95, line 27:
```suggestion
```

PR #60, opened 2023-08-07
Unregister API for input/outputs + refactor

wkozyra95:
- Add unregister API for input/outputs
- Add some basic init options to the Init request (framerate and flag that enables web renderer)
- Refactor video_compositor crate
  - extract API itself from HTTP implementation details (separate http.rs and api.rs)
  - move state and validation to Pipeline (part of the effort to make that crate usable as a standalone library)
  - add PipelineInput trait
- Make sure RtpSender and Receiver are properly stopped when input/output is unregistered
  - RtpSender is an easy case, because the channel that is used to pass the frame will get closed when all senders are dropped, so if RtpSender itself is dropped, the thread will exit
  - RtpReceiver is more complicated
    - Add interrupt callback when creating context. When that callback returns true the PacketIter should return None.
    - When RtpReceiver:start exists consume message from channel to signal drop function that cleanup finished and that it can exit.
    
    
Test plan:

I modified long_ffmpeg example to 
- start with 2 inputs and 2 outputs
- wait 10 seconds
- remove one input and output from scene
- stop ffmpeg command and wait 10 seconds ( also tested without this step)
- unregister input
- unregister output

made sure that the sender and receiver thread have stopped


compositor_common/src/scene.rs
```diff
@@ -22,6 +22,24 @@ pub struct InputId(pub NodeId);
...
+        self.0 .0.fmt(f)
+    }
+}
+
+impl Display for NodeId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
```
WojciechBarczynski, line 42: Consider removing it, deriving `Debug` and using `{0:?}` syntax in error logs:
```rust 
pub enum RegisterInputError {
    #[error("Input {0:?} is already registered.")]
    AlreadyRegistered(InputId),
}
```

I'm not saying that this is necessarily bad. If you find this more descriptive you can leave it as it is.

src/api.rs
```diff
@@ -0,0 +1,119 @@
...
+        id: OutputId,
+    },
+    UpdateScene(Arc<SceneSpec>),
+    RegisterTransformation {
+        key: TransformationRegistryKey,
+        transform: TransformationSpec,
+    },
+    Init(pipeline::Options),
+    Start,
+}
```
WojciechBarczynski, line 51: [NIT] It doesn't matter at all, but `api.rs` file seem like the first one I would checkout out in such a project, so this order makes more sense (especially since out HTTP server is stateful)
```suggestion
#[derive(Serialize, Deserialize)]
#[serde(tag = "type", rename_all = "snake_case")]
pub enum Request {
    Init(pipeline::Options),
    RegisterInput(RegisterInputRequest),
    UnregisterInput {
        id: InputId,
    },
    RegisterOutput(RegisterOutputRequest),
    UnregisterOutput {
        id: OutputId,
    },
    RegisterTransformation {
        key: TransformationRegistryKey,
        transform: TransformationSpec,
    },
    UpdateScene(Arc<SceneSpec>),
    Start,
}
```

noituri (approved):

src/api.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    fn register_input(&mut self, request: RegisterInputRequest) -> Result<()> {
+        let RegisterInputRequest { id, port } = request;
+
+        //if let Some((node_id, _)) = state.inputs.iter().find(|(_, input)| input.port == port) {
+        //    return Err(anyhow!(
+        //        "Failed to register input with {id}. Port {port} is already use by node {node_id}"
+        //    ));
+        //}
```
WojciechBarczynski, line 112: Should we keep this? `pipeline.register_input()` should be responsible for that

src/http.rs
```diff
@@ -1,105 +1,72 @@
...
+                .and_then(Api::new);
+            match result {
+                Ok(new_api) => {
+                    self.send_response(raw_request, Ok(()));
+                    api = Some(new_api);
+                    break;
+                }
+                Err(err) => {
+                    self.send_response(raw_request, Err(err));
+                }
```
WojciechBarczynski, line 42: If you move this logic to separate function (sth like `handle_init_request(..) -> Api`) returning `Api`, then you can remove this `Option` type and unwrap

compositor_common/src/scene.rs
wkozyra95: Those errors will be user facing, so I want to display "Input input1 is already registered." and not "Input InputId(NodeId(input1)) is already registered."

WojciechBarczynski (approved):
GJ 🚀

src/api.rs
wkozyra95: I'll address that in follow-up, I need to think about it. Problem here is that pipeline can't be aware of ports

compositor_pipeline/src/pipeline.rs
```diff
@@ -1,82 +1,148 @@
...
 
-    pub fn start(self: &Arc<Self>) {
+    pub fn start(&self) {
         let (frames_sender, frames_receiver) = unbounded();
-        let pipeline = self.clone();
         let renderer = self.renderer.clone();
+        let outputs = self.outputs.clone();
 
-        pipeline.queue.start(frames_sender);
+        self.queue.start(frames_sender);
```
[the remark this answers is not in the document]
wkozyra95: Probably, I wasn't sure what that TODO refered to :D

src/rtp_receiver.rs
```diff
@@ -4,34 +4,81 @@ use compositor_common::{
...
     }
+}
 
-    fn start(pipeline: Arc<Pipeline>, port: u16, input_id: InputId) -> Result<()> {
+impl Drop for RtpReceiver {
+    fn drop(&mut self) {
+        // - AtomicBool signals to RTP thread that it should abort
+        // - Channel signals to drop method that RTP thread finished
+        self.should_close.store(true, Ordering::Relaxed);
+        self.drop_sender.send(()).unwrap();
```
jerzywilczek, line 71: a bit weird to use a channel for one way communication and an atomic for the other way, but it will work

PR #62, opened 2023-08-08
Rework resolution specification in text renderer

WojciechBarczynski:
Add fixed and fitted resolution options to text renderer


compositor_common/src/scene/text_spec.rs
```diff
@@ -116,3 +122,18 @@ impl From<&TextSpec> for AttrsOwned {
...
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum TextResolution {
+    /// Renders text and "trims" texture to smallest possible size
+    Fitted {
+        /// Must be specified with wrap
+        max_width: Option<u32>,
+        #[serde(default = "default_max_height")]
+        max_height: u32,
+    },
```
wkozyra95, line 135: If you want to keep your current behavior then
```suggestion
    Fitted {
        #[serde(default = "default_max_width")]
        max_width: u32,
        #[serde(default = "default_max_height")]
        max_height: u32,
    },
```
but I don't really like a few things about this approach
- treating our limitations as defaults seems a bit weird
- no way to set the exact width when wrapping

on the other hand, this approach is making an API very simple

so maybe to combine those approaches
```suggestion
    Fitted {
        #[serde(default = "default_max_width")]
        max_width: u32,
        #[serde(default = "default_max_height")]
        max_height: u32,
    },
    ColumnFitted {
        width: u32,
        #[serde(default = "default_max_height")]
        max_height: u32,
    },
```

compositor_common/src/validators.rs
```diff
@@ -249,6 +252,18 @@ impl SceneSpec {
...
         Ok(())
     }
+
+    fn validate_transforms_params(&self) -> Result<(), SpecValidationError> {
+        for TransformNodeSpec {
+            node_id,
+            transform_params,
+            ..
+        } in self.transforms.iter()
+        {
```
wkozyra95, line 262:
```suggestion
        for t in self.transforms.iter() {
            TransformNodeSpec {
                node_id,
                transform_params,
                ..
            } = t;
```
?

compositor_common/src/scene/text_spec.rs
```diff
@@ -116,3 +122,18 @@ impl From<&TextSpec> for AttrsOwned {
         }
     }
 }
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum TextResolution {
```
wkozyra95, line 128: TextDimensions

compositor_common/src/validators.rs
WojciechBarczynski: With `FittedColumn`, the resolution is modeled on the type system, so I removed this check entirely

compositor_common/src/scene.rs
```diff
@@ -94,12 +98,47 @@ pub enum TransformParams {
...
+        resolution: TextResolution,
     },
 }
 
+impl TransformParams {
+    pub fn validate(&self, node: &NodeId) -> Result<(), SpecValidationError> {
+        match self {
+            TransformParams::WebRenderer { .. } => Ok(()),
+            TransformParams::Shader { resolution, .. } => {
+                if resolution.width <= MAX_NODE_RESOLUTION.width
```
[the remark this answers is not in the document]
WojciechBarczynski: With FittedColumn, the resolution is modeled on the type system, so I removed this check entirely

compositor_render/src/renderer.rs
```diff
@@ -33,7 +34,7 @@ pub mod texture;
 
 pub struct Renderer {
     pub wgpu_ctx: Arc<WgpuCtx>,
-    pub text_renderer_ctx: TextRendererCtx,
+    pub text_renderer_ctx: Mutex<TextRendererCtx>,
```
wkozyra95, line 37: Either remove this mutex or those 2 mutexes inside TextRendererCtx:
- `Mutex<TextRendererCtx>` is safer because you don't need to worry about the order of locking on those mutexes
- Mutexes inside TextRendererCtx could be faster, but we are already processing everything in a single pipeline, so it should be fine.

I would go with the first option

wkozyra95 (approved):

compositor_render/src/renderer.rs
WojciechBarczynski: Since `text_renderer.prepare` requires mut ref to `font_system` and `swash_cache`, I need to have mut ref access to both fields of `TextRendererCtx` at the same scope. I couldn't find a way to do this having mutex only on `TextRendererCtx`, so I left mutexes on struct fields.

PR #63, opened 2023-08-08
Add shaders without user-defined parameters.

### LINKED ISSUE #12: Create custom shader transformation node
As a VC user, I want to create a custom transformation node with my shader, to transform frames accordingly to my specific business requirements.

jerzywilczek:
This patch allows shaders that don't take custom parameters.

Closes #12, finally

### DIFF
--- a/compositor_render/examples/silly.wgsl
+++ b/compositor_render/examples/silly.wgsl
@@ -27,7 +27,6 @@ struct CompositorStruct {
 }
 
 @group(0) @binding(0) var textures: binding_array<texture_2d<f32>, 16>;
-@group(1) @binding(0) var<uniform> shaders_custom_buffer: CustomStruct;
 @group(1) @binding(1) var<uniform> parameters_received_from_the_compositor: CompositorStruct;
 @group(2) @binding(0) var sampler_: sampler;
 
--- a/compositor_render/examples/silly.wgsl
+++ b/compositor_render/examples/silly.wgsl
@@ -18,10 +18,6 @@ fn vs_main(input: VertexInput) -> VertexOutput {
     return output;
 }
 
-struct CustomStruct {
-    a: u32,
-}
-
 struct CompositorStruct {
     time: f32,
 }
--- a/compositor_render/src/transformations/shader/node.rs
+++ b/compositor_render/src/transformations/shader/node.rs
@@ -14,18 +14,27 @@ pub struct ShaderNode {
 }
 
 impl ShaderNode {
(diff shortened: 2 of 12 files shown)


compositor_render/src/transformations/shader/node.rs
```diff
@@ -14,18 +14,27 @@ pub struct ShaderNode {
...
+                .device
                 .create_buffer_init(&wgpu::util::BufferInitDescriptor {
-                    label: Some("shader node custom params buffer"),
+                    label: Some("shader node empty custom params buffer"),
+                    contents: &[0],
                     usage: wgpu::BufferUsages::UNIFORM,
-                    contents: &params,
-                });
+                }),
+        };
```
WojciechBarczynski, line 37: With this, uniform in shader still has to be defined, right? I don't think defining `custom_params` (or sth like this) uniform in the shader when no parameters are used is a good solution. Maybe we should have two different bind groups layouts etc. for that.

wkozyra95 (approved):

compositor_render/src/transformations/shader/node.rs
jerzywilczek: i was a bit surprised by this, but the shader works without defining this uniform

jerzywilczek: also, [the spec](https://www.w3.org/TR/webgpu/#abstract-opdef-validating-gpuprogrammablestage) seems to confirm that shaders are allowed to not define bindings for some of the provided bind group entries.

This explains how bindings defined in a shader are validated:
> - For each binding that is [statically used](https://www.w3.org/TR/webgpu/#statically-used) by descriptor:
>    - [validating shader binding](https://www.w3.org/TR/webgpu/#abstract-opdef-validating-shader-binding)(binding, layout) must return true.

But the bindings that are not used by the shader are not validated, so they can be present and actually set to anything really

WojciechBarczynski (approved):

compositor_render/src/transformations/shader/node.rs
WojciechBarczynski: Ok, I didn't know/noticed that :)

compositor_render/examples/silly.wgsl
```diff
@@ -27,7 +27,6 @@ struct CompositorStruct {
 }
 
 @group(0) @binding(0) var textures: binding_array<texture_2d<f32>, 16>;
-@group(1) @binding(0) var<uniform> shaders_custom_buffer: CustomStruct;
```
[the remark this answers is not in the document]
jerzywilczek: oh, true

noituri (approved):

Issue #47, opened 2023-08-03
Start pipeline without waiting for a first frame on input


wkozyra95:
This is needed to support cases where e.g. there are no inputs because all frames are generated e.g. via web renderer.

PR #66, opened 2023-08-09
Fix text text_renderer example

wkozyra95:
Current validation does not allow unused inputs, but we need at least one input to start pipeline (until https://github.com/membraneframework/video_compositor/issues/47 is addressed)

So to fix the example I added second output that is there just to pass input through

### DIFF
--- a/compositor_common/src/validators/test.rs
+++ b/compositor_common/src/validators/test.rs
@@ -40,16 +40,16 @@ fn scene_validation_finds_cycle() {
 
     let b = TransformNodeSpec {
         node_id: b_id.clone(),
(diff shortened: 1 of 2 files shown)


compositor_common/src/validators/test.rs
```diff
@@ -40,16 +40,16 @@ fn scene_validation_finds_cycle() {
 
     let b = TransformNodeSpec {
         node_id: b_id.clone(),
-        input_pads: vec![a_id.clone()],
```
wkozyra95, line 43: clippy

WojciechBarczynski (approved):

PR #67, opened 2023-08-09
Provide fallback color for streams that did not start yet

wkozyra95:
The current implementation is sometimes crashing with more than output because the scene can start using some input before it starts sending frames.

Fix: Ignore the missing frame, set initial color of a texture to black or a user-provided value.

Potential future work could add support for e.g. falling back to the image or maybe even any other node in a scene.


compositor_render/src/renderer.rs
```diff
@@ -176,8 +177,11 @@ impl WgpuCtx {
         let (device, queue) = pollster::block_on(adapter.request_device(
             &wgpu::DeviceDescriptor {
                 label: Some("Video Compositor's GPU :^)"),
-                limits: Default::default(),
-                features: wgpu::Features::TEXTURE_BINDING_ARRAY,
+                limits: wgpu::Limits {
+                    max_push_constant_size: 128,
```
wkozyra95, line 181: no real reason for 128 specifically, for current implementation just 4 would be enough

compositor_common/src/scene.rs
```diff
@@ -65,6 +65,7 @@ pub struct SceneSpec {
 pub struct InputSpec {
     pub input_id: InputId,
     pub resolution: Resolution,
+    pub fallback_color: Option<(u8, u8, u8)>,
```
WojciechBarczynski, line 68: Maybe allow specifying alpha? Someone may want to use transparent
```suggestion
    pub fallback_color_rgba: Option<(u8, u8, u8, u8)>,
```

wkozyra95: we don't support alpha channel on inputs, because YUV does not support it

compositor_render/src/renderer/color_converter_pipeline.rs
```diff
@@ -234,3 +236,91 @@ impl RGBAToYUVConverter {
...
+            .create_command_encoder(&wgpu::CommandEncoderDescriptor {
+                label: Some("Fill R8 texture with color command encoder"),
+            });
+
+        {
+            let mut render_pass = encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
+                label: Some("Fill R8 texture with color render pass"),
+                color_attachments: &[Some(wgpu::RenderPassColorAttachment {
+                    ops: wgpu::Operations {
+                        load: wgpu::LoadOp::Clear(wgpu::Color::BLACK),
```
WojciechBarczynski, line 301: Just use this

WojciechBarczynski: https://docs.rs/wgpu/latest/wgpu/enum.LoadOp.html

WojciechBarczynski: Passing `fill color` as https://docs.rs/wgpu/latest/wgpu/struct.Color.html  instead of `wgpu::Color::BLACK` does the whole job :)

compositor_common/src/scene.rs
WojciechBarczynski: If we `Clear` "input node" output texture (which is in RGBA format) with user-provided color, we can easily support alpha :)

wkozyra95: yes, but what would it mean? This is a fallback for input, if the input does not support the alpha channel, then supporting that in the alpha channel is a bit redundant.

compositor_render/src/renderer/color_converter_pipeline.rs
wkozyra95: responded on slack

wkozyra95:
I decided to change how color are passed via API, now instead of number array it expects "#RRGGBB"

jerzywilczek (approved):
:rocket:

WojciechBarczynski (approved):

PR #69, opened 2023-08-10
Implement query API

wkozyra95:
Add some basic queries e.g.
- get scene
- get registered inputs
- get registered output
(we could add bunch of other queries like that, I just picked those for start)

Add blocking query:
- WaitForNextFrame will block the response until the next frame will not show up in the queue

Break out some queue code into separate files.

~~Currently, I'm responding when a frame is enqueued. I think at this point we know that next render will include a frame (but not necessarily that one). I'm open 2 suggestions, I also considered:~~
- ~~calling callback when pushing frames to the renderer~~
- ~~have 2 different types of callback (one when enqueueing and one when pushing to renderer)~~

Response for input listener is sent when we send a frame from queue to renderer

Few small changes:
- drop unnecessary Arc around internal_queue
- switch Arc<String> to Arc<str> in RtpSender

### DIFF
--- a/compositor_pipeline/src/queue/internal_queue.rs
+++ b/compositor_pipeline/src/queue/internal_queue.rs
@@ -10,6 +10,7 @@ use super::QueueError;
 pub struct InternalQueue {
     /// frames are PTS ordered. PTS include timestamps offsets
     inputs_queues: HashMap<InputId, Vec<Arc<Frame>>>,
+    inputs_listeners: HashMap<InputId, Vec<Box<dyn FnOnce() + Send>>>,
     timestamp_offsets: HashMap<InputId, Duration>,
     output_framerate: Framerate,
     pub sent_batches_counter: u32,
@@ -19,6 +20,7 @@ impl InternalQueue {
     pub fn new(output_framerate: Framerate) -> Self {
         InternalQueue {
             inputs_queues: HashMap::new(),
+            inputs_listeners: HashMap::new(),
             timestamp_offsets: HashMap::new(),
             output_framerate,
             sent_batches_counter: 0,
@@ -130,5 +132,23 @@ impl InternalQueue {
 
     pub fn get_next_output_buffer_pts(&self) -> Duration {
         Duration::from_secs_f64(self.sent_batches_counter as f64 / self.output_framerate.0 as f64)
+    }
+
+    pub fn subscribe_input_listener(
+        &mut self,
+        input_id: InputId,
+        callback: Box<dyn FnOnce() + Send>,
+    ) {
+        self.inputs_listeners
+            .entry(input_id)
+            .or_insert_with(Vec::new)
+            .push(callback)
+    }
+
+    pub fn call_input_listeners(&mut self, input_id: &InputId) {
+        let callbacks = self.inputs_listeners.remove(input_id).unwrap_or_default();
+        for cb in callbacks.into_iter() {
+            cb()
+        }
     }
 }
--- a/src/api.rs
+++ b/src/api.rs
@@ -6,6 +6,7 @@ use compositor_common::{
     transformation::{TransformationRegistryKey, TransformationSpec},
 };
 use compositor_pipeline::pipeline;
+use crossbeam_channel::{bounded, Receiver};
 use serde::{Deserialize, Serialize};
 
 use crate::{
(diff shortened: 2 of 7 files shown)


compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -131,4 +134,22 @@ impl InternalQueue {
...
+        input_id: InputId,
+        callback: Box<dyn FnOnce() + Send>,
+    ) {
+        self.inputs_listeners
+            .entry(input_id)
+            .or_insert_with(std::vec::Vec::new)
+            .push(callback)
+    }
+
+    fn call_input_listeners(&mut self, input_id: &InputId) {
```
WojciechBarczynski, line 149: [NIT]
Since we use terminology from the observer pattern (`listener`), consider using it here as well:
`call` -> `notify`
`register` -> `subscribe`

src/api.rs
```diff
@@ -47,9 +48,48 @@ pub enum Request {
...
+    Scene,
+    Inputs,
+    Outputs,
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum Response {
+    Ok {},
+    WaitForFirstFrame { input_id: InputId },
```
WojciechBarczynski, line 68: What this is for?

compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -36,6 +38,7 @@ impl InternalQueue {
 
     pub fn enqueue_frame(&mut self, input_id: InputId, mut frame: Frame) -> Result<(), QueueError> {
         let next_output_buffer_pts = self.get_next_output_buffer_pts();
+        self.call_input_listeners(&input_id);
```
WojciechBarczynski, line 41: Frames queueing from the user perspective is an implementation detail. 
From the user's perspective, it's more important when frames from certain streams are being composed, rather than enqueued. 
With current queueing logic, input frames will always be used in the next `frame set`,  meaning this implementation has the same effect, as checking the output of the queue, but relaying on that seems weird to me. What do you think?

compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -131,4 +134,22 @@ impl InternalQueue {
...
     }
+
+    pub fn register_input_listener(
+        &mut self,
+        input_id: InputId,
+        callback: Box<dyn FnOnce() + Send>,
+    ) {
+        self.inputs_listeners
+            .entry(input_id)
+            .or_insert_with(std::vec::Vec::new)
```
noituri, line 145: Is `std::vec::` part needed here?

compositor_pipeline/src/queue/internal_queue.rs
wkozyra95: no objection to the subscribe, but notify implies async action which is not true in this case

src/api.rs
wkozyra95: copied accidentally and missed it

compositor_pipeline/src/queue/internal_queue.rs
wkozyra95: I don't think there is anything wrong with this approach if it gives the correct result. It might make a difference if we want to make sure that the next frame is what will be used for next render, but supporting that would be a lot harder (it would not be enough to just move that call to different location)

compositor_pipeline/src/queue/internal_queue.rs
wkozyra95: I added `vec![]` and clippy probably changed it

noituri (approved):

src/api.rs
```diff
@@ -47,9 +48,47 @@ pub enum Request {
...
+    WaitForNextFrame { input_id: InputId },
+    Scene,
+    Inputs,
+    Outputs,
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum Response {
+    Ok {},
```
noituri, line 67: Is `{}` needed here?

wkozyra95: yes I want to return `{}` instead of null

noituri: Oh okay, makes sense

WojciechBarczynski (approved):

PR #72, opened 2023-08-11
Add video count to `CommonShaderParameters` and pass it as a `push_constant`

WojciechBarczynski:


WojciechBarczynski:
Fix `inputs`, change to `push constants`

compositor_render/examples/silly/silly.wgsl
```diff
@@ -18,19 +18,26 @@ fn vs_main(input: VertexInput) -> VertexOutput {
     return output;
 }
 
-struct CompositorStruct {
+struct CommonParams {
     time: f32,
+    textures_count: u32,
 }
```
WojciechBarczynski, line 24: I kept it in a single struct since there aren't many parameters passed here yet and passing multiple separate `push_constants` complicated `pipeline_layout`

wkozyra95: I don't have a strong opinion, but I wouldn't merge those 2 values. Until textures count needs to be passed to all shader it's fine to leave it this way

wkozyra95 (approved):

compositor_render/examples/silly/silly.wgsl
jerzywilczek: I don't like this a lot. Those are updated separately, the time once per frame and the textures count once per node render. once-per-frame buffer pushes are fine, i would prefer to limit push-constants to values that change many times per frame

PR #74, opened 2023-08-16
Add support for static images

wkozyra95:
Supports:
- PNG
- JPEG
- GIF(only as static image)
- SVG(subset of SVG supported by resvg)

### DIFF
--- a/compositor_render/src/transformations/image_renderer.rs
+++ b/compositor_render/src/transformations/image_renderer.rs
@@ -33,7 +33,7 @@ impl Image {
                 ImageRenderer::BitmapAsset(asset)
             }
             ImageSpec::Svg { resolution, .. } => {
-                let asset = SVGAsset::new(&ctx.wgpu_ctx, file, resolution)?;
+                let asset = SvgAsset::new(&ctx.wgpu_ctx, file, resolution)?;
                 ImageRenderer::Svg(asset)
             }
             ImageSpec::Gif { .. } => {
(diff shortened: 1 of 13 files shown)


compositor_render/src/transformations/image_renderer.rs
```diff
@@ -0,0 +1,260 @@
...
+
+    fn download_file(url: &str) -> Result<bytes::Bytes, ImageError> {
+        // TODO: support local files
+        let response = reqwest::blocking::get(url)?;
+        let response = response.error_for_status()?;
+        Ok(response.bytes()?)
+    }
+}
+
+pub struct ImageNode {
```
wkozyra95, line 56: This struct is not needed, but we will want to support images that are not just a static bitmap e.g.
- GIF with animations
- SVG with resolution specified at the update scene level, and not during the transformation registration)

wkozyra95: outdated

compositor_render/src/transformations/image_renderer.rs
```diff
@@ -0,0 +1,255 @@
...
+    image: Arc<Image>,
+    was_rendered: Mutex<bool>,
+}
```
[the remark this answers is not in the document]
wkozyra95: I did that initially, this was the error I mentioned.
If I moved it there then was_rendered would be a one value per transformation, which would not work if more than one node used it. Also, I wouldn't be refreshed when scene is updated.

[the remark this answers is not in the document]
WojciechBarczynski: Ahh, right 😃

WojciechBarczynski (approved):

compositor_render/src/transformations/image_renderer.rs
```diff
@@ -0,0 +1,255 @@
...
+        // TODO: support local files
+        let response = reqwest::blocking::get(url)?;
+        let response = response.error_for_status()?;
+        Ok(response.bytes()?)
+    }
+}
+
+pub struct ImageNode {
+    image: Arc<Image>,
+    was_rendered: Mutex<bool>,
```
noituri, line 58: Maybe consider using `AtomicBool` instead?

PR #77, opened 2023-08-17
Start pipeline without waiting for the first frame

wkozyra95:


compositor_pipeline/src/queue.rs
```diff
@@ -19,58 +21,100 @@ pub enum QueueError {
...
 
-        internal_queue.enqueue_frame(input_id.clone(), frame)?;
-        internal_queue.drop_input_useless_frames(input_id)?;
+        internal_queue.enqueue_frame(input_id.clone(), frame, self.clock_start)?;
+
+        // We don't know when pipeline is started, so we can't resolve real_next_pts,
+        // but we can remove frames based on estimated PTS. This only works if queue
+        // is able to push frames in real time and is never behind more than one frame.
+        let framerate_tick = Duration::from_secs_f64(1.0 / self.output_framerate.0 as f64);
+        let estimated_pts = self.clock_start.elapsed() - framerate_tick;
```
wkozyra95, line 114: I'm not really sure if one tick is enough, technically queue does not do anything heavy, so I don't see a reason why it would be behind with pushing frames, but if we want to be safe we could add e.g.

```suggestion
        let estimated_pts = self.clock_start.elapsed() - framerate_tick * 10;
```

This drop_frame is not strictly necessary, when pipeline is running because queue thread will clean it up, but if pipeline did not start yet that cleanup in QueueThread would not be scheduled

compositor_pipeline/src/queue/internal_queue.rs
```diff
@@ -36,29 +35,37 @@ impl InternalQueue {
...
+        Ok(())
     }
 
-    /// Pops frames closest to buffer pts.
+    /// Gets frames closest to buffer pts.
+    ///
     /// Implementation assumes that "useless frames" are already dropped
     /// by [`drop_useless_frames`] or [`drop_pad_useless_frames`]
-    pub fn get_frames_batch(&mut self, buffer_pts: Duration) -> FrameSet<InputId> {
+    pub fn get_frames_batch(&self, buffer_pts: Duration) -> FrameSet<InputId> {
```
WojciechBarczynski, line 68: We can simplify this and remove this assumption entirely, but dropping frames right before getting it from input queue. Sth like this:
```rust
    pub fn get_frames_batch(&mut self, buffer_pts: Duration) -> FrameSet<InputId> {
        let mut frames_batch = FrameSet::new(buffer_pts);

        let inputs: Vec<InputId> = self.inputs_queues.keys().cloned().collect();

        for input_id in inputs {
            if let Some(nearest_frame) = self
                .get_nearest_frame_by_input_id(buffer_pts, &input_id)
                .unwrap()
            {
                frames_batch
                    .frames
                    .insert(input_id.clone(), nearest_frame.clone());
            }
        }

        frames_batch
    }

    fn get_nearest_frame_by_input_id(
        &mut self,
        buffer_pts: Duration,
        input_id: &InputId,
    ) -> Result<Option<Frame>, QueueError> {
        self.drop_frames_by_input_id(input_id, buffer_pts)?;
        let input_queue = self
            .inputs_queues
            .get(input_id)
            .ok_or_else(|| QueueError::UnknownInputId(input_id.clone()))?;
        Ok(input_queue.first().cloned())
    }
```

compositor_pipeline/src/queue.rs
```diff
@@ -19,58 +21,100 @@ pub enum QueueError {
...
+            .lock()
+            .unwrap()
+            .input_offset(&input_id)
+            .is_none();
+        if is_first_frame {
+            // Sleep here ensures that we will buffer `self.buffer_duration` on each input.
+            // It also makes calculation easier because PTS of frames will be already offset
+            // by a correct value.
+            thread::sleep(self.buffer_duration);
+        }
```
WojciechBarczynski, line 104: 1. At first, I understood `is_first_frame` as a first frame from all inputs (first frame enqueued).
2. I don't like to rely on inserting `input_offset` on the first frame. 

We can solve this by adding some `has_received_frame` function in the `internal queue`

PR #78, opened 2023-08-17
Add common transformations API, add `convert resolution` transformation

WojciechBarczynski:
Add a skeleton for registering and using common transformations. Implements transformation covering resolution, with 3 different strategies: `stretch`, `crop_scale` and `fill_scale`


compositor_common/src/scene/common_transformations.rs
```diff
@@ -0,0 +1,22 @@
...
+pub enum ConvertResolutionParams {
+    /// Rescales input in both axis to match output resolution
+    Stretch,
+    /// Scales input preserving aspect ratio and cuts equal parts
+    /// from both sides in "sticking out" dimension
+    CropScale,
+    /// Scales input preserving aspect ratio and
+    /// fill the rest of the texture with the provided color
+    FillScale(RGBAColor),
+}
```
WojciechBarczynski, line 22: Not sure about this naming. Any better ideas?

compositor_common/src/scene/common_transformations.rs
```diff
@@ -0,0 +1,22 @@
+use serde::{Deserialize, Serialize};
+
+use crate::util::RGBAColor;
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(tag = "type", content = "params", rename_all = "snake_case")]
```
wkozyra95, line 6:
```suggestion
#[serde(tag = "transformation", content = "params", rename_all = "snake_case")]
```
or operation, in general, let's try to avoid type if possible

compositor_common/src/scene/common_transformations.rs
```diff
@@ -0,0 +1,22 @@
...
+pub enum CommonTransformation {
+    ConvertResolution(ConvertResolutionParams),
```
wkozyra95, line 8: EnsureResolution/ChangeResolution/TransformToResolution

compositor_common/src/scene/common_transformations.rs
wkozyra95: FillScale - this is not correct, fill in this context usually means that the image would fill the entire view which would be the same as Crop

I would name it just Fill(instead of CropScale) and Fit(instead of FillScale). The scale does not provide any info.

compositor_render/src/renderer.rs
```diff
@@ -82,14 +86,16 @@ pub enum RendererRegisterTransformationError {
...
         Ok(Self {
-            wgpu_ctx: Arc::new(WgpuCtx::new()?),
+            wgpu_ctx: wgpu_ctx.clone(),
             text_renderer_ctx: TextRendererCtx::new(),
             electron_instance: Arc::new(ElectronInstance::new(9002, init_web)?), // TODO: make it configurable
             scene: Scene::empty(),
             web_renderers: TransformationRegistry::new(RegistryType::WebRenderer),
             shader_transforms: TransformationRegistry::new(RegistryType::Shader),
             image_registry: TransformationRegistry::new(RegistryType::Image),
+            common_transformations_registry: CommonTransformationsRegistry::new(&wgpu_ctx),
```
wkozyra95, line 98: wouldn't call it registry this, this is a static structure so it's not really correct

examples/common_transformations.rs
```diff
@@ -0,0 +1,151 @@
...
+            {
+                "node_id": "fitted",
+                "type": "common",
+                "transformation": {
+                    "type": "convert_resolution",
+                    "params": {
+                        "strategy": "fill_scale",
+                        "color": "#FFFFFF00"
+                    },
+                },
```
wkozyra95, line 113:
```suggestion
                "type": "built-in:convert_resolution",
                "params": {
                    "strategy": "fill_scale",
                    "color": "#FFFFFF00"
                },
```

try to avoid nesting if it's not necessarry, I would even consider
```suggestion
                "type": "built-in:convert_resolution",
                "strategy": "fill_scale",
                "color": "#FFFFFF00"
```

compositor_render/src/transformations.rs
```diff
@@ -1,3 +1,4 @@
+pub mod common;
```
wkozyra95, line 1: if you change name in api, chaneg the code too to use that name. e.g.
```suggestion
pub mod builtin;
```
or
```suggestion
pub mod builtin_transforms;
```

examples/common_transformations.rs
WojciechBarczynski: Since implementing this would require nontrivial parsing, I implemented it as:
```json
"node_id": "fitted",
"type": "built-in",
"transformation": "transform_to_resolution",
"strategy": "fit",
"color": "#FFFFFF00",
"resolution": { "width": VIDEO_RESOLUTION.width, "height": VIDEO_RESOLUTION.height },
"input_pads": ["image"],
```

examples/text_renderer.rs
```diff
@@ -81,13 +81,11 @@ fn start_example_client_code() -> Result<()> {
...
-                    "font_size": 100.0,
-                    "font_family": "Comic Sans MS",
-                    "align": "center",
-                    "wrap": "word",
-                },
+                "content": "VideoCompositor🚀\nSecond Line\nLorem ipsum dolor sit amet consectetur adipisicing elit. Soluta delectus optio fugit maiores eaque ab totam, veritatis aperiam provident, aliquam consectetur deserunt cumque est? Saepe tenetur impedit culpa asperiores id?",
+                "font_size": 100.0,
+                "font_family": "Comic Sans MS",
+                "align": "center",
+                "wrap": "word",
```
WojciechBarczynski, line 88: Flatten this for consistency across transformations

compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -0,0 +1,22 @@
...
+#[serde(tag = "strategy", content = "color", rename_all = "snake_case")]
+pub enum TransformToResolution {
+    /// Rescales input in both axis to match output resolution
+    Stretch,
+    /// Scales input preserving aspect ratio and cuts equal parts
+    /// from both sides in "sticking out" dimension
+    Fill,
+    /// Scales input preserving aspect ratio and
+    /// fill the rest of the texture with the provided color
+    Fit(RGBAColor),
```
wkozyra95, line 21: as we discussed offline, add align option (here or in follow up PR).
e.g. 
```
{
  "strategy": "fit",
  "background_color": "#FFFFFFFF"
  "align": "top" | "bottom" | "top-left" | "top-right" | "left" | "right"
 }
 ```
 e.g. top implies that it's centered horizontally
```

compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -0,0 +1,22 @@
...
+use crate::util::RGBAColor;
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(tag = "transformation", rename_all = "snake_case")]
+pub enum BuiltinTransformation {
+    TransformToResolution(TransformToResolution),
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(tag = "strategy", content = "color", rename_all = "snake_case")]
```
wkozyra95, line 12: background_color or background_color_rgba

compositor_render/src/renderer.rs
```diff
@@ -264,13 +273,18 @@ impl WgpuCtx {
 pub struct CommonShaderParameters {
     time: f32,
     textures_count: u32,
+    output_texture_resolution: [u32; 2],
 }
 
 impl CommonShaderParameters {
-    pub fn new(time: Duration, textures_count: u32) -> Self {
+    pub fn new(time: Duration, textures_count: u32, output_texture_resolution: Resolution) -> Self {
```
wkozyra95, line 280:
```suggestion
    pub fn new(time: Duration, textures_count: u32, output_resolution: Resolution) -> Self {
```

compositor_render/src/renderer/scene.rs
```diff
@@ -22,13 +22,12 @@ use super::{
...
 };
 
-pub struct InputNode {}
-
 pub enum TransformNode {
     Shader(ShaderNode),
     WebRenderer { renderer: Arc<WebRenderer> },
     TextRenderer(TextRendererNode),
     ImageRenderer(ImageNode),
+    Common(ShaderNode),
```
wkozyra95, line 30: rename

compositor_render/src/renderer.rs
```diff
@@ -52,6 +55,7 @@ pub struct RenderCtx<'a> {
     pub(crate) shader_transforms: &'a TransformationRegistry<Arc<Shader>>,
     pub(crate) web_renderers: &'a TransformationRegistry<Arc<WebRenderer>>,
     pub(crate) image_registry: &'a TransformationRegistry<Arc<Image>>,
+    pub(crate) common_transforms: &'a BuiltinTransformations,
```
wkozyra95, line 58: rename

compositor_render/src/renderer/scene.rs
```diff
@@ -49,6 +48,19 @@ impl TransformNode {
...
                     shader_params.as_ref(),
+                    None,
+                )),
+                *resolution,
+            )),
+            TransformParams::Builtin {
+                transformation,
+                resolution,
+            } => Ok((
+                TransformNode::Common(ShaderNode::new(
```
wkozyra95, line 59: I susspect that doing just wrapper around ShaderNode will not be enough for long, but on the other hand solutions written this way are easy to copy-paste by user and modify them for their own usecases

wkozyra95 (approved):

compositor_render/src/renderer/scene.rs
WojciechBarczynski: When we'll need to use different transforms here, we can always wrap it in some struct.
Since there aren't any transformations so far, that couldn't be implemented as shaders, I would leave it as it is right now

compositor_common/src/scene/builtin_transformations.rs
WojciechBarczynski: I would prefer to do it this way in the follow up PR:
```
{
  "strategy": "fit",
  "background_color": "#FFFFFFFF",
  "align_x": "left" | "center" (default) | "right",
  "align_y": "top" | "center" (default) | "bottom",
 }
```

compositor_render/src/renderer/scene.rs
WojciechBarczynski: I think with some abstraction layer over the shader pipeline (single shader transformation pipeline), we could make a few things more straightforward / more efficient and don't use ShaderNode here (e.g. calculate matrixes on CPU and send them via uniform), but I would do it here when it's useful.

compositor_render/src/transformations/builtin/transform_to_resolution/fill.wgsl
```diff
@@ -0,0 +1,78 @@
...
+    // If the output ratio is larger, then the texture is scaled up,
+    // such that input_width = output_width.
+    // Analogously:
+    // scale_factor_x_clip_space = 1.0 (input x coords are already fitted)
+    // scale_factor_y_clip_space = output_ratio / input_ratio
+    if input_ratio >= output_ratio {
+        x_scale = input_ratio / output_ratio;
+    } else {
+        y_scale = output_ratio / input_ratio;
+    }
```
jerzywilczek, line 65: I don't think this works, we should discuss this in person

jerzywilczek: how can this preserve the aspect ratio if you multiply one coordinate by some number != 1 and the other by 1?

compositor_render/src/transformations/builtin/transform_to_resolution/fit.wgsl
```diff
@@ -0,0 +1,78 @@
...
+    // If the output ratio is larger, then the texture is scaled up,
+    // such that input_height = output_height.
+    // Analogously:
+    // scale_factor_x_clip_space = input_ratio / output_ratio
+    // scale_factor_y_clip_space = 1.0 (input y coords are already fitted)
+    if input_ratio >= output_ratio {
+        y_scale = output_ratio / input_ratio;
+    } else {
+        x_scale = input_ratio / output_ratio;
+    }
```
jerzywilczek, line 65: same here

compositor_render/src/transformations/shader/node.rs
```diff
@@ -11,10 +11,16 @@ pub struct ShaderNode {
...
+    clear_color: Option<wgpu::Color>,
 }
 
 impl ShaderNode {
-    pub fn new(ctx: &WgpuCtx, shader: Arc<Shader>, params: Option<&ShaderParam>) -> Self {
+    pub fn new(
+        ctx: &WgpuCtx,
+        shader: Arc<Shader>,
+        params: Option<&ShaderParam>,
+        clear_color: Option<wgpu::Color>,
```
jerzywilczek, line 22: I feel a bit that this does not belong here, but can't find a better place to pass this.

examples/builtin_transformations.rs
```diff
@@ -0,0 +1,147 @@
...
+        "init_web_renderer": false,
+    }))?;
+
+    info!("[example] Start listening on output port.");
+    let output_sdp = write_example_sdp_file("127.0.0.1", 8002)?;
+    Command::new("ffplay")
+        .args(["-protocol_whitelist", "file,rtp,udp", &output_sdp])
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .spawn()?;
```
jerzywilczek, line 59: I feel like most of this code should be extracted somewhere at this point. It's ugly and, I think, very similar in every example

jerzywilczek (changes requested):

compositor_render/src/transformations/builtin/transform_to_resolution/fill.wgsl
WojciechBarczynski: Since those are output texture coords :)

compositor_render/src/transformations/builtin/transform_to_resolution/fit.wgsl
WojciechBarczynski: Same

compositor_render/src/renderer.rs
```diff
@@ -85,14 +89,17 @@ pub enum RendererRegisterTransformationError {
...
         Ok(Self {
-            wgpu_ctx: Arc::new(WgpuCtx::new()?),
+            wgpu_ctx: wgpu_ctx.clone(),
             text_renderer_ctx: TextRendererCtx::new(),
             electron_instance: Arc::new(ElectronInstance::new(9002, init_web)?), // TODO: make it configurable
             scene: Scene::empty(),
             web_renderers: TransformationRegistry::new(RegistryType::WebRenderer),
             shader_transforms: TransformationRegistry::new(RegistryType::Shader),
             image_registry: TransformationRegistry::new(RegistryType::Image),
+            // TODO fix unwrap
```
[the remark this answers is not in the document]
WojciechBarczynski: Done

compositor_render/src/transformations/shader/node.rs
WojciechBarczynski: I don't want to specify this on pipeline creation since the same pipeline might be used with different clear colors. I couldn't find a better place to put it either

examples/builtin_transformations.rs
WojciechBarczynski: Since those are examples, not tests, I would avoid putting abstraction here and keep it as simple as possible

jerzywilczek (approved):
:rocket:

PR #79, opened 2023-08-17
Stop creating textures every frame

wkozyra95:
Empty textures are created for every render, so I moved that from the render function.

I looked into optimizing other parts that are created per frame, but I suspect we can't easily optimize it. `input_textures_bg` is created based on inputs and those can change on every render e.g. because resolution changed. We could maybe keep track of what was used for previous renders and cache that, but this seems a bit hacky, so unless it's the source of performance issues I would prefer to avoid it.

Nothing is logged anymore by wgpu, so I suspect this was only part that shouldn't be done per frame.


WojciechBarczynski (approved):
This texture could be created once for all shader transformations, but it would complicate code, so we can leave it as it is (created once per transformation).

PR #82, opened 2023-08-18
Add support for animated GIFs

wkozyra95:
Create separate textures for each frame and add hard coded limit to max 1000 frames.

### DIFF
--- a/compositor_render/src/transformations/image_renderer.rs
+++ b/compositor_render/src/transformations/image_renderer.rs
@@ -1,6 +1,4 @@
 use std::{
-    collections::BTreeMap,
-    ops::Bound,
     str::{from_utf8, Utf8Error},
     sync::{Arc, Mutex},
     time::Duration,
@@ -285,8 +283,13 @@ pub struct AnimatedNodeState {
 }
 
 pub struct AnimatedAsset {
-    frames: BTreeMap<Duration, RGBATexture>,
+    frames: Vec<AnimationFrame>,
     animation_duration: Duration,
+}
+
+struct AnimationFrame {
+    texture: RGBATexture,
+    pts: Duration,
 }
 
 impl AnimatedAsset {
@@ -297,7 +300,7 @@ impl AnimatedAsset {
         };
 
         let mut animation_duration: Duration = Duration::ZERO;
-        let mut frames = BTreeMap::new();
+        let mut frames = vec![];
         for frame in decoded_frames {
             let frame = &frame?;
             let buffer = frame.buffer();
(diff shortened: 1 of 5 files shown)


compositor_render/src/transformations/image_renderer.rs
```diff
@@ -233,6 +278,140 @@ impl SvgAsset {
...
+                pts: animation_duration,
+            });
+
+            if frames.len() > 1000 {
+                return Err(AnimatedError::TooMuchFrames);
+            }
+        }
+
+        let Some(first_frame) = frames.first() else {
+            return Err(AnimatedError::NoFrames)
```
WojciechBarczynski, line 329:
```suggestion
            return Err(AnimatedError::NoFrames);
```

WojciechBarczynski (approved):

compositor_render/src/transformations/image_renderer.rs
jerzywilczek, line 1: There is a lot of duplication in this code, which i would prefer to eliminate, but I guess it is fine

jerzywilczek: If you feel like eliminating the duplication it would be nice to do that, but I think it can be merged as is if you don't want to

jerzywilczek (approved):

compositor_render/src/transformations/image_renderer.rs
WojciechBarczynski, line 1: This file is quite large. Maybe we should extract some logic from here?

compositor_render/src/transformations/image_renderer.rs
```diff
@@ -233,6 +278,140 @@ impl SvgAsset {
...
+
+        let animation_pts = Duration::from_nanos(
+            ((pts.as_nanos() - first_pts.as_nanos()) % self.animation_duration.as_nanos()) as u64,
+        );
+
+        let closest_frame = self
+            .frames
+            .iter()
+            .min_by_key(|frame| u128::abs_diff(frame.pts.as_nanos(), animation_pts.as_nanos()))
+            .unwrap();
```
WojciechBarczynski, line 379: [NIT]
For n frames in animated assets, this takes O(n^2) for all frames (calling `render` n times). Since `frames` is already sorted, you can just use bin search to bring it to O(nlogn) or use BTreeMap. With a stateful iterator for each node and sorted vec, you can get it down O(n) (using sth similar to cyclic buffer).
IDK how much important it's to optimize it, but you can consider it since this runs on every render call

WojciechBarczynski: You already have `AnimatedNodeState`, so it's quite simple to use optimal iter here

wkozyra95: > For n frames in animated assets, this takes O(n^2) for all frames (calling render n times).

but we are not calling render n time, so this O(n), but you are right this could easilly be O(log n) and if we want to even constant.

My reasoning was that we have a hard limit on frames and the creation of frames is relatively expensive, so I treated the number of frames as constant

Overall I think the performance difference is negligible (Iterating over list of up to 1000 elements is nothing).I don't want to implement my own binary search or custom iterator, because it would affect readability, but I'll try using BTreeMap

compositor_render/src/transformations/image_renderer.rs
wkozyra95: I think it's fine, especially that adding support to any new format will be just one or 2 lines using the existing 3 implementations. If we ever need to add another mode, then yes this will need to be broken up.

compositor_render/src/transformations/image_renderer.rs
jerzywilczek: I would stick with naive iteration here. The time this would take is probably too short to notice compared to uploading a frame to the GPU.

jerzywilczek: and if this was a performance bottleneck, it's easy to notice through profiling and fix

jerzywilczek: oh, I just noticed you changed it already. nevermind then.

wkozyra95: yes, after implementing we decided that increased complexity is not worth it

PR #89, opened 2023-08-23
Add frame embedding for web renderer

### LINKED ISSUE #28: Add video sources to web renderer
As a VC user, I would like to send video sources to web renderer and compose it there

noituri:
This PR implements basic support for fame embedding. Currently communication compositor <-> chrome looks like that:

```mermaid
flowchart TD
    C[Compositor] -->|Send info about input frames via IPC| R[Chrome renderer subprocess]
    C -->|Put frame RGBA data| S(Shared memory)
    R -->|Create JS array buffers| W[Website]
    R -->|Point array buffers to shared mem| S
```

User API is still work in progress, I will implement it in a separate PR.

`Browser` and `Shmem` do not implement `Send` and `Sync` so we can't move them between threads. I used a separate thread for storing them.

Things to check:
- [x] Overall performance
- [x] Test resize
- [x] Test various resolutions
- [x] Test multiple inputs

Closes #28


compositor_render/src/renderer/scene.rs
```diff
@@ -29,9 +29,13 @@ pub enum RenderNode {
 }
 
 impl RenderNode {
-    fn new(ctx: &RenderCtx, spec: &NodeParams) -> Result<Self, GetError> {
+    fn new(ctx: &RenderCtx, spec: &NodeParams, inputs: &[Arc<Node>]) -> Result<Self, GetError> {
         match spec {
             NodeParams::WebRenderer { instance_id } => {
+                for input in inputs {
```
wkozyra95, line 35: move to render

src/bin/process_helper.rs
```diff
@@ -16,33 +24,158 @@ impl cef::App for App {
...
+}
+
 // Subprocess used by chromium
 fn main() -> Result<(), Box<dyn Error>> {
     env_logger::init_from_env(
         env_logger::Env::default().filter_or(env_logger::DEFAULT_FILTER_ENV, "info"),
     );
 
+    let app = App {
+        states: Rc::new(RefCell::new(HashMap::new())),
```
wkozyra95, line 176: Arc

compositor_render/src/renderer/texture/base.rs
```diff
@@ -31,7 +36,11 @@ impl Texture {
 
         let view = texture.create_view(&wgpu::TextureViewDescriptor::default());
 
-        Self { texture, view }
+        Self {
+            texture,
+            view,
+            download_buffer: None,
```
wkozyra95, line 42: We don't want download buffers to be a part of texture, because not all textures need them. Using Option avoids unnecessary entities, but adding them to texture is a hack.

It's fine if e.g. texture has an implementation for creating download buffer for that texture, but buffer itself should be stored separately. Check how it's implemented in scene.

WojciechBarczynski (approved):

compositor_render/src/renderer/scene.rs
```diff
@@ -34,7 +41,8 @@ impl RenderNode {
         match spec {
             NodeParams::WebRenderer { instance_id } => {
                 let renderer = ctx.web_renderers.get(instance_id)?;
-                Ok(Self::Web { renderer })
+                let buffers = Mutex::new(HashMap::new());
```
wkozyra95, line 44: We want to keep the code here minimal, consider creating WebRenderNode and moving this logic there.
Also you don't need mutex, you can change &self to &mut self in render method

compositor_render/src/renderer/texture/base.rs
```diff
@@ -55,4 +57,45 @@ impl Texture {
             self.texture.size(),
         );
     }
+
+    /// Returns `None` for some depth formats
+    pub(super) fn block_size(&self) -> Option<u32> {
+        self.texture.format().block_size(None)
+    }
+
+    pub(super) fn download_buffer(&self, ctx: &WgpuCtx) -> wgpu::Buffer {
```
wkozyra95, line 66:
```suggestion
    pub(super) fn new_download_buffer(&self, ctx: &WgpuCtx) -> wgpu::Buffer {
```

compositor_render/src/transformations/web_renderer/non_sync_send_handler.rs
```diff
@@ -0,0 +1,190 @@
...
+
+/// Handles structs which are not `Sync` and `Send`
+pub(super) struct NonSyncSendHandler {
+    embed_sources_sender: Sender<Vec<(NodeId, Resolution)>>,
+    shmem_update_sender: Sender<(NodeId, Arc<wgpu::Buffer>, wgpu::Extent3d)>,
+    /// Used for synchronizing buffer map and unmap operations
+    unmap_signal_receiver: Receiver<()>,
+}
+
+impl NonSyncSendHandler {
```
wkozyra95, line 24: WebRendererSender or ChromiumSender

compositor_render/src/transformations/web_renderer/non_sync_send_handler.rs
```diff
@@ -0,0 +1,190 @@
...
+        self.shmem_update_sender
+            .send((source_id, buffer, size))
+            .unwrap();
+
+        // Wait until buffer unmap is possible
+        self.unmap_signal_receiver.recv().unwrap();
+    }
+}
+
+struct InnerHandler {
```
wkozyra95, line 84: WebRendererSendThread or ChromiumSenderThread or ChromiumCommunicationThread

jerzywilczek (changes requested):
I I hope I understood everything correctly.

Names of `InnerHandler` and `NonSendSyncHandler` need to change and we need at least a `TODO` about synchronizing the access to shared memory.

compositor_render/src/transformations/web_renderer/node.rs
```diff
@@ -0,0 +1,62 @@
...
+            let size = (4 * pad_to_256(size.width) * size.height) as u64;
+
+            let recreate_buffer = match self.buffers.get(id) {
+                Some(buffer) => buffer.size() != size,
+                None => true,
+            };
+
+            if recreate_buffer {
+                self.buffers.insert(
+                    (*id).clone(),
```
wkozyra95, line 48:
```suggestion
                    id.clone(),
```
wouldn't this work?

compositor_render/src/transformations/web_renderer/node.rs
```diff
@@ -0,0 +1,62 @@
...
+        ctx: &mut RenderCtx,
+        sources: &[(&NodeId, &NodeTexture)],
+        target: &mut NodeTexture,
+    ) {
+        for (id, texture) in sources {
+            let Some(texture_state) = texture.state() else {
+                continue;
+            };
+
+            let texture = texture_state.rgba_texture();
```
wkozyra95, line 37: You could move the code that makes sure the buffer is the correct size to RGBATexture code, or maybe to separate function in WebRendererNode

wkozyra95 (approved):

compositor_render/src/transformations/web_renderer/node.rs
noituri: I would have to use `.clone()` twice

compositor_render/src/transformations/web_renderer.rs
```diff
@@ -12,13 +17,18 @@ use compositor_common::{
...
+mod chromium_sender;
+mod chromium_sender_thread;
+pub(crate) mod node;
+
+pub const EMBED_SOURCE_FRAMES_MESSAGE: &str = "EMBED_SOURCE_FRAMES";
+pub const UNEMBED_SOURCE_FRAMES_MESSAGE: &str = "UNEMBED_SOURCE_FRAMES";
 
-mod browser;
-pub mod chromium;
+pub const SHMEM_FOLDER_PATH: &str = "shmem";
```
jerzywilczek, line 31: wouldn't placing this in `/tmp` make sense?

jerzywilczek (approved):
looks good

compositor_render/src/transformations/web_renderer.rs
noituri: Yeah, I will change it to tmp now

PR #90, opened 2023-08-24
Pass missing frame through pipeline

wkozyra95:
- Instead of generating fake 2x2 texture make texture "Optional". The intention is that each component should decide how to handle missing frame e.g. shaders will convert mising frame to 1x1, but in the future we can e.g. change that behavior if there is only one input.
- If an empty frame is passed all the way to the output, we are filling it with black color
- If frame pts are older by more than the specified time (default 1 second) stop passing old frames to the pipeline. 
- Fix scene update: If scene update failed in old implementation it could break an old scene by removing all the input nodes. Now if scene update fails, old scene is not affected.

### Hacky part

- InputTexture and NodeTexture now have  a method sate that returns Texture and BindingGroup.
- I wanted to avoid dealing with multiple optional values when user needs both of them. For the same reason resize already returns non opional value because at that time we already know that there will be a texture there
- In case of InputTexture it's fine because it's a reference
- In case of NodeTexture it's a bit unsafe because technically caller could take that value and store it somehow. The main reason for that issue is that NodeTexture have internal mutability, so we can either 
  - return copy as it is done in this PR
  - expose mutex so calling code can lock it on it's own
  - use sth like owning-ref crate to return "refernce" to sth that owns a guard
  - Ideally we could rewrite scene to not wrap Nodes into Arcs and all of those issues would go away
  
 ### Follow up PR
 
- implement fallback mechanism
- maybe consider rewriting scene to stop using Arc<Node>

### DIFF
--- a/compositor_render/src/render_loop.rs
+++ b/compositor_render/src/render_loop.rs
@@ -2,36 +2,52 @@ use std::{collections::HashMap, sync::Arc, time::Duration};
 
 use compositor_common::{
     scene::{InputId, OutputId},
+    util::RGBColor,
     Frame,
 };
 use log::error;
 
-use crate::renderer::{
-    scene::{Node, Scene},
-    RenderCtx,
+use crate::{
+    frame_set::FrameSet,
+    renderer::{
+        scene::{Node, Scene},
+        RenderCtx,
+    },
 };
 
 pub(super) fn populate_inputs(
     ctx: &RenderCtx,
     scene: &mut Scene,
-    frames: &mut HashMap<InputId, Frame>,
+    frame_set: &mut FrameSet<InputId>,
 ) {
-    for (input_id, (node, input_textures)) in &mut scene.inputs {
-        let Some(frame) = frames.remove(input_id) else {
+    for (input_id, (_, input_textures)) in &mut scene.inputs {
+        let Some(frame) = frame_set.frames.remove(input_id) else {
+            input_textures.clear();
             continue;
         };
-        node.output.ensure_size(ctx.wgpu_ctx, frame.resolution);
+        if Duration::saturating_sub(frame_set.pts, ctx.stream_fallback_timeout) > frame.pts {
+            input_textures.clear();
+            continue;
+        }
+
         input_textures.upload(ctx.wgpu_ctx, frame);
     }
 
     ctx.wgpu_ctx.queue.submit([]);
 
     for (node, input_textures) in scene.inputs.values() {
-        ctx.wgpu_ctx.format.convert_yuv_to_rgba(
-            ctx.wgpu_ctx,
-            (input_textures.yuv_textures(), input_textures.bind_group()),
-            &node.output.rgba_texture(),
-        );
+        if let Some(input_textures) = input_textures.state() {
+            let node_texture = node
+                .output
+                .ensure_size(ctx.wgpu_ctx, input_textures.resolution());
+            ctx.wgpu_ctx.format.convert_yuv_to_rgba(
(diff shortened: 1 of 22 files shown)


compositor_render/src/renderer/texture.rs
WojciechBarczynski, line 1: I get the impression, that this is overcomplicated. Now we have: `InputTextureState`, `InputTexture`, `NodeTextureState`, `NodeTexture`, and `InnerNodeTexture` (4 of which are public).  Can't we simplify it?

compositor_render/src/renderer/texture.rs
```diff
@@ -93,46 +124,82 @@ impl InnerNodeTexture {
...
+    }
+
+    pub fn resolution(&self) -> Resolution {
+        texture_size_to_resolution(&self.texture.size())
+    }
+}
+
+struct InnerNodeTexture {
+    state: Option<NodeTextureState>,
+    is_empty: bool,
```
WojciechBarczynski, line 143: Instead of having `Option` here, use custom Enum with 3 possible values, representing 3 possible states, and implement protected access to it (like getters). `Option` would be suitable, if there was only one possible reason form the lack of access to `NodeTextureState, while in fact there are two.

compositor_render/src/render_loop.rs
```diff
@@ -2,36 +2,52 @@ use std::{collections::HashMap, sync::Arc, time::Duration};
...
+        let Some(frame) = frame_set.frames.remove(input_id) else {
+            input_textures.clear();
             continue;
         };
-        node.output.ensure_size(ctx.wgpu_ctx, frame.resolution);
+        if Duration::saturating_sub(frame_set.pts, ctx.stream_fallback_timeout) > frame.pts {
+            input_textures.clear();
+            continue;
+        }
+
```
WojciechBarczynski, line 32: I don't like the idea of clearing input textures without logging anything / notifying the user. It makes it harder to find whether you are getting the black screen on the output due to some rendering or network issue.

noituri (approved):

compositor_render/src/renderer/texture.rs
```diff
@@ -25,66 +29,90 @@ pub type YUVTextures = yuv::YUVTextures;
...
+            let textures = YUVTextures::new(ctx, new_resolution);
+            let bind_group = textures.new_bind_group(ctx, ctx.format.yuv_layout());
+            InputTextureState {
+                textures,
+                bind_group,
+            }
         }
-        self.textures.upload(ctx, &frame.data)
+
+        self.0 = match self.0.replace(OptionalState::None) {
```
wkozyra95, line 81: flagging this so it does not get missed during review. I couldn't find an approach where I could avoid this.
I don't need that for NodeTexture because I can clone that state.

compositor_render/src/render_loop.rs
wkozyra95: What would you suggest?
- We can't log it here every time frame is missing because it produce to much logs
- We don't really know what is the problem here regardless. We can only distinguish missing frames and frames that are too old.

WojciechBarczynski: We could keep track of which inputs we cleared last time and log only new ones. It would reduce log count, while we could still notify the user, that some frames are missing / too old.

WojciechBarczynski (approved):

compositor_render/src/render_loop.rs
wkozyra95: add issue https://github.com/membraneframework/video_compositor/issues/102

PR #98, opened 2023-08-25
Add fallback nodes

wkozyra95:
Add fallback nodes:
- Fallback can be defined via fallback_id field in the nodes list in the scene_update request
- There is no way currently to define fallback for input explicitly, you need to have some pass-through transformation.
- There is no way to define fallback on output directly, you need to set it on the last node.
- Currently, we do not have any nodes that can result in missing texture, even if the input is missing, so for now I just added hack to the shader node that clears texture if input is empty and only one.
- Replaced node iterator in render_loop with a recursive walk through the scene, we can't use old approach because without knowledge whether render worked or not we do not know whether we need to render fallback nodes at all.

### DIFF
--- a/compositor_render/src/render_loop.rs
+++ b/compositor_render/src/render_loop.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, time::Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    time::Duration,
+};
 
 use compositor_common::{
     scene::{InputId, NodeId, OutputId},
@@ -10,7 +13,7 @@ use log::error;
 use crate::{
     frame_set::FrameSet,
     renderer::{
-        scene::{Node, Scene, SceneError},
+        scene::{Node, Scene, SceneError, SceneNodesSet},
         RenderCtx,
     },
 };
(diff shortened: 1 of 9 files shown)


compositor_render/src/render_loop.rs
```diff
@@ -123,48 +126,60 @@ pub(super) fn run_transforms(
...
-        enqueue_nodes(child, queue, scene)?;
+    // render all inputs
+    {
+        let input_ids: Vec<_> = nodes.node(node_id)?.inputs.to_vec();
+        for input_id in input_ids {
+            render_node(ctx, nodes, pts, &input_id, already_rendered)?;
+        }
+    }
+    // render node `node_id` if render had an effect fallback_id returned
+    // from the block will be set to None even if fallback is defined
```
jerzywilczek, line 154:
```suggestion
    // render node `node_id`. If render had an effect return `None`.
    // If rendering failed and the node has `fallback_id`, return `Some(fallback_id)`
```
Had a problem understanding what this means for a solid 10mins 😅

jerzywilczek (approved):
:rocket:

WojciechBarczynski (approved):

PR #103, opened 2023-08-28
Shader validation part 1

jerzywilczek:
This PR adds validation to user-provided shaders. It checks whether the user has declared global variables with the same type and binding information as the ones we're passing to them and the correctness of the vertex shader argument type.

I want to add input validation to those custom shaders in a subsequent PR

### DIFF
--- a/compositor_render/src/renderer.rs
+++ b/compositor_render/src/renderer.rs
@@ -12,10 +12,9 @@ use crate::{
     render_loop::{populate_inputs, read_outputs},
     transformations::{
         image_renderer::ImageError,
-        text_renderer::TextRendererCtx,
-        web_renderer::chromium::{ChromiumContext, ChromiumContextError},
         shader::ShaderNewError,
         text_renderer::TextRendererCtx,
+        web_renderer::chromium::{ChromiumContext, ChromiumContextError},
     },
     WebRendererOptions,
 };
@@ -85,7 +84,7 @@ pub enum RendererInitError {
     FailedToInitChromiumCtx(#[from] ChromiumContextError),
 
     #[error("failed to initialize builtin transformation")]
-    BuiltInTransformationsInitError(#[from] WgpuError),
+    BuiltInTransformationsInitError(#[from] ShaderNewError),
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -251,9 +250,9 @@ impl WgpuCtx {
 
         let shader_header =
             naga::front::wgsl::parse_str(include_str!("transformations/shader_header.wgsl"))
-                .expect("shader header"); // TODO: handle
-
-        std::fs::write("header.module", format!("{shader_header:#?}")).unwrap();
+                .expect(
+                    "couldn't find shader header file, which is necessary for shader validation",
+                );
 
         let scope = WgpuErrorScope::push(&device);
 
--- a/compositor_render/src/renderer.rs
+++ b/compositor_render/src/renderer.rs
@@ -250,9 +250,7 @@ impl WgpuCtx {
 
         let shader_header =
             naga::front::wgsl::parse_str(include_str!("transformations/shader_header.wgsl"))
-                .expect(
-                    "couldn't find shader header file, which is necessary for shader validation",
-                );
+                .expect("failed to parse the shader header file");
 
         let scope = WgpuErrorScope::push(&device);
 
--- a/compositor_render/src/renderer.rs
+++ b/compositor_render/src/renderer.rs
@@ -14,6 +14,8 @@ use crate::{
         image_renderer::ImageError,
         text_renderer::TextRendererCtx,
         web_renderer::chromium::{ChromiumContext, ChromiumContextError},
+        shader::ShaderNewError,
+        text_renderer::TextRendererCtx,
     },
     WebRendererOptions,
 };
@@ -92,7 +94,7 @@ pub enum RendererRegisterError {
     RendererRegistry(#[from] registry::RegisterError),
 
     #[error("failed to to initialize the shader")]
-    Shader(#[from] WgpuError),
+    Shader(#[from] ShaderNewError),
 
     #[error("failed to create web renderer instance")]
     WebRendererInstance(#[from] WebRendererNewError),
@@ -176,11 +178,10 @@ pub enum RenderError {
 }
 
 pub struct WgpuCtx {
-    #[allow(dead_code)]
     pub device: wgpu::Device,
-
-    #[allow(dead_code)]
     pub queue: wgpu::Queue,
+
+    pub shader_header: naga::Module,
 
     pub format: TextureFormat,
     pub utils: TextureUtils,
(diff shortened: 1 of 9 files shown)


compositor_render/src/renderer.rs
```diff
@@ -248,6 +248,12 @@ impl WgpuCtx {
             None,
         ))?;
 
+        let shader_header =
+            naga::front::wgsl::parse_str(include_str!("transformations/shader_header.wgsl"))
+                .expect(
+                    "couldn't find shader header file, which is necessary for shader validation",
```
wkozyra95, line 254: include_str reads the file at compile time, so this error can only happen if the file exists but it's not valid. If that is the case, this error is a bit misleading.

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -0,0 +1,418 @@
+use naga::{ArraySize, Constant, ConstantInner, Handle, Module, ShaderStage, Type};
+
+#[derive(Debug, thiserror::Error)]
+pub enum ShaderValidationError {
+    #[error("a global that should be declared in the shader is not declared")]
+    GlobalNotFound(naga::GlobalVariable),
```
wkozyra95, line 6: You are not using that variable in the error message

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -0,0 +1,418 @@
+use naga::{ArraySize, Constant, ConstantInner, Handle, Module, ShaderStage, Type};
+
+#[derive(Debug, thiserror::Error)]
+pub enum ShaderValidationError {
```
wkozyra95, line 4: Error messages are inconsistent in the project in general. Sometimes we write full sentences starting with capital letters and ending with `.` and sometimes all lowercase.

Given that error messages sometimes will have more than one sentence the only consistent way I see to display them is to use full sentences starting with capital letters and `.` at the end.

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    let vertex_input = vertex.function.arguments[0].ty;
+    let vertex_input_type = &shader.types[vertex_input];
+
+    let (header_vertex_input, _) = header
+        .types
+        .iter()
+        .find(|(_, ty)| ty.name == vertex_input_type.name)
+        .ok_or(ShaderValidationError::VertexShaderBadInputTypeName(
```
wkozyra95, line 79:
```suggestion
        .ok_or_else(|| ShaderValidationError::VertexShaderBadInputTypeName(
```

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -0,0 +1,418 @@
...
+    CompositeTypeAsArrayLen(ConstantInner),
+}
+
+#[derive(Debug)]
+pub enum ArraySizeOrConstant {
+    ArraySize(ArraySize),
+    Constant(Constant),
+}
+
+fn type_equivalent(
```
wkozyra95, line 111: check_type_equivalent
or
validate_type_equivalent
or 
is_type_equivalen (and return type to bool)

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -0,0 +1,418 @@
...
+    let ti1 = if let Some(t) = type1.inner.canonical_form(&mod1.types) {
+        t
+    } else {
+        type1.inner.clone()
+    };
+    let ti2 = if let Some(t) = type2.inner.canonical_form(&mod2.types) {
+        t
+    } else {
+        type2.inner.clone()
+    };
```
wkozyra95, line 136:
```suggestion
    let ti1 = if let Some(ref t) = type1.inner.canonical_form(&mod1.types) {
        t
    } else {
        &type1.inner
    };
    let ti2 = if let Some(ref t) = type2.inner.canonical_form(&mod2.types) {
        t
    } else {
        &type2.inner
    };
```
not tested

also consider match instead of if else, it's one line shorter and at least in my opinion more readable

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -0,0 +1,418 @@
...
+                base: base2,
+                size: size2,
+                stride: stride2,
+            } = ti2
+            {
+                if stride1 != stride2 {
+                    return Err(TypeEquivalenceError::TypeStructureMismatch(ti1, ti2));
+                }
+
+                array_size_equivalent(size1, mod1, size2, mod2)?;
```
wkozyra95, line 171: rename, the same pattern as with type_equivalent comment

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -0,0 +1,418 @@
...
+            } else {
+                return Err(TypeEquivalenceError::TypeStructureMismatch(
+                    ti1.clone(),
+                    ti2.clone(),
+                ));
+            }
+        }
+
+        naga::TypeInner::Pointer { .. } => {
+            panic!("this should never happen bc of canonicalization")
```
noituri, line 224:
```suggestion
            unreachable!("this should never happen bc of canonicalization")
```

jerzywilczek: I'm afraid to do that, bc executing the unreachable! macro is UB and I'm not 100% certain canonicalization makes this impossible

compositor_render/src/transformations/shader/validation.rs
jerzywilczek: won't work, canonical_form returns a value that would not have an owner with this solution.

WojciechBarczynski (approved):

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -0,0 +1,416 @@
...
+fn validate_globals(
+    header: &naga::Module,
+    shader: &naga::Module,
+) -> Result<(), ShaderValidationError> {
+    for (_, global) in header.global_variables.iter() {
+        let Some((_, global_in_shader)) = shader.global_variables.iter().find(|(_, s_global)| {
+            s_global.space == global.space && s_global.binding == global.binding
+        }) else {
+            return Err(ShaderValidationError::GlobalNotFound(global.clone()));
+        };
```
wkozyra95, line 44:
```suggestion
        let (_, global_in_shader) = shader.global_variables
          .iter()
          .find(|(_, s_global)| {
            s_global.space == global.space && s_global.binding == global.binding
          })
          .ok_or_else(||ShaderValidationError::GlobalNotFound(global.clone()));
```

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -0,0 +1,416 @@
...
+        } => {
+            if let naga::TypeInner::BindingArray {
+                base: base2,
+                size: size2,
+            } = ti2
+            {
+                validate_array_size_equivalent(size1, mod1, size2, mod2)?;
+                return validate_type_equivalent(base1, mod1, base2, mod2);
+            }
+            return Err(TypeEquivalenceError::TypeStructureMismatch(ti1, ti2));
```
wkozyra95, line 187:
```suggestion
            let naga::TypeInner::BindingArray {
                base: base2,
                size: size2,
            } = ti2 else  {
                validate_array_size_equivalent(size1, mod1, size2, mod2)?;
                return validate_type_equivalent(base1, mod1, base2, mod2);
            }
            return Err(TypeEquivalenceError::TypeStructureMismatch(ti1, ti2));
```
I had to read this few times to realize this is a block for if statement. Maybe sth like this would be cleaner?

wkozyra95 (approved):

noituri (approved):

PR #115, opened 2023-08-31
Add process message handler to `cef::Client`

noituri:
Process message handler in `cef::Client` allows receiving messages back from the renderer subprocess

### DIFF
--- a/compositor_chromium/src/client.rs
+++ b/compositor_chromium/src/client.rs
@@ -1,4 +1,8 @@
+use std::os::raw::c_int;
+
 use crate::{
+    browser::Browser,
+    cef::{Frame, ProcessId, ProcessMessage},
     cef_ref::{CefRefData, CefStruct},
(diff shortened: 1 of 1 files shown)


compositor_chromium/src/client.rs
```diff
@@ -13,6 +17,16 @@ pub trait Client {
...
+
+    fn on_process_message_received(
+        &mut self,
+        _browser: &Browser,
+        _frame: &Frame,
+        _source_process: ProcessId,
+        _message: &ProcessMessage,
+    ) -> bool {
+        false
+    }
```
WojciechBarczynski, line 29: From function signature alone, I can't tell what return bool value means.

noituri: I added docs 🙂

WojciechBarczynski (approved):

wkozyra95 (approved):

PR #116, opened 2023-09-01
Improve error handling - part 2

wkozyra95:
Changes (both introduced and proposed new patterns):

- Stop using names like RendererNewError, instead sth like InitRendererError or CrateRendererError, or if you really want to name it the same as method NewRendererError.
This is very subjective, but at least for me that word order is very confusing. If anyone disagrees I won't oppose it, but we need to decide specific naming convention. e.g. check here https://rust-lang.github.io/api-guidelines/naming.html#names-use-a-consistent-word-order-c-word-order
- This PR fixes issues introduced by a combination of PRs that added layout components and shader validation. (One PR introduced changes to shaders structure and the other still validated the old structure.
- This is not the final PR, but this already grew to much plus it's moving a lot of top level types, so wanted to merge it sooner.

The main idea behind error handling:
- Extract top-level errors to error.rs in crate
- Each error is expected to provide all necessary information and context e.g. we can't just say that expected a name x and received y. We need to identify e.g.  shader where it happened.
- In most cases error should be either transparent or repeat message of an inner error. e.g.
```
#[error("Some error x. {inner_err}")]
```
- If `#[error(msg)]` does not provide any usefull info over err.source() does then change it to transparent.

### DIFF
--- a/compositor_common/src/validators.rs
+++ b/compositor_common/src/validators.rs
@@ -1,4 +1,7 @@
-use std::collections::{HashMap, HashSet};
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Display,
+};
 
 use crate::scene::{
     builtin_transformations::InvalidBuiltinTransformationSpec, NodeId, NodeParams, NodeSpec,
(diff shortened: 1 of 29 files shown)


compositor_common/src/validators.rs
```diff
@@ -112,12 +136,24 @@ impl SceneSpec {
...
         defined_node_ids: I,
-    ) -> Result<(), SpecValidationError> {
+        registered_inputs: &HashSet<&NodeId>,
+    ) -> Result<(), SceneSpecValidationError> {
         let mut nodes_ids: HashSet<&NodeId> = HashSet::new();
 
         for node_id in defined_node_ids {
             if !nodes_ids.insert(node_id) {
-                return Err(SpecValidationError::DuplicateNames(node_id.clone()));
+                match registered_inputs.contains(node_id) {
```
WojciechBarczynski, line 145: Use if

compositor_pipeline/src/error.rs
```diff
@@ -113,34 +117,94 @@ impl From<UnregisterOutputError> for PipelineError {
...
+            UpdateSceneError::NoNodeWithIdError(_) => {
+                // ServerError because it should be validated is spec validation
+                PipelineError::new(MISSING_NODE_WITH_ID, err, ErrorType::ServerError)
+            }
+            UpdateSceneError::WgpuError(err) => err.into(),
+            UpdateSceneError::UnknownResolutionOnOutput(_) => PipelineError::new(
+                UNKNOWN_RESOLUTION_ON_OUTPUT_NODE,
+                err,
+                ErrorType::ServerError,
+            ),
```
WojciechBarczynski, line 143: I think we also need a single-node prams validation (e.g. builtins validations, shaders params validation).
Keep in mind, that sometimes to determine if specs are valid we need to know the input count (e.g. when param specifies how single input should be rendered)

compositor_render/src/registry.rs
```diff
@@ -2,15 +2,9 @@ use std::{collections::HashMap, sync::Arc};
...
-pub enum GetError {
-    #[error("a {0} with a key {1} could not be found")]
-    KeyNotFound(&'static str, Arc<str>),
-}
-
 #[derive(Debug, thiserror::Error)]
 pub enum RegisterError {
-    #[error("a {0} with a key {1} is already registered")]
+    #[error("Failed to register a {0}, The \"{1}\" {0} was already registered.")]
     KeyTaken(&'static str, Arc<str>),
```
noituri, line 8: I know it wasn't changed in this PR but maybe use named fields here instead

compositor_render/src/registry.rs
```diff
@@ -43,15 +37,8 @@ impl<T: Clone> RendererRegistry<T> {
...
-    pub(crate) fn get(&self, key: &RendererId) -> Result<T, GetError> {
-        match self.registry.get(key) {
-            Some(val) => Ok(val.clone()),
-            None => Err(GetError::KeyNotFound(
-                self.registry_type.registry_item_name(),
-                key.0.clone(),
-            )),
-        }
+    pub(crate) fn get(&self, key: &RendererId) -> Option<T> {
+        self.registry.get(key).map(Clone::clone)
```
noituri, line 41:
```suggestion
        self.registry.get(key).cloned()
```

compositor_render/src/transformations/shader_header.wgsl
```diff
@@ -1,14 +1,16 @@
 struct VertexInput {
     @location(0) position: vec3<f32>,
     @location(1) tex_coords: vec2<f32>,
+    @location(2) video_id: i32,
```
WojciechBarczynski, line 4: I know that in the previous PR I used multiple, inconsistent names for this field (my bad. Now when we validate it, we should agree on a single, suitable name. 
I would prefer to use the `texture_id` name, since this id "tells" to which "texture plane" vertex belongs. This texture doesn't have to be video (it can be an image, text, basically anything node can produce). `input_id` on the other hand can be confused with compositor input streams.

WojciechBarczynski: Remember to rename it in other shaders as well

wkozyra95: sure I wasn't sure which was the prefered way, video_id was in more places so I assumed that was the intention

WojciechBarczynski (approved):
🚀 GJ

noituri (approved):

PR #119, opened 2023-09-04
Fix example + add image definition via local path

wkozyra95:
- Fix example (expired url to image)
- Increase timeout on example request (so the code in compositor timeouts quicker than the code in example)
- Improve error messages in image renderer


compositor_render/src/transformations/image_renderer.rs
```diff
@@ -59,11 +61,22 @@ impl Image {
...
+            }
+            Err(err) => {
+                match fs::read(path_or_url) {
+                    Ok(file) => Ok(Bytes::from(file)),
+                    // Fallback to url parsing error if reading
+                    // from disk fails.
+                    Err(_) => Err(err.into()),
+                }
+            }
+        }
```
WojciechBarczynski, line 79: I think it would be cleaner to handle that differently. Instead of this, we can accept the path provided by two options: "url" or "path"/"file"/"file_path"/etc. Passing the file path as "url" seems wrong to me, since it's usually reserved for web resources (https://en.wikipedia.org/wiki/URL).

WojciechBarczynski (approved):

compositor_render/src/transformations/image_renderer.rs
noituri: What about using `file://<path>`? Chromium uses file protocol for opening files. We could use the same approach to make it consistent with web renderer. Although, I'm not sure if web renderer/chromium should dictate how the rest of the compositor looks like

wkozyra95: - Passing the local path seems wrong to me, so I wanted to "hide" this feature a bit and only use that for our examples.
- file:// is not intuitive in my opinion especially when using it to define relative paths. I never remember how many slashes needs to be ther

I like the idea of having 2 fields path and url

PR #118, opened 2023-09-03
Stateful builtin transformations

WojciechBarczynski:
This PR makes builitins transformations stateful, allowing to perform specific calculations only on input updates (node inputs resolutions change) and send them via uniforms. 

The builtin node is no longer a wrapper over the shader node, allowing it to specify different fallback strategies for different transformations and dynamically adjust the output resolution of the node (allowing it to implement transformations such as `flip` in the future).

PR #122, opened 2023-09-04
Generic fallback strategies

WojciechBarczynski:
Allows specifying fallback strategies in a generic way. 
Adds optional fallback strategy specification on shader transformation register.

Should be merged after #118


WojciechBarczynski:
Moved into #127

PR #124, opened 2023-09-04
Rewrite V8Value part 1

noituri:
Currently V8Value contains methods for every JavaScript value. It requires developer to know what the underlying type is.

In this PR I separated value-specific methods into smaller structs which have to be pattern matched before they can be accessed. It removes the possibility of using wrong method on a wrong type.


wkozyra95 (approved):

compositor_chromium/src/v8/object.rs
```diff
@@ -0,0 +1,49 @@
...
+        let inner = self.0.get()?;
+        let key = CefString::new_raw(key);
+        unsafe {
+            let set_value = (*inner).set_value_bykey.unwrap();
+            Ok(set_value(inner, &key, value.get_raw()?, attribute as u32) == 1)
+        }
+    }
+
+    /// Returns `true` if value was deleted successfully
+    pub fn delete(&mut self, key: &str) -> Result<bool, V8ValueError> {
```
WojciechBarczynski, line 33: Since we return the result here, shouldn't unsuccessful value deletion be `Err`? 
Then this function typing would be `Result<(), DeletionError>`

WojciechBarczynski (approved):

compositor_chromium/src/v8/object.rs
noituri: Yeah, good idea

jerzywilczek (approved):
:rocket:

Issue #108, opened 2023-08-30
Use typestate pattern for scene spec validation


WojciechBarczynski:
Currently, we decode SceneSpec and validate it. This approach requires remembering to call `validate` on a decoded scene, before passing it to the renderer. We could use instead separate `NonValidatedScene` and `ValidatedScene`,  which would prevent the use of invalidated scene spec in the renderer

WojciechBarczynski:
Not useful anymore. It's highly improbable that we would forget to call `validate` on the most important struct :) 

This only adds unnecessary complexity and may introduce some misunderstandings, since scene validation isn't a single-step process (no entity currently has all the information necessary to perform validation - e.g. only the renderer has access to registered shaders, only the pipeline has access to registered inputs/outputs).

PR #125, opened 2023-09-06
Refactor scene validation code

wkozyra95:
Initially, this PR was aiming to address https://github.com/membraneframework/video_compositor/issues/108, but while implementing it I realized that it would complicate the implementation without any significant upside. It's very unlikely that we ever forgot to validate the scene when this is only done in place and it's one of the most important endpoints in our API.

Changes here are mostly moving stuff around, improving error messages and small refractors here and there. I'll add more in comments.

### DIFF
--- a/compositor_common/src/error.rs
+++ b/compositor_common/src/error.rs
@@ -1,0 +1,57 @@
+use std::{collections::HashSet, fmt::Display};
+
+use crate::scene::{NodeId, OutputId};
+
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum SceneSpecValidationError {
+    #[error("Unknown node \"{missing_node}\" used as an input in the node \"{node}\". Node is not defined in the scene and it was not registered as an input.")]
+    UnknownInputPadOnNode { missing_node: NodeId, node: NodeId },
+    #[error("Unknown node \"{missing_node}\" is connected to the output stream \"{output}\".")]
+    UnknownInputPadOnOutput {
+        missing_node: NodeId,
+        output: OutputId,
+    },
+    #[error(
+        "Unknown output stream \"{0}\". Register it first before using it in the scene definition."
+    )]
+    UnknownOutput(NodeId),
+    #[error("Invalid node id. There is more than one node with the \"{0}\" id.")]
+    DuplicateNodeNames(NodeId),
+    #[error("Invalid node id. There is already an input stream with the \"{0}\" id.")]
+    DuplicateNodeAndInputNames(NodeId),
+    #[error("Cycles between nodes are not allowed. Node \"{0}\" depends on itself via input_pads or fallback option.")]
+    CycleDetected(NodeId),
+    #[error(transparent)]
+    UnusedNodes(#[from] UnusedNodesError),
+    #[error("Invalid params for node \"{1}\". {0}")]
+    InvalidNodeSpec(#[source] NodeSpecValidationError, NodeId),
+}
+
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub struct UnusedNodesError(pub HashSet<NodeId>);
+
+impl Display for UnusedNodesError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut unused_nodes: Vec<String> = self.0.iter().map(ToString::to_string).collect();
+        unused_nodes.sort();
+        write!(
+            f,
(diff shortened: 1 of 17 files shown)


compositor_common/src/error.rs
```diff
@@ -0,0 +1,57 @@
+use std::{collections::HashSet, fmt::Display};
+
+use crate::scene::{NodeId, OutputId};
+
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
```
wkozyra95, line 5: All the errors grouped together are far more readable,

compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -17,41 +18,6 @@ pub enum BuiltinTransformationSpec {
     },
 }
 
-impl BuiltinTransformationSpec {
```
wkozyra95, line 20: I would suggest keeping types at the beginning of the file and `impl` at the end

compositor_common/src/scene/validation.rs
```diff
@@ -71,22 +21,22 @@ impl SceneSpec {
...
             .collect();
 
-        self.validate_input_pads_are_defined_on_node(&defined_node_ids)?;
-        self.validate_input_pads_are_defined_on_output(&defined_node_ids)?;
-        self.validate_outputs_registered(registered_outputs)?;
-        self.validate_node_ids_uniqueness(defined_node_ids_iter, registered_inputs)?;
-        self.validate_cycles(&transform_nodes)?;
-        self.validate_nodes_are_used(&transform_nodes)?;
-        self.validate_builtin_transformations()?;
+        Self::validate_input_pads_are_defined_on_node(&self.nodes, &defined_node_ids)?;
```
wkozyra95, line 24: the fact that validation methods sometimes used arguments and sometimes self, was hiding complexity. I would even consider removing those functions from `impl` and making them regular unassociated functions.

compositor_common/src/scene/validation_test.rs
```diff
@@ -58,9 +58,10 @@ fn scene_validation_finds_cycle() {
     let registered_inputs = HashSet::from([&input_id]);
     let registered_outputs = HashSet::from([&output_id]);
 
+    let validation_result = scene_spec.validate(&registered_inputs, &registered_outputs);
     assert_eq!(
-        scene_spec.validate(&registered_inputs, &registered_outputs),
-        Err(SceneSpecValidationError::CycleDetected(c_id.clone()))
+        validation_result.err(),
+        Some(SceneSpecValidationError::CycleDetected(c_id.clone()))
```
WojciechBarczynski, line 64:
```suggestion
        validation_result,
        Err(SceneSpecValidationError::CycleDetected(c_id.clone()))
```
?

compositor_common/src/scene/validation_test.rs
```diff
@@ -128,7 +129,9 @@ fn scene_validation_finds_unused_nodes() {
     let registered_outputs = HashSet::from([&output_id]);
 
     assert_eq!(
-        scene_spec.validate(&registered_inputs, &registered_outputs),
-        Err(UnusedNodesError(unused_nodes).into())
+        scene_spec
+            .validate(&registered_inputs, &registered_outputs)
+            .err(),
+        Some(UnusedNodesError(unused_nodes).into())
```
WojciechBarczynski, line 135:
```suggestion
        scene_spec
            .validate(&registered_inputs, &registered_outputs),
        Err(UnusedNodesError(unused_nodes).into())
```
?

WojciechBarczynski (approved):
🚀

compositor_common/src/scene/validation_test.rs
wkozyra95: err() returns Option

compositor_common/src/scene/validation_test.rs
wkozyra95: err() returns Option

compositor_common/src/scene.rs
```diff
@@ -1,15 +1,19 @@
...
 pub mod text_spec;
+mod validation;
+
+#[cfg(test)]
+mod validation_test;
+
+pub type NodeId = id::NodeId;
+pub type InputId = id::InputId;
+pub type OutputId = id::OutputId;
+pub type NodeParams = node::NodeParams;
```
jerzywilczek, line 16: why `type` instead of `use`?

jerzywilczek (approved):

PR #127, opened 2023-09-06
Add generic fallback strategies

WojciechBarczynski:


compositor_common/src/renderer_spec.rs
```diff
@@ -13,6 +13,25 @@ impl Display for RendererId {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+#[serde(rename_all = "snake_case")]
+pub enum FallbackStrategy {
+    NeverFallback,
+    FallbackIfAllInputsMissing,
+    FallbackIfAnyInputsMissing,
+    FallbackIfOnlyInputMissing,
```
jerzywilczek, line 22: i'd remove this, this is a weird option

compositor_common/src/renderer_spec.rs
```diff
@@ -13,6 +13,25 @@ impl Display for RendererId {
...
+}
+
+impl FallbackStrategy {
+    pub fn default_shader_fallback_strategy() -> FallbackStrategy {
+        FallbackStrategy::FallbackIfOnlyInputMissing
```
jerzywilczek, line 27:
```suggestion
        FallbackStrategy::FallbackIfAllInputsMissing
```

compositor_common/src/renderer_spec.rs
```diff
@@ -13,6 +13,25 @@ impl Display for RendererId {
...
+    }
+
+    pub fn default_shader_web_renderer_strategy() -> FallbackStrategy {
+        FallbackStrategy::FallbackIfOnlyInputMissing
```
jerzywilczek, line 31:
```suggestion
        FallbackStrategy::FallbackIfAllInputsMissing
```

compositor_render/src/renderer/scene.rs
```diff
@@ -125,6 +131,35 @@ impl RenderNode {
...
+                .iter()
+                .any(|(_, node_texture)| node_texture.is_empty()),
+            FallbackStrategy::FallbackIfOnlyInputMissing => {
+                sources.len() == 1
+                    && sources
+                        .iter()
+                        .all(|(_, node_texture)| node_texture.is_empty())
+            }
+        }
+    }
```
jerzywilczek, line 151: hsouldn't this be defined in `impl FallbackStrategy`? it seems weird to have it defined in here

jerzywilczek (changes requested):

compositor_render/src/renderer/scene.rs
WojciechBarczynski: We can't move it to impl FallbackStrategy, since it has to be located in `compositor_common`, which should be independent of `compositor_renderer`.
We would need to define some entity that would represent the state of input texture (whether it's empty), map sources to it and then run that method it `compositor_common`. IMO it's not worth it.

noituri (approved):
🙌

jerzywilczek (approved):
looks good, check out my response to ur comment

compositor_render/src/renderer/scene.rs
WojciechBarczynski: I can't see any upsides of creating this trait. It's so specific, that it would be hard to use it anywhere else (we would need to have access to `sources`, which are very specific for node).

I can implement mapping, but not sure whether it's useful in any way.
What do you think @wkozyra95 ?

compositor_common/src/renderer_spec.rs
```diff
@@ -13,6 +13,24 @@ impl Display for RendererId {
...
+
+impl FallbackStrategy {
+    pub fn default_shader_fallback_strategy() -> FallbackStrategy {
+        FallbackStrategy::FallbackIfAllInputsMissing
+    }
+
+    pub fn default_shader_web_renderer_strategy() -> FallbackStrategy {
+        FallbackStrategy::FallbackIfAllInputsMissing
+    }
+}
```
wkozyra95, line 32:
```suggestion
fn default_shader_fallback_strategy() -> FallbackStrategy {
    FallbackStrategy::FallbackIfAllInputsMissing
}

fn default_shader_web_renderer_strategy() -> FallbackStrategy {
    FallbackStrategy::FallbackIfAllInputsMissing
}
```
I assume serder does not need those methods to be public?

compositor_render/src/renderer/scene.rs
jerzywilczek: as I said, it's up to you. I like making things like this reusable in the future, but it's fine to leave it as-is

compositor_render/src/sync_renderer.rs
```diff
@@ -29,7 +29,11 @@ impl SyncRenderer {
         let mut guard = self.0.lock().unwrap();
         match spec {
             RendererSpec::Shader(spec) => {
-                let shader = Arc::new(Shader::new(&ctx.wgpu_ctx, spec.source)?);
+                let shader = Arc::new(Shader::new(
+                    &ctx.wgpu_ctx,
+                    spec.source,
```
wkozyra95, line 34: maybe pass entire spec object instead?

wkozyra95 (approved):

compositor_render/src/sync_renderer.rs
WojciechBarczynski: Can't do that, since builtins use this as well. I don't want to pass `ShaderSpec` to create Shader. `ShaderSpec` specifies `ShaderNode` params, while `Shader` is just abstraction over a single shader graphical pipeline.

PR #130, opened 2023-09-06
Rewrite `V8Value` part 2

noituri:
Part 2 of #124

I added some of the missing V8 types


compositor_chromium/src/v8/bool.rs
```diff
@@ -0,0 +1,31 @@
...
+
+        let inner = unsafe { chromium_sys::cef_v8value_create_bool(value) };
+        Self(Validated(inner))
+    }
+
+    pub fn get(&self) -> Result<bool, V8ValueError> {
+        let inner = self.0.get()?;
+        unsafe {
+            let get_value = (*inner).get_bool_value.unwrap();
+            Ok(get_value(inner) == 1)
```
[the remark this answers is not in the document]
noituri: `get_value` returns `1` if value is `true`

WojciechBarczynski (approved):

compositor_chromium/src/v8/array.rs
```diff
@@ -0,0 +1,87 @@
+use crate::validated::{Validated, ValidatedError};
+
+use super::value::{V8Value, V8ValueError};
+
+pub struct V8Array(pub(super) Validated<chromium_sys::cef_v8value_t>);
+
+impl V8Array {
+    pub fn has(&self, index: usize) -> Result<bool, V8ArrayError> {
+        let inner = self.0.get()?;
```
wkozyra95, line 9: is this usefull for anything? len should be enough to verify if elements in array exists

compositor_chromium/src/v8/array.rs
```diff
@@ -0,0 +1,87 @@
...
+    #[error(transparent)]
+    V8ValueError(#[from] V8ValueError),
+
+    #[error("V8Array element not found at index \"{0}\".")]
+    ElementNotFound(usize),
+
+    #[error("Failed to set V8Array at index \"{0}\".")]
+    SetFailed(usize),
+
+    #[error("Failed to delete V8Array element at index \"{0}\".")]
```
wkozyra95, line 85: I assume this can only happen if there is no element at that index? If that is the case I would expect error to state  sth like no element at the index or some sort of out-of-bounds error. Information that delete failed on it's own is a little supicious

wkozyra95 (approved):

compositor_chromium/src/v8/array.rs
noituri: `len()` returns max size of an array. The array itself can have less elements than its `len`. I think the name is confusing. I will rename it

compositor_chromium/src/v8/array.rs
noituri: I checked and the method returns `Ok(())` even if the index is out of bounds. According to CEF method comment deletion fails when  _"method is called incorrectly, deletion fails or an exception is thrown"_. I couldn't find anything more concrete

PR #134, opened 2023-09-07
Improve shader errors

wkozyra95:
Improve errors especially focusing on type comparisons. Serialization for values is not intended to cover everything in some cases I just left some simplified representation or default debug format. I'm open to suggestions on what to do with cases that we are not serializing exactly, but I don't think we should handle all the cases.

Example error:

before:
```
Type internal structure doesn't match:
    Expected:
    BindingArray {
        base: [9],
        size: Constant(
            [1],
        ),
    }

    Actual:
    Sampler {
        comparison: false,
    }.
```

after
```
Type mismatch (expected: binding_array<texture_2d<f32>, 16>, actual: sampler)
```

### DIFF
--- a/compositor_render/src/transformations/shader/error.rs
+++ b/compositor_render/src/transformations/shader/error.rs
@@ -2,13 +2,15 @@ use compositor_common::scene::shader::ShaderParam;
 
 use super::{USER_DEFINED_BUFFER_BINDING, USER_DEFINED_BUFFER_GROUP, VERTEX_ENTRYPOINT_NAME};
 
+const HEADER_DOCS_URL: &str = "https://docs.placeholder.com/shader_header";
+
 #[derive(Debug, thiserror::Error)]
 pub enum ShaderValidationError {
-    #[error("A global that should be declared in the shader is not declared: \n{0:#?}.")]
-    GlobalNotFound(naga::GlobalVariable),
+    #[error("A global \"{0}\" should be declared in the shader. Make sure to include the compositor header code inside your shader. Learn more: {HEADER_DOCS_URL}.")]
+    GlobalNotFound(String),
 
     #[error("A global in the shader has a wrong type.")]
-    GlobalBadType(#[source] TypeEquivalenceError),
+    GlobalBadType(#[source] TypeEquivalenceError, String),
 
     #[error("Could not find a vertex shader entrypoint. Expected \"fn {VERTEX_ENTRYPOINT_NAME}(input: VertexInput)\"")]
     VertexShaderNotFound,
@@ -16,9 +18,7 @@ pub enum ShaderValidationError {
     #[error("Wrong vertex shader argument amount: found {0}, expected 1.")]
     VertexShaderBadArgumentAmount(usize),
 
-    // TODO: do we enforce type name from header?
-    // #[error("The input type of the vertex shader has a name that cannot be found in the header.")]
-    #[error("The input type of the vertex shader (\"{0}\") was not declared.")]
+    #[error("The input type of the vertex shader has to be named \"VertexInput\".")]
     VertexShaderBadInputTypeName(String),
 
     #[error("The vertex shader input has a wrong type.")]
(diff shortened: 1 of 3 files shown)


compositor_render/src/transformations/shader/validation.rs
```diff
@@ -556,6 +563,95 @@ impl TypeInnerExt for naga::TypeInner {
...
+                let size: Option<&naga::Constant> = match size {
+                    ArraySize::Constant(size) => Some(&module.constants[*size]),
+                    ArraySize::Dynamic => None, // TODO: not sure how to handle this
+                };
+                let base: &naga::Type = &module.types[*base];
+                format!(
+                    "binding_array<{}, {}>",
+                    base.inner.to_string(module),
+                    size.map(|t| t.inner.to_string(module))
+                        .unwrap_or("_".to_string())
```
wkozyra95, line 616: unwrap will happen when BindingArray would have dynamic size which should not happen as far as I understand

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -556,6 +563,95 @@ impl TypeInnerExt for naga::TypeInner {
...
+
+impl ConstantInnerExt for naga::ConstantInner {
+    fn to_string(&self, _module: &naga::Module) -> String {
+        match self {
+            ConstantInner::Scalar { value, .. } => match value {
+                naga::ScalarValue::Sint(v) => format!("{}", v),
+                naga::ScalarValue::Uint(v) => format!("{}", v),
+                naga::ScalarValue::Float(v) => format!("{}", v),
+                naga::ScalarValue::Bool(true) => "true".to_string(),
+                naga::ScalarValue::Bool(false) => "true".to_string(),
```
noituri, line 635: `false` here

wkozyra95:
```suggestion
                naga::ScalarValue::Bool(false) => "false".to_string(),
```

compositor_render/src/transformations/shader/error.rs
```diff
@@ -2,23 +2,23 @@ use compositor_common::scene::shader::ShaderParam;
 
 use super::{USER_DEFINED_BUFFER_BINDING, USER_DEFINED_BUFFER_GROUP, VERTEX_ENTRYPOINT_NAME};
 
+const HEADER_DOCS_URL: &str = "https://docs.placeholder.com/shader_header";
```
WojciechBarczynski, line 5: This doesn't work

compositor_render/src/transformations/shader/error.rs
```diff
@@ -2,23 +2,23 @@ use compositor_common::scene::shader::ShaderParam;
...
     #[error("Could not find a vertex shader entrypoint. Expected \"fn {VERTEX_ENTRYPOINT_NAME}(input: VertexInput)\"")]
     VertexShaderNotFound,
 
     #[error("Wrong vertex shader argument amount: found {0}, expected 1.")]
     VertexShaderBadArgumentAmount(usize),
 
-    // TODO: do we enforce type name from header?
-    // #[error("The input type of the vertex shader has a name that cannot be found in the header.")]
-    #[error("The input type of the vertex shader (\"{0}\") was not declared.")]
+    #[error("The input type of the vertex shader has to be named \"VertexInput\".")]
```
WojciechBarczynski, line 21: Use that `String` or remove it

compositor_render/src/transformations/shader/error.rs
```diff
@@ -147,3 +143,37 @@ pub enum ParametersValidationError {
...
+            naga::AddressSpace::Storage { access } => {
+                let access = match access.bits() {
+                    1 => "read",
+                    2 => "write",
+                    3 => "read_write",
+                    _ => "",
+                };
+                format!("<storage, {}>", access)
+            }
+            naga::AddressSpace::Handle => "".to_string(),
```
WojciechBarczynski, line 173: Not sure what it means, but I got this error when missing "<uniform>" for the user-provided params (binding 1). Make sure this type of error is informative

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -131,8 +137,8 @@ fn validate_type_equivalent(
         | naga::TypeInner::ValuePointer { .. } => {
             if ti1 != ti2 {
                 return Err(TypeEquivalenceError::TypeStructureMismatch {
-                    expected: type1.inner.clone(),
-                    actual: type2.inner.clone(),
+                    expected: type1.inner.to_string(mod1),
+                    actual: type2.inner.to_string(mod2),
```
WojciechBarczynski, line 141: `type1` and `type2` are pretty bad names. Errors assume that those values have other meanings. Use `expected` and `actual`.

wkozyra95: I agree, but I don't really want to through all that code

WojciechBarczynski (approved):

PR #135, opened 2023-09-07
Add mirror-image transformation

wkozyra95:
- Add mirror image transformation with option to flip input vertically or horizontally
- default to horizontal
- output resolution is dynamic and depends on input resolution

### DIFF
--- a/compositor_common/src/scene/builtin_transformations.rs
+++ b/compositor_common/src/scene/builtin_transformations.rs
@@ -1,4 +1,7 @@
+use std::{fmt::Display, str::FromStr};
+
 use serde::{Deserialize, Serialize};
+use serde_with::{DeserializeFromStr, SerializeDisplay};
 
 use crate::{
     error::BuiltinSpecValidationError,
@@ -31,6 +34,10 @@ pub enum BuiltinSpec {
         tile_aspect_ratio: (u32, u32),
         resolution: Resolution,
     },
+    MirrorImage {
+        #[serde(default)]
+        mode: MirrorMode,
+    },
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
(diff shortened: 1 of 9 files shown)


compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -50,6 +57,38 @@ pub struct TextureLayout {
...
+    #[default]
+    Horizontal,
+    Vertical,
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("\"mode\" field can only be set to \"vertical\" or \"horizontal\".")]
+pub struct MirrorModeParseError;
+
+impl Display for MirrorMode {
```
wkozyra95, line 71: I had to implement custom serialization and deserizliation becuase flatten and default don't work together

compositor_render/src/transformations/builtin.rs
```diff
@@ -28,20 +28,30 @@ impl Builtin {
...
             BuiltinSpec::TransformToResolution { resolution, .. } => resolution,
             BuiltinSpec::FixedPositionLayout { resolution, .. } => resolution,
+            BuiltinSpec::MirrorImage { .. } => input_resolutions
+                .first()
+                .copied()
+                .flatten()
+                .unwrap_or(Resolution {
+                    width: 1,
+                    height: 1,
+                }),
```
WojciechBarczynski, line 46: We should probably create sth like fallbacks, where each of those transformations would only specify foxed resolution or resolution of input or other similar options.

compositor_render/src/transformations/builtin/mirror_image.wgsl
```diff
@@ -0,0 +1,53 @@
...
+}
+
+struct MirrorParams {
+    mode: u32,
+}
+
+var<push_constant> common_params: CommonShaderParameters;
+
+@group(0) @binding(0) var textures: binding_array<texture_2d<f32>, 16>;
+@group(1) @binding(0) var<uniform> mirror_params: MirrorParams;
```
WojciechBarczynski, line 26: Why not just u32?

WojciechBarczynski (changes requested):

compositor_render/src/transformations/builtin/node.rs
```diff
@@ -20,17 +20,18 @@ struct ParamsBuffer {
...
 impl ParamsBuffer {
-    pub fn new(mut content: bytes::Bytes, wgpu_ctx: &WgpuCtx) -> Self {
-        if content.is_empty() {
-            content = bytes::Bytes::copy_from_slice(&[0]);
-        }
+    pub fn new(content: bytes::Bytes, wgpu_ctx: &WgpuCtx) -> Self {
+        let content_or_zero = match content.is_empty() {
+            true => bytes::Bytes::copy_from_slice(&[0]),
+            false => content.clone(),
+        };
```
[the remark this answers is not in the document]
wkozyra95: I think match is a lot more readable in this case

> This copy is useless. Create buffer init needs only ref

initially I had `let content` shadow original content, and forgot to remove that clone after

compositor_render/src/transformations/builtin/params.rs
```diff
@@ -76,6 +78,12 @@ impl BuiltinParams {
             }
             BuiltinParams::Fit(fit_params) => fit_params.shader_buffer_content(),
             BuiltinParams::Fill(fill_params) => fill_params.shader_buffer_content(),
+            BuiltinParams::MirrorMode(MirrorMode::Horizontal) => {
+                bytes::Bytes::copy_from_slice(&0_u32.to_le_bytes())
+            }
+            BuiltinParams::MirrorMode(MirrorMode::Vertical) => {
+                bytes::Bytes::copy_from_slice(&1_u32.to_le_bytes())
+            }
```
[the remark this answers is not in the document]
wkozyra95: the point of extracting it to functions is to tie binary structure to the rust type, but here we don't have a type that we can add method too. So:
- I don't think that just extracting that to function is helping with anything.
- Given the simplicity I decided that this is good enough middle ground.
- I could create separate type that wrapps MirrorMode and add appropriate method on it. I think that it's a an overkill in this case, but if you feel strongly about it I could add that wrapper type

compositor_render/src/transformations/builtin/mirror_image.wgsl
wkozyra95: no reason, just my preference

compositor_render/src/transformations/builtin/node.rs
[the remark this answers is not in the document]
wkozyra95: ok nevermind. I need that clone because I can't return borrowed value because somone needs to own that 1-element slice

WojciechBarczynski (approved):

PR #136, opened 2023-09-07
Add tiled layout builtin

WojciechBarczynski:
Adds tile layout, that maximizes space covered by input textures on output texture


WojciechBarczynski:
Review suggestions in the follow up PR

compositor_common/src/error.rs
```diff
@@ -64,6 +64,12 @@ pub enum BuiltinSpecValidationError {
     FixedLayoutLeftRightOnlyOne,
     #[error("Nodes that use transformation \"transform_to_resolution\" need to have exactly one input pad.")]
     TransformToResolutionExactlyOneInput,
+    #[error(
+        "Nodes that use transformation \"tiled_layout\" need to have at most {TILED_LAYOUT_MAX_INPUTS_COUNT}."
```
wkozyra95, line 68: "Transformation \"tiled_layout\" supports up to {TILED_LAYOUT_MAX_INPUTS_COUNT} inputs."

compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -22,6 +24,13 @@ pub enum BuiltinSpec {
         #[serde(default)]
         background_color_rgba: RGBAColor,
     },
+    TiledLayout {
+        #[serde(default)]
+        background_color_rgba: RGBAColor,
```
wkozyra95, line 29: Is default fully transparent?

WojciechBarczynski: Yep

WojciechBarczynski: I planned to add an option to set the background as a node, allowing to set anything that node can produce as a background, but I'm not sure whether mixing it in the next transformation (sth like blend), wouldn't be better. What do you think?

wkozyra95: Yes, this should be the responsibility of a different component, either using fixed position layout or we could make an Overlay component with a dynamic resolution that takes 2 inputs  where (resolution is based on one of them)

PR #140, opened 2023-09-08
Implement unregistering renderers

### LINKED ISSUE #56: API for unregistering inputs/outputs/transformations

wkozyra95:
Closes #56

### DIFF
--- a/compositor_render/src/registry.rs
+++ b/compositor_render/src/registry.rs
@@ -11,6 +11,14 @@ pub enum RegisterError {
     },
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error("Failed to unregister a {item_type}. The \"{renderer_id}\" {item_type} does not exist.")]
+pub struct UnregisterError {
+    item_type: &'static str,
+    renderer_id: RendererId,
+}
+
+#[derive(Debug, Clone, Copy)]
 pub enum RegistryType {
     Shader,
     WebRenderer,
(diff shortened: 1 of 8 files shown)


compositor_render/src/registry.rs
```diff
@@ -11,6 +11,14 @@ pub enum RegisterError {
     },
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error("Failed to unregister a {item_type}. The \"{renderer_id}\" {item_type} does not exist.")]
+pub struct UnregisterError {
+    item_type: &'static str,
+    renderer_id: RendererId,
+}
```
WojciechBarczynski, line 19:
```suggestion
pub struct UnregisterItemNotFoundError {
    item_type: &'static str,
    renderer_id: RendererId,
}
```

There are a lot of other unregister errors

compositor_render/src/registry.rs
```diff
@@ -56,4 +64,14 @@ impl<T: Clone> RendererRegistry<T> {
...
     }
+
+    pub(crate) fn unregister(&mut self, id: &RendererId) -> Result<(), UnregisterError> {
+        match self.registry.remove(id) {
+            Some(_) => Ok(()),
+            None => Err(UnregisterError {
+                item_type: self.registry_type.registry_item_name(),
+                renderer_id: id.clone(),
+            }),
+        }
```
WojciechBarczynski, line 75: Use `.ok_or`

WojciechBarczynski (approved):

compositor_render/src/registry.rs
wkozyra95: I would prefer to leave it like this for consistency with RegisterError.

compositor_render/src/registry.rs
wkozyra95: ok_or maps None to Err, but here I need both map None to Err and Some(value) to `Ok(())`

PR #143, opened 2023-09-08
Corners rounding

WojciechBarczynski:


compositor_render/src/transformations/builtin/corners_rounding.wgsl
WojciechBarczynski, line 1: It's so weird to avoid using inefficient if statements

compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -166,6 +169,13 @@ impl BuiltinSpec {
                     return Err(BuiltinSpecValidationError::MirrorImageExactlyOneInput);
                 }
 
+                Ok(())
+            }
+            BuiltinSpec::CornersRounding { .. } => {
+                if node_spec.input_pads.len() != 1 {
+                    return Err(BuiltinSpecValidationError::CornerRadiusExactlyOneInput);
+                }
```
WojciechBarczynski, line 177: I'll create generic inputs count validation in next PR

compositor_render/src/transformations/builtin.rs
```diff
@@ -23,19 +23,19 @@ impl Builtin {
...
                 ..
-            } => Some(background_color_rgba),
+            } => Some(rgba_to_wgpu_color(background_color_rgba)),
+            BuiltinSpec::CornersRounding { .. } => Some(wgpu::Color::TRANSPARENT),
             BuiltinSpec::MirrorImage { .. } => None,
         }
-        .map(rgba_to_wgpu_color)
     }
 
     pub fn output_resolution(&self, input_resolutions: &[Option<Resolution>]) -> Resolution {
```
wkozyra95, line 41:
```suggestion
    pub fn output_resolution(&self, input_resolutions: &[Option<Resolution>]) -> Resolution {
        fn first_input(input_resolutions: &[Option<Resolution>]) -> Resolution {
            input_resolutions
                    .first()
                    .copied()
                    .flatten()
                    .unwrap_or(Resolution {
                        width: 1,
                        height: 1,
                    })
        }
```

compositor_render/src/transformations/builtin.rs
```diff
@@ -51,6 +51,14 @@ impl Builtin {
...
                     height: 1,
                 }),
+            BuiltinSpec::CornersRounding { .. } => input_resolutions
+                .first()
+                .copied()
+                .flatten()
+                .unwrap_or(Resolution {
+                    width: 1,
+                    height: 1,
+                }),
```
wkozyra95, line 61:
```suggestion
            BuiltinSpec::CornersRounding { .. } => first_input(input_resolutions),
```

wkozyra95 (approved):
We should consider what exactly this component needs to do, If we are rounding the corners then some forms of styling become impossible. e.g. If you want to define e.g. border style this needs to be done here, you can't write a separate component for that.

WojciechBarczynski:
> We should consider what exactly this component needs to do, If we are rounding the corners then some forms of styling become impossible. e.g. If you want to define e.g. border style this needs to be done here, you can't write a separate component for that.

Yep, I fought about it too. Probably we would need to create a builitn transformation for all boarder related stuff

PR #145, opened 2023-09-11
Break up utils

WojciechBarczynski:
I wanted to add some params to transformations and found out, that `utils.rs` is too large. It has implementations of a lot of traits and tests for widely different structs. I think it's easier to maintain it, while everything is split.


compositor_common/src/util/degree.rs
WojciechBarczynski, line 1: Even though those files are fairly small, I think it's easier to navigate thought them, when they are separated in a consistent way.

wkozyra95:
We are hitting this problem regularly and in some cases moving type is a better option, but I'm wondering whether this approach is the correct or idiomatic one.

One alternative that is worth considering is keeping type definitions in one file and moving impl to a different file.

wkozyra95 (approved):

PR #146, opened 2023-09-11
Fix fallback on output

wkozyra95:
The fallback on output did not work. It should be triggered when an empty texture is passed all the way to the output node and no fallback returns a real texture. When this happens compositor should send frames with black stream, but instead it was  not sending any frames and ffplay shown the last frame.


noituri (approved):

WojciechBarczynski (commented):
LGTM

PR #147, opened 2023-09-11
Add some tests for parameters validation

jerzywilczek:


noituri (approved):

WojciechBarczynski (changes requested):
IMO those tests are poorly written.
They don't check whether the returned error is the correct one (you can use [`assert_matches`](https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html) for that).
Values names are bad. It's hard to spot what is the expected value and what the "received"/"checked" one.

wkozyra95 (commented):
I agree with other suggestion(checking for actual error + actual/expected variable naming)

Also, I think the test module is too big to be still in validation.rs

jerzywilczek:
> (you can use [`assert_matches`](https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html) for that)

This is unfortunately an experimental API that can't be used in stable ;(

WojciechBarczynski:
> > (you can use [`assert_matches`](https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html) for that)
> 
> This is unfortunately an experimental API that can't be used in stable ;(

Ohh, I haven't noticed that :)

wkozyra95 (approved):

WojciechBarczynski (approved):
You don't need to change it here, but in future in such a check, you could verify the error exactly. 

e.g. In test `array_len()` we only check that validation returned `Err(ShaderValidationError::GlobalBadType(_, _))`, but values inside `ShaderValidationError::GlobalBadType` don't need to match. The error message provided to the user still can be misleading in such cases.

I know that in some cases you can't do it, because some types don't have `PartialEq` derived, but overall this should be done whenever it's possible.

PR #149, opened 2023-09-11
Fit align params

WojciechBarczynski:
Rewrites fit transformation. Adds horizontal and vertical align to fit to resolution

### DIFF
--- a/compositor_common/src/scene/builtin_transformations.rs
+++ b/compositor_common/src/scene/builtin_transformations.rs
@@ -58,8 +58,8 @@ pub enum TransformToResolutionStrategy {
     Fit {
         #[serde(default)]
         background_color_rgba: RGBAColor,
-        horizontal_align: HorizontalAlign,
-        vertical_align: VerticalAlign,
+        horizontal_alignment: HorizontalAlign,
+        vertical_alignment: VerticalAlign,
     },
 }
 
--- a/compositor_common/src/scene/builtin_transformations.rs
+++ b/compositor_common/src/scene/builtin_transformations.rs
@@ -4,6 +4,8 @@ use serde::{Deserialize, Serialize};
 use serde_with::{DeserializeFromStr, SerializeDisplay};
 
 use crate::error::BuiltinSpecValidationError;
+use crate::util::align::HorizontalAlign;
+use crate::util::align::VerticalAlign;
 use crate::util::colors::RGBAColor;
 use crate::util::coord::Coord;
 use crate::util::degree::Degree;
@@ -56,6 +58,8 @@ pub enum TransformToResolutionStrategy {
     Fit {
         #[serde(default)]
         background_color_rgba: RGBAColor,
+        horizontal_align: HorizontalAlign,
+        vertical_align: VerticalAlign,
     },
 }
 
--- a/compositor_render/src/transformations/builtin/box_layout.rs
+++ b/compositor_render/src/transformations/builtin/box_layout.rs
@@ -2,7 +2,6 @@ use compositor_common::{
     scene::Resolution,
     util::align::{HorizontalAlign, VerticalAlign},
 };
-use log::info;
 use nalgebra_glm::{rotate_z, scale, translate, vec3, Mat4, Vec3};
 
 #[derive(Debug)]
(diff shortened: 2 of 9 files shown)


compositor_render/src/transformations/builtin/box_layout.rs
```diff
@@ -10,6 +14,45 @@ pub struct BoxLayout {
...
+        BoxLayout {
+            top_left_corner: (
+                self.top_left_corner.0 + left_padding,
+                self.top_left_corner.1 + top_padding,
+            ),
+            width: self.width - x_padding,
+            height: self.height - y_padding,
+            rotation_degrees: self.rotation_degrees,
+        }
+    }
```
WojciechBarczynski, line 54: This probably has less numerical precision, than the old approach, but since it's significantly simpler, self-documenting, and doesn't require any thinking at all, I decided to leave it this way. With floats, this shouldn't have any visible effect anyway.

compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -56,6 +58,8 @@ pub enum TransformToResolutionStrategy {
     Fit {
         #[serde(default)]
         background_color_rgba: RGBAColor,
+        horizontal_align: HorizontalAlign,
+        vertical_align: VerticalAlign,
```
WojciechBarczynski, line 62: Question to reviewers: any better names?
I thought about using the API most compliant with CSS, but after analyzing how unreadable this is in CSS, I left it as it is.

compositor_render/src/transformations/builtin/params/transform_to_resolution.rs
```diff
@@ -1,43 +1,33 @@
...
-    /// scale_factor_x_clip_space = 1.0 (input x coords are already fitted)
-    /// scale_factor_y_clip_space = scale_factor_pixels * input_height / output_height =
-    /// = (output_width * input_height) / (output_height * input_width)
-    /// = output_ratio / input_ratio
-    ///
-    /// If the output ratio is larger, then the texture is scaled up,
-    /// such that input_height = output_height.
-    /// Analogously:
-    /// scale_factor_x_clip_space = input_ratio / output_ratio
-    /// scale_factor_y_clip_space = 1.0 (input y coords are already fitted)
```
WojciechBarczynski, line 27: download

compositor_render/src/transformations/builtin/box_layout.rs
```diff
@@ -10,6 +14,45 @@ pub struct BoxLayout {
 }
 
 impl BoxLayout {
+    /// Returns box representing position of input frame fitted into `self`,     
+    /// Fitted means scaled to input resolution without cropping or changing aspect ratio
+    pub fn fit(
```
wkozyra95, line 19: This is great, but I worry a bit about how this will scale, you will add fit and stretch next etc. maybe some other transform too.

If fit and stretch are the only parts we will ever add here then it's fine, but if we are planning to add anything else I would suggest a different abstraction e.g. trait BoxLayoutTransformation or maybe just a standalone function that are not associated to the BoxLayout type

wkozyra95 (approved):

compositor_render/src/transformations/builtin/box_layout.rs
WojciechBarczynski: Yep, so far I haven't planned to add anything other than `fit`/`fill` here (`stretch` happens automatically). I think we can break it down once we need to add more stuff here (probably into functions).

compositor_common/src/scene/builtin_transformations.rs
jerzywilczek: I think this is fine, but I would change '\*align\*' -> '\*alignment\*'

jerzywilczek (approved):
cool

noituri (approved):

compositor_common/src/scene/builtin_transformations.rs
WojciechBarczynski: In text renderer we use `align` to be consistent with CSS `text-align`, so we have to decide to either:
1. Use everywhere one - `align` or `alignment`
2. Use both `align` and `alignment`, picking the one most suitable / sounding better to a specific case

I have no preference, but if this changes your decision / doesn't seem right to you, let me know. Otherwise, I will use `alignment` here and keep `align` in the text renderer.

PR #151, opened 2023-09-12
Constraints system

WojciechBarczynski:
This PR adds constraints validation system. I ultimately decided to implement API for constraints definition in this PR, because it was hard to clarify why validation must happen in `compositor_renderer` (`compositor_common` shouldn't depend on `compositor_renderer`).

## API for defining constraints:
Constraints can be defined as a list of objects on shader or web renderer registration, like:
```rust
    common::post(&json!({
        "type": "register",
        "entity_type": "web_renderer",
        ...,
        "constraints": [
            {
                "type": "inputs_count",
                "lower_bound": 1,
                "upper_bound": 5,
            }
        ]
    }))?;
```

Constratints for `Image`, `Text`, `Builtins` and defaults for `Shader` and `WebRenderer` are predefined. 

## Error messages:
Example error messages for unsatisfied constraints:
```console
    Error stack:
     - Constraints for node "mirrored_image" are not satisfied.
     - "mirror_image" builtin transformation expects exactly one input pad, but none input pads were specified.
```

```console
    Error stack:
     - Constraints for node "layout" are not satisfied.
     - "tiled_layout" builtin transformation expects at least 1 and at most 16 input pads, but 24 input pads were specified.
    
```

## Important changes in `renderer`
I had to refactor some essential parts of the renderer to keep things cleaner. Most notably: 
- `Shader` was split into two parts: `Shader` and `GpuShader` (WIP name). `Shader` represents a single shader renderer, while `GpuShader` is an abstraction over a single wgpu shader - common part for `Bultin` and `Shader`
- Constraints validation happens in `renderer` on scene update, because `Shaders` and `WebRenderers` keep user-defined constraints.
- `GpuShader` doesn't hold `clear_color` and `fallback_strategy` anymore
- `Builtin` wrapper now holds `GpuShader`, like `Shader`


compositor_common/src/error.rs
```diff
@@ -62,19 +62,12 @@ pub enum BuiltinSpecValidationError {
...
-        TILED_LAYOUT_MAX_INPUTS_COUNT
-    )]
-    TiledLayoutTooManyInputs,
-    #[error("Nodes that use transformation \"tiled_layout\" need to have at least one input pad.")]
-    TiledLayoutNoInputs,
-    #[error("Nodes that use transformation \"mirror_image\" need to have exactly one input pad.")]
-    MirrorImageExactlyOneInput,
-    #[error("Nodes that use transformation \"corner_radius\" need to have exactly one input pad.")]
-    CornerRadiusExactlyOneInput,
+    #[error("Nodes that use transformation \"{}\" expects {} input pads, but {} were provided", transformation_name, valid_input_pads_count.error_message_inputs_count(), defined_input_pads_count)]
```
wkozyra95, line 65:
```suggestion
    #[error("Nodes that use transformation \"{}\" expect {} input pads, but {} were provided.", transformation_name, valid_input_pads_count.error_message_inputs_count(), defined_input_pads_count)]
```
I would prefer if you kept error messages for exactly on input separate from errors where expected number of inputs is constant but more than one.

Also, there is sth wrong with that wording "Nodes that use transformation \"{}\" expect" (Yes, I know that I'm probably an author)

compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -105,15 +108,48 @@ impl FromStr for MirrorMode {
 }
 
 impl BuiltinSpec {
-    pub fn validate(&self, node_spec: &NodeSpec) -> Result<(), BuiltinSpecValidationError> {
+    fn transformation_name(&self) -> String {
```
wkozyra95, line 111:
```suggestion
    fn transformation_name(&self) -> &'static str {
```

compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -105,15 +108,48 @@ impl FromStr for MirrorMode {
...
+            BuiltinSpec::FixedPositionLayout { .. } => "fixed_position_layout".to_owned(),
+            BuiltinSpec::TiledLayout { .. } => "tiled_layout".to_owned(),
+            BuiltinSpec::MirrorImage { .. } => "mirror_image".to_owned(),
+            BuiltinSpec::CornersRounding { .. } => "corners_rounding".to_owned(),
+        }
+    }
 
-                Ok(())
-            }
+    fn valid_inputs_count(&self) -> ValidNodeInputsCount {
```
wkozyra95, line 121: This is a good solution for input count but it does not scale. I'm fine if you want to lond it, but this will need to be reworked to implement constraint, so I don't know if it makes sense to take this approach.

Also I don't think this should be limited only to builtin types (although it most useful here)

compositor_common/src/error.rs
WojciechBarczynski: > I would prefer if you kept error messages for exactly on input separate from errors where expected number of inputs is constant but more than one.
 
Implementing a separate variant for 1 input without implementing a separate one for 0 inputs as well seems weird to me. I don't see benefits coming from treating one input differently than any other fixed input count value. We will have builtins that require other specific counts of inputs (e.g. fadings).
Do you have in mind any specific benefits related to separating it? I think that the error message would be similar / the same.

> Also, there is sth wrong with that wording "Nodes that use transformation "{}" expect" (Yes, I know that I'm probably an author)

Do you mean those parentheses are weird or sth else? I can remove them, no preference for that

compositor_common/src/scene/builtin_transformations.rs
WojciechBarczynski: I thought about returning the struct holding all node constraints and then checking. This would need to take place in the renderer, during scene update (to have access to user-defined constraints of shaders/web renderers). I can't come up with constraints different than input count and static/dynamic resolution. I think implementing static/dynamic resolution validation is a wrong idea because it makes API more difficult for newcomers. 

Can you think of any other potential constraints? I don't want to unnecessarily overcomplicate it.

compositor_common/src/error.rs
WojciechBarczynski: Ok, I'll try to fix it :)

compositor_common/src/scene/builtin_transformations.rs
wkozyra95: > I think implementing static/dynamic resolution validation is a wrong idea because it makes API more difficult for newcomers.

This is an argument to not have those constraints on our components, but the user might want to add constraints like this on their own shaders to protect them from mistakes.

Potential constraints:
- input count
  - exact value (0, 1, n)
  - less than n
  - at least n
  - has to be one of the specific values
- resolution
  - exact resolution
  - constant resolution
  - max input size (either on specific dimension or overall)
- static content (only allow inputs that are static and need to be rendered only once)
- infallible (always produces an output either via fallback or on its own)
- You can't pass the same input more than once (not sure what's the use case)
- Allow only nodes of specific type e.g. only inputs/ only images /...
- Allow only inputs named according to specific naming conventions e.g. via regexp or some prefix/suffix

WojciechBarczynski: Ok, I'll add shaders and web render inputs count specification, make builtins return validation struct, and validate it in renderer on scene update.
Having some validation in compositor common makes sense to me in transformations, e.g. `FixedPositionLayout`, where each input requires some params (we can provide more better error message). Other checks should be uniform across all transformations.

Other constraints are outside of the scope of this PR.

wkozyra95: > I'll add shaders and web render inputs count specification

I think it's outside of scope of this PR too. Unless you want to do this.

I would just create a Constraint abstraction attach it to the node (e.g. method that returns constraints) and run constraint validation on scene update.

compositor_render/src/renderer/scene/constraints.rs
```diff
@@ -0,0 +1,46 @@
...
+    for (node_id, node_constraints) in scene
+        .nodes
+        .iter()
+        .map(|node| (&node.node_id, node_constraints(node)))
+    {
+        node_constraints.validate(scene, node_id)?
+    }
+
+    Ok(())
+}
```
WojciechBarczynski, line 19: This function will need to take renderer contexts when web renderer and shader constraints will be user / API configurable.
I haven't put it in this PR, to firstly review the general idea for the constraints system.

compositor_common/src/error.rs
```diff
@@ -44,6 +44,58 @@ impl Display for UnusedNodesError {
...
+        maximal_inputs_expected,
+        defined_input_pads_count
+    )]
+    InvalidBoundedInputsRequired {
+        transformation_name: Arc<str>,
+        minimal_inputs_expected: u32,
+        maximal_inputs_expected: u32,
+        defined_input_pads_count: u32,
+    },
+}
```
WojciechBarczynski, line 97: @wkozyra95  
1. Do you still think we should have 4 different errors? IMO first 3 are basically the same and we could have just 2 errors defined here
2. Would you put those errors here or move them into a separate module?

compositor_common/src/error.rs
```diff
@@ -44,6 +44,58 @@ impl Display for UnusedNodesError {
...
+    InvalidInputsPads(#[source] InputsCountValidationError, NodeId),
+}
+
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum InputsCountValidationError {
+    #[error(
+        "Nodes that use transformation \"{}\" doesn't expect input pads, but {} input pads were specified.",
+        transformation_name,
+        defined_input_pads_count
+    )]
```
wkozyra95, line 59:
```suggestion
    #[error("Nodes that use transformation \"{transformation_name}\" don't support any inputs.")]
```
Also, "Nodes that use transformation" makes sense for builtins, but wouldn't e.g. for image renderer

compositor_common/src/error.rs
```diff
@@ -44,6 +44,58 @@ impl Display for UnusedNodesError {
...
+    NoInputsRequired {
```
wkozyra95, line 60:
```suggestion
    NoInputsAllowed {
```

compositor_common/src/error.rs
```diff
@@ -44,6 +44,58 @@ impl Display for UnusedNodesError {
     }
 }
 
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum ConstraintsValidationError {
+    #[error("Specification of node \"{1}\" doesn't satisfy node input pads count constraint.")]
```
wkozyra95, line 49: I would merge InputsCountValidationError into ConstraintsValidationError. If you prefer to keep them separate then make it transparent.

I would like the user to see sth like this:
- Specification of node \"{1}\" is invalid / Specification of node \"{1}\" failed constraint validation.
- Node does not expect any input pads. / Node does not expect any input pads, but some were provided.

compositor_common/src/error.rs
wkozyra95: 1. I don't have an opinion on that, but I want to be able to specify error messages for those cases in a more elastic way. Whether it's one error or not does not matter to me.
2. Up to you, maybe move it down a bit. (My idea is to keep top-level types higher, but this is subjective) . If error does not have any lower-level source errors then it's fine to move it somewhere else.

I raised my concerns about wording "Nodes that use transformation" before and that was subjective, but when you moved it here it's no longer the case, because now it's not correct anymore, because those errors are no longer builtin specific. Try to reword it, but if you don't have a better idea of how to phrase it then:
 - Rename errors to clarify that they are specific to built-in transformation. (and leave current wording)
 - Remove all the variants that are not currently used.
 - Add an issue to create generic error messages for other types (you can assign me to it)

compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -153,31 +160,26 @@ impl BuiltinSpec {
...
+            BuiltinSpec::TiledLayout { .. }
+            | BuiltinSpec::MirrorImage { .. }
+            | BuiltinSpec::CornersRounding { .. }
+            | BuiltinSpec::TransformToResolution { .. } => Ok(()),
+        }
+    }
 
-                Ok(())
-            }
+    pub fn inputs_constrains(&self) -> InputsCountConstraint {
```
wkozyra95, line 170:
```suggestion
    pub fn constrains(&self) -> InputsCountConstraint {
```

compositor_common/src/error.rs
```diff
@@ -44,6 +44,58 @@ impl Display for UnusedNodesError {
     }
 }
 
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum ConstraintsValidationError {
```
wkozyra95, line 48: The name is a bit misleading when this error is happening it means that constraints were not satisfied. It's fare to say that it's a validation error, but if that is the cases then it's scene validation error.

You could name it e.g. UnsatisfiedConstraintsError

compositor_common/src/error.rs
```diff
@@ -44,6 +44,58 @@ impl Display for UnusedNodesError {
...
+        "Nodes that use transformation \"{}\" doesn't expect input pads, but {} input pads were specified.",
+        transformation_name,
+        defined_input_pads_count
+    )]
+    NoInputsRequired {
+        transformation_name: Arc<str>,
+        defined_input_pads_count: u32,
+    },
+    #[error(
+        "Nodes that use transformation \"{}\" expect single input pad, but {} input pads were specified.",
```
wkozyra95, line 65:
```suggestion
        "Nodes that use transformation \"{}\" require exactly one input pad.",
```

compositor_common/src/error.rs
```diff
@@ -44,6 +44,58 @@ impl Display for UnusedNodesError {
...
+        expected_input_pads_count,
+        defined_input_pads_count
+    )]
+    ExactNumberOfInputsRequired {
+        transformation_name: Arc<str>,
+        expected_input_pads_count: u32,
+        defined_input_pads_count: u32,
+    },
+    #[error(
+        "Nodes that use transformation \"{}\" expect at least {} and at most {} input pads, but {} were specified",
```
wkozyra95, line 85: There should be different message for case where you need more than x and when you need less than x.

compositor_common/src/scene/builtin_transformations.rs
wkozyra95: It should return some wrapper type not input specifically, plus single node can have more than one constraint e.g. `Vec<NodeConstraint>`

compositor_common/src/scene/validation/inputs.rs
```diff
@@ -0,0 +1,66 @@
+use std::sync::Arc;
+
+use crate::error::InputsCountValidationError;
+
+pub enum InputsCountConstraint {
```
wkozyra95, line 5: I would move it under constraint

compositor_render/src/renderer/scene/constraints.rs
```diff
@@ -0,0 +1,46 @@
...
+
+fn node_constraints(node: &NodeSpec) -> NodeConstraints {
+    // TODO: make web renderer and shader constraints API configurable
```
wkozyra95, line 22: I would prefer not to have any node-specific code here. I would either:
- move the entire function to  NodeSpec or NodeParams
- leave a match statement here, but only call to some method on each of those types.

This file should not contain any logic specific to a type

WojciechBarczynski: In the case of user-defined constraints, we need to access them somehow in runtime. We can't move into `comspoitor_common`, since this function will require access to some renderer contexts, and `compositor_common` shouldn't depend on `compositor_renderer`

compositor_render/src/renderer/scene/constraints.rs
wkozyra95: > This function will need to take renderer contexts when web renderer and shader constraints will be user / API configurable.

This function can't take that as param if you define it in common crate.

compositor_common/src/error.rs
```diff
@@ -44,6 +44,23 @@ impl Display for UnusedNodesError {
     }
 }
 
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum UnsatisfiedConstraintsError {
+    #[error(
+        "Invalid input pads specification for node \"{}\". {} requires {}. {}",
```
wkozyra95, line 50: I think we decided that we don't want to chain error like this, and we are using stacktrace-like approach. See my comment https://github.com/membraneframework/video_compositor/pull/151#discussion_r1324367394

compositor_common/src/scene/validation/constraints.rs
```diff
@@ -0,0 +1,33 @@
...
+};
+
+use self::input_count::InputsCountConstraint;
+
+pub mod input_count;
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct Constraints {
+    pub inputs_count: InputsCountConstraint,
+}
```
wkozyra95, line 15:
```suggestion
pub enum Constraint {
    InputCount(InputsCountConstraint),
}
```

compositor_common/src/scene/validation/constraints/input_count.rs
```diff
@@ -0,0 +1,56 @@
...
+            Err(UnsatisfiedConstraintsError::InvalidInputsCount {
+                node_id: node_spec.node_id.clone(),
+                identification_name: node_spec.identification_name(),
+                input_count_constrain: self.clone(),
+                defined_input_pads_count,
+            })
+        }
+    }
+
+    pub fn required_inputs_message(&self) -> Rc<str> {
```
wkozyra95, line 40: methods that return string that will be concatenated with sth else are very sketch, it's very easy for  that to break, if you need sth like that it should be as private as possible and close to the place where it's used. Yes, there is sth similar in registry implementation, but I would like to get rid of that too.

compositor_render/src/renderer/renderers.rs
```diff
@@ -12,15 +12,15 @@ use crate::{
...
 use super::WgpuCtx;
 
-pub(crate) struct Renderers {
-    pub(crate) shaders: RendererRegistry<Arc<Shader>>,
-    pub(crate) web_renderers: RendererRegistry<Arc<WebRenderer>>,
-    pub(crate) images: RendererRegistry<Image>,
-    pub(crate) builtin: BuiltinTransformations,
+pub struct Renderers {
+    pub shaders: RendererRegistry<Arc<Shader>>,
+    pub web_renderers: RendererRegistry<Arc<WebRenderer>>,
```
wkozyra95, line 17: this should not be public

compositor_render/src/transformations/shader_executor.rs
```diff
@@ -0,0 +1,182 @@
+use crate::renderer::{
```
wkozyra95, line 1: why? what is shader_executor? What is suppose to be in this module?

compositor_common/src/error.rs
```diff
@@ -100,3 +104,13 @@ impl<'a> Iterator for ErrorStack<'a> {
         })
     }
 }
+
+impl UnsatisfiedConstraintsError {
```
wkozyra95, line 108: ErrorStack is not a part of this error hierarchy, so I would prefer to keep it either at the very begining or end of the file

compositor_render/src/registry.rs
```diff
@@ -35,7 +35,7 @@ impl RegistryType {
     }
 }
 
-pub(crate) struct RendererRegistry<T: Clone> {
+pub struct RendererRegistry<T: Clone> {
```
noituri, line 38: Just like `Renderers`, it shouldn't be public. I made it `pub(crate)` on my local branch and there were no compile errors so I'm assuming it's not needed

compositor_common/src/error.rs
```diff
@@ -44,6 +44,23 @@ impl Display for UnusedNodesError {
...
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum UnsatisfiedConstraintsError {
+    #[error(
+        "Invalid input pads specification for node \"{}\". {} requires {}. {}",
+        node_id,
+        identification_name,
+        input_count_constrain.required_inputs_message(),
+        Self::defined_inputs_message(defined_input_pads_count)
+    )]
+    InvalidInputsCount {
```
wkozyra95, line 56: consider sth like this

```rust
pub enum UnsatisfiedConstraintsError {
  #[error(transparent)]
  InvalidInputCount(#[from] InvalidInputCountError)
}

#[derive(*)]
struct InputCountValidationError {
        identification_name: Rc<str>,
        input_count_constrain: InputsCountConstraint,
        defined_input_pads_count: u32,
}

impl Display for InputCountValidationError {

}
```

and try to avoid functions that return some string that you will embed in error message

compositor_common/src/scene/node.rs
```diff
@@ -30,10 +35,35 @@ pub enum NodeParams {
...
-    pub fn validate(&self) -> Result<(), NodeSpecValidationError> {
+    pub fn validate_params(&self) -> Result<(), NodeSpecValidationError> {
         match &self.params {
-            NodeParams::Builtin { transformation, .. } => Ok(transformation.validate(self)?),
+            NodeParams::Builtin { transformation } => Ok(transformation.validate_params(self)?),
             _ => Ok(()),
         }
     }
+
+    pub fn identification_name(&self) -> Rc<str> {
```
wkozyra95, line 45: Also, I would avoid Rc<str> there, any time you call that function that string will be immediately used to create an error, so it's never copied. Also because of that you are making errors that don't use it not Send.

I think that w either should use:
- Arc<str> for stuff like ids that we are passing around
- Arc<str> for absolutely any string that is not &'static str

Anything other than that will be confusing when you are reading the code.

compositor_render/src/transformations/shader_executor.rs
WojciechBarczynski: Renamed to gpu_shader

compositor_common/src/scene/builtin_transformations.rs
```diff
@@ -110,16 +114,48 @@ impl FromStr for MirrorMode {
...
+        )]);
+    static ref MIRROR_IMAGE_CONSTRAINTS: NodeConstraints =
+        NodeConstraints(vec![Constraint::InputsCount(
+            InputsCountConstraint::Exact { fixed_count: 1 }
+        )]);
+    static ref CORNERS_ROUNDING_CONSTRAINTS: NodeConstraints =
+        NodeConstraints(vec![Constraint::InputsCount(
+            InputsCountConstraint::Exact { fixed_count: 1 }
+        )]);
+}
```
WojciechBarczynski, line 144: Can't use normal static, due to heap memory allocation required for `vec`. `[Constraints]` won't work, because `Constraints` isn't sized. `Box` also would allocate on `heap`, requiring runtime to instantiate.

compositor_common/src/scene/constraints.rs
```diff
@@ -0,0 +1,57 @@
...
+
+pub mod input_count;
+
+// TODO validate constraints aren't self-contradictory
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct NodeConstraints(pub(crate) Vec<Constraint>);
+
+impl NodeConstraints {
+    pub fn check(
```
WojciechBarczynski, line 20: `check` instead of `validate`, because `validate` can mean that constraints spec validation

compositor_common/src/scene/node.rs
```diff
@@ -30,10 +33,29 @@ pub enum NodeParams {
...
+        InputsCountConstraint::Exact { fixed_count: 0 }
+    )]);
+    static ref IMAGE_CONSTRAINTS: NodeConstraints = NodeConstraints(vec![Constraint::InputsCount(
+        InputsCountConstraint::Exact { fixed_count: 0 }
+    )]);
+}
+
+impl NodeParams {
+    pub fn text_constraints() -> &'static NodeConstraints {
+        &TEXT_CONSTRAINTS
```
wkozyra95, line 55: I would just put it here instead of using lazy_static

Also make it return Vec instead

compositor_render/src/gpu_shader.rs
wkozyra95, line 1: drop `gpu_`

wkozyra95 (approved):

PR #156, opened 2023-09-12
Implement `V8Function`

noituri:
This PR makes it possible to create and call JS functions. I intend to use this later on for retrieving canvas position


compositor_chromium/src/v8/context.rs
```diff
@@ -33,6 +36,24 @@ impl V8Context {
             Ok(V8Value::from_raw(get_global(ctx)))
         }
     }
+
+    pub fn eval(&self, code: &str) -> Result<V8Value, V8ContextError> {
```
wkozyra95, line 40: based on offline discussion this is not needed, so I would remove it.

wkozyra95 (approved):

WojciechBarczynski (approved):

compositor_chromium/src/v8/function.rs
```diff
@@ -0,0 +1,168 @@
...
+        Self(Validated::new(inner))
+    }
+
+    pub fn call(
+        &self,
+        args: &[V8Value],
+        ctx_entered: &V8ContextEntered,
+    ) -> Result<V8Value, V8FunctionError> {
+        self.inner_call(None, args, ctx_entered)
+    }
```
jerzywilczek, line 34: If F panics, the unwinding panic will cross into C++ land, which is a big no-no. 
read [this](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) and catch the panic.

jerzywilczek (changes requested):
:crab:

jerzywilczek (approved):
:rocket:

PR #160, opened 2023-09-14
Add common JS objects

noituri:
_I promise it's the last bindings-related PR (for now)_ 😅

As I was implementing frame embedding I noticed that I created multiple helper structs for some JS objects (global, document, element). I think they should be included as a part of `chromium_bindings`.


.gitignore
```diff
@@ -14,8 +14,9 @@ target/
 
 /target
 
-# VsCode 
+# IDE
 .vscode
```
wkozyra95, line 18: vscode is not IDE :D

wkozyra95 (approved):

WojciechBarczynski (approved):

jerzywilczek (approved):

PR #161, opened 2023-09-14
Extract the encoder from the rtp sender into the pipeline.

jerzywilczek:
I will add better error handling in a subsequent PR. I also plan to extract the decoders in a similar way later.

### DIFF
--- a/compositor_pipeline/src/pipeline.rs
+++ b/compositor_pipeline/src/pipeline.rs
@@ -6,7 +6,7 @@ use std::time::Duration;
 
 use compositor_common::error::ErrorStack;
 use compositor_common::renderer_spec::{RendererId, RendererSpec};
-use compositor_common::scene::{InputId, OutputId, SceneSpec};
+use compositor_common::scene::{InputId, OutputId, Resolution, SceneSpec};
 use compositor_common::{Frame, Framerate};
 use compositor_render::error::{
     InitRendererEngineError, RegisterRendererError, UnregisterRendererError,
@@ -17,6 +17,8 @@ use compositor_render::renderer::RendererOptions;
 use compositor_render::WebRendererOptions;
 use compositor_render::{error::UpdateSceneError, Renderer};
 use crossbeam_channel::unbounded;
+use crossbeam_channel::Sender;
+use ffmpeg_next::{Codec, Packet};
 use log::{error, warn};
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DurationMilliSeconds};
@@ -26,11 +28,17 @@ use crate::error::{
 };
 use crate::queue::Queue;
 
-pub trait PipelineOutput: Send + Sync + 'static {
-    type Opts;
+use self::encoder::{Encoder, EncoderSettings};
 
-    fn send_frame(&self, frame: Frame);
-    fn new(opts: Self::Opts) -> Self;
+pub mod encoder;
+
+pub trait PipelineOutputReceiver: 'static {
+    type Opts: Send + Sync + 'static;
+    type Identifier: Send + Sync + 'static;
+
+    fn send_packet(&mut self, packet: Packet);
+    fn new(opts: Self::Opts, codec: Codec) -> Self;
+    fn identifier(&self) -> Self::Identifier;
 }
 
 pub trait PipelineInput: Send + Sync + 'static {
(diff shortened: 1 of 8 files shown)


compositor_pipeline/src/pipeline.rs
```diff
@@ -26,11 +27,17 @@ use crate::error::{
 };
 use crate::queue::Queue;
 
-pub trait PipelineOutput: Send + Sync + 'static {
-    type Opts;
+use self::encoder::{Encoder, EncoderSettings};
+
+pub mod encoder;
+
+pub trait PipelineOutputReceiver: 'static {
```
noituri, line 34: I find this name slightly confusing. It's a `Receiver` but it has a `send` method and no `receive` method

compositor_pipeline/src/pipeline.rs
```diff
@@ -39,9 +46,56 @@ pub trait PipelineInput: Send + Sync + 'static {
...
 }
 
-pub struct Pipeline<Input: PipelineInput, Output: PipelineOutput> {
+pub struct OutputOptions<Receiver: PipelineOutputReceiver> {
+    pub receiver_options: Receiver::Opts,
+    pub encoder_settings: EncoderSettings,
+    pub resolution: Resolution,
+}
+
+pub struct PipelineOutput<Receiver: PipelineOutputReceiver> {
```
wkozyra95, line 55: This could be part of the Encoder implementation, if you want to keep them as a separate types I would:
- call this Encoder
- rename Encoder to LibavH264Encoder
- move this code to separate file (add as little as possible to the pipeline.rs)

compositor_pipeline/src/pipeline.rs
```diff
@@ -39,9 +46,56 @@ pub trait PipelineInput: Send + Sync + 'static {
...
+            sender: tx,
+            receiver_identifier: id_rx.recv().unwrap(),
+        })
+    }
+
+    fn send_frame(&self, frame: Frame) {
+        self.sender.send(frame).unwrap();
+    }
+
+    pub fn identifier(&self) -> &Receiver::Identifier {
```
wkozyra95, line 91: I don't think an identifier is a good name. Yes it's correct, but not really what is supporse to mean. I think that instead of using identifier you should use what is currently named PipelineOutputReceiver.

compositor_pipeline/src/pipeline.rs
```diff
@@ -221,22 +278,22 @@ impl<Input: PipelineInput, Output: PipelineOutput> Pipeline<Input, Output> {
 
     pub fn with_outputs<F, R>(&self, f: F) -> R
     where
-        F: Fn(OutputIterator<'_, Output>) -> R,
+        F: Fn(OutputIterator<'_, PipelineOutput<Receiver>>) -> R,
```
wkozyra95, line 281:
```suggestion
        F: Fn(OutputIterator<'_, Receiver>) -> R,
```

PipelineOutput should be a purely internal implementation, or if you want to expose both rtp and encode entities, maybe create sth on the fly that gives access to both.

compositor_pipeline/src/pipeline.rs
```diff
@@ -39,9 +46,56 @@ pub trait PipelineInput: Send + Sync + 'static {
...
-pub struct Pipeline<Input: PipelineInput, Output: PipelineOutput> {
+pub struct OutputOptions<Receiver: PipelineOutputReceiver> {
+    pub receiver_options: Receiver::Opts,
+    pub encoder_settings: EncoderSettings,
+    pub resolution: Resolution,
+}
+
+pub struct PipelineOutput<Receiver: PipelineOutputReceiver> {
+    sender: Sender<Frame>,
+    receiver_identifier: Receiver::Identifier,
```
wkozyra95, line 57:
```suggestion
    receiver: Receiver,
```

compositor_pipeline/src/pipeline/encoder.rs
```diff
@@ -0,0 +1,228 @@
...
+pub(crate) struct Encoder {
+    encoder: ffmpeg_next::codec::encoder::video::Encoder,
+    codec: Codec,
+    resolution: Resolution,
+}
+
+// TODO: errors in this file are atrocious
+
+impl Encoder {
+    pub fn new(settings: EncoderSettings, resolution: Resolution) -> Result<Self, String> {
```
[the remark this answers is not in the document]
jerzywilczek: errors are already handled in #162, I think I'd leave this as-is

compositor_pipeline/src/pipeline.rs
jerzywilczek: It seems very unintuitive to me that an `Encoder` would also take care of sending the encoded data out. I agree the naming is quite unfortunate here, but I don't like this suggestion either.

compositor_pipeline/src/pipeline.rs
jerzywilczek: I don't think I get what you mean

compositor_pipeline/src/pipeline.rs
jerzywilczek: This is proving to be quite difficult in our current architecture. I think I would have to remove the generic type parameter from `OutputIterator` and replace it with something concrete, bc doing anything else with the current `OutputIterator` is quite difficult.

jerzywilczek: do you think this would be fine?

compositor_pipeline/src/pipeline/encoder.rs
```diff
@@ -0,0 +1,272 @@
...
+pub struct Encoder<Output: PipelineOutput> {
+    sender: Sender<Frame>,
+    output: Output,
+}
+
+impl<Output: PipelineOutput> Encoder<Output> {
+    pub fn new(opts: OutputOptions<Output>) -> Result<Self, String> {
+        let mut encoder = LibavH264Encoder::new(opts.encoder_settings, opts.resolution)?;
+        let (frame_sender, frame_receiver) = crossbeam_channel::unbounded();
+        let (output_sender, output_receiver) = crossbeam_channel::bounded(0);
```
wkozyra95, line 242:
```suggestion
        // channel used to return information about Encoder/RtpSender initialization back to the API thread.
        let (output_sender, output_receiver) = crossbeam_channel::bounded(0);
```

wkozyra95 (approved):

WojciechBarczynski (approved):

PR #163, opened 2023-09-14
Fix multiple inputs with the same ID not working in web renderer

noituri:
## Issue 1
Currently web renderer identifies its inputs by their `NodeId`. For that reason this node won't work:
```json
{
    ...,
    "input_pads": [
        "input_1",
        "input_1" <-- duplicated node
    ]
}
```
In this PR I fixed this by identifying inputs by their `index` in `input_pads` vector.

## Issue 2
While working on the issue 1 I found another problem. What if there are two different nodes that use the same inputs:

```json
[
    {
        "node_id":  "test_node1",
        "input_pads": [
            "input_1"
        ]
    },
    {
        "node_id":  "test_node2",
        "input_pads": [
            "input_1" <-- uses the same input as test_node1
        ]
    }
]
```

This will break web renderer because both nodes will try to create and use shared memory at `/tmp/input_1`. I decided to fix it by namespacing every input so that shared memory for each node respectively is created at `/tmp/video_compositor/instance_<random_chars>/test_node1/0` and `/tmp/video_compositor/instance_<random_chars>/test_node2/0` _(`0` is the index of the input)_.

I was also thinking about creating 1 global shared memory per input. This approach would minimise the amount of copies but it would make synchronisation way more complicated.


compositor_render/src/transformations/web_renderer/chromium_sender_thread.rs
```diff
@@ -64,10 +61,23 @@ impl ChromiumSenderThread {
             return;
         };
 
+        let mut state = ThreadState {
+            browser,
+            shared_memories: HashMap::new(),
```
wkozyra95, line 66:
```suggestion
            shared_memory: Vec::new(),
```

wkozyra95 (commented):
You will still have an issue if someone starts 2 compositors with the same name. I recommond using paths like `/tmp/compositor-{random_value}/{node_id}-{input_index}-{random_value)` where the first random value is global for a single compositor instance

compositor_render/src/transformations/web_renderer/chromium_sender_thread.rs
noituri: I think it should be `HashMap<NodeId, Vec<..>`. Web renderer instance can be used by multiple nodes

compositor_render/src/transformations/web_renderer/chromium_sender_thread.rs
```diff
@@ -161,42 +173,30 @@ impl ChromiumSenderThread {
 
 struct ThreadState {
     browser: cef::Browser,
-    shared_memories: HashMap<(NodeId, usize), Shmem>,
+    shared_memory: HashMap<NodeId, Vec<SharedMemory>>,
```
noituri, line 176: I wasn't a fan of `shared_memories` name but the plural form helped distinguishing a collection of shared memory from a single shared memory. On the other hand I can't think of better name 😅

compositor_render/src/renderer.rs
```diff
@@ -66,15 +73,23 @@ pub struct RenderCtx<'a> {
...
 
 impl Renderer {
     pub fn new(opts: RendererOptions) -> Result<Self, InitRendererEngineError> {
         let wgpu_ctx = Arc::new(WgpuCtx::new()?);
+        let renderer_id = thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(30)
+            .map(char::from)
+            .collect::<String>()
+            .into();
```
WojciechBarczynski, line 89: [NIT] IMO it's more readable with separate function
```suggestion
        let renderer_id = rand_str()
```

WojciechBarczynski (approved):

PR #166, opened 2023-09-18
Improve shader validation errors

wkozyra95:
- Follow up to PR when we switched back to stack trace approach (remove some concatenated errors)
- Improve wording, fix some typos
- Improve serialization for errors (mostly focused on comparing types from shader with ShaderParam)
- Replace unwraps in validation code with safer alternative

`.unwrap_with("<unnamed>")` is just to make it shorter. Without this I would have to write
`.clone().unwrap_or_else(|| "<unnamed>".to_string())` everywhere.

### DIFF
--- a/compositor_render/src/gpu_shader/validation.rs
+++ b/compositor_render/src/gpu_shader/validation.rs
@@ -3,7 +3,7 @@ use naga::{ArraySize, ConstantInner, Handle, Module, ScalarKind, ShaderStage, Ty
 
 use super::{
     error::{
-        ConstArraySizeEvalError, ParametersValidationError, ShaderGlobalVariableExt,
+        BindingExt, ConstArraySizeEvalError, ParametersValidationError, ShaderGlobalVariableExt,
         ShaderValidationError, TypeEquivalenceError,
     },
     USER_DEFINED_BUFFER_BINDING, USER_DEFINED_BUFFER_GROUP,
@@ -35,7 +35,7 @@ fn validate_globals(
             |err| {
                 ShaderValidationError::GlobalBadType(
                     err,
-                    global.name.clone().unwrap_or("value".to_string()),
+                    global_in_shader.name.clone().unwrap_with("<unknown>"),
                 )
             },
         )?;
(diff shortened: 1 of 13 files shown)


compositor_render/src/gpu_shader/validation.rs
```diff
@@ -87,10 +87,7 @@ fn validate_vertex_input(
         .find(|(_, ty)| ty.name == vertex_input_type.name)
         .ok_or_else(|| {
             ShaderValidationError::VertexShaderBadInputTypeName(
-                vertex_input_type
-                    .name
-                    .clone()
-                    .unwrap_or("<unknown>".to_string()),
+                vertex_input_type.name.clone().unwrap_with("<unknown>"),
```
noituri, line 90: `.clone()` is not necessary because `.unwrap_with()` already clones `self`

compositor_render/src/gpu_shader/validation.rs
```diff
@@ -109,11 +106,11 @@ fn validate_type_equivalent(
...
 
-    if expected_type.name != provided_type.name {
-        return Err(TypeEquivalenceError::TypeNameMismatch(
-            expected_type.name.clone(),
-            provided_type.name.clone(),
-        ));
+    if expected_type.name != provided_type.name && expected_type.name.is_some() {
+        return Err(TypeEquivalenceError::TypeNameMismatch {
+            expected: expected_type.name.clone().unwrap_with("<unknown>"),
+            actual: provided_type.name.clone().unwrap_with("<unknown>"),
```
noituri, line 112: unnecessary `.clone()`

compositor_render/src/gpu_shader/validation.rs
```diff
@@ -685,5 +703,45 @@ impl ScalarKindExt for naga::ScalarKind {
     }
 }
 
+trait ShaderParamExt {
+    fn to_string(&self) -> String;
+}
+
+impl ShaderParamExt for ShaderParam {
```
noituri, line 710: Unless you plan on changing the `to_string`'s signature I would just use rust's `ToString` trait

wkozyra95: one of those types needs to be defined in this crate, neither ToString nor ShaderParam is defined here

noituri (approved):
👌

WojciechBarczynski (approved):

PR #178, opened 2023-09-25
Release pipeline

wkozyra95:
Release process for Linux binaries + docs

### DIFF
--- a/examples/web_renderer.rs
+++ b/examples/web_renderer.rs
@@ -1,5 +1,5 @@
 use anyhow::Result;
-use compositor_chromium::cef;
+use compositor_chromium::cef::bundle_for_development;
 use compositor_common::{scene::Resolution, Framerate};
 use log::{error, info};
 use serde_json::json;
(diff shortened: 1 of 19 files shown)


wkozyra95:
Merging because I need to verify some things on master, I'll address any comments in follow up.

examples/web_renderer.rs
```diff
@@ -38,8 +38,11 @@ fn main() {
         .unwrap()
         .join("..");
 
-    if cef::bundle_app(target_path).is_err() {
-        panic!("Build process helper first: cargo build --bin process_helper");
+    if let Err(err) = bundle_for_development(target_path) {
+        panic!(
+            "Build process helper first: cargo build --bin process_helper. {:?}",
+            err
```
WojciechBarczynski, line 44: Maybe add note about `-r`

src/bin/package_for_release/bundle_macos.rs
```diff
@@ -0,0 +1,79 @@
...
+            "-czvf",
+            output_name,
+            "video_compositor.app",
+        ])
+        .spawn()?
+        .wait()?
+        .code();
+    if exit_code != Some(0) {
+        return Err(anyhow!("Command tar failed with exit code {:?}", exit_code));
+    }
```
WojciechBarczynski, line 76: The exit code itself isn't really helpful. Can we collect the result and add it here? (like the first element returned from: https://hexdocs.pm/elixir/System.html#cmd/3)

WojciechBarczynski (commented):
LGTM

src/bin/package_for_release/bundle_macos.rs
wkozyra95: stdout and stderr is already printed to the terminal.

WojciechBarczynski: Ok

PR #179, opened 2023-09-25
Add alternative web frame embedding

noituri:
This PR adds an alternative way of embedding frames.  The user has now access to 3 possible embedding methods:
- `WebEmbeddingMethod::ChromiumEmbedding` - render inputs via HTML canvas
- `WebEmbeddingMethod::EmbedOnTop` - render inputs on top of the website (Default)
- `WebEmbeddingMethod::EmbedBelow` - render inputs below the website 

Rendering inputs via HTML canvas offers a lot of customisation as the inputs are basically parts of the website. The problem with this approach is performance because every input has to be copied from GPU to shared memory.

I added 2 additional embedding methods which render inputs by simply placing them in the correct positions via shader which is magnitudes faster but doesn't offer as much flexibility as the canvas.

### To Do:
- [ ] Write docs


compositor_render/src/transformations/shader_params.rs
```diff
@@ -18,15 +18,15 @@ impl ParamsBuffer {
         let buffer = wgpu_ctx
             .device
             .create_buffer_init(&wgpu::util::BufferInitDescriptor {
-                label: Some("builtin node params buffer"),
+                label: Some("shader params buffer"),
```
WojciechBarczynski, line 21: This is misleading. I know, that you refer to GPU shader, rather than to shader "transformation", but users will see in logs that suggest that they're using `shader` "transformation".

compositor_render/src/transformations/shader_params.rs
```diff
@@ -2,13 +2,13 @@ use wgpu::util::DeviceExt;
...
-pub(super) struct ParamsBuffer {
-    bind_group: wgpu::BindGroup,
+pub(super) struct ShaderParamsBuffer {
+    pub bind_group: wgpu::BindGroup,
     buffer: wgpu::Buffer,
     content: bytes::Bytes,
 }
 
-impl ParamsBuffer {
+impl ShaderParamsBuffer {
```
WojciechBarczynski, line 11: Misleading, might be interpreted as params buffer for `shader` transformation.

compositor_render/src/transformations/shader_params.rs
WojciechBarczynski, line 1: This should be in `gpu_shader.rs` or in `gpu_shader/params.rs`.

compositor_render/src/transformations/web_renderer.rs
```diff
@@ -43,38 +56,47 @@ impl Default for WebRendererOptions {
...
-    _bgra_bind_group_layout: wgpu::BindGroupLayout,
-    bgra_bind_group: wgpu::BindGroup,
-    bgra_to_rgba: BGRAToRGBAConverter,
+    spec: WebRendererSpec,
+    frame_data: FrameData,
+    source_transforms: SourceTransforms,
+    embedding_helper: EmbeddingHelper,
+
+    website_texture: BGRATexture,
+    shader: GpuShader,
```
WojciechBarczynski, line 65: Use a name that suggests what this shader is responsible for.

compositor_render/src/transformations/web_renderer.rs
```diff
@@ -85,25 +107,69 @@ impl WebRenderer {
...
+            }
+            WebEmbeddingMethod::EmbedBelow => {
+                [source_textures, vec![Some(self.website_texture.texture())]].concat()
+            }
+            WebEmbeddingMethod::ChromiumEmbedding => vec![Some(self.website_texture.texture())],
+        };
+
+        let textures_info = match self.spec.embedding_method {
+            WebEmbeddingMethod::EmbedOnTop => {
+                [TextureInfo::website(), source_transforms].concat().into()
```
WojciechBarczynski, line 153: [NIT]
`.append()` doesn't work?

compositor_render/src/transformations/web_renderer/render_website.wgsl
```diff
@@ -0,0 +1,68 @@
...
+struct CommonShaderParameters {
+    time: f32,
+    texture_count: u32,
+    output_resolution: vec2<u32>,
+}
+
+struct TextureInfo {
+     is_website_texture: i32,
+     transformation_matrix: mat4x4<f32>,
+}
```
WojciechBarczynski, line 22: How does memory alignment work here? Have you tested it?

noituri: I've had some troubles with getting it right but from what I found out each field in this struct needs 16 byte alignment. `i32` is 4 bytes so I had to add 12 byte padding between `is_website_texture` & `transformation_matrix`

WojciechBarczynski (approved):

compositor_render/src/gpu_shader.rs
```diff
@@ -182,3 +178,58 @@ impl GpuShader {
         validate_params(params, ty, &self.shader)
     }
 }
+
+pub(super) struct ParamsBuffer {
```
wkozyra95, line 182: extract that to separate file

compositor_render/src/gpu_shader.rs
```diff
@@ -114,7 +114,7 @@ impl GpuShader {
     pub fn render(
         &self,
         params: &wgpu::BindGroup,
-        sources: &[(&NodeId, &NodeTexture)],
+        textures: &[Option<&Texture>],
```
wkozyra95, line 117: This is a very large change, and I think it's done for the wrong reasons. You are changing it just because it does not fit the API you want for web renderer and this requires addinging workarounds to all other components. This is not a good tradeoff.

I see 2 approaches:
- Ideally, any changes you are doing here should not require any logic changes in other components. You can achieve that by implementing other render method and reusing some code.
- You could not reuse this Shader struct
- You could change this API to accept Option<&Texture>, but calling `sources_to_textures` in every component is not a good solution

compositor_common/src/renderer_spec.rs
```diff
@@ -50,12 +50,28 @@ pub struct WebRendererSpec {
...
+pub enum WebEmbeddingMethod {
+    /// Send frames to chromium directly and render it on canvas
+    ChromiumEmbedding,
+
+    /// Render sources on top of the rendered website
+    EmbedOnTop,
+
+    /// Render sources below the website.
+    /// The website's background has to be transparent
+    EmbedBelow,
```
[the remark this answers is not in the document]
noituri: I like the 1st option but on API side it looks slightly weird:
```json
"embedding_method": {
    "native_embedding": {
        "render_over_content": false
     }
}
```

```json
"embedding_method": "chromium_embedding"
```

In my opinion, `OverContent` and `UnderContent` are bit more descriptive compared to `Background` and `Foreground`

compositor_render/src/gpu_shader.rs
```diff
@@ -121,7 +136,7 @@ impl GpuShader {
     ) {
         let ctx = &self.wgpu_ctx;
 
-        // TODO: sources need to be ordered
+        // TODO: sources/textures need to be ordered
```
noituri, line 139: Is this TODO still needed? Aren't sources already ordered?
cc: @WojciechBarczynski

compositor_render/src/transformations/web_renderer.rs
```diff
@@ -85,25 +107,77 @@ impl WebRenderer {
...
+        &'a self,
+        sources: &'a [(&NodeId, &NodeTexture)],
+    ) -> (Vec<Option<&'a Texture>>, Bytes) {
+        let mut source_textures = sources_to_textures(sources);
+        let source_transforms = TextureInfo::sources(&self.source_transforms.lock().unwrap());
+
+        let mut textures = Vec::new();
+        match self.spec.embedding_method {
+            WebEmbeddingMethod::NativeEmbeddingOverContent => {
+                textures.push(Some(self.website_texture.texture()));
```
wkozyra95, line 144: it's not great, but I understand why it's this way, lets maybe land it this way, but add an issue to rework this in the future

compositor_render/src/renderer/texture/utils.rs
```diff
@@ -10,3 +13,15 @@ pub fn texture_size_to_resolution(size: &wgpu::Extent3d) -> Resolution {
         height: size.height as usize,
     }
 }
+
+pub fn sources_to_textures<'a>(sources: &'a [(&NodeId, &NodeTexture)]) -> Vec<Option<&'a Texture>> {
```
wkozyra95, line 17: inline that code, it's only used in 2 places
this is sth we want to avoid and eventually refactor out

wkozyra95 (approved):
Hold on with merging this until next week

WojciechBarczynski:
@wkozyra95 @noituri Shouldn't this be merged?

PR #181, opened 2023-09-26
Implement user web embedding API

noituri:
Currently I have 3 ideas of the API:
```html
<canvas id="input_0"></canvas>
<canvas id="input_1"></canvas>
```
```js
// Option 1
// Insert frames automatically in input_<index>
// Pros: user doesn't have to configure anything
// Cons: it does not allow any extra configuration from the JS side

// Option 2
register_inputs(
    "input_0",
    { id: "input_1", overrideCanvasSize: false }
);

// Option 3
// Same as option 2 but more explicit. I find it ugly though
register_inputs({
    0: "input_0",
    1: "input_ 1"
});
```

Update:
I went with option 2

### DIFF
--- a/examples/web_renderer.html
+++ b/examples/web_renderer.html
@@ -39,35 +39,13 @@
 
 <body>
 <h2>Frame embedding:</h2>
-<canvas id="input_0"></canvas>
+<canvas id="input_1"></canvas>
 </body>
 
 <script>
-    // TODO: Figure out embedding API
-    // NOTE TO REVIEWERS: The section below is not part of this PR
-    async function run(sourceID, interval) {
-        const sourceCanvas = document.getElementById(sourceID);
-        const ctx = sourceCanvas.getContext("2d");
-
-        setInterval(() => {
-            const data = window[`${sourceID}_data`];
-            const width = window[`${sourceID}_width`];
-            const height = window[`${sourceID}_height`];
-            if (data == undefined || width == undefined || height == undefined) {
-                return;
-            }
-            if (width == 1 || height == 1) {
-                return;
-            }
-
-            sourceCanvas.width = width;
-            sourceCanvas.height = height;
-            const imageView = new ImageData(new Uint8ClampedArray(data), width, height);
-            ctx.putImageData(imageView, 0, 0);
-        }, 16);
-    }
-
-    run("input_0");
+    register_inputs(
+        "input_1"
+    );
 </script>
 
 </html>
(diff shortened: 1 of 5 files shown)


src/bin/process_helper/handler.rs
```diff
@@ -1,31 +1,51 @@
...
+    fn on_context_created(
+        &mut self,
+        _browser: &cef::Browser,
+        _frame: &cef::Frame,
+        context: &cef::V8Context,
+    ) {
+        let mut global = context.global().unwrap();
+        let ctx_entered = context.enter().unwrap();
+
+        context.eval(include_str!("render_frame.js")).unwrap();
```
noituri, line 25: I wanted to write everything in rust but I had some problems with rewriting `render_frame.js` to rust so I decided to simply eval it

WojciechBarczynski (approved):

src/bin/process_helper/state.rs
```diff
@@ -1,45 +1,93 @@
...
+use std::sync::Arc;
 use std::{collections::HashMap, sync::Mutex};
 
 use compositor_chromium::cef;
-use shared_memory::Shmem;
+use shared_memory::{Shmem, ShmemConf};
 
 pub struct State {
-    sources: Mutex<HashMap<PathBuf, Source>>,
+    input_mappings: Mutex<Vec<Arc<str>>>,
```
wkozyra95, line 10: consider extracting this Mutext to Mutex<State>

src/bin/process_helper/state.rs
```diff
@@ -1,45 +1,93 @@
...
+                data_ptr,
+                (4 * frame_info.width * frame_info.height) as usize,
+                ctx_entered,
+            )
+        }
+        .into();
+        let width = cef::V8Uint::new(frame_info.width).into();
+        let height = cef::V8Uint::new(frame_info.height).into();
+
+        #[allow(clippy::arc_with_non_send_sync)]
```
wkozyra95, line 49: add comment that clippy suggest using Rc

wkozyra95 (approved):

examples/web_renderer.html
```diff
@@ -39,35 +39,13 @@
...
-
-            sourceCanvas.width = width;
-            sourceCanvas.height = height;
-            const imageView = new ImageData(new Uint8ClampedArray(data), width, height);
-            ctx.putImageData(imageView, 0, 0);
-        }, 16);
-    }
-
-    run("input_0");
+    register_inputs(
```
wkozyra95, line 46: will this function be always available before first render?

noituri: afaik, the function should be available. However, I'm basing this on bits of information I found scattered in CEF source code

PR #187, opened 2023-10-03
Fix docker example

wkozyra95:
Fix docker example

I also tested a web renderer, but it still does not work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,7 @@ FROM ubuntu:mantic-20230712
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 ARG USERNAME=compositor
-ARG NODE_VERSION=18.12.1
-ARG NVM_VERSION=0.39.1
-ARG RUST_VERSION=1.70
+ARG RUST_VERSION=1.72
 
 ENV DEBIAN_FRONTEND=noninteractive
 ENV WEB_RENDERER_PATH=/home/$USERNAME/project/web_renderer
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -15,7 +13,7 @@ RUN apt-get update -y -qq && \
     build-essential curl pkg-config libssl-dev libclang-dev git sudo \
     libnss3 libatk1.0-0 libatk-bridge2.0-0 libgdk-pixbuf2.0-0 libgtk-3-0 \
     xvfb \
-    libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers \
+    libegl1-mesa-dev libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers \
```
wkozyra95, line 16: even though we have a hardcoded image it broke for some reason

docker/entrypoint.sh
```diff
@@ -1,13 +1,13 @@
...
 source ~/.cargo/env
 
 sudo dbus-daemon --system
 sudo Xvfb :99 -screen 0 640x480x8 -nolisten tcp &
 
-sleep 1
+sleep 2
 
-video_compositor
+exec "$MEMBRANE_VIDEO_COMPOSITOR_MAIN_EXECUTABLE_PATH"
```
wkozyra95, line 13: exec to make sure signals sent to the parent process are passed to the compositor process

examples/docker.rs
```diff
@@ -66,8 +66,6 @@ fn build_and_start_docker(skip_build: bool) -> Result<()> {
             "-it",
             "--device",
             "/dev/dri", // expose gpu to container
-            "--cap-add",
```
wkozyra95, line 69: we don't need that after replacing electron

Dockerfile
```diff
@@ -3,9 +3,7 @@ FROM ubuntu:mantic-20230712
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 ARG USERNAME=compositor
-ARG NODE_VERSION=18.12.1
-ARG NVM_VERSION=0.39.1
-ARG RUST_VERSION=1.70
```
wkozyra95, line 8: we don't need that after replacing electron

WojciechBarczynski (approved):

PR #190, opened 2023-10-04
Check if pipeline already started

wkozyra95:
starting the pipeline twice gives a very cryptic error becaus muxer receives non monotonic dts from encoder


jerzywilczek (approved):
:fire:

WojciechBarczynski (commented):
Shoundn't start request return error in such case?

PR #191, opened 2023-10-04
Validate resolutions are divisible by 2

jerzywilczek:


compositor_pipeline/src/error.rs
```diff
@@ -20,6 +20,9 @@ pub enum RegisterOutputError {
 
     #[error("Encoder error while registering output stream for stream \"{0}\".")]
     EncoderError(OutputId, #[source] OutputInitError),
+
+    #[error("Failed to register input stream. Stream \"{0}\" has an odd resolution.")]
```
wkozyra95, line 24: What about the output stream?
Don't use odd here, it sounds weird, maybe "resolution is not divisible by 2" or
```
Failed to register input stream \"{0}\". Resolution in each dimension has to be divisible by 2."
```

compositor_pipeline/src/error.rs
```diff
@@ -90,6 +93,7 @@ impl From<&RegisterInputError> for PipelineErrorInfo {
 
 const OUTPUT_STREAM_ALREADY_REGISTERED: &str = "OUTPUT_STREAM_ALREADY_REGISTERED";
 const ENCODER_ERROR: &str = "ENCODER_ERROR";
+const ODD_RESOLUTION: &str = "ODD_RESOLUTION";
```
wkozyra95, line 96: maybe sth more generic e.g. UNSUPPORTED_RESOLUTION ?

wkozyra95 (approved):

compositor_pipeline/src/error.rs
WojciechBarczynski: This should be `output_stream`. We can't return error in case of input stream, since resolution of input stream is dynamic

PR #194, opened 2023-10-06
New type layer for API

wkozyra95:
I tried to avoid any changes in this PR, but I introduce some to either test new API or avoid unnecessary work in this PR that would be rewerted in the next one.

Changes:
- builtin transformations are now defined using `type` field e.g. `"type": "builtin:transofrm_to_resolution"`
- Coord values can now be either number or sting (if it's a number we assume the value is in pixels)


src/types/node.rs
```diff
@@ -0,0 +1,240 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+#[serde(tag = "strategy", rename_all = "snake_case", deny_unknown_fields)]
+pub enum TransformToResolution {
+    /// Rescales input in both axis to match output resolution
+    Stretch { resolution: Resolution },
+    /// Scales input preserving aspect ratio and cuts equal parts
+    /// from both sides in "sticking out" dimension
+    Fill { resolution: Resolution },
+    /// Scales input preserving aspect ratio and
+    /// fill the rest of the texture with the provided color]
```
noituri, line 180:
```suggestion
    /// Scales input preserving aspect ratio and
    /// fills the rest of the texture with the provided color
```

src/types/from_node.rs
```diff
@@ -0,0 +1,357 @@
...
+            padding: layout.padding.unwrap_or(0),
+            horizontal_alignment: layout
+                .horizontal_alignment
+                .unwrap_or(HorizontalAlign::Center)
+                .into(),
+            vertical_alignment: layout
+                .vertical_alignment
+                .unwrap_or(VerticalAlign::Center)
+                .into(),
+        });
```
WojciechBarczynski, line 329: I think serde defaults are more readable than manual default implementations on parsing. Using serde, default values are already visible in the type definition, but with this approach, it's unclear what they are looking only at types.

src/types/node.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub mode: Option<MirrorMode>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub enum MirrorMode {
+    #[serde(rename = "horizontal")]
+    Horizontal,
+    #[serde(rename = "vertical")]
+    Vertical,
+    #[serde(rename = "horizontal-vertical")]
```
WojciechBarczynski, line 230: I think using multiple different cases in API only makes it more difficult to use. Consider using a snake case.

src/types/from_node.rs
wkozyra95: I would agree if there was a better way to do that more universally, but if we are using literals in some places, functions in other and in some cases it will need to be done inside `from()` function then I prefer to have that consistent at least.

>  Using serde, default values are already visible in the type definition

No, they are next to type definition (and if you are using functions that return default not even that), but they are not part of type definition. 

Also all those serde annotations are really affecting readability and size. It's especially important that top level API like e.g. entire Node structure would be visible without jumping between bunch of files. e.g. currently to see how to create node e.g. fixed_position_layout you need to check at least 4 files in compositor_common package to understand it's structure.

src/types/node.rs
wkozyra95: As mentioned in the description, I'm trying to avoid any api changes

> I think using multiple different cases in API only makes it more difficult to use. 

Note here that this value is not a key, this is a value, so we are not mixing a case here in my opinion. (The same situation as `built-in` that was removed in ths PR)

src/types/node.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub struct ShaderParamStructField {
+    pub field_name: String,
+    #[serde(flatten)]
+    pub value: ShaderParam,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
```
wkozyra95, line 81: One additional approach I considered was to e.g. wrap all related types into mod. In this case it would be Text,TextStyle,TextWrapMode,TextWeight, TextDimensions

- It would group related types together (from_node.rs is already doing it, but inside wrap function)
- We could consider using serde defaults in that case because all those default_stomehting functions would be encapsulted inside the module
- If we use serde defaults that info would be included inside json schema

src/types/from_node.rs
wkozyra95: There is alternative approach I considered that would work with your suggestion, but I'm still slightly leaning towards current approach.

What do you think about https://github.com/membraneframework/video_compositor/pull/194/files#r1360454689 ?

noituri (approved):

src/types/from_node.rs
WojciechBarczynski: I get your point. It's hard for me to tell whether wrapping everything in types and implementing defaults on them is better than implementing it in `from` functions. I leave it up to you.

src/api.rs
WojciechBarczynski, line 1: Should this be in `src`? I think it's quite confusing that we have:
- `src` that defines API, RTP stuff, building stuff...
- `compositor_common` - types used by `compositor_renderer`, to which modules in `src` translate API received types
- `compositor_renderer` - actual rendering implementation

Doesn't it make sense to define a separate subcrate/folder for API? Modules like `src/error` are confusing. Are those API errors, RTP errors, or building errors? I can't tell just looking at project structure.

src/types/from_node.rs
jerzywilczek: I think the current approach is better

wkozyra95: > I get your point. It's hard for me to tell whether wrapping everything in types and implementing defaults on them is better than implementing it in from functions. I leave it up to you.

That is not what I suggested, an alternative could be using `mod` blocks that wrap multiple types and using serde default mechanism not rust  Default trait

src/types/from_node.rs
```diff
@@ -0,0 +1,357 @@
...
+                TextureLayout {
+                    left: Some(_),
+                    right: Some(_),
+                    ..
+                } => return Err(anyhow!("Fields \"left\" and \"right\" are mutually exclusive, you can only specify one in texture layout in \"fixed_position_layout\" transformation.")),
+                _ => (),
+            };
+
+        Ok(Self {
+            top: value.top.map_or(Ok(None), |v| v.try_into().map(Some))?,
```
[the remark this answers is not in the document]
wkozyra95: Not sure I understand the question?
We have a code that can both return an error and operates on optional value, so retruned value has to be a combination of Result and Option
I prefer to have Result<Option<x>> because I can use `?` to handle it.

src/types/into_node.rs
```diff
@@ -0,0 +1,267 @@
...
+            font_size: spec.font_size,
+            dimensions: spec.dimensions.into(),
+            line_height: spec.line_height,
+            color_rgba: Some(spec.color_rgba.into()),
+            background_color_rgba: Some(spec.background_color_rgba.into()),
+            font_family: Some(spec.font_family),
+            style: Some(style),
+            align: Some(spec.align.into()),
+            wrap: Some(wrap),
+            weight: Some(weight),
```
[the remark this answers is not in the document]
wkozyra95: no idea what you mean by that, please clarify

src/types/node.rs
```diff
@@ -0,0 +1,238 @@
...
+#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
+pub enum NodeParams {
+    WebRenderer(WebRenderer),
+    Shader(Shader),
+    Image(Image),
+    Text(Text),
+    Transition(Transition),
+    #[serde(rename = "builtin:transform_to_resolution")]
+    TransformToResolution(TransformToResolution),
```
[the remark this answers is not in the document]
wkozyra95: I did not intend to change that, if you suggest a change please explain what you would like to switch to. (I still plan to change internally, but I'm not sure about the API yet)

src/types/from_node.rs
wkozyra95: > I think wrapping in types is slightly better since we still have the possibility to automatically generate types/schema for API.

Not sure what you mean by that

src/api.rs
wkozyra95: > Doesn't it make sense to define a separate subcrate/folder for API?

If anything, I would consider removing the common crate. Those types define HTTP API, this is the only place where those types can be used.

Even if we decide to write Rust SDK at some point, I'm not sure if we should share those types. (don't really want to go into detail here, we can talk offline about it if you want)

> Modules like src/error are confusing. Are those API errors, RTP errors, or building errors?

Not sure I see why?

API errors - yes, also it's a short file that includes only a type called ApiError
RTP errors - if we add sth to handle RTP errors it could be there
building errors - what does it even mean?

> I can't tell just looking at project structure.

- This is not a library.
- when you look at the project structure and see error.rs I would expect that you would assume that it contains top-level types around error handling.

src/types/node.rs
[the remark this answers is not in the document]
WojciechBarczynski: I mean using `FitToResolution`, `FillToResolution`, and `StretchToResolution` instead of a single `TransformToResolution`, as I suggested on Thursday.

src/types/into_node.rs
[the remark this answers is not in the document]
WojciechBarczynski: In other places, we define defaults on types translation. I don't see the reason to do this differently in this case.

src/types/from_node.rs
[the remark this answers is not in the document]
WojciechBarczynski: Ok, now I get it. I missed `?`. 

When we have type separation, I would change `builtin_transformations::TextureLayout`, such that the type system provides the guarantee, that only a single top/bottom and left/right are defined. I think we should make invalid types unrepresentable on the renderer/common level.

src/types/from_node.rs
WojciechBarczynski: Why can't we just use `struct FieldName(type)` and `impl Default for FieldName` on the API level? 
We could avoid implementing `try_into` everywhere (only when the field default value depends on other fields values we can't do that). 
I think some tools could automatically generate schema with default values for us.

src/types/from_node.rs
```diff
@@ -0,0 +1,357 @@
...
+            NodeParams::Shader(node) => node.into(),
+            NodeParams::Image(node) => node.into(),
+            NodeParams::Text(node) => node.try_into()?,
+            NodeParams::Transition(node) => node.try_into()?,
+            NodeParams::TransformToResolution(node) => scene::NodeParams::Builtin {
+                transformation: node.try_into()?,
+            },
+            NodeParams::FixedPositionLayout(node) => scene::NodeParams::Builtin {
+                transformation: node.try_into()?,
+            },
```
WojciechBarczynski, line 35: With a new API layer, `transformation` is useless. It was only used for the correct JSON type, but we can use `scene::NodeParams::Builtin(BuiltinSpec)` now

src/types/node.rs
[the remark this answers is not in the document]
wkozyra95: ok we can discuss it, but either way it won't be done in this PR

src/types/into_node.rs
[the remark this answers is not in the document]
WojciechBarczynski: Ohh, right :) I did :P Nevermind

src/types/from_node.rs
wkozyra95: yes it is, but I don't want to change anything in API (both http and renderer) in this PR if I don't have to

src/types/from_node.rs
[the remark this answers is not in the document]
wkozyra95: yes I plan to do that, but not in this PR

src/types/from_node.rs
wkozyra95: > Why can't we just use struct FieldName(type) and impl Default for FieldName on the API level?

We can, but why do you think it's better?

> We could avoid implementing try_into everywhere (only when the field default value depends on other fields values we can't do that).

This would not help with try_into, defaults are not related in any way to when we use From vs TryFrom trait

> I think some tools could automatically generate schema with default values for us.

Did you see my comment linked above https://github.com/membraneframework/video_compositor/pull/194/files#r1360454689 ?

WojciechBarczynski (approved):

jerzywilczek (approved):

PR #201, opened 2023-10-10
Validate builtin spec inside transition

wkozyra95:
specs defined inside transition were not validated


WojciechBarczynski (commented):
LGTM

PR #205, opened 2023-10-10
Add shared memory resize sync

noituri:


compositor_render/src/transformations/web_renderer/chromium_sender_thread.rs
```diff
@@ -87,28 +98,31 @@ impl ChromiumSenderThread {
...
+            let shared_memory = state.shared_memory(&node_id, source_idx)?;
+            let (width, height) = if shared_memory.is_accessible() {
+                let Resolution { width, height } = resolution.unwrap_or_else(|| Resolution {
+                    width: 0,
+                    height: 0,
+                });
+                (width, height)
+            } else {
+                (0, 0)
+            };
```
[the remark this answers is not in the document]
noituri: When shared memory is not accessible (it's currently being resized), I want width and height to be always `0`. The process helper does not render zero-sized frames so the shared memory is not accessed.

WojciechBarczynski (approved):

compositor_render/src/transformations/web_renderer/shared_memory.rs
```diff
@@ -92,6 +133,28 @@ impl SharedMemory {
     }
 }
 
+enum SharedMemoryState {
+    Unloaded,
```
wkozyra95, line 137: No idea what any of those state mean.

compositor_render/src/transformations/web_renderer/shared_memory.rs
```diff
@@ -21,53 +23,92 @@ impl SharedMemory {
...
+            SharedMemoryState::Loaded(shared_memory)
+            | SharedMemoryState::PendingResize { shared_memory, .. } => shared_memory,
+            SharedMemoryState::Unloaded => return Err(SharedMemoryError::SharedMemoryInaccessible),
+        };
+        self.state = SharedMemoryState::PendingResize {
+            shared_memory,
+            new_size: size,
+        };
+
+        let mut msg = cef::ProcessMessage::new(UNEMBED_SOURCE_MESSAGE);
```
wkozyra95, line 79: why is function on an shared memory object sending cef messages? That logic should not be mixed together.

compositor_render/src/transformations/web_renderer/shared_memory.rs
```diff
@@ -21,53 +23,92 @@ impl SharedMemory {
...
+        msg.write_string(0, self.to_path_string());
+        msg.write_string(1, self.node_id.to_string());
+        msg.write_int(2, self.source_idx as i32);
+        frame.send_process_message(cef::ProcessId::Renderer, msg)?;
+
+        Ok(())
+    }
 
+    pub fn resolve_pending_resize(&mut self) -> Result<(), SharedMemoryError> {
```
wkozyra95, line 88: resolve is very confusing here, if I understand correctly this function is "resizing" shared memory  if current state is already in pending resize state. Is that correct?

- You are mixing here logic on how to resize with when to resize.
- Name `resolve` is not a correct word here `finalize_pending_resize` or `execute_pending_resize` would be more appropriate.
- I might be missing sth important here, but is the fact that there is a resize pending sth that should even be part of the shared memory abstraction?

PR #206, opened 2023-10-11
Fix fitted column text

WojciechBarczynski:


compositor_render/src/transformations/text_renderer.rs
```diff
@@ -193,48 +194,52 @@ impl TextRendererNode {
...
-                    height: max_height as usize,
-                },
-                &text_params,
-            ),
-            TextDimensions::FittedColumn { width, max_height } => Self::layout_fitted(
-                buffer,
-                font_system,
+            } => {
+                buffer.set_size(font_system, max_width as f32, max_height as f32);
+
```
wkozyra95, line 207: empty newlines can improve visibility to grup stuff, but if you add it between each line it has only negative effect

wkozyra95 (approved):

PR #212, opened 2023-10-11
Fix letters cutting in text rendering

### LINKED ISSUE #211: Fix cutting letters like "p"

WojciechBarczynski:
Closes #211

### DIFF
--- a/compositor_render/src/transformations/text_renderer.rs
+++ b/compositor_render/src/transformations/text_renderer.rs
@@ -255,7 +255,8 @@ impl TextRendererNode {
             }
         }
 
-        let height = (lines_count as f32 * line_height.ceil() + font_size / 4.0) as usize;
+        let last_line_padding = font_size / 5.0;
+        let height = (lines_count as f32 * line_height.ceil() + last_line_padding) as usize;
         Resolution { width, height }
     }
 }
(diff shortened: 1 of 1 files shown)


compositor_render/src/transformations/text_renderer.rs
```diff
@@ -242,15 +250,12 @@ impl TextRendererNode {
...
                 }
             }
         }
 
-        let height = lines_count * line_height.ceil() as u32;
-        Resolution {
-            width: width as usize,
-            height: height as usize,
-        }
+        let height = (lines_count as f32 * line_height.ceil() + font_size / 4.0) as usize;
```
wkozyra95, line 258: 1/4 font_size minimal value? Seems like a lot

Also maybe extract that to separate variable that have a meaningful name e.g
`let last_line_padding = font_size / 4.0`

wkozyra95 (approved):

compositor_render/src/transformations/text_renderer.rs
WojciechBarczynski: With 1/6 `j` is cut almost always, with 1/5 I didn't find the configuration that cuts `j` or `p`, but I left 1/4 as a margin of error. I'll use 1/5 and change it if we encounter problems in the future.

PR #214, opened 2023-10-12
Descriptive errors on failed server startup

WojciechBarczynski:


compositor_chromium/examples/simple.rs
```diff
@@ -87,7 +87,7 @@ fn main() {
         .join("..");
 
     if cef::bundle_for_development(target_path).is_err() {
-        panic!("Build process helper first: cargo build --bin process_helper");
+        panic!("Build process helper first. For release profile use: cargo build -r --bin process_helper");
```
WojciechBarczynski, line 90: `-r` is recommended, so I think we should use it in error messages

noituri (approved):

src/http.rs
```diff
@@ -14,19 +14,30 @@ use crate::{
     error::ApiError,
 };
 
+pub const API_PORT_ENV: &str = "MEMBRANE_VIDEO_COMPOSITOR_API_PORT";
+const PORT_AVAILABLE_MESSAGE: &str = "Make sure that used port is available. For specifying port use: MEMBRANE_VIDEO_COMPOSITOR_API_PORT environment variable.";
```
wkozyra95, line 18: I'm not really sold on that error message. If the issue is different then it will only confuse users.

wkozyra95 (approved):

wkozyra95 (approved):

Issue #200, opened 2023-10-10
Setup snapshot tests for render result


wkozyra95:
- Input for tests should be JSON files written as standalone files (not embedded in the rust)
- Single test case definition should be very minimal e.g sth like this
```
test("file_name", vec![Durration::from_secs(0), Durration::from_secs(0)])
```
- Test should generate images and write them to disk. I would prefer it to be an uncompressed image, but in the format that browser can display.
- Output results should not be stored inside this repo. I'm considering using git lfs or a submodule for that, but leaning toward a submodule.
- When test fails on CI it should upload the expected and actual image as a job artifact.

wkozyra95:
When you remove output from scene definition we should make sure to delete images from a submodule even if all other snapshot match

PR #218, opened 2023-10-16
Initial support for snapshot tests (part 1)

noituri:
This PR adds the initial implementation of snapshot tests. The tests are defined in `src/snapshot_tests.rs`.
The `test` function accepts the following arguments:
- Test name
- Contents of a json file containing `Vec<RegisterRequests>`
- Contents of a json file containing `Scene`
- Vector of timestamps

Updating snapshots:
`cargo run --bin update_snapshots`

Web renderer is not supported since it does not rely on PTS.

~The saved snapshots are saved in `yuv` format. Later on I intend to change it to a format that is viewable by a web browser.~

Issue: #200


src/bin/snapshot_tester/tests.rs
```diff
@@ -0,0 +1,10 @@
+use std::time::Duration;
+
+use crate::utils::snapshot_test;
+
+pub fn run_tests() {
+    snapshot_test(
+        "snapshot_tests/basic.json",
+        vec![Duration::from_secs(2), Duration::from_secs(4)],
+    );
```
noituri, line 9: I forgot to remove it. I'll leave it for now since it can serve as an example

wkozyra95 (commented):
We want to run that code in tests too, so I suspect you will need to move most of that code from bin directory to the package.

src/bin/snapshot_tester/test_runner.rs
```diff
@@ -0,0 +1,131 @@
...
+                ))?;
+
+                match answer {
+                    true => {
+                        fs::write(&save_path, new_snapshot)?;
+                    }
+                    false => {
+                        println!("Snapshot check skipped");
+                    }
+                }
```
wkozyra95, line 90: I sometimes suggest using match over if/else for boolean because it might be compact but if it's formated like that I would use If/else

In my opinion match is better when cases are short enough to produce e.g.
```
true => fs::write(&save_path, new_snapshot)?,
```

src/bin/snapshot_tester/test_runner.rs
```diff
@@ -0,0 +1,131 @@
...
+            let new_snapshot = frame_to_bytes(output_frame);
+            let save_path = self.snapshot_save_path(&pts, &spec.output_id);
+
+            if self.run_mode == RunMode::UpdateSnapshots || !save_path.exists() {
+                fs::write(&save_path, new_snapshot)?;
+                continue;
+            }
+
+            let old_snapshot = fs::read(&save_path)?;
+            if Self::compare_snapshots(&old_snapshot, &new_snapshot) {
```
wkozyra95, line 72:
```suggestion
            if Self::are_snapshots_equal(&old_snapshot, &new_snapshot) {
```

src/bin/snapshot_tester/tests.rs
```diff
@@ -0,0 +1,10 @@
+use std::time::Duration;
+
+use crate::utils::snapshot_test;
+
+pub fn run_tests() {
+    snapshot_test(
+        "snapshot_tests/basic.json",
```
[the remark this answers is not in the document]
noituri: I use the file name for snapshots' names. If I'd use `include_str`, I'd lose the extra information.
I see 2 possible options:
- I could pass the name separately as an argument `snapshot_test(include_str!(...), name, ...)`
- or I could create a simple macro `snapshot_test!()` which would extract the file name and then it'd use `include_str` underneath:

```rust
macro_rules! snapshot_test {
    ($file:expr, ...) => {{
        let snapshot_name = (extract file name from $file)...;
        run_snapshot_test(include_str!($file), snapshot_name, ...)
    }};
}
```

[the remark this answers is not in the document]
wkozyra95: I would go with second argument

src/snapshot_tests/utils.rs
```diff
@@ -0,0 +1,138 @@
...
+            fs::create_dir_all(&snapshots_path).unwrap();
+        }
+        let out_file_name = format!("{}_{}_{}.yuv", self.test_name, pts.as_millis(), output_id);
+        snapshots_path.join(out_file_name)
+    }
+}
+
+pub fn test(
+    test_name: &str,
+    register_renderers_json: &str,
```
wkozyra95, line 87:
```suggestion
    register_renderers_json: Vec<&str>,
```

wkozyra95 (approved):

WojciechBarczynski (approved):

PR #221, opened 2023-10-18
Simplify types in renderer API

wkozyra95:
Follow up after adding type layer for http:
- remove unnecessary transformation field in Builtin transformation
- modify TextureLayout type to be more type safe


jerzywilczek (approved):
:rocket:

noituri (approved):

PR #225, opened 2023-10-19
Add error handling to the new extracted decoder

jerzywilczek:


compositor_pipeline/src/error.rs
```diff
@@ -59,6 +62,15 @@ pub enum OutputInitError {
     OutputError(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum InputInitError {
+    #[error(transparent)]
+    FfmpegError(#[from] ffmpeg_next::Error),
```
WojciechBarczynski, line 68: [NIT]
```suggestion
    FFmpegError(#[from] ffmpeg_next::Error),
```

WojciechBarczynski (approved):

compositor_pipeline/src/error.rs
```diff
@@ -80,13 +92,18 @@ impl PipelineErrorInfo {
 }
 
 const INPUT_STREAM_ALREADY_REGISTERED: &str = "INPUT_STREAM_ALREADY_REGISTERED";
+const DECODER_ERROR: &str = "INPUT_STREAM_DECODER_ERROR";
```
wkozyra95, line 95: Can we detect port already in use error?

compositor_pipeline/src/error.rs
```diff
@@ -59,6 +62,15 @@ pub enum OutputInitError {
     OutputError(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum InputInitError {
+    #[error(transparent)]
+    FfmpegError(#[from] ffmpeg_next::Error),
+
+    #[error(transparent)]
+    InputError(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),
```
wkozyra95, line 71: I'm fine with leaving it as Box<dyn Error>, but if we are implementing automatic From it's too risky. Add some wrapper-type or remove #[from] and map this error manually. (I slightly prefer the second option, but it's not a strong opinion)

The same applies to the encoder

src/rtp_receiver.rs
```diff
@@ -99,47 +101,68 @@ impl RtpReceiver {
         port: u16,
```
wkozyra95, line 101: not introduced in this PR, but start is a bad name, `run` would be better

src/rtp_receiver.rs
```diff
@@ -99,47 +101,68 @@ impl RtpReceiver {
...
-        )?;
-        let input_ctx = input_with_dictionary_and_interrupt(
-            &sdp_filepath,
-            Dictionary::from_iter([("protocol_whitelist", "file,udp,rtp")]),
-            || should_close.try_recv().is_ok(),
-        )?;
-
-        let input = input_ctx
+
+        let result = File::create(&sdp_filepath)
```
wkozyra95, line 108: It's up to you, but I would extract all that code to separate function that returns a Result<> to avoid bunch of those `and_then` calls and matches on Results

noituri (approved):
👍

compositor_pipeline/src/error.rs
WojciechBarczynski: This would be useful to search for available ports on compositor side #227

compositor_pipeline/src/error.rs
jerzywilczek: makes sense

jerzywilczek: I prefer the second option too, but if I mark the box as `#[source]` instead of `#[from]`, the error cannot be `transparent` :(

I guess we have to stick with the wrapper type

compositor_pipeline/src/error.rs
jerzywilczek: Yes, unfortunately it's ugly ;(

compositor_pipeline/src/error.rs
jerzywilczek: i decided not to

compositor_pipeline/src/error.rs
```diff
@@ -44,6 +47,26 @@ pub enum UnregisterOutputError {
     StillInUse(OutputId),
 }
 
+pub struct CustomError(pub Box<dyn std::error::Error + Send + Sync + 'static>);
```
wkozyra95, line 50:
```suggestion
pub struct ExternalError(pub Box<dyn std::error::Error + Send + Sync + 'static>);
```

src/api.rs
```diff
@@ -242,14 +245,46 @@ impl Api {
...
+        Self::check_port_not_available(&result, port)?;
+
+        result?;
 
         Ok(())
     }
+
+    /// Returns Ok(()) if there isn't an error or the error is not a port already in use error.
+    /// Returns Err(ApiError) if the error is a port already in use error.
+    fn check_port_not_available<T>(
```
wkozyra95, line 265:
```suggestion
    fn check_port_not_available_error<T>(
```

wkozyra95 (approved):

src/api.rs
```diff
@@ -242,14 +245,46 @@ impl Api {
...
+        port: u16,
+    ) -> Result<(), ApiError> {
+        if let Err(RegisterInputError::DecoderError(ref id, InputInitError::InputError(ref err))) =
+            register_input_error
+        {
+            if let Some(err) = err.0.downcast_ref::<rtp_receiver::InitError>() {
+                match err {
+                    rtp_receiver::InitError::FfmpegError(ffmpeg_next::Error::Other { errno: ffmpeg_next::error::EADDRINUSE })
+                    | rtp_receiver::InitError::FfmpegError(ffmpeg_next::Error::Other { errno: ffmpeg_next::error::EADDRNOTAVAIL }) =>
+                        Err(ApiError::new(
```
wkozyra95, line 276: fix format

PR #230, opened 2023-10-23
Add support for schemas in vscode

wkozyra95:
I don't use vscode, so not sure if it's correct. Tried opening it in stock vscode and it worked, but I would recommend testing in on your setups to make sure.


WojciechBarczynski (approved):

noituri (approved):

WojciechBarczynski:
Tested on snapshot tests JSON-s and works fine for me :)

PR #233, opened 2023-10-23
Snapshot tests (part 3)

noituri:
Fixed CI tests

Issue: #200

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,8 +21,8 @@ jobs:
 
                   echo ttf-mscorefonts-installer msttcorefonts/accepted-mscorefonts-eula select true | sudo debconf-set-selections
 
-                  sudo apt-get update -y -qq
-                  sudo apt-get install -y libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev ffmpeg libavcodec-dev libavformat-dev libavfilter-dev libavdevice-dev ttf-mscorefonts-installer
+                  apt-get update -y -qq
+                  apt-get install -y libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev ffmpeg libavcodec-dev libavformat-dev libavfilter-dev libavdevice-dev ttf-mscorefonts-installer
 
             - name: 🔧 Install the rust toolchain
               uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,18 +12,17 @@ concurrency:
 jobs:
     rust:
         runs-on: ubuntu-latest
+        container:
+            image: ubuntu:mantic-20231011
         steps:
             - name: 🛠 Install system dependencies
               run: |
                   set -e
 
-                  wget -qO - https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo apt-key add -
-                  sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-jammy.list https://packages.lunarg.com/vulkan/lunarg-vulkan-jammy.list
-                  sudo add-apt-repository ppa:kisak/kisak-mesa
                   echo ttf-mscorefonts-installer msttcorefonts/accepted-mscorefonts-eula select true | sudo debconf-set-selections
 
                   sudo apt-get update -y -qq
-                  sudo apt-get install -y libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev vulkan-sdk mesa-vulkan-drivers ffmpeg libavcodec-dev libavformat-dev libavfilter-dev libavdevice-dev ttf-mscorefonts-installer
+                  sudo apt-get install -y libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev ffmpeg libavcodec-dev libavformat-dev libavfilter-dev libavdevice-dev ttf-mscorefonts-installer
 
             - name: 🔧 Install the rust toolchain
               uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -72,7 +72,7 @@ jobs:
               if: failure()
               uses: actions/upload-artifact@v3
               with:
-                name: produced_snapshot.png
+                name: produced_snapshot
                 path: failed_snapshot_tests/produced.png
                 retention-days: 2
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -60,21 +60,13 @@ jobs:
             - name: 🧪 Run tests
               run: cargo nextest run --no-fail-fast --workspace
 
-            - name: 📦 Upload original snapshot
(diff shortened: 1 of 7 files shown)


src/snapshot_tests/tests.rs
```diff
@@ -309,6 +312,7 @@ pub fn tiled_layout_tests() -> Vec<TestCase> {
                 input14.clone(),
                 input15.clone(),
             ],
+            allowed_error: 25.29,
```
wkozyra95, line 315: change default to 30 and only add those custom values to those 300 errors

.github/workflows/rust.yml
```diff
@@ -17,8 +17,13 @@ jobs:
               run: |
                   set -e
 
+                  wget -qO - https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo apt-key add -
```
wkozyra95, line 20: don't use additional ppa for drivers, instead just use ubuntu docker container 
See https://github.com/membraneframework/video_compositor/commit/ce7f9572fb02c800bb790956c25f1103f7444e5b how to use containers

.github/workflows/rust.yml
```diff
@@ -17,8 +17,13 @@ jobs:
...
+                  sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-jammy.list https://packages.lunarg.com/vulkan/lunarg-vulkan-jammy.list
+                  sudo add-apt-repository ppa:kisak/kisak-mesa
+                  echo ttf-mscorefonts-installer msttcorefonts/accepted-mscorefonts-eula select true | sudo debconf-set-selections
```
wkozyra95, line 23: do we need ms fonts for veranda too?

.github/workflows/rust.yml
```diff
@@ -12,16 +12,23 @@ concurrency:
...
             - name: 🛠 Install system dependencies
               run: |
                   set -e
 
-                  sudo apt-get update -y -qq
-                  sudo apt-get install -y libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev ffmpeg libavcodec-dev libavformat-dev libavfilter-dev libavdevice-dev
+                  echo ttf-mscorefonts-installer msttcorefonts/accepted-mscorefonts-eula select true | debconf-set-selections
+
+                  apt-get update -y -qq
+                  apt-get install -y build-essential curl pkg-config git libssl-dev libclang-dev libnss3 libatk1.0-0 libatk-bridge2.0-0 libgdk-pixbuf2.0-0 libgtk-3-0
```
wkozyra95, line 25:
```suggestion
                  # required when job is running in docker container
                  apt-get install -y build-essential curl pkg-config git libssl-dev libclang-dev libnss3 libatk1.0-0 libatk-bridge2.0-0 libgdk-pixbuf2.0-0 libgtk-3-0
```

.github/workflows/rust.yml
```diff
@@ -45,11 +52,19 @@ jobs:
...
               run: cargo clippy --workspace --all-targets --all-features -- -D warnings
 
-            - name: Generate JSON schema
+            - name: 📄 Generate JSON schema
               run: cargo run --bin generate_json_schema
 
             - name: 🧪 Run tests
               run: cargo nextest run --no-fail-fast --workspace
 
+            - name: 📦 Upload failed snapshot test artifacts
```
wkozyra95, line 61: I checked all the past failed test and did not see any one where anything was uploaded. Did you test that?

.gitignore
```diff
@@ -27,6 +27,7 @@ target/
 
 # Generated during runtime
 /shmem
+/failed_snapshot_tests
```
wkozyra95, line 30: It would be fine to copy those files only when running on CI, so this would not be needed, but if it's already implemented I'm fine with leaving it for non ci envs too.

src/snapshot_tests.rs
```diff
@@ -12,22 +16,66 @@ mod utils;
 
 #[test]
 fn test_snapshots() {
-    if env::var("CI").is_ok() {
-        return;
+    let artifact_save_path = artifacts_path();
```
wkozyra95, line 19:
```suggestion
    let failed_snapshot_path = failed_snapshot_path();
```

src/snapshot_tests.rs
```diff
@@ -12,22 +16,66 @@ mod utils;
...
-    let unused_snapshots = find_unused_snapshots(&produced_snapshots, snapshots_path());
-    if !unused_snapshots.is_empty() {
-        panic!("Some snapshots were not used: {unused_snapshots:#?}")
+    // Check for unused snapshots
+    let snapshot_paths = test_cases
+        .iter()
+        .flat_map(TestCase::snapshot_paths)
+        .collect::<HashSet<_>>();
+    for path in find_unused_snapshots(&snapshot_paths, snapshots_path()) {
+        println!("Removed unused snapshot {path:?}");
```
wkozyra95, line 39: use either println or eprintln consistently

src/snapshot_tests.rs
```diff
@@ -12,22 +16,66 @@ mod utils;
...
+        fs::remove_file(path).unwrap();
+    }
+}
+
+fn handle_error(err: Error) {
+    let Some(test_case_err) = err.downcast_ref::<TestCaseError>() else {
+        panic!("{err}");
+    };
```
wkozyra95, line 47:
```suggestion
    let test_case_err = err.downcast_ref::<TestCaseError>().unwrap();
```
also I would prefer to avoid that downcast, but given it's a test I'm fine with leaving it if it's problematic

src/snapshot_tests.rs
```diff
@@ -12,22 +16,66 @@ mod utils;
...
+    let TestCaseError::Mismatch {
+        snapshot_from_disk,
+        produced_snapshot,
+    } = test_case_err
+    else {
+        panic!("{err}");
+    };
+
+    let artifact_save_path = artifacts_path();
```
wkozyra95, line 56: misleading naming, the same as above

.github/workflows/rust.yml
noituri: I tested it and it works. I force-pushed my changes to this branch so the test history is overwritten. 
Here's one of the tests: https://github.com/membraneframework/video_compositor/actions/runs/6694201788

wkozyra95: I see that you are removing any context from file name. and user needs to identify failed test from logs. Dropping that information does not make sense if you are not gaining anything by it. Either
- add info that identifies tests to file names
- leave current naming and use the fact that you know exact paths and change the code here to upload those 2 paths separately, so we can't avoid zip file

I would prefer the second option, but don't have strong opinions about it.

.gitignore
noituri: I found it convenient to have failed test snapshots when writing tests locally

.github/workflows/rust.yml
noituri: I tested the second option and it looks like github zips individual files even if I upload them separately.
https://github.com/actions/upload-artifact/issues/3#issuecomment-620722206

wkozyra95 (approved):

PR #236, opened 2023-10-24
Fix new web frame embbeder

noituri:


WojciechBarczynski (approved):

compositor_render/src/wgpu/shader.rs
```diff
@@ -162,7 +162,7 @@ impl WgpuShader {
         });
 
         let common_shader_params =
-            CommonShaderParameters::new(pts, texture_views.len() as u32, target.resolution());
+            CommonShaderParameters::new(pts, textures.len() as u32, target.resolution());
```
wkozyra95, line 165: from what I see original code had `sources.len()`, any reason why you are changing that to textures?

wkozyra95 (approved):
When PR is reverted I would recommend following workflow:
- create a new branch
- revert the revert on that branch
- fix that issue in a separate commit
- add in the description of a PR which commit is just a revert and which includes new changes

In this case, it was hard to review because I had to search for the context of these changes.

compositor_render/src/wgpu/shader.rs
WojciechBarczynski: This function accepts `textures` instead of `sources`, since it's `pub` and can be used separately (without `node` abstraction over it), making it possible to use in the web renderer :)
```rust
sources: &[(&NodeId, &NodeTexture)]
```

```rust
let textures = sources
            .iter()
            .map(|(_, node_texture)| node_texture.state())
            .collect::<Vec<_>>();
```

wkozyra95: I missed that texture now is argument(probably misread the diff from the original PR)

PR #238, opened 2023-10-26
Add better checking for unused nodes & compare snapshots with error margin

noituri:


src/bin/update_snapshots/main.rs
```diff
@@ -12,16 +12,25 @@ mod test_case;
...
 fn main() {
+    let mut produced_snapshots = HashSet::new();
+
     println!("Updating snapshots:");
     for snapshot_test in snapshot_tests() {
-        if snapshot_test.run().is_ok() {
-            continue;
+        let was_test_successful = snapshot_test.run().is_ok();
+        if !was_test_successful {
+            println!("Test \"{}\"", snapshot_test.name);
```
wkozyra95, line 24:
```suggestion
        if was_test_successful {
            println!("PASS: \"{}\"", snapshot_test.name);
        } else 
            println!("UPDATE: \"{}\"", snapshot_test.name);
```

src/bin/update_snapshots/main.rs
```diff
@@ -12,16 +12,25 @@ mod test_case;
...
         }
 
-        println!("Test \"{}\"", snapshot_test.name);
         for snapshot in snapshot_test.generate_snapshots().unwrap() {
```
wkozyra95, line 27: you are rendering everything twice?
- it's inefficient
- if we hit sth that shows non deterministically then different image will be used to detect changes and a different one will be written to disk

src/bin/update_snapshots/main.rs
```diff
@@ -12,16 +12,25 @@ mod test_case;
 
 use tests::snapshot_tests;
 
+use crate::utils::{find_unused_snapshots, snapshots_path};
+
 fn main() {
+    let mut produced_snapshots = HashSet::new();
```
wkozyra95, line 18: I would prefer if gathering those snapshots would not be a part of loop that runs tests. You could add a method to TestCase that returns paths for all tests that should be produced.

wkozyra95 (approved):

WojciechBarczynski (approved):

wkozyra95 (approved):

PR #242, opened 2023-10-31
Add generic layout component

wkozyra95:
- Copy WgpuShader and Pipeline implementation + extract some common parts to avoid duplication
- Create a new Layout node (it's not constructible via API)

I tested that by hardcoding layout inside some fake implentation

### DIFF
--- a/compositor_pipeline/src/error.rs
+++ b/compositor_pipeline/src/error.rs
@@ -212,6 +212,7 @@ impl From<&UpdateSceneError> for PipelineErrorInfo {
 const WGPU_INIT_ERROR: &str = "WGPU_INIT_ERROR";
 const WEB_RENDERER_INIT_ERROR: &str = "WEB_RENDERER_INIT_ERROR";
 const BUILTIN_INIT_ERROR: &str = "BUILTIN_INIT_ERROR";
+const LAYOUT_INIT_ERROR: &str = "LAYOUT_INIT_ERROR";
 
 impl From<&InitRendererEngineError> for PipelineErrorInfo {
     fn from(err: &InitRendererEngineError) -> Self {
@@ -224,6 +225,9 @@ impl From<&InitRendererEngineError> for PipelineErrorInfo {
             }
             InitRendererEngineError::BuiltInTransformationsInitError(_) => {
                 PipelineErrorInfo::new(BUILTIN_INIT_ERROR, ErrorType::ServerError)
+            }
+            InitRendererEngineError::LayoutTransformationsInitError(_) => {
+                PipelineErrorInfo::new(LAYOUT_INIT_ERROR, ErrorType::ServerError)
             }
         }
     }
--- a/compositor_render/src/renderer/renderers.rs
+++ b/compositor_render/src/renderer/renderers.rs
@@ -5,6 +5,7 @@ use crate::{
     transformations::{
         builtin::{error::InitBuiltinError, transformations::BuiltinTransformations},
         image_renderer::Image,
+        layout::LayoutRenderer,
         shader::Shader,
         web_renderer::WebRenderer,
     },
@@ -17,6 +18,7 @@ pub(crate) struct Renderers {
     pub(crate) web_renderers: RendererRegistry<Arc<WebRenderer>>,
     pub(crate) images: RendererRegistry<Image>,
     pub(crate) builtin: BuiltinTransformations,
+    pub(crate) layout: LayoutRenderer,
 }
 
 impl Renderers {
@@ -26,6 +28,8 @@ impl Renderers {
             web_renderers: RendererRegistry::new(RegistryType::WebRenderer),
             images: RendererRegistry::new(RegistryType::Image),
             builtin: BuiltinTransformations::new(&wgpu_ctx)?,
+            layout: LayoutRenderer::new(&wgpu_ctx)
+                .map_err(InitBuiltinError::ApplyTransformationMatrix)?,
         })
     }
 }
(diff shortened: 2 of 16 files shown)


compositor_render/src/renderer/renderers.rs
```diff
@@ -26,6 +28,8 @@ impl Renderers {
             web_renderers: RendererRegistry::new(RegistryType::WebRenderer),
             images: RendererRegistry::new(RegistryType::Image),
             builtin: BuiltinTransformations::new(&wgpu_ctx)?,
+            layout: LayoutRenderer::new(&wgpu_ctx)
+                .map_err(InitBuiltinError::ApplyTransformationMatrix)?,
```
WojciechBarczynski, line 32: It uses `apply_layouts.wgsl`. This error suggests that compiling `apply_transformation_matrix.wgsl` failed.

compositor_render/src/transformations/layout.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct Layout {
+    pub top_left_corner: (f32, f32),
+    pub width: f32,
+    pub height: f32,
+    pub rotation_degrees: f32,
+    pub background_color: RGBAColor,
+    pub embeded_child_index: Option<usize>,
```
WojciechBarczynski, line 39: Enum

compositor_render/src/transformations/layout/apply_layouts.wgsl
```diff
@@ -0,0 +1,62 @@
...
+    @location(0) tex_coords: vec2<f32>,
+    @location(1) @interpolate(flat) layout_id: i32,
+}
+
+
+struct Layout {
+    position_transformation: mat4x4<f32>,
+    texture_id: i32,
+    background_color: vec4<f32>,
+}
```
WojciechBarczynski, line 18: Some comments explain that `background_color` is only used when `texture_id==-1` would be helpful.
Also `background_color` is probably not the most suitable name here. Without looking at the implementation, I expect that this shader receives n textures with backgrounds and renders those backgrounds behind if texture size doesn't match/fill the whole position of a single layout/rectangle.

`position_transformation` is also misleading a bit, since this not only changes the position of the box, but also its rotation. I would use `box_transformation`.

Since `wgsl` doesn't support union types, ideally I would define [const](https://www.w3.org/TR/WGSL/#const-decls) for `fill_color_id`
```wgsl
const fill_color_id : i32 = -1;

struct Layout {
    box_transtionation: mat4x4<f32>,
    texture_id: i32,
    fill_color: vec4<f32> // used only when texture_id == fill_color_id
}
```

compositor_render/src/transformations/layout/params.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        result[0..64].copy_from_slice(&mat4_shader_buffer_content(transformation_matrix));
+        result[64..68].copy_from_slice(&texture_id.to_ne_bytes());
+        // 12 bytes padding
+        result[80..84].copy_from_slice(&from_u8_color(background_color.0));
+        result[84..88].copy_from_slice(&from_u8_color(background_color.1));
+        result[88..92].copy_from_slice(&from_u8_color(background_color.2));
+        result[92..96].copy_from_slice(&from_u8_color(background_color.3));
+        result
```
WojciechBarczynski, line 98: [NIT]
I don't like naming variables a `result`. I would prefer to use some descriptive name, like `buffer_content`, even if it matches the function name.

compositor_render/src/wgpu/shader/pipeline.rs
```diff
@@ -72,33 +75,45 @@ impl Pipeline {
...
-            sampler,
-            surfaces,
-            textures_bgl,
-        }
+    pub fn create_texture_bind_group_layout(
+        device: &wgpu::Device,
+        texture_count: u32,
+    ) -> wgpu::BindGroupLayout {
+        device.create_bind_group_layout(&wgpu::BindGroupLayoutDescriptor {
+            label: Some("shader transformation textures bgl"),
```
WojciechBarczynski, line 105: `WgpuShader` uses that. It's not specific for `shader transformation`. This may produce misleading logs/errors, since this doesn't have to be `shader transformation` stuff.

WojciechBarczynski (approved):

compositor_pipeline/src/error.rs
```diff
@@ -212,6 +212,7 @@ impl From<&UpdateSceneError> for PipelineErrorInfo {
 const WGPU_INIT_ERROR: &str = "WGPU_INIT_ERROR";
 const WEB_RENDERER_INIT_ERROR: &str = "WEB_RENDERER_INIT_ERROR";
 const BUILTIN_INIT_ERROR: &str = "BUILTIN_INIT_ERROR";
+const LAYOUT_INIT_ERROR: &str = "LAYOUT_INIT_ERROR";
```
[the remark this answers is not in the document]
wkozyra95: layout is a separate node, builtins will be removed

compositor_render/src/transformations/layout/params.rs
wkozyra95: Why? The function name already describes what the result is well, if it was idiomatic I would use a single letter there. Also value is used immediately before return without any conditions or loops.

compositor_render/src/wgpu/shader/pipeline.rs
wkozyra95: Not added in this PR + currently that code is used for everything so probably the only correct label here would be None :D

PR #243, opened 2023-11-02
Synchronize shared memory resize

noituri:
Resizing step by step:
- `ChromiumSenderThread` creates a `PendingResize` and sends an unembed request to process helper
- Process helper unembeds the source (stops using the shared memory) and sends a confirmation to `BrowserClient`
- `BrowserClient` asks `ChromiumSenderThread` to finalize the resize
- `ChromiumSenderThread` resizes the shared memory

During the resize procedure, shared memory can't be accessed.

### DIFF
--- a/compositor_render/src/transformations/web_renderer.rs
+++ b/compositor_render/src/transformations/web_renderer.rs
@@ -22,13 +22,12 @@ use log::{error, info};
 pub mod browser_client;
 pub mod chromium_context;
 mod chromium_sender;
-mod chromium_sender_thread;
 mod embedder;
 pub(crate) mod node;
 mod shared_memory;
 
-pub const EMBED_SOURCE_FRAMES_MESSAGE: &str = "EMBED_SOURCE_FRAMES";
-pub const UNEMBED_SOURCE_FRAMES_MESSAGE: &str = "UNEMBED_SOURCE_FRAMES";
+pub const EMBED_SOURCES_MESSAGE: &str = "EMBED_SOURCES";
+pub const UNEMBED_SOURCE_MESSAGE: &str = "UNEMBED_SOURCE";
 pub const GET_FRAME_POSITIONS_MESSAGE: &str = "GET_FRAME_POSITIONS";
 
 pub(super) type FrameData = Arc<Mutex<Bytes>>;
--- a/compositor_render/src/transformations/web_renderer/browser_client.rs
+++ b/compositor_render/src/transformations/web_renderer/browser_client.rs
@@ -1,19 +1,23 @@
 use std::sync::{Arc, Mutex};
 
-use crate::GET_FRAME_POSITIONS_MESSAGE;
+use crate::{GET_FRAME_POSITIONS_MESSAGE, UNEMBED_SOURCE_MESSAGE};
 use bytes::Bytes;
 use compositor_chromium::cef;
 use compositor_common::scene::Resolution;
-use log::error;
+use crossbeam_channel::Sender;
+use log::{debug, error};
 
 use crate::transformations::builtin::box_layout::BoxLayout;
 use crate::transformations::web_renderer::{FrameData, SourceTransforms};
+
+use super::chromium_sender::ChromiumSenderMessage;
 
 #[derive(Clone)]
 pub(super) struct BrowserClient {
     frame_data: FrameData,
     source_transforms: SourceTransforms,
     resolution: Resolution,
+    chromium_thread_sender: Option<Sender<ChromiumSenderMessage>>,
 }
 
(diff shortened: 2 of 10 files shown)


compositor_render/src/transformations/web_renderer/chromium_sender/thread_state.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    pub fn shared_memory_mut(
+        &mut self,
+        node_id: &NodeId,
+        source_idx: usize,
+    ) -> Result<&mut SharedMemory, ChromiumSenderThreadError> {
+        let node_shared_memory = self
+            .shared_memory
+            .get_mut(node_id)
```
WojciechBarczynski, line 54: Why `mut`?

WojciechBarczynski (approved):

compositor_render/src/transformations/web_renderer/chromium_sender/thread_state.rs
noituri: I want to return a mutable reference to `SharedMemory` which is stored in `HashMap<NodeId, Vec<SharedMemory>>`. I have to borrow mutably `Vec<SharedMemory>` first so that I could return a mutable reference to one of the elements

compositor_render/src/transformations/web_renderer/browser_client.rs
```diff
@@ -81,6 +53,61 @@ impl BrowserClient {
...
+            error!("Expected shared memory path");
+            return;
+        };
+        debug!("Process helper successfuly unembedded source. \"{shared_memory_path}\" is no longer used");
+        if let Some(chromium_thread_sender) = &self.chromium_thread_sender {
+            chromium_thread_sender
+                .send(ChromiumSenderMessage::FinalizePendingResize {
+                    shared_memory_path: shared_memory_path.into(),
+                })
+                .unwrap()
```
wkozyra95, line 110: is this safe?

compositor_render/src/transformations/web_renderer/browser_client.rs
```diff
@@ -81,6 +53,61 @@ impl BrowserClient {
...
             resolution,
+            chromium_thread_sender: None,
+        }
+    }
+
+    pub fn set_chromium_thread_sender(&mut self, sender: Sender<ChromiumSenderMessage>) {
+        self.chromium_thread_sender.replace(sender);
+    }
+
+    fn update_source_transforms(&self, msg: &cef::ProcessMessage) {
```
wkozyra95, line 64: You are parsing the message in a function that is named update_sth. I get that trying to maintain single responsibility for everything is often overkill, but this is a textbook case where it should not be mixed.

compositor_render/src/transformations/web_renderer/browser_client.rs
```diff
@@ -81,6 +53,61 @@ impl BrowserClient {
...
+        let mut transforms_matrices = Vec::new();
+        for i in (0..msg.size()).step_by(4) {
+            let Some(x) = msg.read_double(i) else {
+                error!("Expected \"x\" value of DOMRect");
+                continue;
```
wkozyra95, line 69: Is this the correct way to handle an error here?

compositor_render/src/transformations/web_renderer.rs
```diff
@@ -22,13 +22,12 @@ use log::{error, info};
...
 mod chromium_sender;
-mod chromium_sender_thread;
 mod embedder;
 pub(crate) mod node;
 mod shared_memory;
 
-pub const EMBED_SOURCE_FRAMES_MESSAGE: &str = "EMBED_SOURCE_FRAMES";
-pub const UNEMBED_SOURCE_FRAMES_MESSAGE: &str = "UNEMBED_SOURCE_FRAMES";
+pub const EMBED_SOURCES_MESSAGE: &str = "EMBED_SOURCES";
+pub const UNEMBED_SOURCE_MESSAGE: &str = "UNEMBED_SOURCE";
```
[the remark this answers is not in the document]
noituri: Unembeding stops the process helper from using the shared memory and removes any references to it

compositor_render/src/transformations/web_renderer/browser_client.rs
noituri: I'm not 100% sure. I will add error checking here.

compositor_render/src/transformations/web_renderer/browser_client.rs
noituri: Now that I look at it, I think `return` would be more suitable. There's also a possible issue with synchronization here when the amount of inputs changes

PR #244, opened 2023-11-06
Fix stretch to resolution

noituri:
Fixes: #234 
Snapshot tests: https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/3


wkozyra95 (commented):
make that PR to master instead, in the new api stretch will be removed so there is no point fixing it there, but we need it fixed in the old one because we will be publishing a new version with the old API for the membrane plugin

wkozyra95 (commented):
you did not uncomment all the stretch tests

wkozyra95 (approved):

WojciechBarczynski (commented):
LGTM

PR #247, opened 2023-11-09
Handle unregistering web renderer

noituri:


jerzywilczek (approved):
I mean it looks fine, but I might be missing something bc I don't understand the big picture

WojciechBarczynski (approved):

wkozyra95 (approved):

wkozyra95:
after merging cherry pick that to new-api too

wkozyra95 (changes requested):
we most likelly won't be releasing any new version for old API, so either switch this PR to v0.1 branch or rebase to new API. I don't think there is a need to fix that in the old API, so I would suggest rebasing.

PR #248, opened 2023-11-09
Add layout layer

wkozyra95:
Concepts:
- RenderNode - element in render graph with its own texture to render on.
- RenderGraph - graph of RenderNodes. Each node has a set of input pads that define list of nodes that will be passed to it as textures when rendering a node.
- Scene - tree of components
- Component - a base fragment of a scene. A single node is responsible for rendering one or more components.
- BaseNode - intermediate representation when building Node tree.
- Node - represents a group of components. Node tree is constructed from a Component tree and it's used to build a new RenderGraph where each Node represents params used to create a RenderNode.
- LayoutComponent - a subset of components that define the layout of elements. If the layout component is a child of a different layout component it can be flattened and rendered as one node.

How building scene works:
- base_node() method returns Node struct that converts the component tree into the Node tree
- That generated tree can be used to build RenderGraph where:
  - Node.children are nodes that should be connected to the current node via input_pads
  - Node.kind represents params that should be passed to when creating RenderNode. 
    - For nodes like shader nothing really is changing, because NodeKind:Shader is just passing the same prams it was always passed.
    - For LayoutNode we are passing the root of a component subtree that will be recalculated for each frame

How render works:
- For non-layout components nothing changed
- For layout components:
  - We start with a component that is the root of the component sub-tree.
  - We go down the component tree until we find a component that is not a LayoutComponent and finishes the tree traversal there
  - If the layout component is a direct child of another layout component we know that traversing the tree this way will never get us to components that are rendered by a different layout node.
  - `layout()` method on each LayoutComponent will return NestedLayout that represents specific rectangles that need to be rendered, but it still preserves the parent-children structure
  - NestedLayout can be flattened to just a list of rectangles, by recalculating coordinates and reindexing child nodes.


compositor_render/src/scene/view_component.rs
```diff
@@ -0,0 +1,188 @@
...
+                            }
+                        }
+                        _ => {
+                            NestedLayout {
+                                layout: layout.clone(),
+                                children: vec![NestedLayout {
+                                    layout: Layout {
+                                        top: 0.0,
+                                        left: 0.0,
+                                        content: LayoutContent::ChildNode(0), // TODO: this will be recalculated latter
```
wkozyra95, line 88: This is probably the most hacky part of the entire approach. I'm relaying here on the fact that when traversing the component tree I will visit all children components that represent nodes in the same order as when traversing the Node tree.

In particular, this means that this approach will only work if all children nodes will be always used in the layout. If we introduce component that has children that only conditionally show up in layout this will stop working.

compositor_render/src/scene/layout.rs
```diff
@@ -0,0 +1,141 @@
...
+                    input.size = input_resolutions[child_index_offset];
+                    child_index_offset += 1;
+                }
+                Component::Shader(_) => {
+                    child_index_offset += 1; // no state
+                }
+                Component::Layout(layout) => {
+                    let node_children = layout.node_children().len();
+                    layout.update_state(
+                        &input_resolutions[child_index_offset..child_index_offset + node_children],
```
noituri, line 106: Is it guaranteed that `child_index_offset + node_children <= input_resolutions.len()`?

noituri (approved):
I'm still trying to wrap my head around everything but lgtm so far

jerzywilczek (approved):
Looks good

wkozyra95:
I'll talk about the comments in follow-up PR. Feel free to keep commenting here if you have any other feedback.

compositor_render/src/scene/view_component.rs
```diff
@@ -0,0 +1,188 @@
...
+            height: height as f32,
+            rotation_degrees: 0.0,
+            content: LayoutContent::Color(RGBAColor(0, 0, 0, 0)),
+        }
+    }
+
+    /// Calculate size of a dynamically sized child component. Returned value
+    /// represents width if the direction is `ViewChildrenDirection::Row` or height if
+    /// the direction is `ViewChildrenDirection::Column`.
+    fn dynamic_size_for_static_children(&self, size: Resolution) -> usize {
```
WojciechBarczynski, line 141: Make it `static_child_size`

WojciechBarczynski (commented):
GJ

PR #250, opened 2023-11-10
Add relative positioning support for layout components

wkozyra95:
- https://github.com/membraneframework/video_compositor/pull/250/commits/1917333ffb0d09e7dd8b1c9b7f3b21a1a0b9ac84 Add relative positioning to view (it's intended to be supported in other layout components)
- https://github.com/membraneframework/video_compositor/pull/250/commits/929739d11363d7a6eaae2714088f7a651dd14060 Fix counting child nodes (old implementation wouldn't work for components where not all children are part of a layout tree)
- Fix rendering background on component that is a root of a node

https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/5

### DIFF
--- a/compositor_render/src/scene/view_component.rs
+++ b/compositor_render/src/scene/view_component.rs
@@ -1,4 +1,4 @@
-use compositor_common::{scene::Resolution, util::colors::RGBAColor};
+use compositor_common::scene::Resolution;
 
 use crate::{
     scene::ViewChildrenDirection,
@@ -52,14 +52,15 @@ impl ViewComponent {
         self.children.iter_mut().collect()
     }
 
-    pub(super) fn layout(&self, size: Resolution) -> Vec<NestedLayout> {
+    pub(super) fn layout(&self, size: Resolution) -> NestedLayout {
         let dynamic_child_size = self.dynamic_size_for_static_children(size);
 
         // offset along x or y direction (depends on self.direction) where next
         // child component should be placed
         let mut static_offset = 0;
 
-        self.children
+        let children: Vec<_> = self
+            .children
             .iter()
             .map(|child| {
                 let position = match child {
@@ -90,7 +91,19 @@ impl ViewComponent {
                     }
                 }
             })
-            .collect()
+            .collect();
+        NestedLayout {
+            layout: Layout {
+                top: 0.0,
+                left: 0.0,
+                width: size.width as f32,
+                height: size.height as f32,
+                rotation_degrees: 0.0,
+                content: LayoutContent::Color(self.background_color),
+            },
+            child_nodes_count: children.iter().map(|l| l.child_nodes_count).sum(),
+            children,
+        }
     }
 
     fn layout_static_child(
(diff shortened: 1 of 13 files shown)


src/types/component.rs
```diff
@@ -47,11 +47,44 @@ pub struct InputStream {
...
+
     pub width: Option<usize>,
     pub height: Option<usize>,
+
+    /// Direction defines how static children are positioned inside the View.
+    /// "row" - Children positioned from left to right.
+    /// "column" - Children positioned from top to bottom.
+    pub direction: Option<ViewDirection>,
+
+    /// Distance between the top edge of this component and the top edge of it's parent. If this
```
noituri, line 59:
```suggestion
    /// Distance between the top edge of this component and the top edge of its parent. If this
```
Change `it's` to `its` in other comments too

noituri (approved):
I think it's ok

src/types/component.rs
```diff
@@ -47,11 +47,44 @@ pub struct InputStream {
...
+    /// Height of a component in pixels. Required when using absolute positioning.
     pub height: Option<usize>,
+
+    /// Direction defines how static children are positioned inside the View.
+    /// "row" - Children positioned from left to right.
+    /// "column" - Children positioned from top to bottom.
+    pub direction: Option<ViewDirection>,
+
+    /// Distance between the top edge of this component and the top edge of its parent.
+    /// If this field is defined, then component will ignore a layout defined by its parent.
```
jerzywilczek, line 62: I don't think this line is super clear. How about `If this field is defined, this element will be absolutely positioned, instead of being laid out by it's parent`?

compositor_render/src/scene/view_component.rs
```diff
@@ -1,11 +1,25 @@
...
+    /// offset inside parent component
+    static_offset: usize,
+    /// For direction=row defines width of a dynamically sized component
+    /// For direction=column defines height of a dynamically sized component
+    dynamic_child_size: usize,
+    parent_size: Resolution,
+}
 
 impl ViewComponent {
     pub(super) fn base_node(&self) -> Result<BaseNode, BuildSceneError> {
```
jerzywilczek, line 25: I forgot to add that in the previous review:
The name `BaseNode` threw me off quite a bit. It confused me for an embarrassingly long time when I was looking over the last PR. I think the name should be changed, but I still don't fully understand what this is so I can't really think of a better name.

jerzywilczek (approved):
Mostly looks good, but I think we should take a more general look at the naming of some of the structs and enums and rethink it. The thing I mentioned here is just the tip of the iceberg.

compositor_render/src/scene/view_component.rs
```diff
@@ -1,11 +1,25 @@
...
+
+#[derive(Debug)]
+struct StaticChildLayoutOpts {
+    width: Option<usize>,
+    height: Option<usize>,
+    /// offset inside parent component
+    static_offset: usize,
+    /// For direction=row defines width of a dynamically sized component
+    /// For direction=column defines height of a dynamically sized component
+    dynamic_child_size: usize,
```
WojciechBarczynski, line 20: `dynamic` is quite unintuitive for me

compositor_render/src/scene/view_component.rs
wkozyra95: Did you see the comment on the BaseNode type?
```
/// Intermediate representation of a node tree while it's being constructed.
```

> but I still don't fully understand what this is so I can't really think of a better name.

BaseNode is an intermidiate representation. When we are creating a Node tree from the Component tree we go through the component tree recursively merging components into nodes if possible. We need that intermediate type to add special logic for layout components that are node roots.

WojciechBarczynski (approved):

PR #251, opened 2023-11-10
Fix child indexing in LayoutNode

wkozyra95:


wkozyra95:
Merged into https://github.com/membraneframework/video_compositor/pull/250

PR #255, opened 2023-11-17
Add `overflow=hidden` for `View` component

wkozyra95:
https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/11/files

tested only cropping views, to support it fully we need RenderLayout to support cropping

### DIFF
--- a/compositor_render/src/transformations/layout/flatten.rs
+++ b/compositor_render/src/transformations/layout/flatten.rs
@@ -1,0 +1,124 @@
+use compositor_common::util::colors::RGBAColor;
+
+use super::{Crop, LayoutContent, NestedLayout, RenderLayout, RenderLayoutContent};
+
+impl NestedLayout {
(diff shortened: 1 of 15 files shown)


compositor_render/src/transformations/layout/flatten.rs
```diff
@@ -0,0 +1,124 @@
...
+                        // The same factor holds for translations from `self.layout`
+                        let horizontal_scale_factor = child_crop.width / layout.width;
+                        let vertical_scale_factor = child_crop.height / layout.height;
+
+                        let crop = Crop {
+                            top: child_crop.top + (top_diff * vertical_scale_factor),
+                            left: child_crop.left + (left_diff * horizontal_scale_factor),
+                            width: crop_width * horizontal_scale_factor,
+                            height: crop_height * vertical_scale_factor,
+                        };
```
WojciechBarczynski, line 86: Doesn't this ignore `child_crop` `width` and `height`?

PR #259, opened 2023-11-20
Add `overflow=fit` option

wkozyra95:
https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/12

### DIFF
--- a/compositor_render/src/scene/view_component/layout.rs
+++ b/compositor_render/src/scene/view_component/layout.rs
@@ -32,14 +32,21 @@ impl ViewComponentParam {
         pts: Duration,
     ) -> NestedLayout {
         let static_child_size = self.static_child_size(size, children, pts);
-        let crop = match self.overflow {
-            Overflow::Visible => None,
-            Overflow::Hidden => Some(Crop {
-                top: 0.0,
-                left: 0.0,
-                width: size.width,
-                height: size.height,
-            }),
+        let (scale, crop) = match self.overflow {
+            Overflow::Visible => (1.0, None),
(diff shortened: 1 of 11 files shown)


wkozyra95:
> Should I review this?

Yes, sorry forgot to add reviewers

compositor_render/src/scene/view_component/layout.rs
```diff
@@ -175,6 +188,32 @@ impl ViewComponentParam {
...
+        size: Size,
+        children: &[StatefulComponent],
+        pts: Duration,
+    ) -> f32 {
+        let sum_size = self.sum_static_children_sizes(children, pts);
+        let (max_size, max_alternative_size) = match self.direction {
+            super::ViewChildrenDirection::Row => (size.width, size.height),
+            super::ViewChildrenDirection::Column => (size.height, size.width),
+        };
+        let max_alternative_size_for_child = Self::static_children_iter(children, pts)
```
wkozyra95, line 202: If direction is row then I consider size to be widht and alternative size to be height. I couldn't find a better name, I'm open to suggestions

compositor_render/src/transformations/layout.rs
```diff
@@ -70,6 +70,9 @@ pub struct NestedLayout {
     pub width: f32,
     pub height: f32,
     pub rotation_degrees: f32,
+    pub scale_x: f32,
+    pub scale_y: f32,
```
WojciechBarczynski, line 74: What is this? Shouldn't the scale be already included in `width` / `height`?

noituri (approved):

PR #264, opened 2023-11-23
Configure multi-staged docker build

wkozyra95:
It will produce smaller images e.g. not target directory or other dependencies that are needed only for build

On macos do not try to connect GPU - default to software-based rendering.


WojciechBarczynski (approved):
Doesn't work on mac with arm. idk why

PR #266, opened 2023-11-23
Setup docs website

wkozyra95:
Deployed to https://docs.membrane.stream/video_compositor/

- CI workflow to build and lint js
- CI workflow to deploy automatically after merging to master
- Generated from template
- Copied a few pages from the wiki to see how the content looks there. Those are not intended as a final content


noituri (approved):

docs/static/img/logo.png
WojciechBarczynski, line 1: IMO it's too generic. I can't tell what this tool is supposed to do. With the `live compositor` name, I think it would be beneficial, that the logo has some relationship with video processing, such that newcomers can understand what this project is supposed to do just by looking at the name and logo.

WojciechBarczynski (approved):

docs/static/img/logo.png
wkozyra95: As mentioned on slack this is placeholder. I would prefer sth better, but I also think that it's good enough if we don't come up with anything better.

PR #267, opened 2023-11-24
Cleanup code from old API

wkozyra95:


compositor_render/src/transformations/layout/transformation_matrices.rs
WojciechBarczynski, line 1: Perfectly, now you are on git blame for this atrocity :P

WojciechBarczynski (commented):
LGTM

PR #268, opened 2023-11-24
Add documentation for new API Components

wkozyra95:


src/types/component.rs
```diff
@@ -166,22 +216,34 @@ pub enum TextStyle {
...
+    /// Wraps at a glyph level.
     Glyph,
+    /// Preserve words when wrapping.
     Word,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum TextWeight {
+    /// Thin weight (100), the thinnest value.
```
noituri, line 230: I'm not sure about it but I think `the thinnest value` and `the thickest value` sound a bit off. The font can be thin/thick but the value is small or large.

noituri (approved):

src/types/component.rs
wkozyra95: those comments are copied from glyphone, but I agree this sound a bit "not yes" :D

src/types/component.rs
```diff
@@ -36,7 +38,7 @@ pub struct View {
     /// Height of a component in pixels. Required when using absolute positioning.
     pub height: Option<f32>,
 
-    /// Direction defines how static children are positioned inside the View.
+    /// Direction defines how static children are positioned inside the View component.
```
WojciechBarczynski, line 41: Doesn't define what `static` means.

src/types/component.rs
```diff
@@ -18,6 +18,8 @@ pub enum Component {
     Tiles(Tiles),
 }
 
+/// Component that represents incoming RTP stream. Stream is identified
+/// by an `input_id` that was defined in an `RegisterInputStream` request.
```
WojciechBarczynski, line 22:
```suggestion
/// Component representing incoming RTP stream. Stream is identified
/// by an `input_id` defined in a `RegisterInputStream` request.
```

src/types/component.rs
```diff
@@ -137,21 +167,41 @@ pub struct Text {
...
+    /// (default=7682) Width of a texture that text will be rendered on. Value is ignored if width
+    /// is defined.
     pub max_width: Option<f32>,
+    /// (default=4320) Height of a texture that text will be rendered on. Value is ignored if
+    /// height is defined.
     pub max_height: Option<f32>,
 
+    /// Font size in pixels.
     pub font_size: f32,
+    /// Distance between lines. Defaults to value of a `font_size` property.
```
WojciechBarczynski, line 187:
```suggestion
    /// Distance between lines in pixels. Defaults to the value of a `font_size` property.
```

src/types/component.rs
```diff
@@ -137,21 +167,41 @@ pub struct Text {
...
+    /// (default="#00000000") Background color in `#RRGGBBAA` format.
     pub background_color_rgba: Option<RGBAColor>,
+    /// (default="Verdana") Font family.
+    ///
+    /// Provide family-name for specific font. "generic-family" values like e.g. "sans-serif" will not work.
     /// https://www.w3.org/TR/2018/REC-css-fonts-3-20180920/#family-name-value   
-    /// use font family name, not generic family name
-    pub font_family: Option<String>, // TODO: Arc<str>
+    pub font_family: Option<Arc<str>>,
+    /// (default="normal") Font style. Font that you selected needs to support the selected weight.
```
WojciechBarczynski, line 198: I would keep it consistent across the whole documentation. I prefer a non-personal form rather than addressing users like this.

```suggestion
    /// (default="normal") Font style. The selected font needs to support the selected weight.
```

src/types/component.rs
```diff
@@ -166,22 +216,34 @@ pub enum TextStyle {
 #[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum TextWrapMode {
+    /// Disable text wrapping. Text that does not fit inside the texture will be cut off.
     None,
+    /// Wraps at a glyph level.
     Glyph,
+    /// Preserve words when wrapping.
```
WojciechBarczynski, line 223:
```suggestion
    /// Wraps at a word level. Prevent splitting words when wrapping.
```

src/types/component.rs
```diff
@@ -137,21 +167,41 @@ pub struct Text {
...
-    /// use font family name, not generic family name
-    pub font_family: Option<String>, // TODO: Arc<str>
+    pub font_family: Option<Arc<str>>,
+    /// (default="normal") Font style. Font that you selected needs to support the selected weight.
     pub style: Option<TextStyle>,
+    /// (default="left") Text align.
     pub align: Option<HorizontalAlign>,
+    /// (default="none") Text wrapping options.
     pub wrap: Option<TextWrapMode>,
+    /// (default="normal") Font weight. Font that you selected needs to support the selected weight.
```
WojciechBarczynski, line 204:
```suggestion
    /// (default="normal") Font weight. Font that you selected needs to support the selected weight.
    /// Based on [OpenType specification](https://learn.microsoft.com/en-gb/typography/opentype/spec/os2#usweightclass).
```

src/types/component.rs
```diff
@@ -137,21 +167,41 @@ pub struct Text {
     pub id: Option<ComponentId>,
     pub text: Arc<str>,
 
+    /// Width of a texture that text will be rendered on. If not provided the resulting texture
+    /// will be sized based on the defined text, but limited to `max_width` value.
```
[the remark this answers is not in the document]
wkozyra95: Please clarify, no key is required here, all combinations are supported except very specific one (unknown width and known height).
And I would argue that one unsupported combination does not really make sense.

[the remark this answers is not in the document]
wkozyra95: Good point, I must have copied it from weight/height and did not notice. The sentence "Width of a texture that text will be rendered on" for max_width was incorrect.

> Based solely on the max_width description one might understand that in the case of specifying max_width output always has output texture width of max_width.

In this case, max_width had incorrect information so it's fair, but I disagree with the take that field like max_width needs to make sense on its own if a user does not read any other field's description.

> /// If text is longer than specified `max_width`, it will be wrapped accordingly
> /// to the `wrap` parameter.

- max width limits width of a component
- width explicitly defines width of a component
- warp defines a behavior based on that width
I don't think we should mention all fields that will be impacted by that field in the description

src/types/component.rs
```diff
@@ -60,20 +62,32 @@ pub struct View {
...
 #[serde(rename_all = "snake_case")]
 pub enum Overflow {
+    /// Components that are outside of their parent area will be rendered.
     Visible,
+    /// Only render parts of the children that are inside their parent area.
     Hidden,
+    /// If children component are to big to fit inside the parent resize everything inside to fit.
+    ///
+    /// Components that have dynamic size will be treated as if they had a size 0 when calculating
+    /// scaling factor.
```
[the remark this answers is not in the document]
WojciechBarczynski: I think sth like this would be better

WojciechBarczynski (commented):
LGTM

PR #269, opened 2023-11-27
Generate wiki from rust types

noituri:
In this PR I implemented basic docs generation from rust types. There are still some imperfections and the code itself needs a refactor. I will fix it in the follow-up PR.

Example documentation page:

# View
```typescript
type View = {
  background_color_rgba?: string,
  bottom?: f32,
  children?: Component[],
  direction?: "row" | "column",
  height?: f32,
  id?: string,
  left?: f32,
  overflow?: "visible" | "hidden" | "fit",
  right?: f32,
  rotation?: f32,
  top?: f32,
  transition?: Transition,
  type: "view",
  width?: f32,
}
```
### Properties
- `background_color_rgba` - Background color of a component in a "#RRGGBBAA" format. Defaults to transparent "#00000000".
- `bottom` - Distance between the bottom edge of this component and the bottom edge of its parent. If this field is defined, this element will be absolutely positioned, instead of being laid out by it's parent.
- `direction` - Direction defines how static children are positioned inside the View. "row" - Children positioned from left to right. "column" - Children positioned from top to bottom.
- `height` - Height of a component in pixels. Required when using absolute positioning.
- `left` - Distance between the left edge of this component and the left edge of its parent. If this field is defined, this element will be absolutely positioned, instead of being laid out by it's parent.
- `right` - Distance between the right edge of this component and the right edge of its parent. If this field is defined, this element will be absolutely positioned, instead of being laid out by it's parent.
- `rotation` - Rotation of a component in degrees. If this field is defined, this element will be absolutely positioned, instead of being laid out by it's parent.
- `top` - Distance between the top edge of this component and the top edge of its parent. If this field is defined, then component will ignore a layout defined by its parent.
- `width` - Width of a component in pixels. Required when using absolute positioning.

## Transition
```typescript
type Transition = {
  duration_ms: f64,
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ target/
 # Generated during runtime
 /shmem
 /failed_snapshot_tests
+/docs
 
 /video_compositor.app
 /video_compositor
--- a/src/bin/generate_wiki/docs_builder.rs
+++ b/src/bin/generate_wiki/docs_builder.rs
@@ -1,0 +1,567 @@
+use std::{collections::HashMap, ops::Deref, rc::Rc};
+
+use schemars::{
+    gen::{SchemaGenerator, SchemaSettings},
+    schema::{InstanceType, RootSchema, Schema, SchemaObject, SingleOrVec, SubschemaValidation},
+    JsonSchema,
+};
+
+#[derive(Debug)]
+pub struct DocPage {
+    title: Rc<str>,
+    // TODO: Consider using HashMap with ordering
+    definitions: Vec<TypeDefinition>,
+}
+
+impl DocPage {
+    pub fn new(title: Rc<str>) -> Self {
+        Self {
+            title,
+            definitions: Vec::new(),
+        }
+    }
+
+    pub fn add_definition(&mut self, def: TypeDefinition) {
+        self.definitions.push(def);
+    }
+
+    pub fn remove_definition(&mut self, name: &Rc<str>) {
+        self.definitions
+            .retain(|def| def.name.as_ref() != Some(name))
+    }
+
+    pub fn definition(&self, name: &Rc<str>) -> Option<&TypeDefinition> {
+        self.definitions
+            .iter()
+            .find(|def| def.name.as_ref() == Some(name))
+    }
+}
+
+pub struct DocsBuilder {
+    schema_generator: SchemaGenerator,
+    pages: Vec<DocPage>,
+}
+
+impl DocsBuilder {
+    pub fn new() -> Self {
+        let mut settings = SchemaSettings::default();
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -28,6 +28,7 @@ target/
 # Generated during runtime
 /shmem
 /failed_snapshot_tests
+/docs
```
wkozyra95, line 31: docs directory is already used on `new-api` branch, website is under that path

src/bin/generate_wiki/docs_builder.rs
wkozyra95, line 1: `generate_wiki` -> `generate_docs`

src/bin/generate_wiki/main.rs
```diff
@@ -0,0 +1,18 @@
...
+mod docs_builder;
+mod types;
+
+fn main() {
+    let docs_path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("docs");
+    if !docs_path.exists() {
+        fs::create_dir(&docs_path).unwrap();
+    }
+
+    let pages = DocsBuilder::new().add_definition::<Component>(true).build();
```
wkozyra95, line 14: given that you are creating a separate page for each component wouldn't be easier just to generate docs for each component explicitly?
e.g.
```
let schema = ...

let view = Docs::generate_component_docs(schema, "View", "view") // I assume you need that 3 argument to detect vairant?
let shader = DocsBuild::generate_compontnt_docs(schema, "Shader", "shader")

// write to file
```

This way you don't need to modify the schema

Also you don't seem to need the Builder abstraction, I assume this is leftover after some differnt approach?

src/bin/generate_wiki/docs_builder.rs
```diff
@@ -0,0 +1,567 @@
...
+        Type::Object(properties) => properties
+            .iter()
+            .for_each(|prop| find_references(&prop.type_def.ty, references)),
+        _ => {}
+    }
+}
+
+fn indent(n: usize) -> String {
+    "  ".repeat(n)
+}
```
wkozyra95, line 546: Maybe make INDENT const instead and use `INDENT.repeat(n)` where you need it?

src/bin/generate_wiki/docs_builder.rs
```diff
@@ -0,0 +1,567 @@
...
+            .collect::<String>();
+        if properties.is_empty() {
+            return String::new();
+        }
+        format!("#### Properties\n{properties}")
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+enum Type {
```
wkozyra95, line 279: I would avoid name like that

PR #270, opened 2023-11-27
Add generated docs to website

wkozyra95:
Now website build requires rust to generate markdown files, so I moved website lint/build job to the same workflow as the regular Rust build.


WojciechBarczynski (commented):
LGTM

PR #272, opened 2023-11-27
Add `resize_content` option

wkozyra95:
Still need to test it. I opened this to get early feedback on API.

The behavior of a `resize_content` option:
- `fit` with  image inside view will produce that image aligned to top left
- `fit` with multiple children where some of them are `View` that do not have width/height.
  - if summary width(direction=row) is larger than parents then rescale to fit. All dynamic components will have width 0.
  - if summary width(direction=row) is smaller than parents do not scale because it will always fit
- `fill` with image inside view will produce that image centered inside the view
- `fill` with multiple children  where some of them are `View`s that do not have width specified
  - if summary width(direction=row) is larger than parents then rescale to fit. All dynamic components will have width 0.
  - if the summary width(direction=row) is smaller than the parents do not scale because the dynamic component will scale.
- Yes I know fit and fill are inconsistent (I'm open to suggestions). The main problem here is that in most cases users will want to align to the center when it's a non-layout component and to the top-left when it's a layout, but I don't want to add logic like that in code, it would be hard to explain in docs.

Alternative approaches:
- Separate component that can only take one child.
  - API is more intuitive. There is an easy way to add more options.
  - A lot more work overall. The new component would be a LayoutComponent, so eventually we would have to support a lot of the same fields as in View e.g. width/height/top/left/right/bottom. Respect top/left/right/bottom in its children etc...
- We could remove the current inconsistency between fit and fill and
   - always center the image (both for fit and fill)
   - always align to the top left
   - provide options on how to align (I want to avoid nested objects in API and it would be hard to name those additional fields so it's clear that they refer to alignment after resizing
- Make `resize_content` only work if it has only one child
  - Pro: Clean API, (overflow=fit and resize_content) has its own use cases and a bit different behavior)
  - Pro: We don't need to worry about dynamically sized components. (We only have one so it's a lot easier to handle all the cases)
  - Pro: We don't need to implement totally new LayoutComponent, so the maintenance burden is not increased
  - Con: I don't like it. It seems wrong. :D

### DIFF
--- a/compositor_render/src/scene/components.rs
+++ b/compositor_render/src/scene/components.rs
@@ -129,6 +129,7 @@ pub struct ViewComponent {
     pub position: Position,
     pub transition: Option<Transition>,
     pub overflow: Overflow,
+    pub resize_content: ResizeMode,
 
     pub background_color: RGBAColor,
 }
@@ -138,6 +139,13 @@ pub enum Overflow {
     Visible,
     Hidden,
     Fit,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum ResizeMode {
+    None,
+    Fit,
+    Fill,
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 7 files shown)


wkozyra95:
Closing in favor of creating separate component

PR #275, opened 2023-11-29
Add `Rescaler` component

wkozyra95:
Resize, or Rescale would be better name, but we should try to use verbs for component names to keep it consistent. Initially, I started with `Scaler`, but Jerzy suggested `Rescaler`. I don't have any ideas for a better name, I'm open to suggestions.

https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/18

### DIFF
--- a/compositor_render/src/scene/rescaler_component.rs
+++ b/compositor_render/src/scene/rescaler_component.rs
@@ -1,12 +1,15 @@
 use std::{ops::Add, time::Duration};
 
-use compositor_common::util::{colors::RGBAColor, ContinuousValue, InterpolationState};
+use compositor_common::util::{
+    align::{HorizontalAlign, VerticalAlign},
+    ContinuousValue, InterpolationState,
+};
 
-use crate::{scene::ViewChildrenDirection, transformations::layout::NestedLayout};
+use crate::transformations::layout::NestedLayout;
 
 use super::{
-    components::ViewComponent, layout::StatefulLayoutComponent, scene_state::BuildStateTreeCtx,
-    Component, ComponentId, IntermediateNode, Overflow, Position, SceneError, Size,
+    components::RescalerComponent, layout::StatefulLayoutComponent, scene_state::BuildStateTreeCtx,
+    Component, ComponentId, IntermediateNode, Position, ResizeMode, SceneError, Size,
     StatefulComponent, Transition,
 };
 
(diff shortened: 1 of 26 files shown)


compositor_render/src/scene/rescaler_component/layout.rs
```diff
@@ -0,0 +1,111 @@
...
+                .map(|child_width| child_width * scale)
+                .unwrap_or(size.width),
+            height: child
+                .height(pts)
+                .map(|child_height| child_height * scale)
+                .unwrap_or(size.height),
+            rotation_degrees: 0.0,
+            scale_x: scale,
+            scale_y: scale,
+            crop: None,
```
wkozyra95, line 105: not sure if rescaler should crop when using fill? currentlly it does not. any opinions?

src/bin/update_snapshots/main.rs
```diff
@@ -20,11 +20,18 @@ use crate::{
 fn main() {
     println!("Updating snapshots:");
 
-    let test: Vec<_> = snapshot_tests()
-        .into_iter()
-        .map(TestCaseInstance::new)
-        .collect();
-    for test in test.iter() {
+    let tests: Vec<_> = snapshot_tests();
+    let has_only_flag = tests.iter().any(|t| t.only);
```
wkozyra95, line 24: This flag is intended for local debuging with test. If you want to run only one test you can set `only: true` on there and run `cargo --bin update_snapshots`. Only selected test will be run and code responsible for removing unused snapshots will not be run.

compositor_render/src/scene/rescaler_component.rs
```diff
@@ -0,0 +1,143 @@
...
+        let transition = self.transition.or_else(|| {
+            let Some(previous_state) = previous_state else {
+                return None;
+            };
+            let Some(duration) = previous_state.remaining_transition_duration(ctx.last_render_pts)
+            else {
+                return None;
+            };
+            previous_state.transition.map(|_| Transition { duration })
+        });
```
WojciechBarczynski, line 125: As I said in person, this might not be correct for non-linear transition methods. I don't think you have to fix this right now. You can leave some comment here about it.

compositor_render/src/scene/rescaler_component/layout.rs
```diff
@@ -0,0 +1,111 @@
...
+        child: &StatefulComponent,
+        pts: Duration,
+        scale: f32,
+    ) -> NestedLayout {
+        let (content, children, child_nodes_count) = match child {
+            StatefulComponent::Layout(layout_component) => {
+                let children_layouts = layout_component.layout(
+                    Size {
+                        width: size.width / scale,
+                        height: size.height / scale,
```
WojciechBarczynski, line 55: [NIT]
I think it's more intuitive to multiply by scale param, than divide by it. You would also need to change the division order in lines 32 and 35 in such case.

compositor_render/src/scene/rescaler_component/layout.rs
WojciechBarczynski: Option `visible` / `hidden`

compositor_render/src/scene/rescaler_component/layout.rs
wkozyra95: yes that is how scale works here, but in this specific case I'm doing transformation in other direction. I have a size after scalling and want to calculate what size it should have before scalling

compositor_render/src/scene/rescaler_component.rs
```diff
@@ -0,0 +1,143 @@
...
+impl StatefulRescalerComponent {
+    /// Generate state of the component for particular pts value.
+    fn transition_snapshot(&self, pts: Duration) -> RescalerComponentParam {
+        let (Some(transition), Some(start)) = (self.transition, &self.start) else {
+            return self.end.clone();
+        };
+        let interpolation_progress = InterpolationState(f64::min(
+            1.0,
+            (pts.as_secs_f64() - self.start_pts.as_secs_f64()) / transition.duration.as_secs_f64(),
+        ));
```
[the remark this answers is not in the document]
wkozyra95: I think this is sth interpolation logic should handle when we have sth other than linear, for now it's fine i think

noituri (approved):
lgtm but I don't really like the name `Rescaler`

compositor_render/src/scene/rescaler_component.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    position: Position,
+    mode: ResizeMode,
+    horizontal_align: HorizontalAlign,
+    vertical_align: VerticalAlign,
+}
+
+impl StatefulRescalerComponent {
+    /// Generate state of the component for particular pts value.
+    fn transition_snapshot(&self, pts: Duration) -> RescalerComponentParam {
```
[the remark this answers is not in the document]
WojciechBarczynski: I don't see any. We can leave it as it is.

WojciechBarczynski (approved):

PR #277, opened 2023-11-29
Remove ffmpeg-based RTP receiver in favor of the rust `rtp` crate

jerzywilczek:


WojciechBarczynski (approved):

src/rtp_receiver.rs
```diff
@@ -39,228 +26,120 @@ pub struct Options {
...
+            .await;
+
+            match received_bytes {
+                Some(n) => {
+                    let packet = match rtp::packet::Packet::unmarshal(&mut &buffer[..n]) {
+                        // https://datatracker.ietf.org/doc/html/rfc5761#section-4
+                        //  Given these constraints, it is RECOMMENDED to follow the guidelines
+                        //  in the RTP/AVP profile [7] for the choice of RTP payload type values,
+                        //  with the additional restriction that payload type values in the range
+                        //  64-95 MUST NOT be used.
```
wkozyra95, line 101:
```suggestion
                        // https://datatracker.ietf.org/doc/html/rfc5761#section-4
                        // Given these constraints, it is RECOMMENDED to follow the guidelines
                        // in the RTP/AVP profile [7] for the choice of RTP payload type values,
                        // with the additional restriction that payload type values in the range
                        // 64-95 MUST NOT be used.
```

src/rtp_receiver.rs
```diff
@@ -39,228 +26,120 @@ pub struct Options {
...
+                    None
+                },
+                async {
+                    let (len, _) = socket.recv_from(&mut buffer).await.unwrap();
+                    Some(len)
+                },
+            )
+            .await;
+
+            match received_bytes {
```
wkozyra95, line 94:
```suggestion
            let Some(received_bytes) = received_bytes else {
                return
            }
```

clearer intention + less indentation in the rest of the code.

technically you could do break, but I think return shows intention better

compositor_pipeline/src/pipeline/decoder.rs
```diff
@@ -48,9 +50,18 @@ impl<Input: PipelineInput> Decoder<Input> {
         let parameters = input.decoder_parameters();
 
         std::thread::spawn(move || {
+            let mut h264_depayloader = rtp::codecs::h264::H264Packet::default();
```
wkozyra95, line 53: I don't like the fact that we are using here anything from rtp package. Packets that we receive here should already in a format that does not depend on any protocol

examples/web_view.rs
```diff
@@ -152,6 +152,8 @@ fn start_example_client_code() -> Result<()> {
             "-an",
             "-c:v",
             "libx264",
+            "-profile:v",
```
wkozyra95, line 155: verify if this works without this

src/rtp_receiver.rs
```diff
@@ -39,228 +26,120 @@ pub struct Options {
...
-                decoder_params_sender
-                    .send(Err(Box::new(InitError::FfmpegError(e))))
-                    .unwrap();
-                return;
-            }
-        };
+        loop {
+            let received_bytes = future::or(
+                async {
+                    should_close_rx.recv().await.unwrap();
```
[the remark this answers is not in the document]
jerzywilczek: It only errors if the other side of the channel is closed, which wont happen

src/rtp_receiver.rs
jerzywilczek: I want to address this in the following PRs, so I'm leaving this as-is for now.

compositor_pipeline/src/pipeline/decoder.rs
jerzywilczek: I want to address this in the following PRs, so I'm leaving this as-is for now.

examples/web_view.rs
jerzywilczek: Removing this, committed it accidentally

wkozyra95 (approved):

PR #278, opened 2023-11-29
FFmpeg 6.1 support in `ffmpeg_next`

WojciechBarczynski:


wkozyra95 (changes requested):
It does not work. I'm getting errors about missing symbols even though I see there were added in fork. I suspect that nothin actually is setting `ffmpeg_6_1` feature.

wkozyra95:
Closing in favor of https://github.com/membraneframework/video_compositor/pull/281

PR #291, opened 2023-12-04
Make input  query return next output frame pts

WojciechBarczynski:


wkozyra95 (changes requested):
Please explain why you are adding that.

My guess is that you want that info to synchronise audio, but if this is the case then I think it won't work. You need to know how input PTS relates to output PTS to do that, and you are not sending either of them here. This is not input PTS, because we are normalizing the value to 0. This is not output PTS, because it would be generated in rtp sender and queue does not have access to it.

This PTS would only be useful if you are doing calculations in shaders based on PTS.

WojciechBarczynski:
The user probably wants to synchronize video with audio. Currently, we don't support audio mixing in VideoCompositor. 

So in most use cases pipeline looks like this:
Screenshot 2023-12-05 at 14 38 55

When the input stream is sent to the compositor, due to compositor frames queueing, networking latency, etc. the user can't determine the PTS of the output frame at which the input stream can start appearing on output. Therefore, it's impossible to synchronize audio and video. If the user could query this PTS mapping / offset, they would be able to perform audio and video synchronization (by properly enqueueing the Audio Mixer inputs).

Overall, I think merging both, audio and video, is such a common use case (likely most VC usage), and performing synchronization outside of the Video Compositor is such painful, that we should add audio mixing ASAP.

wkozyra95:
> If the user could query this PTS mapping / offset, they would be able to perform audio and video synchronization (by properly enqueueing the Audio Mixer inputs).

yes, but the value you are returning is not sth that would allow you to do that. You are returning the wrong PTS if that was the goal

> Overall, I think merging both, audio and video, is such a common use case (likely most VC usage), and performing synchronization outside of the Video Compositor is such painful, that we should add audio mixing ASAP.

agreed

PR #294, opened 2023-12-07
Add query `wait_for_eos`

wkozyra95:
Add a query endpoint that allows you to wait for the end of a stream (EOS)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,0 @@
-{
-  "name": "video_compositor",
-  "lockfileVersion": 3,
-  "requires": true,
-  "packages": {}
-}
--- a/src/api.rs
+++ b/src/api.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use compositor_common::scene;
 use compositor_pipeline::pipeline::{self};
 use compositor_render::{EventLoop, RegistryType};
 use crossbeam_channel::{bounded, Receiver};
@@ -49,6 +50,7 @@ pub enum UnregisterRequest {
 #[serde(tag = "query", rename_all = "snake_case")]
 pub enum QueryRequest {
     WaitForNextFrame { input_id: InputId },
+    WaitForEos { input_id: InputId },
     Inputs,
     Outputs,
 }
@@ -159,6 +161,28 @@ impl Api {
                     .collect()
                 });
                 Ok(ResponseHandler::Response(Response::Outputs { outputs }))
+            }
+            QueryRequest::WaitForEos { input_id } => {
+                let input_id: scene::id::InputId = input_id.into();
+                let Some((_, input)) = self.pipeline.inputs().find(|(id, _)| id == &&input_id)
+                else {
+                    return Err(ApiError::input_stream_not_found(
+                        &input_id.0,
+                        "Failed to subscribe to EOS event.",
+                    ));
+                };
+                let (sender, receiver) = bounded(1);
+                input
+                    .subscribe_eos_listener(Box::new(move || {
+                        sender.send(Ok(Response::Ok {})).unwrap();
+                    }))
+                    .map_err(|_err| {
+                        ApiError::input_stream_not_found(
+                            &input_id.0,
+                            "Failed to subscribe to EOS event.",
+                        )
+                    })?;
+                Ok(ResponseHandler::DeferredResponse(receiver))
             }
         }
     }
--- a/src/error.rs
+++ b/src/error.rs
@@ -36,6 +36,14 @@ impl ApiError {
             StatusCode(400),
         )
     }
+
+    pub fn input_stream_not_found(input_id: &str, msg: &str) -> Self {
+        ApiError::new(
+            "INPUT_STREAM_NOT_FOUND",
+            format!("{msg} Stream \"{input_id}\" does not exist."),
+            StatusCode(404),
+        )
+    }
 }
 
 impl<T> From<T> for ApiError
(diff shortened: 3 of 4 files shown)


WojciechBarczynski (commented):
Maybe `wait_for_next_frame` and `wait_for_eos` should return pts of the next/last composed frame? I think it might be crucial for audio synchronization. What do you think?

wkozyra95:
> Maybe `wait_for_next_frame` and `wait_for_eos` should return pts of the next/last composed frame? I think it might be crucial for audio synchronization. What do you think?

This is the same problem I already described in one of your PRs that added pts to `wait_for_next_frame`. Which PTS would you add? 
If the goal is audio synchronization then we don't need PTS, we need to have an offset between input stream PTS and PTS of all the outputs.
Also, why would we need to do audio synchronization after EOS?

WojciechBarczynski:
1. This query returns immediately after EOS is received on input. In most cases the user is most interested in receiving information, that the last frame from input has been rendered. This is not the same, as we have a buffer in the queue. Currently, this buffer is 50ms, so it only affects a couple of frames. Still, I think this buffer will be customizable in the future (e.g. in typical HLS or RTMP streaming use cases you might want to increase latency up to a few seconds to account for any networking issues). With the current approach, sending some video to VC, sending this query to wait for EOS, and updating the scene after rendering all frames is impossible.

2. What do you think about returning some form of the current VC context in all responses? I mean sth like:
```typescript
type ctx = {
    "inputs": [input],
    "outputs": [outputs],
    "last_rendered_pts": float
...
}

type input = {
    "id": string,
    "pts_offset": float
...
}

type output = {
    "id": string,
    "current_root": root
...
}
```
This would allow users to write functional VC requests handling, like:
```
fn update_scene(vc_ctx, app_state) -> outputs
```

wkozyra95:
> This query returns immediately after EOS is received on input.

At this point I'm not sure if I want to add this endpoint anymore if we don't have a specific use case for it. I think it's better to defer this decision.

> In most cases the user is most interested in receiving information, that the last frame from input has been rendered.

arguable

> Currently, this buffer is 50ms, so it only affects a couple of frames. Still, I think this buffer will be customizable in the future (e.g. in typical HLS or RTMP streaming use cases you might want to increase latency up to a few seconds to account for any networking issues).

You are assuming here that the only way to use that event is to send scene update after after that request returns. Have you considered other approaches?

> What do you think about returning some form of the current VC context in all responses? I mean sth like:

Why? What problem do you want to solve? What use cases would this enable?
Also, what are all those pts values in your example?

WojciechBarczynski:
> At this point I'm not sure if I want to add this endpoint anymore if we don't have a specific use case for it. I think it's better to defer this decision.

I think updating scenes on such events is a good use case :) I don't know if it's the best way to do it (described below). If we don't want to do it this way, then I don't know what is the purpose of adding this query. Maybe we should discuss first what problem we want to solve. 

> You are assuming here that the only way to use that event is to send scene update after after that request returns. Have you considered other approaches?
> 
> For example, if EOS is sent earlier then user would have time to send an update request that should trigger for a specific PTS, so we don't see any black frames.

I think we should support scene updates on events such as input EOS. 
I currently see two ways to tackle this using only one-way HTTP requests:
1. Sending some form of blocking request, that returns when all frames from input have been processed and sending a scene update after that.

Advantages:
- Avoids scheduling / enqueuing scene updates and all ambiguity that it brings (What happens it multiple scene updates are scheduled? How those updates should be validated? When and in which order updates will be applied? How treat component IDs and "component state inheritance"  in such cases? etc.)

Disadvantages:
- Isn't the frame perfect
- Probably not the most intuitive. Those requests probably would need to be async on the user side
- IMO harder to use as the first

2. Some sort of scene update scheduling

Advantages:
- Frame-perfect updates
- Sounds simpler to get started with

Disadvantages:
- Updates scheduling brings a lot of complexity
- A lot of edge cases, that are hard to reason about

Overall, scene update scheduling sounds better to me, since it allows frame-perfect updates, but there are plenty of edge cases that might make it really hard to reason about and ultimately implement.

> Why? What problem do you want to solve? What use cases would this enable?

I don't think it enables any additional use cases, just that writing logic handling VC scene updates is simpler that way. Membrane Plugin provides a similar context. IMO it simplifies handling scene updates. I don't know it would be helpful for other users.

> Also, what are all those pts values in your example?

I thought about the PTS of the output frame on which the corresponding frame is rendered, but since we can have multiple outputs, each potentially can start at a different time. I think this should be "composition PTS" - the time difference between now and the start of the composition

wkozyra95:
Let's maybe just talk about it offline to avoid this back and forth, but to address some of your points

> I think updating scenes on such events is a good use case :) 

It's neither an use case nor it's specific. An example of a use case would be that we need to unregister output when specific input is finished.

> then I don't know what is the purpose of adding this query.

The purpose was to support unregistering output when input stream has ended, but it's not urgent anymore so I'm reconsidering adding this query.

> I think we should support scene updates on events such as input EOS.

I don't understand that statement. I suspect based on the rest of the converstation that you are just saying that we need to have a way for a user to do sth in response to EOS, but this on its own is to generic/not defined to have any productive discussion.

>  IMO it simplifies handling scene updates. I don't know it would be helpful for other users.

"It makes it easier" does not really clarify anything. What specifically it would help with?

> I think this should be "composition PTS" - the time difference between now and the start of the composition

It depends on what you want to use it for, but I have some doubts that this information alone would be usefull for anything on its own.

WojciechBarczynski:
Ok, we can talk about this in detail offline tomorrow

PR #296, opened 2023-12-08
[docs] Replace generated docs + some style changes + update

wkozyra95:
- hide table of contents for generated pages
- change coma to semicolon in the type definitions
- update main routes page (it had old info + links pointing to wiki)
- replace Image renderer definition with custom text
- replace Shader component definition with custom text
- extract common text fragment explaining absolute positioning
- change `(default=sth` to `(**default=sth**)`. I tried to move `**` outside of `()` but one of star is cut of when genering schema.

### DIFF
--- a/docs/pages/api/routes.md
+++ b/docs/pages/api/routes.md
@@ -74,8 +74,8 @@ type RegisterInputStream = {
 Register a new RTP input stream.
 
 - `input_id` - An identifier for the input stream. It can be used in the [`InputStream`](./components/InputStream) component to render the stream content.
-- `port` - UDP port or port range that the compositor should listen for stream. Integer value between 1 and 65535 representing specific port or
-  string in the `START:END` format for a port range.
+- `port` - UDP port or port range on which the compositor should listen for the stream. An integer value between 1 and 65535 that represents a specific port
+or string in the `START:END` format for a port range.
 
 ***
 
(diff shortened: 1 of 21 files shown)


docs/pages/api/routes.md
```diff
@@ -38,78 +43,61 @@ Starts the processing pipeline. If outputs are registered and defined in the sce
...
 ```
 
-Register a new [input stream](https://github.com/membraneframework/video_compositor/wiki/Main-concepts#inputoutput-streams).
+Register a new RTP input stream.
 
-- `input_id` - Identifier that can be used in `UpdateScene` request to connect that stream to transformations or outputs.
-- `port` - UDP port that the compositor should listen for stream.
+- `input_id` - An identifier for the input stream. It can be used in the [`InputStream`](./components/InputStream) component to render the stream content.
+- `port` - UDP port or port range that the compositor should listen for stream. Integer value between 1 and 65535 representing specific port or
+  string in the `START:END` format for a port range.
```
noituri, line 78:
```suggestion
- `port` - UDP port or port range on which compositor should listen for the stream. Integer value between 1 and 65535 representing specific port or
  string in the `START:END` format for a port range.
```

noituri (approved):

WojciechBarczynski (approved):
🚀

PR #297, opened 2023-12-08
[docs] Add query documentation

wkozyra95:
- Add query documentation
- Change format of register input response

### DIFF
--- a/docs/pages/api/query.md
+++ b/docs/pages/api/query.md
@@ -1,0 +1,73 @@
+---
+description: Queries
+---
+
+# Queries
+
+### Inputs
+
+```typescript
+type Query = {
+  query: "inputs";
+}
+
+type Response = {
+  inputs: Input[];
+}
+
+type Input = {
+  id: string;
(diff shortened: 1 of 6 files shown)


docs/pages/api/query.md
```diff
@@ -0,0 +1,73 @@
...
+  outputs: Output[];
+}
+
+type Output = {
+  id: string;
+  port: u16;
+  ip: string;
+}
+
+```
```
WojciechBarczynski, line 43: I think those queries are useless. Should we have them? Users can and probably should keep registered inputs, outputs, and the current scene in their state. I think it's unwise for them to make a request every time. It's a significantly more error-prone solution IMO.

I know this is outside of the scope for this PR, but we should reconsider what types of queries are helpful and get rid of everything else, to make the whole API a little bit smaller / easier to understand.

WojciechBarczynski (approved):

noituri (approved):
👍

PR #298, opened 2023-12-11
[snapshot tests] Add test inputs with patterns

noituri:


src/snapshot_tests/test_case.rs
```diff
@@ -216,6 +217,15 @@ impl TestCaseInstance {
     }
 }
 
+#[allow(dead_code)]
+#[derive(Debug, Clone)]
+pub enum TestInputPattern {
```
wkozyra95, line 222: I don't think we need to allow configuration here, but I think it's important to apply multiple effects at the same time

src/snapshot_tests/test_case.rs
```diff
@@ -224,70 +234,126 @@ pub struct TestInput {
...
+    ) -> Self {
         let color = Self::COLOR_VARIANTS[index];
-        let r = color.0 as f32;
-        let g = color.1 as f32;
-        let b = color.2 as f32;
+        let mut y_plane = vec![0; resolution.width * resolution.height];
+        let mut u_plane = vec![0; (resolution.width * resolution.height) / 4];
+        let mut v_plane = vec![0; (resolution.width * resolution.height) / 4];
+
+        let rgb_color = |x: usize, y: usize| match pattern {
```
wkozyra95, line 303: I would suggest following approach:
- always add border, lets say 10 pixels
- divide input into grid of squares of constant size, for our tests probably 50 px border is fine. Half of the squares would be in normal color, and the other half in the alternative one. 
  - It's up to you, but I don't think you need to take into account border width when calculating grid
- generate alternative colors instead using black. It could be done by adding or subtracting sth from regular color's RGB values. This color does not need to be very distinct from original, we just need it to be visible where they are next to each other.

src/snapshot_tests/test_case.rs
```diff
@@ -224,70 +234,126 @@ pub struct TestInput {
...
+                let is_border_in_y: bool = y <= border_size
+                    || (y <= resolution.height && y >= resolution.height - border_size);
+
+                if is_border_in_x || is_border_in_y {
+                    RGBColor::BLACK
+                } else {
+                    color
+                }
+            }
+            TestInputPattern::Grid => {
```
wkozyra95, line 317: I prefer constant value instead of taking 10% of width, if you take a percent of other dimensions you don't know whether the input has a different size, or whether the component was rescalled.

src/snapshot_tests/test_case.rs
```diff
@@ -224,41 +225,41 @@ pub struct TestInput {
 }
 
 impl TestInput {
-    const COLOR_VARIANTS: [(u8, u8, u8); 17] = [
+    const COLOR_VARIANTS: [RGBColor; 17] = [
         // RED, input_0
-        (255, 0, 0),
+        RGBColor(255, 0, 0),
         // BLUE, input_1
```
wkozyra95, line 231: change blue and green with each other or, make blue brighter

wkozyra95 (approved):

PR #299, opened 2023-12-11
Send RTCP Goodbye when unregistering an input.

wkozyra95:
Send RTCP goodbye packet when RTP output stream is unregistered. This solution assumes that RTP and RTCP are multiplexed on the same port.

I tested this by unregistering an output, but to verify if it's correct we would need to test that with membrane plugin

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3137,6 +3137,7 @@ dependencies = [
  "image",
  "lazy_static",
  "log",
+ "rand",
  "reqwest",
  "rtcp",
  "rtp",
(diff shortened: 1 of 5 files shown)


wkozyra95:
> IDK if we should multiplex RTP with RTCP in a single socket/port. There's an RFC for that, but usually, RTCP is sent via a different port/socket.

Would that make it easier on the membrane side?
Do you think it's worth all the complications of allocating another port, configuring on both sides of connection which port should be used, error handling from that additional logic just to send that one message?

Also I'm not changing the behvir here, we were doing that already in ffmpeg implementation, and our rtp_receiver is handling RTCP packets.

PR #302, opened 2023-12-12
Cleanup public API for `compositor_render` crate

wkozyra95:
Clanup API + some internal refactor

- rename `renderer` module to `state` (renderer is confusing because web/shader/image are also called renderers)
- export 3 modules `web_renderer`, `shader`, and `image` for types that are needed for those renderers. Note that anything that is used in the scene definition should be inside `scene` module, so e.g. `image` mod should only have types that are strictly related to the renderer.
- combine `sync_renderer` and `renderer` into one file. Over time the logic inside the main instance was significantly thin, so I think it's fine to put those 2 fields together.
- Fix a regression where the lock was held for too long when registering a renderer.


WojciechBarczynski (approved):
LGTM

PR #308, opened 2023-12-14
Web rendering pipeline separation

WojciechBarczynski:


compositor_render/src/transformations/web_renderer/shader.rs
```diff
@@ -0,0 +1,199 @@
...
+            render_pass.set_bind_group(2, &self.sampler.bind_group, &[]);
+
+            self.wgpu_ctx
+                .plane_cache
+                .plane(input_id)
+                .unwrap()
+                .draw(&mut render_pass);
+        };
+
+        if common_params.texture_count == 0 {
```
wkozyra95, line 188: this does not really make sense for web renderer, you will never have zero textures

compositor_render/src/transformations/web_renderer/shader.rs
```diff
@@ -0,0 +1,199 @@
...
+                    push_constant_ranges: &[wgpu::PushConstantRange {
+                        stages: wgpu::ShaderStages::VERTEX_FRAGMENT,
+                        range: 0..CommonShaderParameters::push_constant_size(),
+                    }],
+                });
+
+        let pipeline = wgpu_ctx
+            .device
+            .create_render_pipeline(&wgpu::RenderPipelineDescriptor {
+                label: Some("shader transformation pipeline :^)"),
```
wkozyra95, line 67: change label

compositor_render/src/transformations/web_renderer/shader.rs
```diff
@@ -0,0 +1,199 @@
...
+        })
+    }
+
+    pub(super) fn render(
+        &self,
+        params: &wgpu::BindGroup,
+        textures: &[Option<&Texture>],
+        target: &NodeTextureState,
+        pts: Duration,
+        clear_color: Option<wgpu::Color>,
```
wkozyra95, line 117: you don't use this option

compositor_render/src/transformations/web_renderer/shader.rs
```diff
@@ -0,0 +1,199 @@
...
+            pipeline,
+            textures_bgl,
+            sampler,
+        })
+    }
+
+    pub(super) fn render(
+        &self,
+        params: &wgpu::BindGroup,
+        textures: &[Option<&Texture>],
```
wkozyra95, line 114: I would change the signature of this method to fit the web renderer, e.g. separately pass texture from web renderer and textures from children, also modify shader code.

compositor_render/src/transformations/web_renderer/shader.rs
```diff
@@ -0,0 +1,199 @@
...
+        target: &NodeTextureState,
+        pts: Duration,
+        clear_color: Option<wgpu::Color>,
+    ) {
+        let mut texture_views: Vec<&wgpu::TextureView> = textures
```
wkozyra95, line 119: make the same change as in the layout node, and pass only one texture per draw call. If you want to do that in follow-up PR it's fine, but I think there is not much point in separating that.

compositor_render/src/transformations/web_renderer/shader.rs
```diff
@@ -0,0 +1,199 @@
...
+            sampler,
+        })
+    }
+
+    pub(super) fn render(
+        &self,
+        params: &wgpu::BindGroup,
+        textures: &[Option<&Texture>],
+        target: &NodeTextureState,
+        pts: Duration,
```
wkozyra95, line 116: you don't need pts here

WojciechBarczynski:
Almost all suggestions are already applied in the follow-up PR. I will merge all those PRs to a separate branch and create the one PR with all changes bundled together.

PR #311, opened 2023-12-15
Create separate shader pipeline for each use case

WojciechBarczynski:
Changes/reasons:

General:
- remove `WgpuShader` - It wasn't configurable enough. It's more straightforward to use separate pipelines
- separate pipelines for web rendering and shaders - same as above
- remove `texture_id` from the vertex, remove `PlaneCache`, and add `plane_id` to push constant params in the shader - `texture_id` was used only in the shader. Having this in vertices requires creating buffers for each plane (as before) or creating a large buffer with many vertices and slicing it on binding. Both are less elegant than using it in push constant. This removes 128 info  logs about buffer creation on init :)
- remove `ParamsBuffer` and `U32Uniform` - Isn't useful anymore
- `CommonShaderParameters` -> `BaseShaderParameters` - They aren't common in the sense of internal implementation. For user-facing API, I don't have strong feelings about this. `Base` sounds a little bit better for me.

WebRenderer:
- simplify shader with new pipeline
- embed each frame in a separate draw call - Remove artifacts on AMD.
- texture array -> single texture, uniform texture info -> push constant - Removes limit of input textures. In shader, we still want to use an array, since the user might want to sample from multiple inputs simultaneously (e.g. mask effect).
- fixed embedding - accidentally tbh

GL rebasing after this XD


compositor_render/src/wgpu/common_pipeline.rs
```diff
@@ -12,21 +12,38 @@ pub const PRIMITIVE_STATE: wgpu::PrimitiveState = wgpu::PrimitiveState {
...
 
-pub const MAX_PLANES_COUNT: u32 = 128;
+use super::{validation::ShaderValidationError, WgpuError};
+
+pub const VERTEX_ENTRYPOINT_NAME: &str = "vs_main";
+pub const FRAGMENT_ENTRYPOINT_NAME: &str = "fs_main";
+
+pub const USER_DEFINED_BUFFER_GROUP: u32 = 1;
+pub const USER_DEFINED_BUFFER_BINDING: u32 = 0;
+
```
WojciechBarczynski, line 22: This probably shouldn't be here. I'll move it

compositor_render/src/transformations/layout/shader.rs
```diff
@@ -71,19 +68,47 @@ impl LayoutShader {
...
                 });
 
-        let pipeline = pipeline::Pipeline::create_render_pipeline(
-            &wgpu_ctx.device,
-            &pipeline_layout,
-            &shader_module,
-        );
-
-        let empty_texture = Texture::empty(wgpu_ctx);
+        let pipeline = wgpu_ctx
```
wkozyra95, line 71: Any reason why you are not using `pipeline::Pipeline::create_render_pipeline`?  I assumed this code will be the same for all pipelines?

compositor_render/src/transformations/shader/node.rs
```diff
@@ -33,6 +34,7 @@ impl ShaderNode {
             _custom_params_buffer: custom_params_buffer,
             shader,
             resolution: *resolution,
+            wgpu_ctx: ctx.wgpu_ctx.clone(),
```
wkozyra95, line 37: Why are you adding it? ctx is passed to the render method, you don't need to store it in node.

compositor_render/src/transformations/web_renderer/shader.rs
```diff
@@ -0,0 +1,157 @@
...
+                .create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
+                    label: Some("Web renderer pipeline layout"),
+                    bind_group_layouts: &[&texture_bgl, &sampler.bind_group_layout],
+                    push_constant_ranges: &[wgpu::PushConstantRange {
+                        stages: wgpu::ShaderStages::VERTEX_FRAGMENT,
+                        range: 0..RenderInfo::size(),
+                    }],
+                });
+
+        let pipeline = wgpu_ctx
```
wkozyra95, line 57: the same as for layout, couldn't this call be reused?

compositor_render/src/wgpu/ctx.rs
```diff
@@ -76,7 +77,24 @@ impl WgpuCtx {
             format,
             utils,
             shader_parameters_bind_group_layout,
-            plane_cache,
+            plane,
+            empty_texture,
+        })
+    }
+
+    fn parameters_bind_group_layout(device: &wgpu::Device) -> wgpu::BindGroupLayout {
```
wkozyra95, line 85: it does not look like good place for this function. Probably correct module, but making it asociated function to WgpuCtx is confusing

docs/pages/concept/shaders.md
```diff
@@ -24,9 +24,12 @@ The videos are represented in vertex shaders as two triangles, aligned like so:
 
 The rectangle formed by these triangles spans the whole clip space, i.e. [-1, 1] X [-1, 1].
 
-Each video passed in as input gets a separate rectangle, with the field `texture_id` describing which video this is.
+Each video passed in as input gets a separate rectangle.
+`plane_id` in `base_params` `push_constant` represents number of currently rendered plane (texture).
+
+If there are no input textures, `plane_id` is equal to -1 and a single rectangle is passed into shader. It's useful for shaders only generating something in fragment shader.
```
wkozyra95, line 30:
```suggestion
If there are no input textures, `plane_id` is equal to -1 and a single rectangle is passed to the shader. It is only useful for shaders that generate something in the fragment shader.
```

examples/silly.wgsl
```diff
@@ -19,7 +18,8 @@ fn vs_main(input: VertexInput) -> VertexOutput {
     return output;
 }
 
-struct CommonShaderParameters {
+struct BaseShaderParameters {
+    plane_id: i32,
```
wkozyra95, line 22: Did you verify that conditional sampling based on push constant works? maybe add some snapshot tests?

WojciechBarczynski: It should work, but I'll add some tests for user shaders because we don't have any reasonable ones yet (we are only verifying them only manually in examples)

compositor_render/src/transformations/shader/node.rs
WojciechBarczynski: I first added it when prototyping and forgot to remove it later. Good catch

examples/silly.wgsl
WojciechBarczynski: I think though, that adding shader rendering tests is outside of the scope of this PR. I'll do it in the next one.

wkozyra95 (approved):

examples/silly.wgsl
wkozyra95: not that by modifying a struct, some alignment might have change and e.g. silly shader does not use any of those filelds, so even if data there is wrong we wouldn't know just based on that

PR #313, opened 2023-12-15
Add animations to `Tiles` component

wkozyra95:
https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/23

https://github.com/membraneframework/video_compositor/assets/9753141/546f257e-2764-448b-bc79-536abee0cb03

### DIFF
--- a/compositor_render/src/transformations/layout.rs
+++ b/compositor_render/src/transformations/layout.rs
@@ -161,3 +161,24 @@ impl LayoutNode {
             .render(ctx.wgpu_ctx, self.params.bind_group(), &textures, target);
     }
 }
+
+impl NestedLayout {
+    /// NestedLayout that can won't ever be rendered. It's intended to optimized out
+    /// in the flattening process. Its only propose is to keep track of child nodes that are not
+    /// currently used, so the index offset can be calculated correctly.
+    pub(crate) fn child_nodes_placeholder(child_nodes_count: usize) -> Self {
+        Self {
+            top: 0.0,
+            left: 0.0,
+            width: 0.0,
+            height: 0.0,
+            rotation_degrees: 0.0,
+            scale_x: 1.0,
+            scale_y: 1.0,
+            crop: None,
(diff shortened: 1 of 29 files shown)


examples/tiles.rs
```diff
@@ -0,0 +1,170 @@
...
+            "outputs": [
+                {
+                    "output_id": "output_1",
+                    "root": scene_with_inputs(i),
+                }
+            ]
+        }))?;
+
+        thread::sleep(Duration::from_secs(1));
+    }
```
WojciechBarczynski, line 168: NIT last scene could have more then one tile, since this is `tiles` example :)

WojciechBarczynski (approved):

compositor_render/src/transformations/layout.rs
```diff
@@ -161,3 +161,24 @@ impl LayoutNode {
             .render(ctx.wgpu_ctx, self.params.bind_group(), &textures, target);
     }
 }
+
+impl NestedLayout {
+    /// NestedLayout that can won't ever be rendered. It's intended to optimized out
+    /// in the flattening process. Its only propose is to keep track of child nodes that are not
+    /// currently used, so the index offset can be calculated correctly.
```
noituri, line 168:
```suggestion
    /// NestedLayout that won't ever be rendered. It's intended to be optimized out
    /// in the flattening process. Its only purpose is to keep track of child nodes that are not
    /// currently used so that the index offset can be calculated correctly.
```

examples/tiles.rs
```diff
@@ -0,0 +1,170 @@
...
+    }))?;
+
+    let shader_source = include_str!("./silly.wgsl");
+    info!("[example] Register shader transform");
+    common::post(&json!({
+        "type": "register",
+        "entity_type": "shader",
+        "shader_id": "example_shader",
+        "source": shader_source,
+    }))?;
```
noituri, line 90: It's not used anywhere

PR #319, opened 2023-12-18
Improve latency when reading from UDP buffer

wkozyra95:
Fixes an issue with lost frames when the UDP buffer is small.

The main fix is to change the channel of size 1 to an unbound channel. Decoding was a bottleneck for reading. Not sure if adding `Arc` helps much.


noituri (approved):

WojciechBarczynski (commented):
LGTM

PR #321, opened 2023-12-19
Fix register json schema

WojciechBarczynski:


noituri (approved):

wkozyra95:
I don't think this is correct, I suspect only deny_unknown_fields on image was necessary

PR #325, opened 2023-12-20
Add base params shader tests, fix memory layout for base params

WojciechBarczynski:
https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/24

I'll add the user params test in the next PR


src/snapshot_tests/tests.rs
```diff
@@ -18,10 +20,143 @@ pub fn snapshot_tests() -> Vec<TestCase> {
...
+        })
+        .to_string();
+        static ref OUTPUT_RESOLUTION_SHADER: String = json!({
+            "type": "register",
+            "entity_type": "shader",
+            "shader_id": "base_params_output_resolution",
+            "source": include_str!("../../snapshot_tests/shader/base_params_output_resolution.wgsl")
+        })
+        .to_string();
+    };
```
WojciechBarczynski, line 63: I don't like using static ref here, but the only safe option I know, that would fix this and satisfy borrow checker rules requires changing the typing of the `renderers` field in `TestCase` to `String` and calling clone everywhere. I think this is better

wkozyra95: box leak

snapshot_tests/shader/base_params_output_resolution.scene.json
```diff
@@ -0,0 +1,12 @@
...
+    "root": {
+        "type": "shader",
+        "shader_id": "base_params_output_resolution",
+        "resolution": {
+            "width": 1920,
+            "height": 1080
+        },
+        "children": []
+    }
+}
```
wkozyra95, line 12: newline

snapshot_tests/shader/base_params_plane_id_5_inputs.scene.json
```diff
@@ -0,0 +1,33 @@
...
+                "type": "input_stream",
+                "input_id": "input_4"
+            },
+            {
+                "type": "input_stream",
+                "input_id": "input_5"
+            }
+        ]
+    }
+}
```
wkozyra95, line 33: newline

I think there is  an option in vscode to avoid creating files like this

snapshot_tests/shader/base_params_time.scene.json
```diff
@@ -0,0 +1,17 @@
...
+            "height": 360
+        },
+        "children": [
+            {
+                "type": "input_stream",
+                "input_id": "input_1"
+            }
+        ]
+    }
+}
```
wkozyra95, line 17: newline

src/snapshot_tests/tests.rs
```diff
@@ -18,10 +20,143 @@ pub fn snapshot_tests() -> Vec<TestCase> {
     tests.append(&mut text_snapshot_tests());
     tests.append(&mut tiles_snapshot_tests());
     tests.append(&mut rescaler_snapshot_tests());
+    tests.append(&mut shader_tests());
     tests
 }
 
-pub fn rescaler_snapshot_tests() -> Vec<TestCase> {
+fn shader_tests() -> Vec<TestCase> {
```
wkozyra95, line 27:
```suggestion
fn shader_snapshot_tests() -> Vec<TestCase> {
```

snapshot_tests/shader/base_params_output_resolution.wgsl
```diff
@@ -0,0 +1,40 @@
+struct VertexInput {
```
wkozyra95, line 1: I would prefer if those shaders were named base on what they are doing. If you want to mention in the name that it relies on output_resolution then it's fine to mention it.

snapshot_tests/shader/base_params_output_resolution.wgsl
```diff
@@ -0,0 +1,40 @@
...
+var<push_constant> base_params: BaseShaderParameters;
+
+@vertex
+fn vs_main(input: VertexInput) -> VertexOutput {
+    var output: VertexOutput;
+
+    let scale_factor_x = f32(base_params.output_resolution.x) / 4000.0;
+    let scale_factor_y = f32(base_params.output_resolution.y) / 2000.0;
+
+    output.position = vec4(input.position.x * scale_factor_x, input.position.y * scale_factor_y, input.position.z, 1.0);
```
wkozyra95, line 30: that feels very synthetic, I'm not sure based on this code what would be correct output

src/snapshot_tests/tests.rs
```diff
@@ -18,10 +20,143 @@ pub fn snapshot_tests() -> Vec<TestCase> {
...
+                input1.clone(),
+                input2.clone(),
+                input3.clone(),
+                input4.clone(),
+                input5.clone(),
+            ],
+            ..Default::default()
+        },
+        TestCase {
+            name: "shader/base_params_time_5s",
```
wkozyra95, line 96: remove that 5s, it's confusing, if you want you can add here in a comment or in wgsl code comment that entire animation takes 5 seconds

wkozyra95 (approved):

src/snapshot_tests/tests.rs
WojciechBarczynski: "Memory leakage is considered safe" :P

.vscode/settings.json
```diff
@@ -1,12 +1,17 @@
...
+            "fileMatch": [
+                "*.register.json"
+            ],
             "url": "./schemas/register.schema.json"
         }
-    ]
-}
+    ],
+    "files.insertFinalNewline": true
+}
```
wkozyra95, line 17: Interesting, adding a setting that ensures final newline did not add final new line

rah3of6xp3531

wkozyra95: either way, no strong opinion, but I don't think this should be here. It should be in your personal config not the project one

PR #326, opened 2023-12-20
Add test names uniqueness validation

WojciechBarczynski:


src/snapshot_tests.rs
```diff
@@ -21,7 +21,15 @@ fn test_snapshots() {
         .into_iter()
         .map(TestCaseInstance::new)
         .collect();
+    let mut test_names = HashSet::new();
```
wkozyra95, line 24: I wouldn't mix that code with actual test run. Just extract that code to separate function and iterate over that collection twice.

wkozyra95 (approved):

PR #330, opened 2024-01-02
Add location in parse shader error message

WojciechBarczynski:
Before:
```console
Error stack:
     - Failed to register shader "shader_example_1".
     - Shader parse error: expected ';', found 'return'
```

After:
```console
Error stack:
     - Failed to register shader "shader_example_1".
     - Shader parsing error in line 18 column 5.
     - expected ';', found 'return'
```

The last line is not formatted according to the convention, due to using `naga::front::wgsl::ParseError` as a source (the error message comes directly from naga). I don't want to manually change it, because this might make errors less readable (e.g. first word is case sensitive variable name, the dot at the end interferes with the provided syntax error making it misleading, etc.)

### DIFF
--- a/compositor_render/src/transformations/shader/validation/error.rs
+++ b/compositor_render/src/transformations/shader/validation/error.rs
@@ -27,10 +27,11 @@ impl ShaderParseError {
 impl Display for ShaderParseError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self.parse_error.location(&self.source) {
-            Some(location) => f.write_str(&format!(
+            Some(location) => write!(
+                f,
                 "Shader parsing error in line {} column {}.",
                 location.line_number, location.line_position
-            )),
+            ),
             None => f.write_str("Shader parsing error."),
         }
     }
--- a/compositor_render/src/transformations/shader/validation/error.rs
+++ b/compositor_render/src/transformations/shader/validation/error.rs
@@ -1,3 +1,7 @@
+use std::fmt::Display;
+
+use naga::SourceLocation;
+
 use crate::{
     transformations::shader::pipeline::{USER_DEFINED_BUFFER_BINDING, USER_DEFINED_BUFFER_GROUP},
     wgpu::common_pipeline::VERTEX_ENTRYPOINT_NAME,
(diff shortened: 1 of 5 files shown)


compositor_render/src/transformations/shader/validation/error.rs
```diff
@@ -6,6 +10,34 @@ use crate::{
...
 
+#[derive(Debug, thiserror::Error)]
+pub struct ShaderParseError {
+    parse_error: naga::front::wgsl::ParseError,
+    location: Option<SourceLocation>,
+}
+
+impl ShaderParseError {
+    pub fn new(parse_error: naga::front::wgsl::ParseError, source: &str) -> Self {
+        let location = parse_error.location(source);
```
jerzywilczek, line 21: I think it would be better to move this line to the `Display` implementation. Keeping this value in the `ShaderParseError` makes us have two sources of truth.

compositor_render/src/transformations/shader/validation/error.rs
```diff
@@ -6,6 +10,34 @@ use crate::{
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self.location {
+            Some(location) => f.write_str(&format!(
+                "Shader parse error in line {} column {}: {}.",
+                location.line_number, location.line_position, self.parse_error
+            )),
+            None => f.write_str(&format!("Shader parse error: {}. ", self.parse_error)),
+        }
+    }
+}
```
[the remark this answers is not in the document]
jerzywilczek: I think it's fine to keep it here.

jerzywilczek (approved):
Didn't notice `naga` had the `location` method when I was implementing this. Very cool.

noituri (approved):

compositor_render/src/transformations/shader/validation/error.rs
```diff
@@ -6,6 +8,34 @@ use crate::{
...
+
+impl Display for ShaderParseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self.parse_error.location(&self.source) {
+            Some(location) => f.write_str(&format!(
+                "Shader parsing error in line {} column {}.",
+                location.line_number, location.line_position
+            )),
```
wkozyra95, line 33:
```suggestion
            Some(location) => write!(
                f, "Shader parsing error in line {} column {}.",
                location.line_number, location.line_position
            ),
```

wkozyra95 (approved):

PR #334, opened 2024-01-03
Audio API spike

WojciechBarczynski:
The `simple` example is modified to showcase the new API.

This isn't meant to be merged. It'll be gradually implemented in other audio support PRs.

### DIFF
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -74,21 +74,11 @@ fn start_example_client_code() -> Result<()> {
                 "width": VIDEO_RESOLUTION.width,
                 "height": VIDEO_RESOLUTION.height,
             },
-            "encoder_settings": {
-                "preset": "medium"
-            }
-            // Not necessary
-            // "stream": {
-            //     "rtp_payload_type": 96,
-            //     "clock_rate": 90_000
-            // }
+            "encoder_preset": "medium",
         },
         "audio": {
-            "sample_rate": 44_100
-            // Not necessary
-            // "stream": {
-            //     "clock_rate": 44_000
-            // }
+            "sample_rate": 44_100,
+            "rtp_clock_rate": 48_000,
         }
     }))?;
 
@@ -100,18 +90,11 @@ fn start_example_client_code() -> Result<()> {
         "port": 8004,
         "video": {
             "codec": "h264",
-            // Not necessary
-            // "stream": {
-            //     "rtp_payload_type": 96,
-            //     "clock_rate": 90_000
-            // }
         },
         "audio": {
             "codec": "aac",
-            // Not necessary
-            // "stream": {
-            //     "clock_rate": 44_000
-            // }
+            "sample_rate": 44_100,
+            "rtp_clock_rate": 48_000
         }
     }))?;
 
@@ -142,8 +125,8 @@ fn start_example_client_code() -> Result<()> {
                 ],
                 "resolution": { "width": VIDEO_RESOLUTION.width, "height": VIDEO_RESOLUTION.height },
             },
-            "track": {
-                "children": [{
+            "audio": {
+                "inputs": [{
                     "input_id": "input_1"
                     // some other options per input potentially
                 }],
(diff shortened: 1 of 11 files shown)


src/api.rs
```diff
@@ -25,21 +28,24 @@ pub enum Request {
...
     Query(QueryRequest),
     Start,
 }
 
 #[derive(Serialize, Deserialize, JsonSchema)]
-pub struct UpdateScene {
-    pub outputs: Vec<types::OutputScene>,
+pub struct UpdateCompositionRequest {
+    pub video_outputs: Vec<types::VideoCompositionParams>,
+    pub audio_outputs: Vec<types::AudioMixParams>,
```
WojciechBarczynski, line 39: Any suggestions for those names?

src/types/track_properties.rs
```diff
@@ -0,0 +1,12 @@
...
+
+use super::AudioInputId;
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub struct TrackProperties {
+    input_id: AudioInputId,
+    // Temporal
+    volume: Option<u32>,
+    // ...
+}
```
WojciechBarczynski, line 12: In the first version, I don't want to use any params. All audio tracks would be mixed in the simplest possible way

examples/simple.rs
```diff
@@ -65,25 +65,37 @@ fn start_example_client_code() -> Result<()> {
...
         "input_id": "input_1",
-        "port": 8004
+        "port": 8004,
+        "video": {
+            "codec": "h264",
+        },
+        "audio": {
+            "codec": "aac",
+            "sample_rate": 44_100,
+            "rtp_clock_rate": 48_000
```
jerzywilczek, line 97: cant find the relevant fragment of an rfc, but is having two values here necessary?

src/api/register_request.rs
```diff
@@ -44,11 +42,16 @@ fn register_output(api: &mut Api, request: RegisterOutputRequest) -> Result<(),
...
-        resolution,
-        encoder_settings,
+        video,
         ip,
+        ..
     } = request;
 
+    let Some(video) = video else {
+        // TODO error
+        return Ok(());
```
jerzywilczek, line 52: :smiling_face_with_three_hearts:

jerzywilczek (approved):
:rocket:

src/api/register_request.rs
WojciechBarczynski: Yep, it's just a spike and it's not meant to be merged anyway.
I don't want to spend time fixing implementation when it's not going to be used anyway.
This PR is mostly for API design

examples/simple.rs
WojciechBarczynski: `sample_rate` describes audio params
`rtp_clock_rate` describes is used for RTP timestamps

As I understand, I need `rtp_clock_rate` for correct RTP packet depayloading and sample rate for decoding actual audio (rtp packet payload).

WojciechBarczynski:
Update: 
1. Payload type ranges for our audio and video formats are dynamically defined and their ranges CAN overlap.

When I used ffmpeg to stream audio via RTP, the payload type specified in SDP was 97 (same for opus and aac). For h264 it was 96. We can't however assume that's always the case, because RFC specifies that they can be dynamically defined.

I think we have to require providing `video: {.., payload_type: number}, audio: {.., payload_type: number}` if both `audio` and `video` values are provided.

2. I've ignored audio channels in this API. I think we need it for decoding audio, however, RFC states, that values provided in SDP can't be treated as a source of truth and audio can change from mono to stereo dynamically. We don't process SDP, so I think it's fine to require providing channels count and treat changing it dynamically as a user error.

src/types/register_request.rs
```diff
@@ -36,33 +67,45 @@ pub struct RegisterOutputRequest {
...
+    pub video: Option<OutputVideoOptions>,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
-pub struct EncoderSettings {
-    preset: Option<EncoderPreset>,
+pub struct OutputAudioOptions {
+    sample_rate: u32,
+    // Default aac
+    codec: Option<AudioCodec>,
```
noituri, line 78: Just curious, is there any reason to make aac the default codec over opus?

noituri (approved):

PR #336, opened 2024-01-04
Switch from `env_logger` to `tracing`

wkozyra95:
- Use tracing library
- Default to json logger when not building from source (based on CARGO_MANIFEST_DIR env)
- Add env `FFMPEG_LOG_LEVEL` that controll ffmpeg logs (default to warn)

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -25,7 +25,10 @@ extern "C" fn ffmpeg_log_callback(
     arg1: *mut libc::c_void,
     log_level: libc::c_int,
     fmt: *const libc::c_char,
+    #[cfg(not(all(target_os = "macos", target_arch = "aarch64")))]
     va_list_tag: *mut ffmpeg_next::sys::__va_list_tag,
+    #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+    va_list_tag: ffmpeg_next::sys::va_list,
 ) {
     unsafe {
         match ffmpeg_log_level() {
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -1,0 +1,58 @@
+use std::{env, sync::OnceLock};
+
+#[derive(Debug, Clone, Copy)]
+enum FfmpegLogLevel {
+    Error,
+    Warn,
+    Info,
+    Debug,
+}
+
+fn ffmpeg_log_level() -> FfmpegLogLevel {
+    static LOG_LEVEL: OnceLock<FfmpegLogLevel> = OnceLock::new();
+
+    *LOG_LEVEL.get_or_init(|| match env::var("FFMPEG_LOG_LEVEL") {
+        Ok(debug) if debug == "debug" => FfmpegLogLevel::Debug,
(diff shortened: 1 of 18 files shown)


noituri (commented):
It doesn't compile on MacOS
image

src/logger.rs
```diff
@@ -0,0 +1,58 @@
...
+            FfmpegLogLevel::Debug if log_level <= 48 => {
+                ffmpeg_next::sys::av_log_default_callback(arg1, log_level, fmt, va_list_tag)
+            }
+            _ => (),
+        }
+    }
+}
+
+pub fn init_logger() {
+    if env::var("CARGO_MANIFEST_DIR").is_ok() {
```
jerzywilczek, line 50: Shouldn't we also have a manual switch for this?

jerzywilczek (approved):
:rocket:

src/logger.rs
wkozyra95: I'll be reworking env based configuration in follow-up PR, so I can address it then.

src/logger.rs
```diff
@@ -0,0 +1,61 @@
...
+            }
+            FfmpegLogLevel::Info if log_level <= 32 => {
+                ffmpeg_next::sys::av_log_default_callback(arg1, log_level, fmt, va_list_tag)
+            }
+            FfmpegLogLevel::Debug if log_level <= 48 => {
+                ffmpeg_next::sys::av_log_default_callback(arg1, log_level, fmt, va_list_tag)
+            }
+            _ => (),
+        }
+    }
```
WojciechBarczynski, line 49: I had a hard time understanding, that `log_level` refers to value received from FFmpeg, while result of `ffmpeg_log_level` is value set by user. Changing `ffmpeg_log_level` to `ffmpeg_logging_level` would solve that.

src/logger.rs
```diff
@@ -0,0 +1,61 @@
...
+    static LOG_LEVEL: OnceLock<FfmpegLogLevel> = OnceLock::new();
+
+    *LOG_LEVEL.get_or_init(|| match env::var("FFMPEG_LOG_LEVEL") {
+        Ok(debug) if debug == "debug" => FfmpegLogLevel::Debug,
+        Ok(info) if info == "info" => FfmpegLogLevel::Info,
+        Ok(warn) if warn == "warn" => FfmpegLogLevel::Warn,
+        Ok(error) if error == "error" => FfmpegLogLevel::Error,
+        Ok(_) => FfmpegLogLevel::Warn,
+        Err(_) => FfmpegLogLevel::Warn,
+    })
```
WojciechBarczynski, line 21: Do we want to use env vars for such configurations?
We have more and more things configurable by env vars in similar way. Isn't it better to use some config file instead?

wkozyra95: I addressed that in https://github.com/membraneframework/video_compositor/pull/342 description

WojciechBarczynski: That explanation convinced me 👍

PR #342, opened 2024-01-05
Switch to env based configuration + remove `"init"` request

wkozyra95:
Switch to env base config

### Why envs

Why not init request:
- some values need to be set before e.g. API PORT, so it requires using both methods of configuration
- it's very non-standard config method, even if it was the best option it could be perceived as hack.

Why not config files:
- hard to work with in prod environment, to add config inside k8s or any other container-based platform you would need to: create some volume entity from that file, attach/mount that volume to the container, and pass some env that defines the path where compositor should look for the file. 
- hard to work with in a dev environment. To tests some other configuration you need to modify a file. If we want to persist the config we can always use `.env` file.

### DIFF
--- a/docs/pages/deployment/configuration.md
+++ b/docs/pages/deployment/configuration.md
@@ -48,5 +48,5 @@ Enable web rendering capabilities. With this option disabled, you can not use [`
 
 ### `LIVE_COMPOSITOR_WEB_RENDERER_GPU_ENABLE`
 
-Disable GPU support inside the embedded Chromium instance.
+Enable GPU support inside the embedded Chromium instance.
 
--- a/docs/pages/deployment/configuration.md
+++ b/docs/pages/deployment/configuration.md
@@ -1,0 +1,52 @@
+# Configuration
+
+## Environment variables
+
+### `LIVE_COMPOSITOR_API_PORT`
+
+API port. Defaults to 8001.
+
+### `LIVE_COMPOSITOR_OUTPUT_FRAMERATE`
+
+Output framerate for all output streams. This value can be a number or string in the `NUM/DEN` format , where both `NUM` and `DEN` are unsigned integers.
+
+### `LIVE_COMPOSITOR_STREAM_FALLBACK_TIMEOUT_MS`
+
+A timeout that defines when the compositor should switch to fallback on the input stream that stopped sending frames.
+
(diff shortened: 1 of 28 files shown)


docs/pages/deployment/configuration.md
```diff
@@ -0,0 +1,52 @@
...
+
+See `-loglevel` option in [FFmpeg documentation](https://ffmpeg.org/ffmpeg.html).
+
+### `LIVE_COMPOSITOR_WEB_RENDERER_ENABLE`
+
+Enable web rendering capabilities. With this option disabled, you can not use [`WebView` components](../api/components/WebView) or register [`WebRenderer` instances](../api/renderers/web).
+
+### `LIVE_COMPOSITOR_WEB_RENDERER_GPU_ENABLE`
+
+Disable GPU support inside the embedded Chromium instance.
```
noituri, line 51:
```suggestion
Enable GPU support inside the embedded Chromium instance.
```

The env's name contains `ENABLE` so I think it makes sense to also reflect that in the description

WojciechBarczynski (approved):
LGTM

docs/pages/deployment/configuration.md
```diff
@@ -0,0 +1,52 @@
...
+### `LIVE_COMPOSITOR_LOGGER_FORMAT`
+
+Logger format. Supported options:
+- `json`
+- `compact`
+- `pretty`
+
+:::warning
+This option does not apply to logs produced by `FFmpeg` or the embedded Chromium instance used for web rendering.
+:::
```
jerzywilczek, line 32: what does `:::` do?

jerzywilczek (approved):
:rocket:

docs/pages/deployment/configuration.md
noituri: image

^ This

noituri (approved):

docs/pages/deployment/configuration.md
WojciechBarczynski, line 1: I think all those env vars should have `required` / `optional` next to them (with required being the first ones). For `optional` values provide defaults.

WojciechBarczynski: You can also do two separate sections for that

WojciechBarczynski: I'm pretty sure that output framerate is (or at least should be) required

PR #344, opened 2024-01-08
Add Dockerfile without web renderer

wkozyra95:


.github/workflows/docker.yml
```diff
@@ -25,7 +25,11 @@ jobs:
...
             - uses: hadolint/hadolint-action@v3.1.0
               with:
-                dockerfile: Dockerfile
+                dockerfile: build-tools/docker/slim.Dockerfile
+
+            - uses: hadolint/hadolint-action@v3.1.0
+              with:
+                dockerfile: build-tools/docker/full.Dockerfile
 
             - name: Build image
```
wkozyra95, line 34: I'm building only slim here, I'm wondering whether it make sense to build both. Any opinions?

build-tools/docker/slim.Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+# Runtime image
+FROM ubuntu:mantic-20231011
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+ARG USERNAME=compositor
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV NVIDIA_DRIVER_CAPABILITIES compute,graphics,utility
```
WojciechBarczynski, line 34: Shouldn't this be?
```suggestion
ENV NVIDIA_DRIVER_CAPABILITIES=compute,graphics,utility
```

wkozyra95: this was copy paste from original dockerfile
as far as I remember both syntax are valid, but yest we should be consitent

build-tools/docker/slim.Dockerfile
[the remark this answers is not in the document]
wkozyra95: > Idk if slim is self-explanatory

You are assuming here that slim image means that it's compositor without web renderer :D

It depends what this file is suppose to mean.
- If it is intended to be smaller image with minimal set of feature that has not any significant downsides (e.g. large size, security issues) then this name is self explanatory and consistent with existing naming conventions
- If it is intended to be full compositor with just web renderer disabled then yes name is not self-explantory

I decided to go with slim because from those 2 approaches it seem to me slightly better, but I don't feel strongly about it.

.github/workflows/docker.yml
jerzywilczek: I don't really know. If we're not using too much of the CI time then maybe we should build both versions, but I don't know how to check these limits/the spending

jerzywilczek (approved):

.github/workflows/docker.yml
wkozyra95: I don't think there are limits for public repositories. My concern was about witing for CI, but this job is not always run, so maybe it make sense to run both

PR #349, opened 2024-01-10
Refactor the output architecture.

### LINKED ISSUE #154: Decouple rtp parsing and h264 decoding in RtpSender/RtpReceiver
encoder should be a part of a pipeline crate and RTP handling should stay in top level compositor crate

jerzywilczek:
This patch changes the output architecture in a very similar way to how the input architecture was changed.

This simplifies things a bunch and makes adding additional encoders and outputs simpler.

Same as the inputs refactor, this does not address the problem of having multiple encoders per output (e.g. audio).

I think this closes #154


compositor_pipeline/Cargo.toml
```diff
@@ -16,4 +16,5 @@ rtp = { workspace = true }
 webrtc-util = "0.8.0"
 socket2 = "0.5.5"
 rtcp = "0.10.0"
+rand = "0.8.5"
```
noituri, line 19: I'm not sure what is considered a good practice in this case but `rand` is used in 2 crates so maybe consider making it a workspace dependency.

compositor_pipeline/src/pipeline.rs
```diff
@@ -137,19 +126,26 @@ impl<Output: PipelineOutput> Pipeline<Output> {
...
-        output_opts: OutputOptions<Output>,
+        encoder_opts: EncoderOptions,
+        output_opts: OutputOptions,
     ) -> Result<(), RegisterOutputError> {
         if self.outputs.contains_key(&output_id) {
             return Err(RegisterOutputError::AlreadyRegistered(output_id));
         }
 
-        if output_opts.resolution.height % 2 != 0 || output_opts.resolution.width % 2 != 0 {
+        let EncoderOptions::H264(ref opts) = encoder_opts;
```
noituri, line 136: I thought enums always required `else` branch even if there was only 1 variant. That's cool

compositor_pipeline/src/pipeline/encoder/ffmpeg_h264.rs
```diff
@@ -0,0 +1,295 @@
...
+        let (result_sender, result_receiver) = crossbeam_channel::bounded(0);
+
+        let options2 = options.clone();
+
+        let encoder_thread = std::thread::Builder::new()
+            .name(format!("Encoder thread for output {}", options.output_id))
+            .spawn(move || {
+                match Self::encoder_thread(options2, frame_receiver, packet_sender, &result_sender)
+                {
+                    Ok(_) => todo!(),
```
noituri, line 111: Is this `todo!` supposed to be here?

noituri (approved):
LGTM

compositor_pipeline/src/pipeline/output/rtp.rs
```diff
@@ -116,3 +122,13 @@ impl PipelineOutput for RtpSender {
...
     }
 }
+
+impl Drop for RtpSender {
+    fn drop(&mut self) {
+        match self.sender_thread.take() {
+            Some(handle) => handle.join().unwrap(),
+            None => error!("RTP sender thread was already joined."),
+        }
+        println!("rtp cleaned up nicely")
```
wkozyra95, line 132:
```suggestion
        debug!("RtpSender cleanup finished.")
```
or info

compositor_pipeline/src/pipeline/encoder/ffmpeg_h264.rs
```diff
@@ -0,0 +1,295 @@
...
+                options.resolution.height as u32,
+            );
+
+            frame_into_av(frame, &mut av_frame);
+
+            if let Err(e) = encoder.send_frame(&av_frame) {
+                error!("Encoder error: {e}.")
+            }
+
+            let mut packet = Packet::empty();
```
wkozyra95, line 219: If it is safe to reuse this packet object than maybe move it outside of the loop, and if it's no safe than there should be separate packet instance per `receive_packet` call

compositor_pipeline/src/pipeline/encoder/ffmpeg_h264.rs
```diff
@@ -0,0 +1,295 @@
...
+
+            while encoder.receive_packet(&mut packet).is_ok() {
```
wkozyra95, line 221: what about the errors from receive_packet? Even if we are not doing anything about it we should at least log them

compositor_pipeline/src/pipeline/encoder/ffmpeg_h264.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+}
+
+impl Drop for LibavH264Encoder {
+    fn drop(&mut self) {
+        self.frame_sender.send(Message::Stop).unwrap();
+        match self.encoder_thread.take() {
+            Some(handle) => handle.join().unwrap(),
+            None => error!(
+                "[output {}] Encoder thread was already joined. Something very bad has happened.",
```
wkozyra95, line 244: >  Something very bad has happened.

That does not look good in an error message. In cases where some condition is clearlly a bug use phrases like
- This should not/never happen.
- Internal compositor error
- This is a bug, please report it. (I'm not fan of this one, but saw it in some places)

compositor_pipeline/src/pipeline/encoder/ffmpeg_h264.rs
```diff
@@ -0,0 +1,295 @@
...
+                Pixel::YUV420P,
+                options.resolution.width as u32,
+                options.resolution.height as u32,
+            );
+
+            frame_into_av(frame, &mut av_frame);
+
+            if let Err(e) = encoder.send_frame(&av_frame) {
+                error!("Encoder error: {e}.")
+            }
```
wkozyra95, line 217: Inside frame_into_av and when calling encoder.send_frame you are login in error, but do not change any control flow in response. The resulting behavior is not obvious.

If `frame_into_av` fails you will send empty frame (either green or random if initial frame does not zero its memory)

If `send_frame` fails you will log error and receive_packet will not return anything, so behvior is correct, but I think it would be cleaner to handle those errors explicitlly

compositor_pipeline/src/pipeline/encoder/ffmpeg_h264.rs
```diff
@@ -0,0 +1,295 @@
...
+
+impl LibavH264Encoder {
+    pub fn new(
+        options: Options,
+    ) -> Result<(Self, Box<dyn Iterator<Item = EncodedChunk> + Send>), EncoderInitError> {
+        let (frame_sender, frame_receiver) = crossbeam_channel::unbounded();
+        let (packet_sender, packet_receiver) = crossbeam_channel::unbounded();
+        let (result_sender, result_receiver) = crossbeam_channel::bounded(0);
+
+        let options2 = options.clone();
```
wkozyra95, line 104:
```suggestion
        let options_clone = options.clone();
```
or `_for_thread`

compositor_pipeline/src/pipeline/encoder/ffmpeg_h264.rs
```diff
@@ -0,0 +1,321 @@
...
+        options: Options,
+        frame_receiver: Receiver<Message>,
+        packet_sender: Sender<EncodedChunk>,
+        result_sender: &Sender<Result<(), EncoderInitError>>,
+    ) -> Result<(), EncoderInitError> {
+        let codec = ffmpeg_next::codec::encoder::find(Id::H264).ok_or(EncoderInitError::NoCodec)?;
+
+        let mut encoder = Context::new().encoder().video()?;
+
+        let pts_unit_secs = Rational::new(1, 90000);
```
WojciechBarczynski, line 157: Where does this 90000 come from? I'm pretty sure, that this is the RTP clock rate.

Using `i64` as pts in `EncodedChunk` breaks the separation between input-decoder and encoder-output. Converting from/to duration depends on the RTP clock rate, which not only won't have any sense when other input/output sources will be added, but also is completely different for other RTP payload types (yeah, for video 90kHz is used most frequently, but audio can use 48kHz).

jerzywilczek: I thought about this problem and decided this way of solving it was a bit better, but looking at it today I think I have a different opinion now. Maybe we can talk about this live?

WojciechBarczynski: Sure

PR #350, opened 2024-01-12
Fix interpolation abstraction for non linear transitions

wkozyra95:
- Remove duplicated code around transitions (and recover old transitions)
- Add type that represents interpolation type
- Fix calculation when recovering the transition state from the previous state. Old implementation worked only for linear transitions.

I started implementing transition based on cubic bezier curve, but it appears to be more complicated than I expected so I left it for now with just linear.

### DIFF
--- a/compositor_render/src/scene/transition.rs
+++ b/compositor_render/src/scene/transition.rs
@@ -31,7 +31,7 @@ pub(super) struct TransitionOptions {
 }
 
 impl TransitionState {
-    pub fn from_previous_transition(
+    pub fn new(
         current_transition: Option<TransitionOptions>,
         previous_transition: Option<TransitionState>,
         last_pts: Duration,
--- a/compositor_render/src/scene/transition.rs
+++ b/compositor_render/src/scene/transition.rs
@@ -1,0 +1,96 @@
+use std::time::Duration;
+
+use super::{types::interpolation::InterpolationState, InterpolationKind};
+
+/// Similar concept to InterpolationState, but it represents a time instead.
+/// Values between 0 and 1 represent transition and larger than 1 post transition.
+///
+/// If interpolation_kind is linear then InterpolationState and TransitionProgress
+/// have the same numerical value.
+#[derive(Debug, Clone, Copy)]
+struct TransitionProgress(f64);
+
+#[derive(Debug, Clone)]
+pub(super) struct TransitionState {
+    /// Additional offset for transition. It is non zero if you want to start
+    /// a transition in the middle of the interpolation curve.
+    initial_offset: (TransitionProgress, InterpolationState),
+
+    // PTS of a first frame of transition.
+    start_pts: Duration,
+
+    /// Duration of the transition.
+    duration: Duration,
+
+    interpolation_kind: InterpolationKind,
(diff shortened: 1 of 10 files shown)


compositor_render/src/scene/transition.rs
```diff
@@ -0,0 +1,96 @@
...
+            (Some(current_transition), _) => Some(Self {
+                initial_offset: (TransitionProgress(0.0), InterpolationState(0.0)),
+                start_pts: last_pts,
+                duration: current_transition.duration,
+                interpolation_kind: current_transition.interpolation_kind,
+            }),
+        }
+    }
+
+    pub fn state(&self, pts: Duration) -> InterpolationState {
```
wkozyra95, line 74: I tried, but I don't see any way to simplify this.

compositor_render/src/scene/transition.rs
```diff
@@ -0,0 +1,96 @@
...
+
+pub(super) struct TransitionOptions {
+    pub duration: Duration,
+    pub interpolation_kind: InterpolationKind,
+}
+
+impl TransitionState {
+    pub fn from_previous_transition(
+        current_transition: Option<TransitionOptions>,
+        previous_transition: Option<TransitionState>,
```
noituri, line 36: I don't have strong opinion on this but from the function's name alone I would expect that `previous_transition` is the first argument.

noituri (approved):

compositor_render/src/scene/transition.rs
```diff
@@ -0,0 +1,96 @@
...
+        match (current_transition, previous_transition) {
+            (None, None) => None,
+            (None, Some(previous_transition)) => {
+                let remaining_duration = (previous_transition.start_pts
+                    + previous_transition.duration)
+                    .saturating_sub(last_pts);
+                let progress_offset = TransitionProgress(
+                    1.0 - (remaining_duration.as_secs_f64()
+                        / previous_transition.duration.as_secs_f64()),
+                );
```
WojciechBarczynski, line 54: This only makes sense for the same transition type*. Is this expected?

* I can explain it in more detail in person. Long story short, if transition functions have different values at transition progress (like one is linear, the other is cubic), this would cause discontinuity in the interpolation function (e.g. rapid change of corners rounding radius).

wkozyra95: This is a case `(None, Some(previous_transition))` so scene update where previous transition is still in progress and new update does not have a transition defined yet. So we are only dealing here with one type of transition.

And whether it will work for non linear transition: I think it works, I tested it with some basic examples, but not enough to be sure.

compositor_render/src/scene/transition.rs
wkozyra95: I think arg order is fine here, if that `from` is misleading then we could change the name to just `new`.

WojciechBarczynski (approved):

compositor_render/src/scene/transition.rs
jerzywilczek: it seems a bit misleading to me as well

jerzywilczek (approved):

PR #351, opened 2024-01-14
Audio receiving and decoding

WojciechBarczynski:
I tested it sending the FFmpeg generated RTP opus stream to the compositor. Queue logged receiving samples.


compositor_pipeline/src/pipeline/decoder.rs
```diff
@@ -2,33 +2,142 @@ use std::sync::Arc;
...
+    Video(VideoDecoder),
+    Audio(AudioDecoder),
+    VideoWithAudio {
+        video: VideoDecoder,
+        audio: AudioDecoder,
+    },
+}
+
+#[derive(Debug, Clone)]
+pub enum DecoderOptions {
```
wkozyra95, line 27: This seems like an unnecessary complication, you could just represent it as 
```
struct {
  video: Option<VideoDecoder>,
  audio: Option<AudioDecoder>,
}
```

I suspect your goal was to disallow the representation of an invalid state. Normally we want to do that to avoid maintaining additional logic to handle edge cases that those states present, but here your choice has the opposite effect. Stream without video and audio is perfectly valid and wouldn't require additional maintenance, we want to disallow it on the validation layer, but there is no reason to block it in internal abstraction. By selecting an enum you are forcing yourself to handle 3 cases everywhere instead of 2

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -0,0 +1,119 @@
...
+};
+
+use crate::pipeline::structs::{AudioCodec, EncodedChunk, EncodedChunkKind, VideoCodec};
+
+use super::{DepayloadingError, RtpStream};
+
+pub struct Depayloader {
+    /// (Depayloader, payload type)
+    pub video: Option<(VideoDepayloader, u8)>,
+    pub audio: Option<(AudioDepayloader, u8)>,
```
WojciechBarczynski, line 13: I didn't want to create another struct for that, because naming it would be nightmare

src/types/convert.rs
```diff
@@ -109,3 +113,104 @@ impl TryFrom<register_request::Port> for api::Port {
...
+        match port {
+            api::Port::Range((lower, upper)) => {
+                Ok(compositor_pipeline::pipeline::Port::Range((lower, upper)))
+            }
+            api::Port::Exact(port) => Ok(compositor_pipeline::pipeline::Port::Exact(port)),
+        }
+    }
+}
+
+impl From<crate::types::VideoCodec> for compositor_pipeline::pipeline::structs::VideoCodec {
```
wkozyra95, line 185: see how it's done in competent conversion
- import stuff from this crate directly  e.g `super::*`
- stuff from pipeline import with some qualifier e.g. `pipeline::structs::VideoCodec` (it's fine if it's entire path, but it does not have to be.

src/types/register_request.rs
```diff
@@ -24,6 +23,49 @@ pub enum RegisterRequest {
...
+}
+
+#[derive(Debug, Default, Serialize, Deserialize, Clone, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum VideoCodec {
+    #[default]
+    H264,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
```
wkozyra95, line 45: We do not generate documentation for input streams now, but treat it as if we would. If we plan to add new stream types they will have to be generated

src/types.rs
```diff
@@ -35,10 +35,13 @@ pub use component::View;
...
 
 pub use register_request::Port;
-pub use register_request::RegisterInputRequest;
 pub use register_request::RegisterOutputRequest;
 pub use register_request::RegisterRequest;
 
+pub use register_request::AudioChannels;
+pub use register_request::AudioCodec;
+pub use register_request::VideoCodec;
+
```
wkozyra95, line 44:
```suggestion
```

It does not look like you are using those

src/types/convert.rs
```diff
@@ -109,3 +113,104 @@ impl TryFrom<register_request::Port> for api::Port {
...
+    fn from(value: crate::types::VideoCodec) -> Self {
+        match value {
+            crate::types::VideoCodec::H264 => {
+                compositor_pipeline::pipeline::structs::VideoCodec::H264
+            }
+        }
+    }
+}
+
+impl From<crate::types::AudioCodec> for compositor_pipeline::pipeline::structs::AudioCodec {
```
wkozyra95, line 195: `::structs::` is fine to use inside `compositor_pipeline::pipeline` but name like that should not be visible outside of that crate, or I would even suggest not using it outside of `compositor_pipeline::pipeline` module

compositor_pipeline/src/pipeline.rs
```diff
@@ -31,6 +33,19 @@ pub mod input;
 pub mod output;
 pub mod structs;
```
wkozyra95, line 34:
```suggestion
mod structs;
```

wkozyra95 (commented):
Mostlly focused on API here, I'll make a second pass to look at the core implemntation.

compositor_pipeline/src/pipeline.rs
```diff
@@ -31,6 +33,19 @@ pub mod input;
...
 
+pub struct ExactPort(pub u16);
+
+pub enum Port {
```
wkozyra95, line 38: If you want to distinguish those types maybe change `Port` to `PortRequest`/`RequestedPort`/`PortRange`

compositor_pipeline/src/queue.rs
```diff
@@ -88,6 +90,19 @@ impl Queue {
...
+    pub fn enqueue_samples(
+        &self,
+        input_id: InputId,
+        samples: AudioSamplesBatch,
+    ) -> Result<(), QueueError> {
+        // TODO
+        info!(
+            "Queue received samples: {:?}, for input: {}",
+            samples, input_id
+        );
```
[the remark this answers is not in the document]
WojciechBarczynski: Yep, I'll work on audio queue in the next PR

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
WojciechBarczynski: Firstly I thought about sth like:
```rust
struct SomeName { 
     depayloader: VideoDepayloader,
     payload_type: u8
}
```
because this `u8` isn't self-explanatory. After reading your comment, I realized, that using `PayloadType(u8)` would be.

I don't want to use separate fields here, because that doesn't encapsulate in type system information, that both of those values have to be the same kind (either both `None` or both `Some`).

src/types/register_request.rs
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
...
+    /// Represents audio received on input.
+    /// If set, `input_id` stream can be used in audio mixing.
+    pub audio: Option<Audio>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub struct Video {
+    /// (**default=`"h264"`**) Video codec of input stream.
+    pub codec: Option<VideoCodec>,
+    /// (**default=`"96"`**) Value of payload type field in received RTP packets.
```
wkozyra95, line 43:
```suggestion
    /// (**default=`96`**) Value of payload type field in received RTP packets.
```

src/types/register_request.rs
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
...
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub struct Audio {
+    /// (**default=`"opus"`**) Audio codec of input stream.
+    pub codec: Option<AudioCodec>,
+    /// Sample rate of input audio. If specified sample rate doesn't match
+    /// real sample rate, audio won't be mixed properly.
+    pub sample_rate: u32,
+    /// Audio channels in received audio stream.
+    pub channels: AudioChannels,
+    /// (**default=`"97"`**) Value of payload type field in received RTP packets.
```
wkozyra95, line 64:
```suggestion
    /// (**default=`97`**) Value of payload type field in received RTP packets.
```

src/types/register_request.rs
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
...
 
 #[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+/// Parameters of registered RTP input stream.
+/// Before using input in video composition or output mixing,
+/// input has to be firstly registered using `register_input` request.
+///
+/// At least one of `video` and `audio` has to be defined.
 pub struct RegisterInputRequest {
     pub input_id: InputId,
+    /// Port on which RTP stream is send.
```
wkozyra95, line 29:
```suggestion
    /// Port or range of ports where the compositor will listen for RTP traffic.
```

src/types/register_request.rs
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+/// Parameters of registered RTP input stream.
+/// Before using input in video composition or output mixing,
+/// input has to be firstly registered using `register_input` request.
+///
+/// At least one of `video` and `audio` has to be defined.
```
wkozyra95, line 26:
```suggestion
```

Comment on that type would probably be better to add in markdown directlly

src/types/register_request.rs
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
...
 pub struct RegisterInputRequest {
     pub input_id: InputId,
+    /// Port on which RTP stream is send.
     pub port: Port,
+    /// Represents video received on input.
+    /// If set, `input_id` stream can be used in video composition.
```
wkozyra95, line 32:
```suggestion
    /// Parameters of a video source included in the RTP stream.
```

> If set, `input_id` stream can be used in audio mixing.

It should be valid to use stream without video for video mixing and without audio for audio mixing.


I used "video source" because stream would sound weird if I use "RTP stream" in the same sentence. Also s in SSRC stands for source so it seems fitting

src/types/register_request.rs
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
...
+    pub video: Option<Video>,
+    /// Represents audio received on input.
+    /// If set, `input_id` stream can be used in audio mixing.
```
wkozyra95, line 35:
```suggestion
    /// Parameters of an audio source included in the RTP stream.```

File: src/types/register_request.rs [hunk 16]
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
...
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub struct Audio {
+    /// (**default=`"opus"`**) Audio codec of input stream.
+    pub codec: Option<AudioCodec>,
+    /// Sample rate of input audio. If specified sample rate doesn't match
+    /// real sample rate, audio won't be mixed properly.
+    pub sample_rate: u32,
+    /// Audio channels in received audio stream.
```
wkozyra95, line 62:
```suggestion
    /// Audio channels.
```

src/types/register_request.rs
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
...
+    pub channels: AudioChannels,
+    /// (**default=`"97"`**) Value of payload type field in received RTP packets.
+    /// If the packet payload type is different, packet won't be used as video.
+    pub rtp_payload_type: Option<u8>,
+    /// (**default=`"false"`**) Specifies if received audio stream use forward error correction.
```
wkozyra95, line 67:
```suggestion
    /// (**default=`"false"`**) Specifies whether the stream uses forward error correction.
```
Isn't this opus specific?

src/types/register_request.rs
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
...
+#[serde(rename_all = "snake_case")]
+pub enum VideoCodec {
+    /// H264 video.
+    H264,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub struct Audio {
+    /// (**default=`"opus"`**) Audio codec of input stream.
+    pub codec: Option<AudioCodec>,
```
wkozyra95, line 58: do we want to have a default here?

src/types/register_request.rs
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
...
+    /// If set, `input_id` stream can be used in video composition.
+    pub video: Option<Video>,
+    /// Represents audio received on input.
+    /// If set, `input_id` stream can be used in audio mixing.
+    pub audio: Option<Audio>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub struct Video {
+    /// (**default=`"h264"`**) Video codec of input stream.
```
wkozyra95, line 41:
```suggestion
    /// (**default=`"h264"`**) Video codec.
```
or
```suggestion
    /// (**default=`"h264"`**) Video codec of an input stream.
```

src/types/register_request.rs
```diff
@@ -21,9 +19,69 @@ pub enum RegisterRequest {
...
+pub enum VideoCodec {
+    /// H264 video.
+    H264,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+pub struct Audio {
+    /// (**default=`"opus"`**) Audio codec of input stream.
+    pub codec: Option<AudioCodec>,
+    /// Sample rate of input audio. If specified sample rate doesn't match
```
wkozyra95, line 59:
```suggestion
    /// Sample rate of an input audio. If the specified sample rate doesn't match
```
You seem to use "input"/"input audio"/"audio stream", please unify how you are referring to the same thing.

compositor_pipeline/src/pipeline/input/rtp.rs
```diff
@@ -133,80 +194,115 @@ impl Drop for RtpReceiver {
...
+            .map_or((None, None), |(tx, rx)| (Some(tx), Some(rx)));
+        let (audio_sender, audio_receiver) = depayloader
+            .audio
+            .as_ref()
+            .map(|_| unbounded())
+            .map_or((None, None), |(tx, rx)| (Some(tx), Some(rx)));
+
+        let should_close2 = should_close.clone();
+        let depayloader_thread = std::thread::Builder::new()
+        .name(format!("Depayloading thread for input: {}", input_id.0))
```
wkozyra95, line 230: fix formating

src/types/register_request.rs
WojciechBarczynski: No strong feelings about it. We only support a single codec right now, then requiring specifying it seems boilerplate-ish. If there are more codecs, this can be required. No strong opinion on that.

wkozyra95 (approved):

PR #356, opened 2024-01-19
Remove `NodeId` usage from `ChromiumSenderThread`

noituri:
Initially, when I implemented the web renderer I thought that it would be possible to have one web renderer that is used by multiple web views. In this PR I removed some of the unnecessary stuff that was added because of the assumptions I made when I was implementing it.


compositor_render/src/scene/validation.rs
```diff
@@ -61,3 +64,31 @@ fn validate_component_ids_uniqueness(outputs: &[OutputScene]) -> Result<(), Scen
...
+        component
+            .children()
+            .into_iter()
+            .try_for_each(|c| visit(c, ids))
+    }
+
+    outputs
+        .iter()
+        .try_for_each(|output| visit(&output.root, &mut web_renderer_ids))
+}
```
WojciechBarczynski, line 94: Should we validate this in the renderer?

jerzywilczek (approved):

compositor_render/src/transformations/web_renderer/shared_memory.rs
```diff
@@ -42,22 +36,11 @@ impl SharedMemory {
...
-        frame.send_process_message(cef::ProcessId::Renderer, process_message)?;
-        // -----
+    pub fn len(&self) -> usize {
+        self.inner.as_ref().map(|shmem| shmem.len()).unwrap()
+    }
 
+    pub fn resize(&mut self, size: usize) -> Result<(), SharedMemoryError> {
         // Releasing the current `Shmem` instance to ensure it does not erase the shared memory descriptor from the file system
         // This is critical to ensure when a new `Shmem` is created at the same location, it doesn't conflict with the old descriptor
         drop(self.inner.take());
```
wkozyra95, line 46:
```suggestion
        self.inner.take();
```
wouldn't this be the same?

compositor_render/src/transformations/web_renderer/chromium_sender_thread.rs
```diff
@@ -128,34 +127,46 @@ impl ChromiumSenderThread {
...
+                let mut process_message = cef::ProcessMessage::new(UNEMBED_SOURCE_FRAMES_MESSAGE);
+                process_message.write_string(0, shmem.to_path_string());
+                frame.send_process_message(cef::ProcessId::Renderer, process_message)?;
+                // -----
+
+                shmem.resize(size)?;
             }
         }
 
+        // Create additional shared memory
```
wkozyra95, line 156: what if the number of inputs is smaller, there is no cleanup of already allocated memory

compositor_render/src/transformations/web_renderer/chromium_sender_thread.rs
```diff
@@ -128,34 +127,46 @@ impl ChromiumSenderThread {
...
+        for (resolution, shmem) in resolutions.iter().zip(shared_memory.iter_mut()) {
+            let size = size_from_resolution(*resolution);
+            if shmem.len() != size {
+                // TODO: This should be synchronised
+                let mut process_message = cef::ProcessMessage::new(UNEMBED_SOURCE_FRAMES_MESSAGE);
+                process_message.write_string(0, shmem.to_path_string());
+                frame.send_process_message(cef::ProcessId::Renderer, process_message)?;
+                // -----
+
+                shmem.resize(size)?;
```
wkozyra95, line 152: what if you only resize if shm is too small? Would this avoid potential crashes even without synchronization?

wkozyra95 (approved):
Please verify if this approach works (does not crash) even without validation that only one component is using the renderer.

WojciechBarczynski (approved):

wkozyra95:
I need those changes for the other PR I'm working on and it seems everyone already reviewed it, so I'll merge it. Please take a look at the comments and if there is anything actionable there address that in a follow-up PR.

compositor_render/src/transformations/web_renderer/shared_memory.rs
noituri: This whole line is not needed unless I forgot about something.

compositor_render/src/transformations/web_renderer/chromium_sender_thread.rs
noituri: There is no cleanup since it requires synchronization. It will be added eventually

compositor_render/src/transformations/web_renderer/chromium_sender_thread.rs
noituri: I think it would work

PR #362, opened 2024-01-26
[docs] Fix inconsistent header sizing in routes page

wkozyra95:
Docs for RegisterInputRequest were not supposed to be generated. There are a few problems with that change:
- headers are the wrong size, when other types use `###` the `RegisterInputRequest` is using `##`
- imported markdown files do not show up in markdown

For now, I manually copied generated docs and changed headers to look consistent. This is just a temporary fix, it needs to be addressed better.

### DIFF
--- a/docs/pages/api/routes.md
+++ b/docs/pages/api/routes.md
@@ -2,8 +2,6 @@
 description: API routes to configure the compositor.
 ---
 
-import Docs from "@site/pages/api/generated/renderer-RegisterInputRequest.md"
-
 # Routes
 
 ### Start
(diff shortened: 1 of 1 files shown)


docs/pages/api/routes.md
```diff
@@ -38,7 +36,69 @@ type OutputScene = {
...
+  type: "register";
+  entity_type: "input_stream";
+  input_id: string;
+  port: Port;
+  video?: Video;
+  audio?: Audio;
+}
+```
+
+Parameters of registered RTP input stream. Before using input in video composition or output mixing, input has to be firstly registered using `register_input` request.
```
noituri, line 52:
```suggestion
Parameters of registered RTP input stream. Before using input in video composition or output mixing, input has to be first registered using `register_input` request.
```

docs/pages/api/routes.md
```diff
@@ -38,7 +36,69 @@ type OutputScene = {
 
 ***
 
-<Docs />
+### RegisterInputRequest
```
WojciechBarczynski, line 39: Why can't we use generated docs here?

wkozyra95: I explained the problems in the PR description.

We want to use generated docs here, but docs generation needs to be adjusted to produce sth that will fit otherwise it looks very bad.

PR #363, opened 2024-01-26
Add option to force GPU rendering

wkozyra95:
- Add `LIVE_COMPOSITOR_FORCE_GPU` env
- Improve logging defaults to avoid log spam
- Display useful info about available adpaters

It looks like M1 metal is detected as CPU, but given that this option is mostly useful for deployment purposes then I think it's still fine

### DIFF
--- a/compositor_render/src/wgpu/ctx.rs
+++ b/compositor_render/src/wgpu/ctx.rs
@@ -1,4 +1,4 @@
-use log::error;
+use log::{error, info};
 
 use super::{
     common_pipeline::plane::Plane, format::TextureFormat, texture::Texture, utils::TextureUtils,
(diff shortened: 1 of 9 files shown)


wkozyra95:
@noituri @WojciechBarczynski Can one of you test that env on M1? I'm not sure if it won't be detected as CPU.

noituri:
image

I think it works @wkozyra95

compositor_render/src/wgpu/ctx.rs
```diff
@@ -35,9 +37,19 @@ impl WgpuCtx {
...
+        );
+        if force_gpu && adapter_info.device_type != wgpu::DeviceType::Cpu {
+            error!("Selected adapter is CPU based. Aborting.");
+            return Err(CreateWgpuCtxError::NoAdapter);
+        }
+
         let (device, queue) = pollster::block_on(adapter.request_device(
             &wgpu::DeviceDescriptor {
-                label: Some("Video Compositor's GPU :^)"),
+                label: None,
```
jerzywilczek, line 52: :(

jerzywilczek (approved):
very weird that it detects M1 metal as `Cpu`, but can't really do nothing about it

WojciechBarczynski (commented):
LGTM

PR #364, opened 2024-01-29
Add minimal router + add `/status` route

wkozyra95:
Routes:
- `POST /--/api` main endpoint used for almost everything
- `GET /status` health check - it just returns 200 response


WojciechBarczynski (approved):
LGTM, but I don't know why we use `/--/api`? What those `/--/` are for? Is this some common form of routing pattern, that I'm not aware of?

wkozyra95:
> What those /--/ are for? Is this some common form of routing pattern, that I'm not aware of?

No strong opinion about it, I think just `/api` looks wierd. As for the `/--`, I saw sth like this or `/__` in a lot of places. I think the main reason for prefix like that is if you host your backend code on the same domain as you frontend, so api routes have a prefix that is unlikely cause trouble and easy to redirect to different service using load balancer.

jerzywilczek (approved):
:rocket:

PR #366, opened 2024-01-30
Audio queue

WojciechBarczynski:
Adds audio queue

Update:
While implementing the audio mixer, I decided not to fill the missing samples in the queue, since the logic for dealing with corner cases already has to be implemented in the mixer anyway (starting samples in the middle of the buffer, samples ending in the middle, etc.)

Noticed problems (outside of the PR scope):
- Since the video queue uses `saturating sub` while calculating timestamp offset, streams starting with large initial PTS values (larger than `clock_start.elapsed()`) won't be handled properly. Rust duration doesn't support negative values, so maybe we should use 2 separate offsets: stream-added time offset and starting pts offset. We haven't noticed ant problems yet, since all our test streams start with pts 0.
- The video queue uses `Vec` for keeping enqueued frames, which is inefficient when draining starting elements. We should consider using `VecDeque` (with only a few elements in the queue it's not a problem, but those allocations are unnecessary)


compositor_pipeline/src/pipeline/decoder/opus_decoder.rs
```diff
@@ -58,7 +58,7 @@ impl OpusDecoder {
                 }
                 AudioChannels::Stereo => {
                     let mut samples = Vec::with_capacity(decoded_samples_count / 2);
-                    for i in 0..(decoded_samples_count / 2) {
+                    for i in 0..decoded_samples_count {
```
WojciechBarczynski, line 61: It was a bug

compositor_pipeline/src/queue/audio_queue.rs
WojciechBarczynski, line 1: I just noticed that input buffering isn't correct here. I'll try to fix it.

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -0,0 +1,163 @@
...
+
+        Ok(())
+    }
+
+    pub fn pop(&mut self, length: Duration) -> AudioSamples {
+        let samples_count = (length.as_secs_f64() * self.sample_rate as f64) as usize;
+        let samples_to_take = min(samples_count, self.queue.len());
+
+        match &mut self.queue {
+            SamplesQueue::Mono(queue) => {
```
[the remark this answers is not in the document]
wkozyra95: >Panics if the starting point is greater than the end point or if the end point is greater than the length of the deque.

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -0,0 +1,163 @@
...
+                let AudioSamples::Mono(samples) = samples else {
+                    return Err(QueueError::MismatchedSamplesChannels {
+                        expected: AudioChannels::Mono,
+                        received: samples,
+                    });
+                };
+                // To account for calculation precision errors
+                if missing_samples > 5 {
```
wkozyra95, line 123: why 5, I wouldn't expect numeric errors to cause more than one sample duration error

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    pub fn enqueue_samples(
+        &mut self,
+        samples: AudioSamples,
+        pts: Duration,
+    ) -> Result<(), QueueError> {
+        let expected_previous_samples = (pts.as_secs_f64() * self.sample_rate as f64) as i64;
+        let missing_samples =
```
wkozyra95, line 111: this value can be negative, it seems that in case like that you should either drop sample from queue or from new same batch

compositor_pipeline/src/queue/audio_queue.rs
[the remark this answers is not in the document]
WojciechBarczynski: That's why I used `samples_to_take = min(samples_count, self.queue.len())`

compositor_pipeline/src/queue/audio_queue.rs
WojciechBarczynski: Yep, while testing I only managed to get the numeric error 1 off. I used 5 for safety. In reality, the only reason for missing samples is that the compositor couldn't receive/read some chunks of samples (e.g. lost packet). The usual length of an audio chunk in a single RTP packet is 20ms, meaning that there are hundreds of samples. I guess that's the case for all inputs (receiving or reading only a few samples is very unlikely). I can't find a probable reason for missing a few samples, but calculation errors >1 can happen. Therefore, I want to use some value >1 here. I think that anything in the range [3, 10] is a reasonable threshold here.

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use compositor_render::{AudioSamples, AudioSamplesBatch, AudioSamplesSet, InputId};
+
+use crate::pipeline::AudioChannels;
+
+use super::QueueError;
+
+pub struct AudioQueue {
+    input_queues: HashMap<InputId, InputQueue>,
+    timestamp_offsets: HashMap<InputId, Duration>,
```
jerzywilczek, line 15: Some docs on this field would be nice

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -0,0 +1,163 @@
...
+    ) -> Result<(), QueueError> {
+        let Some(input_queue) = self.input_queues.get_mut(&input_id) else {
+            return Err(QueueError::UnknownInputId(input_id));
+        };
+        let offset = *self
+            .timestamp_offsets
+            .entry(input_id)
+            .or_insert_with(|| samples_batch.pts);
+
+        input_queue.enqueue_samples(samples_batch.samples, samples_batch.pts - offset)?;
```
jerzywilczek, line 48: If I understand correctly, this causes every audio stream's first buffer to have `pts == 0`. I think this is incorrect, shouldn't the inserted offset be `samples_batch.pts - current_queue_pts`, or something similar?

compositor_pipeline/src/queue/audio_queue_thread.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl AudioQueueThread {
+    pub fn new(queue: Arc<Queue>, sender: Sender<AudioSamplesSet>, opts: Options) -> Self {
+        AudioQueueThread {
+            queue,
+            sender,
+            buffer_duration: opts.buffer_duration,
+            chunk_length: opts.pushed_chunk_length,
+            send_chunks: 0,
```
jerzywilczek, line 33:
```suggestion
            sent_chunks: 0,
```

jerzywilczek (commented):
:nerd_face:

noituri (approved):

compositor_pipeline/src/queue/audio_queue_thread.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    pub fn spawn(mut self) -> JoinHandle<()> {
+        thread::spawn(move || self.run())
+    }
+
+    fn run(&mut self) {
+        thread::sleep(self.buffer_duration);
+        let ticker = tick(self.chunk_length);
+        loop {
+            ticker.recv().unwrap();
```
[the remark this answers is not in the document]
WojciechBarczynski: In 2h test I didn't notice error agregation

compositor_pipeline/src/queue/audio_queue.rs
WojciechBarczynski: With the new audio queue approach, this logic is no longer useful

compositor_pipeline/src/pipeline.rs
```diff
@@ -126,13 +131,18 @@ impl Pipeline {
...
         } = register_options;
 
         if self.inputs.contains_key(&input_id) {
             return Err(RegisterInputError::AlreadyRegistered(input_id));
         }
 
+        if self.inputs.contains_key(&input_id) {
+            return Err(RegisterInputError::AlreadyRegistered(input_id));
+        }
+
```
wkozyra95, line 145:
```suggestion
```

compositor_pipeline/src/queue.rs
```diff
@@ -57,55 +95,87 @@ pub struct Queue {
...
             clock_start: Instant::now(),
+            audio_queue: Mutex::new(AudioQueue::new()),
+            audio_chunk_duration: DEFAULT_AUDIO_CHUNK_DURATION,
         }
     }
 
-    pub fn add_input(&self, input_id: InputId) {
-        self.internal_queue.lock().unwrap().add_input(input_id);
+    pub fn add_input(&self, options: InputType) {
+        if options.video.is_some() {
```
wkozyra95, line 109:
```suggestion
        if let Some(_) = options.video {
```

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -0,0 +1,101 @@
...
+use compositor_render::{AudioSamplesBatch, AudioSamplesSet, InputId};
+
+#[derive(Debug)]
+pub struct AudioQueue {
+    /// Enqueued sample batches per output
+    input_queues: HashMap<InputId, VecDeque<AudioSamplesBatch>>,
+    /// Stream added timestamp offset in relation to clock start
+    timestamp_offsets: HashMap<InputId, Duration>,
+    /// Stream starting pts offsets (starting pts doesn't have to be 0)
+    start_pts: HashMap<InputId, Duration>,
```
wkozyra95, line 16: you don't need that value

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -0,0 +1,101 @@
...
+        self.input_queues.remove(input_id);
+    }
+
+    pub fn enqueue_samples(
+        &mut self,
+        input_id: InputId,
+        mut samples_batch: AudioSamplesBatch,
+        clock_start: Instant,
+    ) -> Result<(), QueueError> {
+        let Some(input_queue) = self.input_queues.get_mut(&input_id) else {
```
wkozyra95, line 42: You are syncing here based on the first audio sample, have you considered syncing both video and audio on the first received packet of either of them?

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -0,0 +1,101 @@
...
+    time::{Duration, Instant},
+};
+
+use super::QueueError;
+use compositor_render::{AudioSamplesBatch, AudioSamplesSet, InputId};
+
+#[derive(Debug)]
+pub struct AudioQueue {
+    /// Enqueued sample batches per output
+    input_queues: HashMap<InputId, VecDeque<AudioSamplesBatch>>,
```
wkozyra95, line 12: Why batching? You said you wanted to keep samples in queue?

WojciechBarczynski: Keeping samples requires dealing with missing samples in the queue. As I wrote in the PR description, when I started implementing audio mixing, I noticed that the mixer already has to deal with a lot of corner cases like this and I don't think spreading this responsibility meaningful simplifies code. Mixer will have to deal with cases like:
- Different start pts of samples for different inputs (when input is added)
- Missing samples at the end (when samples haven't been received yet)
- Different input parameters, like sample rate, channels, etc.

Previously I thought that keeping just a vec of samples would simplify implementation, but now I think it just increases the number of silent implementation assumptions between mixer and queue (mixer has to know how the queue deals with missing samples, etc. to work properly). With the current approach (sending batches), the only assumption the mixer has to make about the queue is that it delivers batches with samples in the range of currently mixed fragments when possible. I think this increases testability and makes debugging easier. 

Another benefit of keeping batches is that this reduces copying and allocations, but I don't think that performance is critically important here.

compositor_pipeline/src/queue.rs
WojciechBarczynski: I know that you want this to look consistent with the statement bellow, but:
https://rust-lang.github.io/rust-clippy/master/index.html#/redundant_pattern_matching

wkozyra95: Did you verify this? The pattern match bellow is also redundant, so what is the difference there?
If for some reason there is a differnce than maybe change audio check to is_some?

WojciechBarczynski: Yep, I verified this.
It's interesting, since if I use
```rust
if let Some(_name) = sth {...}
```
Clippy doesn't have any issues, but 
```rust
if let Some(_) - sth {...}
```
throws a warning. 

I guess, that adding a `name` part might turn off the warning because someone wants to more thoroughly explain what is inside (I guess in some cases it might improve readability).

Anyway, using `Some(_audio_options)` when `_audio_options = ()` is a little bit weird, but I can use it if you want (IMO it doesn't make any difference here).

wkozyra95: doesn't matter which approach, but I would prefer the same one for both audio and video

compositor_pipeline/src/queue/audio_queue.rs
WojciechBarczynski: I have considered it, but since the video decoder deals with timestamps differently than the audio decoder, this wouldn't work. It comes down to the same conversation as in the previous suggestion - how the PTS of queue input should look like (should PTS for all inputs start at 0 and increase monotonically).

compositor_pipeline/src/queue/audio_queue.rs
wkozyra95: I understand those issues, but none of those is an argument to keep those hash maps. All of the problems you listed are easily solvable without hacks involving 2 offsets.

PR #372, opened 2024-02-02
Support per-output scene updates in renderer

wkozyra95:
- ensure variant without chrome builds on CI
- pass information about register/unregistering inputs and unregistering output is passed to renderer (register output is not necessary now, but it will be added in follow up)
- removed fallback code. Until now fallback code was left there because we want to address that issue, but don't know how yet, plus it didn't cost us any maintenance overhead, but now it makes it harder so I removed it.
- internals
  - nodes do not have an ids right now (they were only needed to establish edges of the graph)
  - render graph is now a tree

### DIFF
--- a/compositor_render/src/state.rs
+++ b/compositor_render/src/state.rs
@@ -32,8 +32,6 @@ pub mod render_graph;
 mod render_loop;
 pub mod renderers;
 
-pub(crate) use render_loop::NodeRenderPass;
-
 pub struct RendererOptions {
     pub web_renderer: web_renderer::WebRendererInitOptions,
     pub framerate: Framerate,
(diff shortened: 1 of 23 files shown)


compositor_render/src/state.rs
```diff
@@ -192,34 +211,34 @@ impl InnerRenderer {
...
-        self.render_graph.update(
-            &RenderCtx {
-                wgpu_ctx: &self.wgpu_ctx,
-                text_renderer_ctx: &self.text_renderer_ctx,
-                renderers: &self.renderers,
-                stream_fallback_timeout: self.stream_fallback_timeout,
-            },
-            output_nodes,
-        )?;
+        for output_scene in outputs {
```
wkozyra95, line 228: tmp, in a follow up PR the update_scene will be modified to only update one scene

src/snapshot_tests/utils.rs
```diff
@@ -73,6 +74,9 @@ pub(super) fn create_renderer(
         }
         renderer.register_renderer(spec).unwrap();
     }
+    for id in 0..16 {
```
wkozyra95, line 77: this will be reworked in PR with a new API

examples/pass_through.rs
```diff
@@ -0,0 +1,113 @@
...
+        "entity_type": "rtp_input_stream",
+        "input_id": "input_1",
+        "port": 8004,
+        "video": {
+            "codec": "h264"
+        }
+    }))?;
+
+    info!("[example] Update scene");
+    common::post(&json!({
```
wkozyra95, line 79: given the internal changes now case where input goes into output is a special case, so I added an example to monitor for regressions

WojciechBarczynski (approved):
LGTM

noituri (approved):
lgtm 👌

PR #373, opened 2024-02-02
Update wgpu to 0.19.1

jerzywilczek:
Text layouting was very slightly changed in the new `glyphon` version, and one snapshot needs to be updated: membraneframework-labs/video_compositor_snapshot_tests#26

There are some breaking changes in `wgpu`, which force us to change our code:

- in `wgpu::RenderPassColorAttachment`:
  - the `store` field in `wgpu::Operations` is now an enum instead of a bool
  - two new fields were added, `timestamp_writes` and `occlusion_query_set`. Both are used for debugging and performance measurements
- in `naga`, the shader parser:
  - `ArraySize` was simplified a lot. `ArraySize::Constant` no longer holds a generic shader constant (which could be negative for example). It's represented as a `NonZeroU32` now, which allows us to simplify the array size evaluation code. Because of this change we also don't need to deal with constants anymore at all, which means we can get rid of some code for converting constants to strings.
  - an additional `Scalar` struct was added. Previously all functions and structs held information about scalars in two separate fields, `kind` and `width`. This is represented as a single `Scalar` struct now, which holds both of these values.
- in `wgpu::DeviceDescriptor`:
  - `limits` were renamed to `required_limits`
  - `features` were renamed to `required_features`
- `TextureFormat::block_size` was renamed to `TextureFormat::block_copy_size`

### DIFF
--- a/compositor_render/src/transformations/shader/validation.rs
+++ b/compositor_render/src/transformations/shader/validation.rs
@@ -1,4 +1,4 @@
-use naga::{ArraySize, ConstantInner, Handle, Module, ScalarKind, ShaderStage, Type, VectorSize};
+use naga::{ArraySize, Handle, Module, ScalarKind, ShaderStage, Type, VectorSize};
 
 use crate::scene::ShaderParam;
 
@@ -171,12 +171,7 @@ fn validate_type_equivalent(
                 });
             }
 
-            validate_array_size_equivalent(
-                expected_size,
-                expected_module,
-                provided_size,
-                provided_module,
-            )?;
+            validate_array_size_equivalent(expected_size, provided_size)?;
             return validate_type_equivalent(
                 expected_base,
                 expected_module,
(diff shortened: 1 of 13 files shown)


wkozyra95:
- When you are adding PR with snapshot changes link to the PR in snapshot repo
- This is sth that is hard to review based on code. Can you add a description in a PR(or PR comments if it makes it easier) that explains the changes?

compositor_render/src/transformations/shader/validation.rs
```diff
@@ -637,75 +604,99 @@ impl TypeInnerExt for naga::TypeInner {
                 class,
             } => match (dim, arrayed, class) {
                 (naga::ImageDimension::D1, false, naga::ImageClass::Sampled { kind, .. }) => {
-                    format!("texture_1d<{}>", kind.to_string(4))
```
wkozyra95, line 640: By removing ScalarKindExt this function becomes far bigger than it needs to be (especially when just a single case statement generates that much code). Can you rework it to be more readable?

wkozyra95 (commented):
I'm a bit concerned about places where we add `..Default::default()`. If we did not need that before then I assume it's just one or two fields that were added, so unless it's a lot I would prefer to keep that explicit.

noituri (approved):
LGTM

wkozyra95 (approved):

PR #375, opened 2024-02-05
Per output update scene request

wkozyra95:
- Modify scene_update request to update single output at the time
- Add initial_scene field to register_output
- Update docs

### DIFF
--- a/compositor_pipeline/src/error.rs
+++ b/compositor_pipeline/src/error.rs
@@ -33,6 +33,9 @@ pub enum RegisterOutputError {
 
     #[error("Failed to register output stream \"{0}\". Resolution in each dimension has to be divisible by 2.")]
     UnsupportedResolution(OutputId),
+
+    #[error("Failed to initialize the scene when registering output \"{0}\"")]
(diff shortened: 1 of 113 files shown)


WojciechBarczynski (commented):
What are the benefits of using per-output scene updates? I think that the most common multi-output usage scenario is composition for multiple platforms. In such a case, when the user wants to update the composition, I assume that usually they want to update it for every output. What are the benefits of having to send `output_count` requests instead of just one?

wkozyra95:
@WojciechBarczynski 

- Simpler API for us (and cleaner docs)
- Consistent with register output (you pass the same type of data)
- Ability to use a single instance for different purposes (currently every update had to include all outputs)
- I don't have anything against adding an endpoint that allows batching those requests, but I don't see that being too helpful.
- Almost everyone will have just one output
- It makes it easier to add API e.g. `scene_update` instead of defining scenes links to different outputs that will be used without re-rendering. Useful to produce multiple resolutions of the same scene

WojciechBarczynski (approved):

jerzywilczek (approved):
if you like it i like it

PR #377, opened 2024-02-06
Fix initial scene validation

noituri:


compositor_render/src/scene/validation.rs
```diff
@@ -36,13 +36,16 @@ pub(super) fn validate_scene_update(
...
+    let updated_outputs = match old_outputs.is_empty() {
+        true => vec![updated_output],
+        false => old_outputs
+            .iter()
+            .map(|(id, output)| match id {
+                id if id == &updated_output.output_id => updated_output,
+                _ => output,
+            })
+            .collect(),
+    };
```
WojciechBarczynski, line 48: This won't work if there already are some outputs and a new one is added.

WojciechBarczynski (changes requested):

compositor_render/src/scene/validation.rs
wkozyra95: Wojtek is correct, I addressed it in https://github.com/membraneframework/video_compositor/pull/375 so fill free to close it.

Issue #352, opened 2024-01-16
Rewrite `ChromiumSenderThread`


noituri:
- [x] Simplify thread state
- [x] Simplify communication compositor <--> chromium

PR #378, opened 2024-02-07
Remove `register_inputs` from web renderer

noituri:
Currently, users have to specify HTML elements that display frames in the `register_inputs` JS function.
For example:
```html
<body>
  <canvas id="my_video"></canvas>
</body>

<script>
  register_inputs("my_video")
</script>
```

The problems with this approach:
- Code editors don't know about `register_inputs` so they show an error
- You can't test the website in a normal browser because the browser would also show an error 
- The embedding config isn't in one place
- It could cause problems for react or similar libraries

The new approach uses component IDs for specifying IDs of HTML elements.
```ts
{
    "output_id": "output_1",
    "root": {
        "id": "embed_input_on_website",
        "type": "web_view",
        "instance_id": "example_website",
        "children": [
            {
                "id": "my_video", // <-- here
                "type": "input_stream",
                "input_id": "input_1",
            }
        ]
    }
}
```
  
I don't know if using component ID is the correct way to solve this problem. I considered adding a new field to each child but I'm not sure if that's any better.

Issue #352

### DIFF
--- a/compositor_render/src/transformations/web_renderer/chromium_sender.rs
+++ b/compositor_render/src/transformations/web_renderer/chromium_sender.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use crate::{state::RegisterCtx, Resolution};
+use crate::{scene::ComponentId, state::RegisterCtx, Resolution};
 use crossbeam_channel::{Receiver, Sender};
 use log::error;
 
@@ -45,10 +45,17 @@ impl ChromiumSender {
         }
     }
 
-    pub fn embed_sources(&self, sources: &[&NodeTexture]) -> Result<(), ChromiumSenderError> {
+    pub fn embed_sources(
+        &self,
+        sources: &[&NodeTexture],
+        children_ids: Vec<ComponentId>,
+    ) -> Result<(), ChromiumSenderError> {
         let resolutions = sources.iter().map(|texture| texture.resolution()).collect();
         self.message_sender
-            .send(ChromiumSenderMessage::EmbedSources { resolutions })
+            .send(ChromiumSenderMessage::EmbedSources {
+                resolutions,
+                children_ids,
+            })
             .map_err(|_| ChromiumSenderError::MessageChannelDisconnected)
     }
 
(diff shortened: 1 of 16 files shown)


examples/web_view.html
```diff
@@ -1,50 +1,41 @@
...
-</script>
+            50% {
+                margin-left: 250px;
+            }
+        }
+    </style>
 
+    <body>
+        <h2>Frame embedding:</h2>
+        <canvas id="big_bunny_video"></canvas>
```
wkozyra95, line 39: Potential options
- [current] Use component_id directly for html id
- add some prefix e.g. for component id `big_bunny_video` html id could be `live_compositor:big_bunny_video`
  - a bit cleaner, scoped, harder to make collisions
  - can't be used to embed sth on a website you don't own
  
I agree with the current approach, but I'm assuming that it is possible to use this rendering on website you don't own. If there is already a reason why it would not be possible then I would prefer approach 2.

does this element need to be a canvas for all rendering options or just for chromium embeding?

compositor_render/src/transformations/web_renderer/chromium_sender.rs
```diff
@@ -45,10 +45,17 @@ impl ChromiumSender {
         }
     }
 
-    pub fn embed_sources(&self, sources: &[&NodeTexture]) -> Result<(), ChromiumSenderError> {
+    pub fn embed_sources(
+        &self,
+        sources: &[&NodeTexture],
+        children_ids: &[ComponentId],
```
wkozyra95, line 51: you are passing vector as a slice just to convert it back to the vector.

```suggestion
        children_ids: Vec<ComponentId>,
```

examples/web_view.html
noituri: It's possible to use embedding on websites you don't own. I like the idea of using a namespace but I'm not sure if we want to limit the web renderer to websites designed specifically for the compositor.

For chromium embedding canvas element is required. Other embedding methods are not limited to canvases

WojciechBarczynski (approved):

PR #383, opened 2024-02-09
Implement control flow for elements before queue

wkozyra95:
- Refactor queues to store just one HashMap
- use chrono::Duration for offset calculations (supports negative values)
- remove code from mp4 code that simulates real-time
- add queue between decoder and queue
  - decoder <-> queue - bounded channel 10
  - mp4 <-> decoder - bounded channel 10
  - rtp - for now  unbounded
- ~~[REGRESSION] I removed sleep which ensured buffer was filled. (I'm working on that, but I'm not sure if this will be part of this PR or a follow up)~~
  - resolved in https://github.com/membraneframework/video_compositor/pull/383/commits/9c85719dc90b778e256d782b85729ff3144ade27

For now, I just did minimal testing + ~~I'm still looking into replacement for sleep~~, but I think this PR is complete enough for review.

### DIFF
--- a/compositor_pipeline/src/queue.rs
+++ b/compositor_pipeline/src/queue.rs
@@ -9,46 +9,15 @@ use std::{
     time::{Duration, Instant},
 };
 
-use compositor_render::{
-    AudioSamples, AudioSamplesBatch, AudioSamplesSet, Frame, FrameSet, Framerate, InputId,
-};
+use compositor_render::{AudioSamplesSet, FrameSet, Framerate, InputId};
 use crossbeam_channel::{unbounded, Receiver, Sender};
-use log::error;
-use thiserror::Error;
 
-use crate::pipeline::{decoder::DecodedDataReceiver, AudioChannels};
+use crate::pipeline::decoder::DecodedDataReceiver;
 
 use self::{
     audio_queue::AudioQueue, audio_queue_thread::AudioQueueThread, video_queue::VideoQueue,
     video_queue_thread::VideoQueueThread,
 };
-
-#[derive(Debug, Clone)]
-pub struct InputOptions {
-    pub video: Option<Receiver<Frame>>,
-    pub audio: Option<Receiver<AudioSamplesBatch>>,
-}
-
-#[derive(Error, Debug)]
-pub enum QueueError {
-    #[error("the input id `{:#?}` is unknown", 0)]
-    UnknownInputId(InputId),
-    #[error(
-        "expected samples in {:#?} channel format, but received samples {:#?}",
-        expected,
-        received
-    )]
-    MismatchedSamplesChannels {
-        expected: AudioChannels,
-        received: AudioSamples,
-    },
-    #[error(
-        "expected samples with {} sample rate, but received samples of rate {}",
-        expected,
-        received
-    )]
-    MismatchedSampleRate { expected: u32, received: u32 },
-}
 
 const DEFAULT_BUFFER_DURATION: Duration = Duration::from_millis(16 * 5); // about 5 frames at 60 fps
 const DEFAULT_AUDIO_CHUNK_DURATION: Duration = Duration::from_millis(20); // typical audio packet size
--- a/compositor_pipeline/src/queue.rs
+++ b/compositor_pipeline/src/queue.rs
@@ -1,38 +1,32 @@
 mod audio_queue;
 mod audio_queue_thread;
+mod utils;
 mod video_queue;
 mod video_queue_thread;
 
 use std::{
     sync::{Arc, Mutex},
-    thread,
     time::{Duration, Instant},
 };
 
 use compositor_render::{
-    error::ErrorStack, AudioSamples, AudioSamplesBatch, AudioSamplesSet, Frame, FrameSet,
-    Framerate, InputId,
+    AudioSamples, AudioSamplesBatch, AudioSamplesSet, Frame, FrameSet, Framerate, InputId,
 };
 use crossbeam_channel::{unbounded, Receiver, Sender};
 use log::error;
 use thiserror::Error;
 
-use crate::pipeline::AudioChannels;
+use crate::pipeline::{decoder::DecodedDataReceiver, AudioChannels};
 
 use self::{
     audio_queue::AudioQueue, audio_queue_thread::AudioQueueThread, video_queue::VideoQueue,
     video_queue_thread::VideoQueueThread,
 };
 
-pub struct InputType {
-    pub input_id: InputId,
-    pub video: Option<()>,
-    pub audio: Option<AudioOptions>,
-}
-
-pub struct AudioOptions {
-    pub sample_rate: u32,
-    pub channels: AudioChannels,
+#[derive(Debug, Clone)]
+pub struct InputOptions {
+    pub video: Option<Receiver<Frame>>,
+    pub audio: Option<Receiver<AudioSamplesBatch>>,
 }
 
 #[derive(Error, Debug)]
@@ -105,15 +99,18 @@ impl Queue {
         }
     }
 
-    pub fn add_input(&self, options: InputType) {
-        if let Some(_video_options) = options.video {
+    pub fn add_input(&self, input_id: &InputId, receiver: DecodedDataReceiver) {
+        if let Some(receiver) = receiver.video {
             self.video_queue
                 .lock()
                 .unwrap()
-                .add_input(options.input_id.clone());
+                .add_input(input_id, receiver);
         };
-        if let Some(_audio_options) = options.audio {
-            self.audio_queue.lock().unwrap().add_input(options.input_id);
+        if let Some(receiver) = receiver.audio {
+            self.audio_queue
+                .lock()
+                .unwrap()
+                .add_input(input_id, receiver);
         }
     }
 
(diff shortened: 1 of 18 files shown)


compositor_pipeline/src/queue.rs
```diff
@@ -1,38 +1,32 @@
...
-    pub audio: Option<AudioOptions>,
-}
-
-pub struct AudioOptions {
-    pub sample_rate: u32,
-    pub channels: AudioChannels,
+#[derive(Debug, Clone)]
+pub struct InputOptions {
+    pub video: Option<Receiver<Frame>>,
+    pub audio: Option<Receiver<AudioSamplesBatch>>,
```
WojciechBarczynski, line 29: Is this used anywhere?

wkozyra95: Nope, I missed that.
It looks like I defined type like that twice and I'm using the other one.

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -3,99 +3,125 @@ use std::{
...
+        let (start_pts, end_pts) = range;
 
-    pub fn drop_old_samples(&mut self, up_to_pts: Duration) {
-        for input_queue in self.input_queues.values_mut() {
-            while input_queue
-                .front()
-                .map_or(false, |batch| batch.end() < up_to_pts)
-            {
-                input_queue.pop_front();
+        fn is_ready(queue: &VecDeque<AudioSamplesBatch>, range_end_pts: Duration) -> bool {
```
WojciechBarczynski, line 81: `is_ready` is a pretty bad name here. When I first saw function `is_ready` inside `pop_samples`, I thought that checks if enqueued samples should be pushed. Only after a while did I realize it determines if samples from channel should be read. Maybe move the whole reading from the channel part to a separate function. I wouldn't expect from function `pop_sampels` to fetch anything from the channel (even though it makes sense here), especially since the function docs don't even mention that.

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -3,99 +3,125 @@ use std::{
...
             }
         }
+
+        while !is_ready(&self.queue, end_pts) {
+            if self.try_enqueue_samples(clock_start).is_err() {
+                break;
+            }
+        }
+
+        let poped_samples = self
```
WojciechBarczynski, line 94:
```suggestion
        let popped_samples = self
```

compositor_pipeline/src/queue/audio_queue.rs
WojciechBarczynski: Especially since there already exists `is_ready` function in `video_queue` that does the completely different thing

compositor_pipeline/src/queue/video_queue_thread.rs
WojciechBarczynski, line 1: Function
```rust
fn should_push_pts(&self, pts: Duration) -> bool {
        self.opts.clock_start.add(pts) < Instant::now()
    }
```

with current changes works incorrectly and buffering is broken because of it. It doesn't account for buffering, meaning that the `DEFAULT_BUFFER_DURATION` of every stream will be handled improperly (and the beginning of every stream is black screen because of that).

Take buffer duration into account.

compositor_pipeline/src/queue/utils.rs
```diff
@@ -0,0 +1,90 @@
...
+    Ready { offset: chrono::Duration },
+}
+
+impl<Payload: ApplyOffsetExt> InputState<Payload> {
+    pub(super) fn process_new_chunk(
+        &mut self,
+        mut payload: Payload,
+        pts: Duration,
+        clock_start: Instant,
+    ) -> VecDeque<Payload> {
```
WojciechBarczynski, line 24: 1. This can just return iterator
2. If the input is shorter than `DEFAULT_BUFFER_DURATION` it won't work. If the stream freezes after receiving only a few frames, the queue won't enqueue them.
3. This + not accounting for buffer duration in the video queue thread results in black frames for the beginning of the input stream.

compositor_pipeline/src/queue/utils.rs
```diff
@@ -0,0 +1,90 @@
...
+use compositor_render::{AudioSamplesBatch, Frame};
+
+use super::DEFAULT_BUFFER_DURATION;
+
+#[derive(Debug)]
+pub(super) enum InputState<Payload: ApplyOffsetExt> {
+    WaitingForStart,
+    Buffering { buffer: Vec<(Payload, Duration)> },
+    Ready { offset: chrono::Duration },
+}
```
WojciechBarczynski, line 16: Why should we do it in such a weird way? Can't we just keep `offset: Option<chrono::Duration>` and apply it to every enqueued frame/sample? I don't understand how this helps.

WojciechBarczynski (commented):
Overall I like the idea

compositor_pipeline/src/queue/audio_queue.rs
wkozyra95: > I thought that checks if enqueued samples should be pushed

I don't follow that logic. That function checks if queue for input is ready (if it has enough samples to fill the range)

compositor_pipeline/src/queue/utils.rs
wkozyra95: 1. Yes, but it still needs to be an iterator that owns the data, so what would be the point. Plus this way I can just call append on queue
2. Good point, but I don't think this is a problem.
3. It does. Please explain why you think that and I will maybe add a description to clarify that.

compositor_pipeline/src/queue/video_queue_thread.rs
wkozyra95: It does. InputState abstraction is handling that.

compositor_pipeline/src/queue/utils.rs
wkozyra95: No. I calculate offset when switching from buffering to ready state.

compositor_pipeline/src/queue/video_queue_thread.rs
WojciechBarczynski: It should determine whether frames should be forced pushed from the queue or not. In no place, does it take into consideration the buffer duration, so it can't possibly determine that correctly.

Let's walk through it:

This function checks:
```rust
    fn should_push_pts(&self, pts: Duration) -> bool {
        self.opts.clock_start.add(pts) < Instant::now()
    }
```

and is called in 

```rust
    fn on_queue_event(&mut self) {
        let mut internal_queue = self.queue.video_queue.lock().unwrap();
        let next_buffer_pts = self.get_next_output_buffer_pts();

        let ready_to_push = internal_queue
            .check_all_inputs_ready(next_buffer_pts, self.opts.clock_start)
            || self.should_push_pts(next_buffer_pts);
...
```

next_buffer_pts is determined by:
```rust
    fn get_next_output_buffer_pts(&self) -> Duration {
        Duration::from_secs_f64(
            self.sent_batches_counter as f64 * self.opts.output_framerate.den as f64
                / self.opts.output_framerate.num as f64,
        ) + self.output_frame_offset
    }
```

and output_frame_offset is set by:
```rust
    fn run(&mut self) {
        // This is just in case QueueThread is not spawned after the creation
        self.output_frame_offset = self.opts.clock_start.elapsed();
```

How possibly it correctly determines if frames should be pushed already if it doesn't consider the buffer duration? It only forces frames to be pushed every 1 / output_fps second. This results that while inputs are buffering, the queue pushes frames to the renderer.

Let's simplify this and say that buffer duration = 10s every input/output fps = 1 and input streams start exactly at the same moment as the start request.
It means, that the queue buffers 10s of the input stream (in input state code), while this function forces the queue to push frames. Before the buffering phase ends, the queue would already push 10 batches, since this function doesn't take into consideration, that it should wait for those 10s. This means, that the beginning 10s of the composed video wouldn't have input streams included.

My point is, that this should look sth like:
```rust
    fn should_push_pts(&self, pts: Duration) -> bool {
        self.opts.clock_start.add(pts) + self.buffer_duration < Instant::now()
    }
```

wkozyra95: No that is not correct. As I mentioned InputState abstraction is handling that. Value of pts itself is already adjusted to account for that.

noituri (commented):
I think it's ok 👍

compositor_pipeline/src/queue/utils.rs
WojciechBarczynski: 1.
I don't like that the typing of this function depends on the type in which the queue keeps frames/samples.

I thought before about keeping the audio samples in some structure that is ordered based on their pts (like ordered map / BST tree), because the audio decoder doesn't guarantee monotonical timestamps (RTP packets can be received non-monotonically and audio PTS come directly from as RTP packets PTSs).
Using the concrete `VecDeque` type here makes this code less flexible (sth like using `&Vec<T>` instead of `&[T]`)

In any case like this using iter and extend is more flexible, but if this iter would need to own data it comes with some overhead. 

I understand why you might want to keep this instead, so I leave it up to you.

2. If we don't want to use larger buffers in the near future
3. Discussed offline

WojciechBarczynski (approved):

PR #384, opened 2024-02-09
Implement control flow for elements after queue

wkozyra95:
The main change here is changing `send` in queue to `try_send`/`send_deadline`. Other changes in this PR are only partially related to control flow:

- remove code that drops output frames from the pipeline and encoder
- change all unbound queues to bound(5)
- refactor output registration code (just moving stuff around, no changes in behavior)
- for audio there is large buffer, so just try_send is fine
- for video, we have very small buffers, so `send_deadline` needs to be used to check against buffer size how long we can wait.

### DIFF
--- a/compositor_pipeline/src/queue/video_queue_thread.rs
+++ b/compositor_pipeline/src/queue/video_queue_thread.rs
@@ -7,6 +7,7 @@ use std::{
 
 use compositor_render::{FrameSet, Framerate, InputId};
 use crossbeam_channel::{tick, Sender};
+use log::warn;
 
 use super::Queue;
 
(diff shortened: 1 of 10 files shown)


compositor_pipeline/src/queue/video_queue_thread.rs
```diff
@@ -71,7 +72,14 @@ impl VideoQueueThread {
...
         }
-        self.sender.send(frames_batch).unwrap();
+        let send_deadline = self.opts.clock_start.add(frames_batch.pts);
+        if self
+            .sender
+            .send_deadline(frames_batch, send_deadline)
+            .is_err()
+        {
+            warn!("Dropping video frame on queue output.")
+        }
```
WojciechBarczynski, line 82: This deadline looks wrong to me. This exact timestamp looks like the time when output frames rendering should start. What happens if rendering takes some time? Won't this stop sending any frames to output?

wkozyra95: discussed offline

WojciechBarczynski (approved):

noituri (approved):

PR #396, opened 2024-02-16
Implement RTP over TCP on input

wkozyra95:
Add `transport_protocol` field to request that registers rtp input stream.  Option for tcp is called `tcp_server` to indicate the direction, but I don't plan to ever implement `tcp_client` variant.


WojciechBarczynski (commented):
LGTM

PR #397, opened 2024-02-19
Rewrite ChromiumSenderThread

### LINKED ISSUE #352: Rewrite `ChromiumSenderThread`

noituri:
I mostly moved things around and renamed them.

The important changes:
- Shared memory resize is now synchronized
  - `WebRendererThread` asks `process_helper` to stop using the shared memory (drop it) and awaits confirmation
  - `process_helper` drops shared memory handle and sends confirmation back to `WebRendererThread` (the confirmation has to go through `Browser client` before it reaches `WebRendererThread`)
  - `WebRendererThread` resizes the shared memory
- All of the communication with chrome happens through the `WebRendererThread`
  - Requests/messages are sent through channels to `WebRendererThread`
  - Requests that require a response, send response-channel together with the request. `WebRendererThread` uses the response channel to send back data.

Here's a rough sketch of how the communication works:
Untitled-2023-09-05-1308

Closes #352


WojciechBarczynski (approved):

compositor_render/src/transformations/web_renderer/web_renderer_thread/communication.rs
```diff
@@ -0,0 +1,103 @@
...
+use bytes::Bytes;
+use crossbeam_channel::{Receiver, Sender};
+
+use crate::{scene::ComponentId, transformations::layout::Position, Resolution};
+
+pub const EMBED_FRAMES_MESSAGE: &str = "EMBED_FRAMES";
+pub const DROP_SHARED_MEMORY: &str = "DROP_SHARED_MEMORY";
+pub const GET_FRAME_POSITIONS_MESSAGE: &str = "GET_FRAME_POSITIONS";
+
+pub struct ResponseSender<Response> {
```
wkozyra95, line 12: The whole point of adding this wrapper was to hide implementation details, the fact that channel are used internally, while here:
- We still need to create the channel and pass it here
- API is the same 

Using this abstraction is almost no different than just using `Option<Sender<x>>` directly. The problem that this abstraction was supposed to solve was the ability to wait for a result.

Also it does not help that you are making everything public event if its not used outside

compositor_render/src/transformations/web_renderer/renderer.rs
```diff
@@ -75,34 +90,188 @@ impl WebRenderer {
...
+            Ok(())
+        }
+    }
+
+    pub fn frame_data(&self) -> Result<Option<Bytes>, RenderWebsiteError> {
+        let (response_sender, response_receiver) = new_response_channel();
+        self.request_sender
+            .send(WebRendererThreadRequest::GetRenderedWebsite { response_sender })?;
+
+        Ok(response_receiver.recv()?)
```
wkozyra95, line 211:
```suggestion
        let result = self.request_sender.send(WebRendererThreadRequest::GetRenderedWebsite)?;
        Ok(result.wait()?)
```

It should look sth like this

PR #399, opened 2024-02-20
Implement RTP over TCP on output

wkozyra95:
- Implement RTP over TCP on output
- Fix RTP over TCP on input. I realized that stuff like read_exact is not safe when the socket can timeout. We are reasking that some of the data is read and timeout is triggered that discards it.

### DIFF
--- a/compositor_pipeline/src/pipeline/output/rtp.rs
+++ b/compositor_pipeline/src/pipeline/output/rtp.rs
@@ -1,6 +1,12 @@
 use compositor_render::OutputId;
 use log::{debug, error};
-use std::sync::Arc;
+use std::{
+    io::{self, Write},
+    sync::{atomic::AtomicBool, Arc},
+    thread,
+    time::Duration,
+    u16,
+};
 
 use rand::Rng;
 use rtp::packetizer::Payloader;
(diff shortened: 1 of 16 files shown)


compositor_pipeline/src/pipeline/output/rtp.rs
```diff
@@ -47,42 +63,142 @@ impl RtpSender {
...
+fn run_udp_sender_thread(
+    context: &mut RtpContext,
+    packets: Box<dyn Iterator<Item = EncodedChunk> + Send>,
+) {
+    for packet in packets {
         // TODO: check if this is h264
         let EncodedChunk { data, pts, .. } = packet;
 
-        let payloads = match context.payloader.payload(1500, &data) {
+        let payloads = match context.payloader.payload(1400, &data) {
```
WojciechBarczynski, line 201: MTU? I guess that padding is for headers etc., but you could explain where the magic value of 1400 comes from

compositor_pipeline/src/pipeline/rtp.rs
```diff
@@ -0,0 +1,62 @@
...
+    SocketBind(std::io::Error),
+    PortAlreadyInUse(u16),
+    AllPortsAlreadyInUse { lower_bound: u16, upper_bound: u16 },
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RequestedPort {
+    Exact(u16),
+    Range((u16, u16)),
+}
```
WojciechBarczynski, line 21: In some places, the range is a tuple, in others it has 2 named fields `lower_bound` and `upper_bound`. Any reasons for that inconsistency?

wkozyra95: after checking I don't think you are correct. We are using lower_bound and upper_bound only in 2 error types, everywhere else we use tuples to represent range

PR #402, opened 2024-02-21
Implement more interpolation types

noituri:


wkozyra95 (changes requested):
- Add documentation
- Expose interpolation type with just a raw bezier curve and params provided by user
- Some interpolation options will require params, so adjust API properlly

compositor_render/src/scene/transition/cubic_bezier.rs
```diff
@@ -0,0 +1,146 @@
...
+
+trait F64Ext {
+    fn is_close_to(&self, value: Self) -> bool;
+
+    fn clamp_valid_root_in_unit_range(self) -> Self;
+}
+
+impl F64Ext for f64 {
+    fn is_close_to(&self, value: Self) -> bool {
+        (*self - value).abs() < 2.0 * f64::EPSILON
```
jerzywilczek, line 112: are you sure these comparisons involving `f64::EPSILON` are correct? `f64::EPSILON` is the difference between 1.0 and the next representable number. I believe this means that there are no representable numbers (other than `2.0`) in the range `(2.0 - f64::EPSILON, 2.0 + f64::EPSILON)`, since the representable floats get more and more spread apart the further away from `0.0` you are. For a big part of all numbers, this comparison is equivalent to a simple `==` comparison. Am I wrong here? Is this the effect you wanted to achieve?

src/types/convert_util.rs
```diff
@@ -31,29 +31,77 @@ impl From<compositor_render::Resolution> for Resolution {
...
+            "ease_out_quint" => Ok(Self::EaseOutQuint),
+            "ease_in_out_quint" => Ok(Self::EaseInOutQuint),
+            "ease_in_expo" => Ok(Self::EaseInExpo),
+            "ease_out_expo" => Ok(Self::EaseOutExpo),
+            "ease_in_out_expo" => Ok(Self::EaseInOutExpo),
+            "bounce" => Ok(Self::Bounce),
+            "cubic_bezier" => Err(TypeError::new(
+                "Cubic bezier easing function needs control points specified. Please define it as an object with the control points provided."
+            )),
+            _ => Err(TypeError::new("Unknown easing function.")),
```
noituri, line 104: I'm not sure how to implement it better

compositor_render/src/scene/transition/cubic_bezier.rs
noituri: In the cases I tested, `2 * Epsilon` was enough. But I agree that it could be larger. I increased the allowed error to `1e-7`

src/types/convert_util.rs
WojciechBarczynski: Don't check it on serialization. Use a separate type for API and one for implementation. Put this logic in From<APItype> for BackendImplType and return type error on scene update. With 2 types this check is also forced by the type system and you don't have to write this weird deserialization. We do it like this in other places, so I would keep that approach as a project convention.

src/types/util.rs
```diff
@@ -15,6 +15,72 @@ pub struct Resolution {
...
+
+        fn visit_map<A>(self, map: A) -> Result<Self::Value, A::Error>
+        where
+            A: serde::de::MapAccess<'de>,
+        {
+            Deserialize::deserialize(serde::de::value::MapAccessDeserializer::new(map))
+        }
+    }
+
+    deserializer.deserialize_any(Visitor(std::marker::PhantomData))
```
WojciechBarczynski, line 83: Can be removed with approach mentioned above

src/types/convert_util.rs
noituri: The problem is that the enum value can be either represented as a string (for example: `"linear"`) or as an object:
```json
{
    "function_name": "cubic_bezier",
    "points": [0.0, 0.0, 1.0, 1.0]
}
```

I could simplify it by using only the object notation.

WojciechBarczynski: Not if you use:
```rust
#[serde(untagged)]
enum StringOrObject {
    String(String)
    Object(Object)
}
```

WojciechBarczynski: So it would look like:
```rust
#[serde(untagged)]
enum ApiBezierParams {
    BuiltIn(BuiltInInterpolations)
    Custom(CustomBezier)
}

enum BuiltIn {
    Linear,
    Ease,
    EaseIn,
    EaseOut,
    EaseInOut,
    EaseInQuint,
    EaseOutQuint,
    EaseInOutQuint,
    EaseInExpo,
    EaseOutExpo,
    EaseInOutExpo,
    Bounce,
}

struct CustomBezier {
    points: [f64; 4]
}
```

src/types/util.rs
```diff
@@ -15,6 +15,38 @@ pub struct Resolution {
...
+    String(EasingFunction),
+    Object { function_name: EasingFunction },
+    ObjectWithParams(EasingFunctionWithParams),
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+#[serde(tag = "function_name", rename_all = "snake_case")]
+pub enum EasingFunctionWithParams {
+    CubicBezier { points: [f64; 4] },
+}
```
WojciechBarczynski, line 33: I would simplify this:
```suggestion
#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
#[serde(untagged)]
pub enum EasingFunctionDefinition {
    BuiltIn(EasingFunction),
    Custom(CustomCubicBezierEasing),
}

#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
pub struct CustomCubicBezierEasing {
    cubic_bezier_points: [f64; 4],
}
```
such that in API it looks like:
```typescript
type Transition = {
  duration_ms: f64;
  easing_function?: BuiltInEasingFunction | CustomCubicBezierEasing;
}

type BuiltInEasingFunction =
  | "linear"
  | "ease"
  | "ease_in"
  | "ease_in_out"
  | "ease_in_quint"
  | "ease_out_quint"
  | "ease_in_out_quint"
  | "ease_in_expo"
  | "ease_out_expo"
  | "ease_in_out_expo"
  | "bounce"

type CustomCubicBezierEasing = {
    cubic_bezier_points: [f64, f64, f64, f64]
}
```

noituri: I'd prefer to keep the current API:
```json
{
  "function_name": "linear"
}
```
```json
{
  "function_name": "cubic_bezier",
  "control_points": [0, 0, 1, 1]
}
```

I think this is more consistent.

I will simplify the enums though

WojciechBarczynski: Ok

wkozyra95 (approved):

WojciechBarczynski (commented):
LGTM

PR #404, opened 2024-02-26
Non-blocking unregistered requests.

wkozyra95:
When we schedule unregistered requests, they happen on the queue thread, so they are blocking the queue.

This PR modify unregister logic to avoid blocking.

- Make channels on output as small as possible
- Now when output unregister happens
  - frames in channel between queue and renderer will be lost
  - frames in encoder and RTP UDP sender will still be delivered
  - frames in RTP TCP sender will be delivered unless:
    - TCP socket is not connected
    - TCP connection exists but writing takes longer than 30 seconds

### DIFF
--- a/compositor_pipeline/src/pipeline.rs
+++ b/compositor_pipeline/src/pipeline.rs
@@ -314,7 +314,7 @@ impl Pipeline {
             return;
         }
         info!("Starting pipeline.");
-        let (frames_sender, frames_receiver) = bounded(20);
+        let (frames_sender, frames_receiver) = bounded(1);
         // for 20ms chunks this will be 60 seconds of audio
         let (audio_sender, audio_receiver) = bounded(300);
         let renderer = self.renderer.clone();
--- a/compositor_pipeline/src/pipeline.rs
+++ b/compositor_pipeline/src/pipeline.rs
@@ -224,6 +224,7 @@ impl Pipeline {
 
         self.audio_mixer.unregister_output(output_id);
         self.outputs.remove(output_id);
+        self.renderer.unregister_output(output_id);
         Ok(())
     }
 
--- a/compositor_pipeline/src/pipeline/encoder/ffmpeg_h264.rs
+++ b/compositor_pipeline/src/pipeline/encoder/ffmpeg_h264.rs
@@ -5,7 +5,7 @@ use ffmpeg_next::{
     format::Pixel,
     frame, Dictionary, Packet, Rational,
 };
-use log::{error, warn};
+use tracing::{debug, error, warn};
 
 use crate::{
     error::EncoderInitError,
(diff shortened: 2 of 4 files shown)


compositor_pipeline/src/pipeline.rs
```diff
@@ -314,7 +314,7 @@ impl Pipeline {
             return;
         }
         info!("Starting pipeline.");
-        let (frames_sender, frames_receiver) = bounded(20);
+        let (frames_sender, frames_receiver) = bounded(1);
```
WojciechBarczynski, line 317: What about audio?

WojciechBarczynski (commented):
To avoid losing frames wouldn't it be better to pass though pipeline sth like:
```rust
enum Event {
    Frame(Frame)
    EndOfStream,
    ...
}
```
? I mean sth doing similar to what Membrane does.

This way frames/samples won't be lost and scheduling events like scene updates on EOS should be easier.

wkozyra95:
> To avoid losing frames wouldn't it be better to pass though pipeline sth like:
> 
> ```rust
> enum Event {
>     Frame(Frame)
>     EndOfStream,
>     ...
> }
> ```
> 
> ? I mean sth doing similar to what Membrane does.
> 
> This way frames/samples won't be lost and scheduling events like scene updates on EOS should be easier.

Yes we will need that, but what is your point here?
- To implement `Event` like you described we need changes from this PR.
- Adding Event does not help with dropping frames before the renderer.
- Adding Event does not halp with dropping frames after renderer because changes from this PR are already doing it.

I don't disagree with the need for that abstraction, but it has nothing to do with the problem.

compositor_pipeline/src/pipeline.rs
wkozyra95: audio is processed in sync  with video, so it will never buffer too much. It might be good to reconsider when audio support is finished, but for now I don't see point in speculating what should we do

WojciechBarczynski:
> Yes we will need that, but what is your point here?
> 
> To implement Event like you described we need changes from this PR.
> Adding Event does not help with dropping frames before the renderer.
> Adding Event does not halp with dropping frames after renderer because changes from this PR are already doing it.
> I don't disagree with the need for that abstraction, but it has nothing to do with the problem.

It should solve:

> - Now when output unregister happens
> - - frames in channel between queue and renderer will be lost

My point is, that dropping this channel this way is probably incorrect since we lose those frames

compositor_pipeline/src/pipeline.rs
WojciechBarczynski: Then why do those channels have different capacities?

wkozyra95: because sample rate and distance between frame is different, plus it does not hurt us to keep more audio

wkozyra95: also initially audio and video output was not synced, it was necessary to implement event scheduling

WojciechBarczynski (approved):

noituri (approved):

PR #405, opened 2024-02-26
Add EOS for inputs element

wkozyra95:
- Handle EOS in input elements (mostly pass around, no support for RTCP goodbey yet)
- Handle EOS in queue
  - if EOS received input is always marked as ready
  - if EOS received and get_frame returned last frame then remove it.

Output elements and sending EOS from queue will be handled in follow up PR

### DIFF
--- a/build_tools/nix/flake.nix
+++ b/build_tools/nix/flake.nix
@@ -55,6 +55,13 @@
           );
           devDependencies = with pkgs; [
             ffmpeg # to add ffplay
+
+            gst_all_1.gstreamer
+            gst_all_1.gst-plugins-base
+            gst_all_1.gst-plugins-good
+            gst_all_1.gst-plugins-bad
+            gst_all_1.gst-libav
+
             nodejs_18
             rustfmt
             clippy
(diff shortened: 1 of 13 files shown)


build_tools/nix/flake.nix
```diff
@@ -55,6 +55,13 @@
           );
           devDependencies = with pkgs; [
             ffmpeg # to add ffplay
+
+            gst_all_1.gstreamer
```
wkozyra95, line 59: for rtp_tcp example

compositor_pipeline/src/queue.rs
```diff
@@ -69,6 +67,20 @@ pub struct ScheduledEvent {
     callback: Box<dyn FnOnce() + Send>,
 }
 
+pub enum PipelineEvent<T> {
+    Data(T),
+    EOS,
+}
```
WojciechBarczynski, line 73: Shouldn't this be defined in `pipeline`?

WojciechBarczynski (approved):

compositor_pipeline/src/queue.rs
wkozyra95: It could be, I don't have strong opinion. I placed it here to contrast with ScheduledEvent above

noituri (approved):

PR #407, opened 2024-02-28
 Unblock dropping of the mp4 source

jerzywilczek:


compositor_pipeline/src/pipeline/input/mp4.rs
```diff
@@ -133,8 +116,14 @@ impl Mp4 {
...
-        drop(self.audio_thread.take());
+        if let Some(stop_video) = &self.stop_video_thread {
+            stop_video.store(true, std::sync::atomic::Ordering::Relaxed);
+        }
+
+        if let Some(stop_audio) = &self.stop_audio_thread {
+            stop_audio.store(true, std::sync::atomic::Ordering::Relaxed);
+        }
+
         if let Source::Url(_) = self.source {
```
wkozyra95, line 127: What are the risks here? I assume it won't cause crash because we handle read error, but I would prefer if unregistering input in the middle of writing would not cause some read errors/warning to be loged

wkozyra95 (approved):
One concern about file deletion. otherwise LGTM

compositor_pipeline/src/pipeline/input/mp4.rs
jerzywilczek: No risks, on unixes `unlink` will not delete the file while it's open, on windows it will throw an error in that case (which I handled gracefully)

PR #416, opened 2024-03-06
Add EOS on output + refactor

wkozyra95:
- Change some channels to bounded
- Ensure unified pattern / function naming for elements that spwan new threads
- Add some logs/spans
- FIX: The register renderer was blocking the rendering thread
- Remove some Arc/Mutex from the pipeline state.
- Changes related to EOS:
  - Dropping TCP connection on input will send EOS
  - Queue will send EOS
  - Render thread will mark input as EOS received in pipeline, but that information is not used yet
  - EOS will be passed through encoders and to the RTP output
  
Follow-up work:
- handle RTCP Goodbye packet on input
- send RTCP Goodbye packet  on output
- Add API that defines a relation between outputs and input stream (when to send EOS on output)


compositor_pipeline/src/pipeline.rs
```diff
@@ -321,129 +324,115 @@ impl Pipeline {
...
 
-impl<'a> Iterator for OutputIterator<'a> {
-    type Item = (&'a OutputId, &'a PipelineOutput);
+fn run_audio_mixer_thread(
+    pipeline: Arc<Mutex<Pipeline>>,
+    audio_receiver: Receiver<QueueAudioOutput>,
+) {
+    let audio_mixer = pipeline.lock().unwrap().audio_mixer.clone();
+    for mut samples in audio_receiver.iter() {
+        for (input_id, event) in samples.samples.iter_mut() {
```
wkozyra95, line 407: This is only marking input as EOS, but we need to add logic that check the inputs and decides whether to send EOS on output too. I'll add that in the follow-up.

PR #417, opened 2024-03-06
Extract the video reading logic from the mp4 component

jerzywilczek:
More will come in subsequent PRs

### DIFF
--- a/compositor_pipeline/src/pipeline/input/mp4.rs
+++ b/compositor_pipeline/src/pipeline/input/mp4.rs
@@ -1,33 +1,28 @@
 use std::{
-    fs::File,
-    io::Read,
     ops::ControlFlow,
-    os::unix::fs::MetadataExt,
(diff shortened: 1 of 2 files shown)


WojciechBarczynski (approved):

compositor_pipeline/src/pipeline/input/mp4.rs
```diff
@@ -116,10 +122,6 @@ impl Mp4 {
 
 impl Drop for Mp4 {
     fn drop(&mut self) {
-        if let Some(stop_video) = &self.stop_video_thread {
-            stop_video.store(true, std::sync::atomic::Ordering::Relaxed);
```
wkozyra95, line 120: If I understand correctly after this change Drop on Mp4 will happen before the drop on VideoReader, so we will start removing file before we even set should_close to true.

wkozyra95 (approved):

PR #418, opened 2024-03-06
Support RTCP Goodbye

wkozyra95:
- Handle RTCP Goodbye on input
- Handle RTCP Goodbye on output
- Change encoder output type to specify which EOS is sent

### DIFF
--- a/compositor_pipeline/src/pipeline/input/rtp.rs
+++ b/compositor_pipeline/src/pipeline/input/rtp.rs
@@ -12,6 +12,7 @@ use crate::{
 };
 use compositor_render::InputId;
 use crossbeam_channel::{bounded, Receiver, Sender};
+use rtcp::header::PacketType;
 use tracing::{debug, error, span, warn, Level};
 use webrtc_util::Unmarshal;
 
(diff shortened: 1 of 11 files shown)


compositor_pipeline/src/pipeline/input/rtp.rs
```diff
@@ -158,6 +159,27 @@ fn run_depayloader_thread(
     video_sender: Option<Sender<PipelineEvent<EncodedChunk>>>,
     audio_sender: Option<Sender<PipelineEvent<EncodedChunk>>>,
 ) {
+    let mut audio_eos_received = audio_sender.as_ref().map(|_| false);
+    let mut video_eos_received = video_sender.as_ref().map(|_| false);
```
noituri, line 163: I think this could be simplified by using `bool` instead of `Option<bool>`

noituri (approved):
LGTM

WojciechBarczynski (approved):
Some docs should document how compositor handles those messages

compositor_pipeline/src/pipeline/input/rtp.rs
wkozyra95: Yes, but it would only work because we want to do the same action for `None` and `Some(true)`, which are very different states. I prefer to keep the distinction.

PR #419, opened 2024-03-07
Log missing frames + lower stream fallback to 500 ms

wkozyra95:
When debuging slow membrane pipeline it was hard to identify the bottelnecek because we were not sure if compositor receives data or not.

- Add log when we are pusing samples/frames to renderer or mixer before all inputs are ready
- Lower fallback timeout from 2000ms to 500ms. (Unrelated to this issue, I just think this value is to large, it should significantlly larger than any reasonable distance between frames, but small enough so user does not notice the freeze)


WojciechBarczynski (commented):
LGTM

PR #421, opened 2024-03-07
Update docusaurus

### LINKED ISSUE #411: Update input stream docs
https://compositor.live/docs/api/components/InputStream/

Is outdated + href in note is not valid

wkozyra95:
Fixes #411 

The link mentioned in the issue is correct, but sometimes depending on navigation routing was redirecting to the wrong file. I can't reproduce that locally, but given that we used the first release of a new major version I suspect this bug should be fixed. I'll verify this after  release


WojciechBarczynski (commented):
LGTM

PR #422, opened 2024-03-08
Potentially fixes issue with high bitrate videos

wkozyra95:


noituri (approved):

WojciechBarczynski (approved):

wkozyra95:
The issue is happening when `tune` option is set to `zerolatency`, It resolves the issue partially, but for low enough CRF decoder starts throwing errors after some time

PR #425, opened 2024-03-11
Add integration tests

noituri:
Snapshots PR: https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/27


compositor_integration_tests/src/helpers.rs
```diff
@@ -0,0 +1,242 @@
...
+
+pub struct CompositorInstance;
+
+impl Drop for CompositorInstance {
+    fn drop(&mut self) {
+        // TODO(noituri): Clean up compositor inputs and outputs
+    }
+}
+
+impl CompositorInstance {
```
noituri, line 32: This will be changed

compositor_integration_tests/src/helpers.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// TODO(noituri): Implement this properly
+fn validate_packets(expected: &Packet, actual: &Packet) -> Result<()> {
+    if expected.header.payload_type != actual.header.payload_type {
+        return Err(anyhow::anyhow!("Payload type mismatch"));
+    }
+
+    Ok(())
+}
```
noituri, line 173: This is not working properly as of now. I need to decode the packets to make it work reasonably. I wanted to include it in this PR but there's already a lot of code so I'll do that in a follow-up PR.

compositor_integration_tests/src/tests.rs
```diff
@@ -0,0 +1,13 @@
+use anyhow::Result;
+
+use crate::CompositorInstance;
+
+use self::simple::run_simple_test;
+
+mod simple;
+
+pub fn integration_tests(update_dumps: bool) -> Result<()> {
+    run_simple_test(CompositorInstance::start(8000), update_dumps)?;
```
wkozyra95, line 10: I assume all tests are expected to be added to this function. Do we need to enforce running all the test sequentially? Can we maybe support  running them concurrentlly?

compositor_integration_tests/src/compositor_instance.rs
```diff
@@ -0,0 +1,96 @@
...
+            )?,
+        };
+
+        socket.connect(&SocketAddr::new(Ipv4Addr::new(127, 0, 0, 1).into(), port).into())?;
+
+        Ok(Self { protocol, socket })
+    }
+
+    pub fn send(&self, rtp_packets: &[u8]) -> Result<()> {
+        let mut sent_bytes = 0;
```
wkozyra95, line 78: You can use `write_exact` on socket

compositor_integration_tests/src/tests/simple.rs
```diff
@@ -0,0 +1,66 @@
...
+        "type": "start",
+    }))?;
+
+    let packets_dump = input_dump_from_disk("8_colors_input.rtp")?;
+    let sender = PacketSender::new(CommunicationProtocol::Udp, 8004)?;
+    sender.send(&packets_dump)?;
+
+    let new_output_dump = dump_output(
+        8002,
+        CommunicationProtocol::Tcp,
```
wkozyra95, line 52: How would this work for UDP? For udp you need to start listening before you register output, but on the other hand this is blocking. You need to rethink API of those helpers.

compositor_integration_tests/src/bin/update_output_dumps.rs
wkozyra95, line 1: `update_snapshots.rs`

compositor_integration_tests/src/bin/dump_packets_from_file.rs
wkozyra95, line 1: `generate_rtp_from_file.rs`

compositor_integration_tests/src/bin/dump_packets_from_file.rs
```diff
@@ -0,0 +1,23 @@
...
+        return;
+    }
+
+    let input_path = PathBuf::from(&args[1]);
+    let output_path = input_path
+        .parent()
+        .unwrap()
+        .join(input_path.file_stem().unwrap());
+    let dump_packets_command = format!(
+        "gst-launch-1.0 -v funnel name=fn filesrc location={input_path:?} ! qtdemux ! h264parse ! rtph264pay config-interval=1 pt=96  ! .send_rtp_sink rtpsession name=session .send_rtp_src ! fn. session.send_rtcp_src ! fn. fn. ! rtpstreampay ! filesink location={output_path:?}.rtp"
```
wkozyra95, line 16: Using `:?` for sth that is not a log seems wrong, maybe convert it to string explicitlly before

compositor_integration_tests/src/bin/play_dumped_packets.rs
wkozyra95, line 1: play_rtp/play_rtp_from_file/play_rtp_dump

compositor_integration_tests/src/tests.rs
wkozyra95: If we can run them concurrentlly, then I would expect each test to be separate function marked with `#[test]`, but I'm not sure how updating tests would work in that scenario. (Maybe it could be done with feature)

noituri: I did consider marking each test with `#[test]`, but as you said, updating those tests was problematic

compositor_integration_tests/src/compositor_instance.rs
noituri: I still need to track what bytes were sent for UDP because RTP dump includes packet length in front of each packet.  And packet length has to be omitted for UDP.

compositor_integration_tests/src/test_runner.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    use_global_wgpu_ctx();
+
+    let args = Arguments::from_args();
+    let tests = integration_tests()
+        .into_iter()
+        .map(|test| Trial::test(test.name, move || test.run_test().map_err(Into::into)))
+        .collect();
+
+    libtest_mimic::run(&args, tests).exit();
```
noituri, line 21: It's a custom test harness which makes each integration test behave like a normal test:
- `cargo test -- --list` - lists every integration test as a separate test
- It's compatible with `cargo test` and `cargo nextest`
- Each test is run in parallel unless tests are run with `--test-threads=1` flag

wkozyra95: Also is there a reason why you are introducing new lib instead of just using regular rust test?

wkozyra95 (approved):

compositor_integration_tests/src/test_runner.rs
noituri: Sorry, I forgot to commit and push the latest changes. I removed `libtest-mimic` in favor of normal tests.
The problem I had with normal tests was that you can't call test functions like normal functions. You'd have to define a wrapper function for each test. For that reason, I decided to use that lib. It would remove some boilerplate from test definition.

In the end, I decided to use normal tests because they are more readable since there's less 'magic' involved.

Please check the latest commit.

compositor_integration_tests/src/helpers.rs
```diff
@@ -0,0 +1,213 @@
+use anyhow::{Context, Result};
```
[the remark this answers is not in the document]
noituri: I will address this in a follow-up PR

wkozyra95 (approved):

PR #429, opened 2024-03-12
Extract the audio reader

jerzywilczek:
This is done by stopping to use symphonia and using the `mp4` library instead.

The existing video reader code was generalized to avoid code duplication.

I decided to keep all HLS-related PRs and patches on a separate branch and only megre it to `master` when we're sure it's ready and doesn't break current functionalities.


wkozyra95:
It's up to you but I would keep in mind that maintaining the separate branch might be hard (depends on your plans).

An alternative approach you can consider is to add some feature flag that changes the behavior.

Few questions:
- Are you planinng to update this branch (`hls`) with changes from master regularly or just when merging at the end. If yes then via `merge` or `rebase`?
- When you will be merging `hls` branch to master are you plan to squash everything or merge?

Also, hls branch is already few commits behind master, so please rebase it.

jerzywilczek:
@wkozyra95 
> - Are you planning to update this branch (hls) with changes from master regularly or just when merging at the end. If yes then via merge or rebase?

I wanted to only update it at the end, probably using `rebase`

> - When you will be merging hls branch to master are you plan to squash everything or merge?

I wanted to squash

This whole affair shouldn't take long, probably a week or so.

compositor_pipeline/src/pipeline/decoder/fdk_aac.rs
```diff
@@ -88,6 +64,30 @@ fn run_decoder_thread(
...
             }
         };
+
+        let decoder = match &decoder {
+            Some(d) => d,
+            None => {
+                decoder = match Decoder::new(
+                    std::mem::take(&mut options)
+                        .expect("AAC decoder initialization options should be present"),
+                    &chunk,
```
wkozyra95, line 74: refactor this, initializing decoder this way is very hacky

compositor_pipeline/src/pipeline/input/mp4.rs
```diff
@@ -98,183 +97,45 @@ impl Mp4 {
             None => (None, None, None),
         };
 
-        let audio_reader_info = spawn_audio_reader(&input_path, options.input_id.clone())?;
+        let (audio_reader, audio_receiver, audio_deocder_options) = match Mp4FileReader::new_audio(
```
wkozyra95, line 100: extract matched value to variable for readbility

compositor_pipeline/src/pipeline/input/mp4/mp4_file_reader.rs
```diff
@@ -0,0 +1,369 @@
...
+                    ),
+                    dts: Some(dts),
+                    kind: track_info.chunk_kind,
+                };
+
+                let mut chunk = Some(PipelineEvent::Data(chunk));
+                loop {
+                    match sender.send_timeout(chunk.take().unwrap(), Duration::from_millis(50)) {
+                        Ok(()) => {
+                            continue 'samples;
```
wkozyra95, line 344: if you need labels then it's a sign to refactor, especially here. A function that tries to write to a channel while periodically checking for should_close could be useful in few places.

compositor_pipeline/src/pipeline/input/mp4/mp4_file_reader.rs
```diff
@@ -0,0 +1,369 @@
...
+    sample_count: u32,
+    timescale: u32,
+    track_id: u32,
+    decoder_options: DecoderOptions,
+    sample_unpacker: SampleUnpacker,
+    chunk_kind: EncodedChunkKind,
+}
+
+impl Mp4FileReader<AudioDecoderOptions> {
+    pub fn new_audio(
```
wkozyra95, line 43: It's not specific to this function. Try to avoid making everything pub by default, other than options most stuff should have `pub(super)` or `pub(crate)`

compositor_pipeline/src/pipeline/decoder/fdk_aac.rs
jerzywilczek: i refactored it a bit, see if it's better now

wkozyra95: The main problem with that code is that you are initializing that in the loop.

WojciechBarczynski (commented):
LGTM

PR #431, opened 2024-03-12
New audio mixer

WojciechBarczynski:
Adds new audio mixer with 2 available mixing strategies:
- `sum_clip` - Sums input waves and clip result if it's outside of the PCM range
- `sum_scale` - Sums input waves and scales down the result if it's outside of the PCM range

Currently, all audio in the pipeline before the mixer is always in stereo format for simplicity (if the input is mono, then it's duplicated for both channels). The mixer mixes audio and converts it to mono if necessary.

TODO:
- resampling for AAC in different sample rates then set `output_sample_rate`

### DIFF
--- a/compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
+++ b/compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
@@ -142,7 +142,7 @@ impl InternalAudioMixer {
     }
 
     fn samples_in_frame(start: Duration, end: Duration, sample_rate: u32) -> usize {
-        (end.saturating_sub(start).as_secs_f64() * sample_rate as f64).round() as usize
+        (end.saturating_sub(start).as_nanos() * sample_rate as u128 / 1_000_000_000) as usize
     }
 
     fn push_missing_samples(samples_buffer: &mut Vec<(i16, i16)>, samples_count: usize) {
--- a/compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
+++ b/compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
@@ -1,12 +1,15 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, time::Duration};
 
 use compositor_render::{error::UpdateSceneError, InputId, OutputId};
 
 use crate::audio_mixer::{InputParams, MixingStrategy};
 
-use super::types::{
-    AudioChannels, AudioMixingParams, AudioSamples, InputSamplesSet, OutputSamples,
-    OutputSamplesSet,
+use super::{
+    types::{
+        AudioChannels, AudioMixingParams, AudioSamples, InputSamplesSet, OutputSamples,
+        OutputSamplesSet,
+    },
+    InputSamples,
 };
 
 #[derive(Debug)]
@@ -60,7 +63,7 @@ impl InternalAudioMixer {
 
     pub fn mix_samples(&mut self, samples_set: InputSamplesSet) -> OutputSamplesSet {
         let start_pts = samples_set.start_pts;
-        let (input_samples, samples_count) = self.input_samples(samples_set);
+        let (input_samples, samples_count) = self.merge_fill_input_samples(samples_set);
 
         OutputSamplesSet(
             self.outputs
@@ -73,7 +76,7 @@ impl InternalAudioMixer {
         )
     }
 
-    fn input_samples(
+    fn merge_fill_input_samples(
         &self,
         input_samples_set: InputSamplesSet,
     ) -> (HashMap<InputId, Vec<(i16, i16)>>, usize) {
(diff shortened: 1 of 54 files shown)


compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -189,4 +73,200 @@ impl InternalAudioMixer {
...
+    let mixed: Vec<(i16, i16)> = match output_info.audio.mixing_strategy {
+        MixingStrategy::SumClip => summed_samples
+            .into_iter()
+            .map(|(l, r)| (clip(l), clip(r)))
+            .collect(),
+        MixingStrategy::SumScale => {
+            // abs panics in debug if val = i64::MIN, but it would so many i16 samples to sum, that it'll never happen.
+            // Assumes that v is not empty (therefore unwrap is safe)
+            let max_abs =
+                |v: &Vec<(i64, i64)>| v.iter().map(|(l, r)| (l.abs().max(r.abs()))).max().unwrap();
```
wkozyra95, line 235: this is hard to read, extract to separate function, do not try to cramp everything into one line

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -189,4 +73,200 @@ impl InternalAudioMixer {
...
+
+            let scaling_factor = |max_abs: i64| {
+                if max_abs > i16::MAX as i64 {
+                    max_abs as f64 / i16::MAX as f64
+                } else {
+                    1.0
+                }
+            };
+
+            let scale = |v: Vec<(i64, i64)>| {
```
wkozyra95, line 245: You seem to overuse clousure far too much. max_abs and clip are small examples of that but here it does not make sense at all. function scale and scaling factor are only called once, why are you creating closure for that?

src/config.rs
```diff
@@ -160,6 +160,7 @@ fn read_config() -> Result<Config, String> {
         queue_options: QueueOptions {
             ahead_of_time_processing,
             output_framerate: framerate,
+            output_sample_rate,
```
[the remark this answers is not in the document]
WojciechBarczynski: We need that to calculate end pts of input samples batches

compositor_pipeline/src/pipeline/decoder/opus.rs
```diff
@@ -71,21 +62,27 @@ fn run_decoder_thread(
                 }
             };
 
-        let mut decoded_samples = Vec::with_capacity(decoded_samples_count / 2);
+        let mut left = Vec::with_capacity(decoded_samples_count / 2);
+        let mut right = Vec::with_capacity(decoded_samples_count / 2);
         for i in 0..decoded_samples_count {
-            decoded_samples.push((buffer[2 * i], buffer[2 * i + 1]));
+            left.push(buffer[2 * i]);
+            right.push(buffer[2 * i + 1]);
```
noituri, line 69: Are `left` and `right` capacities correct? It looks like you iterate over every decoded sample

WojciechBarczynski: No. Moreover, this is useless. I forgot to remove it

compositor_pipeline/src/audio_mixer/types.rs
```diff
@@ -70,6 +93,18 @@ impl AudioSamples {
     }
 }
 
+impl Debug for InputSamples {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
```
wkozyra95, line 98: There is a helper for struct debug implementations
https://github.com/membraneframework/video_compositor/blob/fa44592411ac7bd4d137ca40f6f690ff6196615a/compositor_render/src/types.rs#L24

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -189,4 +73,204 @@ impl InternalAudioMixer {
...
+    let mut summed_samples = vec![(0i64, 0i64); samples_count];
+
+    for input_params in inputs {
+        let Some(input_samples) = input_samples.get(&input_params.input_id) else {
+            continue;
+        };
+        summed_samples
+            .iter_mut()
+            .zip(input_samples.iter())
+            .for_each(|(sum, s)| {
```
wkozyra95, line 256: This is fine, but I would write a for loop here.

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -189,4 +73,204 @@ impl InternalAudioMixer {
...
+                sum.0 += (s.0 as f64 * input_params.volume as f64) as i64;
+                sum.1 += (s.1 as f64 * input_params.volume as f64) as i64;
+            })
+    }
+
+    summed_samples
+}
+
+fn convert_channels(samples: Vec<(i16, i16)>, channels: AudioChannels) -> AudioSamples {
```
wkozyra95, line 265: I would inline this function

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -189,4 +73,204 @@ impl InternalAudioMixer {
...
+                let max_abs = summed_samples
+                    .iter()
+                    .map(|(l, r)| (l.abs().max(r.abs())))
+                    .max()
+                    .unwrap();
+                if max_abs > i16::MAX as i64 {
+                    max_abs as f64 / i16::MAX as f64
+                } else {
+                    1.0
+                }
```
wkozyra95, line 223:
```suggestion
                let max_abs = summed_samples
                    .iter()
                    .map(|(l, r)| i64::max(l.abs(), r.abs()))
                    .max()
                    .unwrap();
                f64::max(max_abs as f64 / i16::MAX as f64, 1.0)
```

jerzywilczek (approved):

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -189,4 +79,194 @@ impl InternalAudioMixer {
...
+
+/// Mix input samples accordingly to provided specification.
+fn mix(
+    input_samples: &HashMap<InputId, Vec<(i16, i16)>>,
+    output_info: &OutputInfo,
+    samples_count: usize,
+) -> AudioSamples {
+    /// Clips sample to i16 PCM range
+    fn clip_to_i16(sample: i64) -> i16 {
+        i64::min(i64::max(sample, i16::MIN as i64), i16::MAX as i64) as i16
```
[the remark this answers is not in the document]
noituri: Wouldn't it be better to just use `sample.clamp(i16::MIN as i64, i16::MAX as i64)`?

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -189,4 +73,204 @@ impl InternalAudioMixer {
...
+                .iter()
+                .map(|(output_id, output_info)| {
+                    let samples = mix(&input_samples, output_info, samples_count);
+                    (output_id.clone(), OutputSamples { samples, start_pts })
+                })
+                .collect(),
+        )
+    }
+
+    fn save_samples(&mut self, samples: &HashMap<InputId, Vec<InputSamples>>) {
```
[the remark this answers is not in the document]
WojciechBarczynski: I removed that entirely

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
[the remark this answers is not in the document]
WojciechBarczynski: Yep, that looks better :)

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -189,4 +60,184 @@ impl InternalAudioMixer {
...
+            });
+
+        let missing_samples = Self::samples_in_frame(start_pts, end_pts, sample_rate);
+        Self::push_missing_samples(&mut samples_in_frame, missing_samples);
+
+        samples_in_frame
+    }
+
+    fn samples_in_frame(start: Duration, end: Duration, sample_rate: u32) -> usize {
+        (end.saturating_sub(start).as_secs_f64() * sample_rate as f64).round() as usize
```
wkozyra95, line 150:
```suggestion
        end.saturating_sub(start).as_nanos() * sample_rate as u128 / 1000_000_000
```

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -189,4 +60,184 @@ impl InternalAudioMixer {
...
+                    (sample_time_after_frame * sample_rate as f64).round() as usize;
+                let end_index = input_samples.len() - samples_after_frame;
+
+                samples_in_frame.extend(input_samples.samples[start_index..end_index].iter());
+
+                input_samples.end_pts
+            });
+
+        let missing_samples = Self::samples_in_frame(start_pts, end_pts, sample_rate);
+        Self::push_missing_samples(&mut samples_in_frame, missing_samples);
```
wkozyra95, line 144: I don't like how this works. It seems like a catch all case to hide any errors in the code above.

WojciechBarczynski: No, it's here to fill samples in case the last batch for some input isn't pushed from the queue to the mixer. I'm talking about a case in which the input samples set looks like this:
```rust
InputSamplesSet {
    start_pts: 0ms,
    end_pts: 20ms,
    samples: {
           input_1: [Samples{start_pts: 0ms, end_pts: 10ms}]
    }
}
```
where the next batch for input_1 was received yet.

This should fill time [10ms to 20ms].

I also want to guarantee, that this function returns Vec with all samples needed for mixing in [set_start_pts, set_end_pts]. The previous loop can introduce rounding errors resulting in missing a few samples (calculating those indexes isn't (and I'm pretty sure it's extremely hard to make it) numerically safe).


Another approach would be to return the last end pts from this fold and fill the samples missing in [last_input_samples_end_pts, set_end_pts], but it doesn't deal with the index calculation errors mentioned above.

What do you exactly suggest here?

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -189,4 +60,184 @@ impl InternalAudioMixer {
...
+                let sample_time_after_frame =
+                    input_samples.end_pts.saturating_sub(end_pts).as_secs_f64();
+                let samples_after_frame =
+                    (sample_time_after_frame * sample_rate as f64).round() as usize;
+                let end_index = input_samples.len() - samples_after_frame;
+
+                samples_in_frame.extend(input_samples.samples[start_index..end_index].iter());
+
+                input_samples.end_pts
+            });
```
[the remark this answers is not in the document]
WojciechBarczynski: I think that it's better to check:
```RUST
// SOME_SMALL_POSITIVE_VAL is sth like 3 or 5
missing_samples > SOME_SMALL_POSITIVE_VAL 
```
rather than:
```RUST
missing_samples > 0 
```
There are only a few reasons for missing samples being >0 here:
1.  Compositor hasn't received/read some input samples. eg. compositor received/read samples [0ms, 20ms] and [40ms, 60ms] (lost packets, missing audio fragments, etc.)
2. Audio is correct, but some PTSs are a little off, due to converting between different formats and numerically unsafe calculation of missing samples (eg. converting from RTP range timestamps to Duration can introduce small errors in PTS values)

We should fill samples only in the 1st case.

I think that with values like `missing_samples = 1` or `missing_samples = 2` it's significantly more probable that it's the 2nd reason. Nobody sends packets with only a few audio samples or puts 1/44100s of audio into the MP4 segment.

PR #437, opened 2024-03-18
Add video validation for integration tests

noituri:
- Implemented video dump validation
- Split one large file into multiple ones


wkozyra95 (approved):
LGTM, but in follow-up PRs try to write some tests that verify if this API is flexible enough for tests we need to write.

Few examples:
- TCP and UDP on inputs and outputs
- What happens when you send a few packets on input, sleep for a bit, and then continue (with required: true and false)
- What happens if you push a lot of data and immediately unregister

PR #439, opened 2024-03-18
Abort registration if initial scene is not correct

wkozyra95:
Currently, if the initial scene is not correct, the output registration will throw an error but output will still be registered.

### DIFF
--- a/compositor_pipeline/src/pipeline/pipeline_output.rs
+++ b/compositor_pipeline/src/pipeline/pipeline_output.rs
@@ -81,7 +81,18 @@ impl Pipeline {
             }),
         };
 
-        self.outputs.insert(output_id.clone(), output);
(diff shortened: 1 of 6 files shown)


compositor_pipeline/src/pipeline/pipeline_output.rs
```diff
@@ -91,12 +102,7 @@ impl Pipeline {
             );
         }
 
-        self.update_output(
-            output_id.clone(),
-            video.map(|v| v.initial),
-            audio.map(|a| a.initial),
-        )
-        .map_err(|e| RegisterOutputError::SceneError(output_id, e))?;
```
noituri, line 99: Is it not needed?

WojciechBarczynski (approved):

PR #441, opened 2024-03-18
Fix integration tests sometimes failing

noituri:
I couldn't reproduce the problem locally but I believe it's caused by wrongly chosen timestamps.

The test video consists of images where each image appears for 1 second. I used timestamps that were at the transition between images

### DIFF
--- a/compositor_integration_tests/src/tests/simple.rs
+++ b/compositor_integration_tests/src/tests/simple.rs
@@ -64,7 +64,7 @@ pub fn simple_test() -> Result<()> {
     compare_dumps(
         &output_dump_from_disk,
         &new_output_dump,
-        &[Duration::from_secs(1), Duration::from_secs(2)],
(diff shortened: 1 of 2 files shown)


compositor_integration_tests/src/tests/simple.rs
```diff
@@ -64,7 +64,7 @@ pub fn simple_test() -> Result<()> {
     compare_dumps(
         &output_dump_from_disk,
         &new_output_dump,
-        &[Duration::from_secs(1), Duration::from_secs(2)],
+        &[Duration::from_millis(500), Duration::from_millis(1500)],
```
WojciechBarczynski, line 67:
```suggestion
        &[Duration::from_millis(500), Duration::from_millis(2137)],
```

WojciechBarczynski (approved):

wkozyra95 (approved):

PR #445, opened 2024-03-20
improve audio mixer calculations

wkozyra95:


compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -109,45 +126,159 @@ impl InternalAudioMixer {
...
+                start_pts + Duration::from_secs_f64(sample_count as f64 / sample_rate as f64);
+
+            // potentially fill missing spots
+            if expected_next_sample_start_pts + max_error < input_samples.start_pts {
+                let missing_time = input_samples
+                    .start_pts
+                    .saturating_sub(expected_next_sample_start_pts);
+                let missing_samples_count = time_to_sample_count(missing_time);
+                if missing_samples_count < 1 {
+                    warn!(
```
wkozyra95, line 174: if distance is larger than max_error but less than one sample then most likely this is a bug

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -109,45 +126,159 @@ impl InternalAudioMixer {
...
+                        "Decoder produced overlapping samples. Dropping {} samples.",
+                        samples_to_remove_from_start
+                    );
+                }
+                start_range = samples_to_remove_from_start;
+            };
+
+            // check if we need to drop samples at the end
+            let mut end_range = input_samples.len();
+            if input_samples.end_pts > end_pts + max_error {
```
wkozyra95, line 204: I was considering if we should use `input_sample.end_pts` directly or calculate end pts based on `input_sample.start_pts` and amount of samples.

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -109,45 +126,159 @@ impl InternalAudioMixer {
...
+            let expected_next_sample_start_pts =
+                start_pts + Duration::from_secs_f64(sample_count as f64 / sample_rate as f64);
+
+            // potentially fill missing spots
+            if expected_next_sample_start_pts + max_error < input_samples.start_pts {
+                let missing_time = input_samples
+                    .start_pts
+                    .saturating_sub(expected_next_sample_start_pts);
+                let missing_samples_count = time_to_sample_count(missing_time);
+                if missing_samples_count < 1 {
```
WojciechBarczynski, line 173: Did you want to use `<` here? Shoudn't it be `>`?

wkozyra95: no, it's perfectly normal to have a missing samples, but it's a bug if the gap is larger than numerical error but still so small

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -109,45 +126,159 @@ impl InternalAudioMixer {
...
+                let time_to_remove_from_start =
+                    expected_next_sample_start_pts.saturating_sub(input_samples.start_pts);
+                let samples_to_remove_from_start = time_to_sample_count(time_to_remove_from_start);
+                if batch_index != 0 {
+                    // We should only drop samples in the first batch.
+                    warn!(
+                        "Decoder produced overlapping samples. Dropping {} samples.",
+                        samples_to_remove_from_start
+                    );
+                }
```
WojciechBarczynski, line 198: How can you know that at this point? Those PTS values are changed multiple times across the pipeline. Here, in the mixer, in the queue, maybe in the resampler. How do you know that there is not a bug in any of those places, that caused that? Well, I don't. Logging such warnings at this point can point developers to the wrong place. This is even worse than no message at all.

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
WojciechBarczynski, line 204: Same.

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -109,45 +126,159 @@ impl InternalAudioMixer {
...
+                Self::ensure_correct_amount_of_samples(
+                    start_pts,
+                    end_pts,
+                    sample_rate,
+                    &mut samples_in_frame,
+                );
+            }
+        }
+
+        Self::check_input_batch(start_pts, end_pts, sample_rate, &samples_in_frame);
```
WojciechBarczynski, line 225:
```suggestion
        Self::check_samples_count(start_pts, end_pts, sample_rate, &samples_in_frame);
```

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -101,6 +102,22 @@ impl InternalAudioMixer {
             .collect()
     }
 
+    // Produce continuous batch of samples for range (start_pts, end_pts).
```
WojciechBarczynski, line 105: Yep, and does like 10 different input validations along the way. I don't like those validations in the middle of this merging logic. I understand why you want to validate if the queue output is right (if I wrote it, I wouldn't be sure either :P), but this should be done before calling this function, not in the middle of it

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -109,45 +126,159 @@ impl InternalAudioMixer {
...
+    ) {
+        let samples_count_times_1e9 =
+            end_pts.saturating_sub(start_pts).as_nanos() * sample_rate as u128;
+        if samples_count_times_1e9 % 1_000_000_000 != 0 {
+            warn!(
+                "Duration {:?} is not divisible by sample duration (sample rate: {}).",
+                end_pts.saturating_sub(start_pts),
+                sample_rate,
+            )
+        }
```
WojciechBarczynski, line 253: This function is called in one place - the end of the `frame_input_samples` function. I don't think this is the correct place for such validation.

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -109,45 +126,159 @@ impl InternalAudioMixer {
...
+        if samples.len() as u128 != samples_count_times_1e9 / 1_000_000_000 {
+            warn!(
+                "Wrong amount of samples generated. Expected: {}, Actual: {}.",
+                samples_count_times_1e9 / 1_000_000_000,
```
WojciechBarczynski, line 257: This 1e9 value repeats over and over. Create a const like:
```rust
const NANOS_IN_SEC: u128 = 1_000_000_000;
```

compositor_pipeline/src/audio_mixer/internal_audio_mixer.rs
```diff
@@ -109,45 +126,159 @@ impl InternalAudioMixer {
...
+
+        // Fill at the end only if last batch is ending to quickly
+        if let Some(pts) = last_batch_end_pts {
+            if pts < end_pts + max_error {
+                Self::ensure_correct_amount_of_samples(
+                    start_pts,
+                    end_pts,
+                    sample_rate,
+                    &mut samples_in_frame,
+                );
```
WojciechBarczynski, line 221: You call the function that ensures the size and can potentially drain some samples. The if statement above should guarantee that it's not the case, but in similar places, you check for this explicitly and log errors. I don't like logging errors like this, but if you do it, make it consistent.

WojciechBarczynski (commented):
This took me over an hour to understand what it does (and I still don't understand it fully).

I think it's not an improvement. I think that modifying code from the master, just to calculate missing samples at the end using the `last_batch_pts` and check if the samples count in returned vec is in the range [expected_samples_count - 1; expected_samples_count + 1] does pretty much the same as those 150+ lines of hard to understand code.

WojciechBarczynski (approved):

PR #446, opened 2024-03-20
Merge the hls branch

jerzywilczek:
Even if we decide the in-compositor HLS support is unnecessary, this branch contains improvements to the mp4 reader which should be merged.


wkozyra95 (approved):
Maybe change PR to sth like "Add base HLS implementation".

Just to be sure, nothing HLS related is available via API?

WojciechBarczynski (approved):

PR #447, opened 2024-03-20
Add an example demonstrating playing hls video using gstreamer

jerzywilczek:


WojciechBarczynski (commented):
Screenshot 2024-03-20 at 22 03 57

WojciechBarczynski:
> Doesn't work for me Screenshot 2024-03-20 at 22 03 57

I checked for any gstreamer errors, but I could see any. I assume I might be missing some gstreamer packages, but I want this code to inform users in such case exactly what went wrong. The current behaviour looks like a compositor failure and it certainly shouldn't be the first impression we want to deliver with examples.

wkozyra95:
> I checked for any gstreamer errors, but I could see any. I assume I might be missing some gstreamer packages, but I want this code to inform users in such case exactly what went wrong.

This is not really a feasible expectation, we can print the error that command x failed, but that is not really what the exact problem is. We are showing how to use compositor with GStreamer, this is not about providing a better experience for GStreamer or teaching users how to use GStreamer itself.

How would you even detect the problem in a generic way?

> The current behavior looks like a compositor failure and it certainly shouldn't be the first impression we want to deliver with examples.

We can be more clear that gst command failed (if it failed).

noituri (approved):
It does work, but it lags quite a lot

PR #448, opened 2024-03-20
Refactor internal audio mixer abstraction

wkozyra95:
Move stuff around, implement simples methods in AudioMixer directly instead of passing them to InternalAudioMixer


WojciechBarczynski (approved):
I think splitting this is a good idea, but I have suggestions regarding the calculations themselves, which I'll address in the review of #445

PR #454, opened 2024-03-24
Make configuration docs consistent

WojciechBarczynski:
I noticed that configuration docs were inconsistent

Added missing defaults, fixed the wrong ones, recorded config code to match docs (it's easier to check it once it follows the same order)

### DIFF
--- a/docs/pages/deployment/configuration.md
+++ b/docs/pages/deployment/configuration.md
@@ -4,7 +4,7 @@
 
 ### `LIVE_COMPOSITOR_API_PORT`
 
-API port. Defaults to 8001.
+API port. Defaults to 8081.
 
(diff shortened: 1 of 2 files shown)


docs/pages/deployment/configuration.md
```diff
@@ -4,19 +4,19 @@
 
 ### `LIVE_COMPOSITOR_API_PORT`
 
-API port. Defaults to 8001.
+API port. Defaults to 8081.
```
WojciechBarczynski, line 7: I wasn't sure if it should be 8001 or 8081, but it was 8081 in code, so I decided to change docs

wkozyra95 (approved):

PR #462, opened 2024-03-27
Fix prepare audio input error check + add tests

wkozyra95:


compositor_pipeline/src/audio_mixer/prepare_inputs.rs
```diff
@@ -71,8 +76,8 @@ fn frame_input_samples(
     let time_to_sample_count = |duration: Duration| {
         let sample_count = duration.as_secs_f64() * sample_rate as f64;
         // If value is close to the integer then round it, otherwise fallback to standard
-        // integer division behavior.
-        if (sample_count - sample_count.round()).abs() < max_error.as_secs_f64() {
+        // integer division behavior. Close is defined as 1% of a sample (the same as max_error).
+        if (sample_count - sample_count.round()).abs() < 0.01 {
```
wkozyra95, line 80: max error represents the duration that reflects 1% of a sample and in this condition, we compare sample count not duration. 

max_error should be about 1e-8 seconds and sample duration is about 1e-5 seconds so the old implementation was only allowing for errors of the order 1e-13 seconds

compositor_pipeline/src/audio_mixer/prepare_inputs.rs
```diff
@@ -71,8 +76,8 @@ fn frame_input_samples(
     let time_to_sample_count = |duration: Duration| {
         let sample_count = duration.as_secs_f64() * sample_rate as f64;
         // If value is close to the integer then round it, otherwise fallback to standard
-        // integer division behavior.
-        if (sample_count - sample_count.round()).abs() < max_error.as_secs_f64() {
+        // integer division behavior. Close is defined as 1% of a sample (the same as max_error).
```
WojciechBarczynski, line 79: [NIT]
```suggestion
        // Integer division behavior. Close is defined as 1% of a sample (the same as max_error).
```

compositor_pipeline/src/audio_mixer/prepare_inputs/consecutive_frames_tests.rs
```diff
@@ -0,0 +1,246 @@
...
+                    start_pts: third_batch_start,
+                    end_pts: third_batch_start + (4 * sample_duration)
+                }
+            ],
+            sample_rate
+        ),
+        vec![(7, 7), (8, 8), (9, 9), (10, 10), (11, 11), (12, 12)]
+    );
+
+    // shifted by small_error (subtract) + batches overlapping between frames
```
WojciechBarczynski, line 157: Where do those overlap? I don't understand this

WojciechBarczynski (approved):

compositor_pipeline/src/audio_mixer/prepare_inputs/consecutive_frames_tests.rs
wkozyra95: third_batch_start is slightly smaller than second_batch_start, but output is still continous

compositor_pipeline/src/audio_mixer/prepare_inputs.rs
wkozyra95: This is not the start of the sentence, check line above

WojciechBarczynski: Ok, nevermid, my bad :)

PR #463, opened 2024-03-27
Add docs for `/--/ws` endpoint

### LINKED ISSUE #408: New mechanism for sending events back to the client
Currently `WaitForNextFrame` query waits until an input frame is ready to be rendered. That means:
- The decoded input frame is already in the buffer 
- The input's offset <  current PTS
- Pipeline processing is already started (`start` request was sent)

The problem is when we want to check whether an input is ready before the pipeline processing has started. The query request will be blocked until the pipeline is started + the above requirements are met.

We potentially want to support these cases:
1.  Next frame is ready to be rendered 
2. Next frame was decoded and is in the buffer

Things to consider in the 2nd case:
- When should we take an input's offset into account? Is a frame ready when it's decoded, or is it ready when it's decoded + `offset < current PTS`

wkozyra95:
Closes #408

### DIFF
--- a/docs/pages/api/events.md
+++ b/docs/pages/api/events.md
@@ -1,0 +1,81 @@
+---
+description: WebSocket events
+---
+
+# Events
+
+LiveCompositor is using WebSocket connection to send events to the connected clients. Supported events are listed below.
+
+### `VIDEO_INPUT_DELIVERED`
+
+```typescript
+type Event = {
+  type: "VIDEO_INPUT_DELIVERED";
+  input_id: string;
+}
+```
+
+The compositor received the input, and the first frames of that input are ready to be used. If you want to ensure that some inputs are ready before you send the [`start`](./routes.md#start) request, you can wait for those events for specific inputs.
+
+### `VIDEO_INPUT_PLAYING`
+
+```typescript
+type Event = {
+  type: "VIDEO_INPUT_PLAYING";
+  input_id: string;
+}
+```
+
+The compositor received the input and is using the first frame for rendering. This event will not be sent before the [`start`](./routes.md#start) request.
+
+This event is usually sent at the same time as `VIDEO_INPUT_DELIVERED` except for 2 cases:
+- Before [`start`](./routes.md#start) request.
+- If input has the `offset_ms` field defined.
+
+### `VIDEO_INPUT_EOS`
+
+```typescript
+type Event = {
+  type: "VIDEO_INPUT_EOS";
+  input_id: string;
+}
+```
+
+The input stream has ended and all the frames were already processed.
+
+### `AUDIO_INPUT_DELIVERED`
+
+```typescript
+type Event = {
+  type: "AUDIO_INPUT_DELIVERED";
+  input_id: string;
+}
+```
+
+The compositor received the input, and the first samples on that input are ready to be used. If you want to ensure that some inputs are ready before you send the [`start`](./routes.md#start) request, you can wait for those events for specific inputs.
+
+### `AUDIO_INPUT_PLAYING`
+
+```typescript
+type Event = {
+  type: "AUDIO_INPUT_PLAYING";
+  input_id: string;
+}
+```
+
+The compositor received the input and is using the first samples for rendering. This event will not be sent before the [`start`](./routes.md#start) request.
+
+This event is usually sent at the same time as `AUDIO_INPUT_DELIVERED` except for 2 cases:
+- Before [`start`](./routes.md#start) request.
+- If input has the `offset_ms` field defined.
+
+### `AUDIO_INPUT_EOS`
+
+```typescript
+type Event = {
+  type: "AUDIO_INPUT_EOS";
+  input_id: string;
+}
+```
+
+The input stream has ended and all the audio samples were already processed.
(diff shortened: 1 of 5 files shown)


docs/pages/api/events.md
```diff
@@ -0,0 +1,87 @@
...
+}
+```
+
+The compositor received the input and is using the first samples for rendering. This event will not be sent before the [`start`](./routes.md#start) request.
+
+This event is usually sent at the same time as `VIDEO_INPUT_DELIVERED` except for 2 cases:
+- Before [`start`](./routes.md#start) request.
+- If input has the `offset_ms` field defined.### `AUDIO_INPUT_EOS`
+
+### `AUDIO_INPUT_EOS`
```
noituri, line 74:
```suggestion
- If input has the `offset_ms` field defined.

### `AUDIO_INPUT_EOS`
```

wkozyra95: xd
I was wondering at some point where that `AUDIO_INPUT_EOS` header disapeared and added it agin :D

noituri (approved):

docs/pages/api/events.md
```diff
@@ -0,0 +1,87 @@
...
+### `VIDEO_INPUT_DELIVERED`
+
+```typescript
+type Event = {
+  type: "VIDEO_INPUT_DELIVERED";
+  input_id: string;
+}
+```
+
+The compositor received the input, and the first frames of that input are ready to be used. If you want to ensure that some inputs are ready before you send the [`start`](./routes.md#start) request, you can wait for those events for specific inputs.
```
WojciechBarczynski, line 18: I'm unsure if you should use "the first frames of that input are ready to be used" here. "ready" can be understood as: "it will be used for rendering". Without the "inside" knowledge of how compositor works, what queue does, etc. I would interpret it this way.

Consider pointing out that this doesn't guarantee that those input frames could be used in the next render and recommend using `VIDEO_INPUT_PLAYING`.

docs/pages/api/events.md
```diff
@@ -0,0 +1,87 @@
...
+```typescript
+type Event = {
+  type: "VIDEO_INPUT_EOS";
+  input_id: string;
+}
+```
+
+The input stream has ended and all the frames were already processed.
+
+It is possible that samples mixed using this input are still in the compositor pipeline, but after this event, you can safely unregister it.
```
WojciechBarczynski, line 46: What do you mean here by "safely"?

docs/pages/api/events.md
```diff
@@ -0,0 +1,87 @@
...
+```typescript
+type Event = {
+  type: "AUDIO_INPUT_PLAYING";
+  input_id: string;
+}
+```
+
+The compositor received the input and is using the first samples for rendering. This event will not be sent before the [`start`](./routes.md#start) request.
+
+This event is usually sent at the same time as `VIDEO_INPUT_DELIVERED` except for 2 cases:
```
WojciechBarczynski, line 70: You meant `AUDIO`?
```suggestion
This event is usually sent at the same time as `AUDIO_INPUT_DELIVERED` except for 2 cases:
```

docs/pages/api/events.md
```diff
@@ -0,0 +1,87 @@
...
+```typescript
+type Event = {
+  type: "VIDEO_INPUT_PLAYING";
+  input_id: string;
+}
+```
+
+The compositor received the input and is using the first frame for rendering. This event will not be sent before the [`start`](./routes.md#start) request.
+
+This event is usually sent at the same time as `VIDEO_INPUT_DELIVERED` except for 2 cases:
```
WojciechBarczynski, line 31: Technically no. What about buffering?

wkozyra95: It's exactly like that both events are sent after buffering. Buffering is an implementation detail, mentioning it here would be only confusing.

docs/pages/api/events.md
wkozyra95: No frames/samples will be lost if you unregister it at this point.

docs/pages/api/events.md
wkozyra95: > "ready" can be understood as: "it will be used for rendering"

- If compositor is started and there is no offset then this understanding is correct.
- If compositor was not started then distinction between those 2 events is obvious
- If input stream has offset defined then it's also 

> Consider pointing out that this doesn't guarantee that those input frames could be used in the next render and recommend using VIDEO_INPUT_PLAYING.

It does gurantee it if there is no offset.

WojciechBarczynski: Ok, I understood it wrong the first time. I thought that this event was sent as soon as the first frame was sent to the queue, but now I see, that it's emitted on the buffering stage change.

docs/pages/api/events.md
WojciechBarczynski: I wouldn't use the word "safely" in such a context. This might be ambiguous or imply that unregistering without receiving this message is "not safe". Just explain in the docs what you meant here

WojciechBarczynski (approved):

docs/pages/api/events.md
wkozyra95: > unregistering without receiving this message is "not safe".

It is unsafe, you are losing data :D. But yes I get your point, I don't think this would be confusing, but after thinking more about I think it might be better to drop that sentence entierlly.

PR #465, opened 2024-03-27
Support offline processing without any inputs

wkozyra95:
With the current setup if you have:
- ahead of time processing enabled
- there is no inputs
- and machine is not able to encode/render frames in real time

then queue will drop some output frames

### DIFF
--- a/docs/pages/deployment/configuration.md
+++ b/docs/pages/deployment/configuration.md
@@ -77,13 +77,25 @@ Enable GPU support inside the embedded Chromium instance.
 
 Defaults to `true`. Valid values: `true`, `false`, `1`, `0`.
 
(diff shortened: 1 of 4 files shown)


docs/pages/deployment/configuration.md
WojciechBarczynski, line 1: I'm not sure if "deployment" is a good place for such documentation. It covers all important configurations of VideoCompositor, not just once related to "deployment".

WojciechBarczynski (approved):

docs/pages/deployment/configuration.md
wkozyra95: This sth to rethink, but for now deployment section does not have its main content, so in my opinion, it's too early to tell.
Main reason behind puting it here is that those envs will usually be set in k8s yaml or sth lilke that.

PR #472, opened 2024-03-28
[docs] Replace some of the hand-written docs

noituri:
`Shader` and `RtpOutputStream` are generated almost the same as the hand-written one.
I also changed the definition ordering.


WojciechBarczynski (approved):

wkozyra95 (approved):
Can you inline `Audio` inside `OutputAudioOptions`?

wkozyra95:
I merged sth that changed schema, so remember to rebase and regenerate otherwise CI on master will fail.

PR #477, opened 2024-03-29
Fix integration tests sometimes failing

noituri:
- ~Added `LIVE_COMPOSITOR_NEVER_DROP_OUTPUT_FRAMES` env~
- Made output receiver dump longer
- Lowered resolution
- Changed `encoder_preset` to `ultrafast`

Snapshots PR: https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/30

### DIFF
--- a/compositor_integration_tests/src/compositor_instance.rs
+++ b/compositor_integration_tests/src/compositor_instance.rs
@@ -61,6 +61,7 @@ fn init_compositor_prerequisites() {
         return;
     }
 
+    env::set_var("LIVE_COMPOSITOR_NEVER_DROP_OUTPUT_FRAMES", "1");
     env::set_var("LIVE_COMPOSITOR_WEB_RENDERER_ENABLE", "0");
     ffmpeg_next::format::network::init();
     logger::init_logger(LoggerConfig {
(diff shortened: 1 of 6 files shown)


compositor_integration_tests/src/compositor_instance.rs
```diff
@@ -91,6 +91,7 @@ fn get_free_port() -> u16 {
 fn init_compositor_prerequisites() {
     static GLOBAL_PREREQUISITES_INITIALIZED: OnceLock<()> = OnceLock::new();
     GLOBAL_PREREQUISITES_INITIALIZED.get_or_init(|| {
+        env::set_var("LIVE_COMPOSITOR_NEVER_DROP_OUTPUT_FRAMES", "1");
```
wkozyra95, line 94: This should not be necessary if you lowered resolution. Is the issue still happening?

snapshot_tests/snapshots
wkozyra95, line 1: What about input streams shouldn't we use lower resolution for them too?

wkozyra95 (approved):
This is fine, but you are mixing different things without any good reason. In my opinion proper approach here would be to either:
- Only add LIVE_COMPOSITOR_NEVER_DROP_OUTPUT_FRAMES in this PR. Fix issue properly in follow PR
- Fix issue in first PR and do not add LIVE_COMPOSITOR_NEVER_DROP_OUTPUT_FRAMES at all

While this PR adds to both temporary hack and half of the proper solution.

snapshot_tests/snapshots
noituri: Input streams are already low resolutions (640x360). It shouldn't be a problem I think

compositor_integration_tests/src/compositor_instance.rs
noituri: I didn't test without it. But it seems to work without that env

PR #478, opened 2024-03-29
Validate video frames between timestamps in integration tests

noituri:


wkozyra95 (commented):
I don't really see how we could merge that, given that current setup have a lot of issues with race conditions this change would always return huge error.

integration_tests/src/tests/schedule_update.rs
```diff
@@ -124,8 +124,9 @@ pub fn schedule_update() -> Result<()> {
     compare_video_dumps(
         OUTPUT_DUMP_FILE,
         &new_output_dump,
-        &[Duration::from_millis(500), Duration::from_micros(3500)],
+        &[Duration::from_millis(500)..Duration::from_millis(3500)],
         20.0,
+        0.0,
```
wkozyra95, line 129: - pack all optional arguments into struct that implements default
- change argument for allowed invalid frames to define number of frames

src/server.rs
```diff
@@ -21,40 +22,51 @@ pub fn run() {
...
 pub fn run_with_config(config: Config) {
     info!("Starting LiveCompositor with config:\n{:#?}", config);
-    let (api, event_loop) = Api::new(config).unwrap_or_else(|err| {
+    let (event_loop, should_close_sender) = start_api(config);
+
+    let event_loop_fallback = || {
+        let mut signals = Signals::new([consts::SIGINT]).unwrap();
+        signals.forever().next();
+    };
+    if let Err(err) = event_loop.run_with_fallback(&event_loop_fallback) {
```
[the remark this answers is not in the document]
noituri: It's currently not supported. I will implement it in follow up PR

src/server.rs
```diff
@@ -48,14 +43,19 @@ pub fn run_with_config(config: Config) {
     }
 }
 
-async fn run_tokio_runtime(api: ApiState) {
-    let port = api.config.api_port;
-    let app = routes(api);
-    let listener = tokio::net::TcpListener::bind(SocketAddr::from(([0, 0, 0, 0], port)))
+pub fn start_api(state: ApiState, should_close_receiver: Receiver<()>) -> tokio::io::Result<()> {
+    let rt = Runtime::new().unwrap();
+    rt.block_on(async { run_tokio_runtime(state, should_close_receiver).await })
```
wkozyra95, line 48:
```suggestion
    rt.block_on(run_tokio_runtime(state, should_close_receiver))
```
isn't this the same?

src/server.rs
```diff
@@ -48,14 +43,19 @@ pub fn run_with_config(config: Config) {
     }
 }
 
-async fn run_tokio_runtime(api: ApiState) {
-    let port = api.config.api_port;
-    let app = routes(api);
-    let listener = tokio::net::TcpListener::bind(SocketAddr::from(([0, 0, 0, 0], port)))
+pub fn start_api(state: ApiState, should_close_receiver: Receiver<()>) -> tokio::io::Result<()> {
```
wkozyra95, line 46: This function is blocking, using name start is misleading

wkozyra95 (approved):

PR #481, opened 2024-04-02
Downgrade nixpkgs to older xz

wkozyra95:
It does not look like nix packages were affected, but we are downgrading just to be safe.


noituri (approved):
LGTM

PR #484, opened 2024-04-02
Fix sending input stream before start

wkozyra95:
The current implementation for dropping frames before starting did not work correctly.
- Frames/samples were not dropped
- When the compositor started initial frames were treated as zero, which resulted in sending a few initial frames that should have been dropped to the renderer.
- Tested on video

https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/32

### DIFF
--- a/compositor_pipeline/src/queue/audio_queue.rs
+++ b/compositor_pipeline/src/queue/audio_queue.rs
@@ -117,9 +117,9 @@ impl AudioQueue {
         }
     }
 
-    pub(super) fn drop_old_samples(&mut self, pts: Duration, queue_start: Instant) {
+    pub(super) fn drop_old_samples_before_start(&mut self) {
         for input in self.inputs.values_mut() {
-            input.drop_old_samples(pts, queue_start)
+            input.drop_old_samples_before_start()
         }
     }
 }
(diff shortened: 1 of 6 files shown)


compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -192,7 +192,14 @@ impl AudioQueueInput {
             })
             .collect::<Vec<InputSamples>>();
 
-        self.drop_old_samples(pts_range.1, queue_start);
+        // Drop all batches older than `end_pts`. Entire batch (all samples inside) has to be older.
+        while self
```
wkozyra95, line 196: given the assumption that calling code has to have it's cleaner to inline that code

noituri (approved):
LGTM, but you forgot about snapshots PR

wkozyra95:
> In the case of non required stream, we could drop the input frames / samples like this on enqueue, preventing filling up RAM.

- whether stream is  `required` or not, it does not affect this issue only `offset` does.
- If you mean adding that logic in `try_enqueue_samples` then
  - we only need to run that check before start, and that function is always called
  - before start nothing is calling `try_enqueue_samples`, so  we would still need to have function like drop old frames that is triggering that enqueue

> Why should we wait for the start request?

We don't wait for start request. Why do you think we do?

WojciechBarczynski:
@wkozyra95 I misunderstood it before. I overlooked `QueueThreadAfterStart` and `QueueThread` separation. Now it makes sense.

The rest looks fine

PR #485, opened 2024-04-03
Do not mix sample batches older than queue start

wkozyra95:
When starting the compositor sample batch can start before the queue starts and after that. Because the PTS of samples can't be negative number the batch range of (-10ms, 10ms) would be converted to `(0, 10ms)` but still contain 20ms worth of samples. As a result, the mixer is detecting overlapping samples and printing warnings.

Additionally, add some logs + fix incorrect spans

### DIFF
--- a/compositor_pipeline/src/pipeline/decoder/audio.rs
+++ b/compositor_pipeline/src/pipeline/decoder/audio.rs
@@ -1,9 +1,9 @@
-use std::{sync::Arc, time::Duration};
+use std::{fmt, sync::Arc, time::Duration};
 
 use compositor_render::InputId;
 use crossbeam_channel::{bounded, Receiver, Sender};
 use log::{debug, error};
-use tracing::{span, Level};
+use tracing::{span, trace, Level};
 
 extern crate opus as lib_opus;
 use crate::{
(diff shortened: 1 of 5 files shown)


compositor_pipeline/src/pipeline/decoder/audio.rs
```diff
@@ -57,7 +66,8 @@ pub fn start_audio_decoder_thread(
                 Level::INFO,
                 "Audio decoder",
                 input_id = input_id.to_string()
-            );
+            )
+            .entered();
```
wkozyra95, line 70: This was not working before, span has to be entered to have effect https://docs.rs/tracing/latest/tracing/span/index.html

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -173,7 +173,16 @@ impl AudioQueueInput {
             .queue
             .iter()
             // start_pts and end_pts are already in units of this input
-            .filter(|batch| batch.start_pts <= end_pts && batch.end_pts >= start_pts)
+            .filter(|batch| {
+                let batch_in_range = batch.start_pts <= end_pts && batch.end_pts >= start_pts;
```
WojciechBarczynski, line 177:
```suggestion
                let is_in_range = batch.start_pts <= end_pts && batch.end_pts >= start_pts;
```

compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -173,7 +173,16 @@ impl AudioQueueInput {
...
+                match self.offset {
+                    Some(_offset) => batch_in_range,
+                    // We need to check if batch.start_pts does not represent time before queue
+                    // start. PTS can only be positive, so without this check we could produce
+                    // incorrect start_pts value on output framerate.
```
WojciechBarczynski, line 182: No idea what you meant here:
`we could produce incorrect start_pts value on output framerate.`

I guess you meant:
`we could produce incorrect start_pts value on output.`

Still, handling this here sounds like mixing logic for handling specific corner cases with common / "expected" flow.

wkozyra95: > Still, handling this here sounds like mixing logic for handling specific corner cases with common / "expected" flow.

I don't understand this point?

wkozyra95:
@WojciechBarczynski merging for now let me know if sth is still unclear

WojciechBarczynski (commented):
LGTM

PR #488, opened 2024-04-03
Separate API routes

wkozyra95:
Still missing (will be add in this PR after first round of review):
- documentation
- rework examples
- rework integration and snapshot tests


examples/hls.rs
```diff
@@ -37,54 +37,56 @@ fn start_example_client_code() -> Result<()> {
...
+        &json!({
+            "type": "rtp_stream",
+            "transport_protocol": "tcp_server",
+            "port": OUTPUT_PORT,
+            "video": {
+                "resolution": {
+                    "width": VIDEO_RESOLUTION.width,
+                    "height": VIDEO_RESOLUTION.height,
+                },
+                "encoder_preset": "medium",
```
WojciechBarczynski, line 73: This is too slow probably

WojciechBarczynski (approved):

PR #504, opened 2024-04-10
Generic API for  output/input streams

wkozyra95:
- Extract encoder specific options to enums `VideoEncoderOptions` and `AudioEncoderOptions`
- Add enum for input RTP codec specific options (for now it's only OPUS)
- Move `video.initial` to `video.initial.root` in register output request and `video` to `video.root` in the update output request
- Make some fileds required:
  - RegisterRtpInput `video.decoder` / `audio.decoder` (we assume that all the video/audio specific options on the rtp input are decoder specific)
  - RegisterRtpOutput `video.encoder.type` / `audio.encoder.type`
- Move stuff around, rename `convert_*` files to `from_` and remove serialization code

Most changes in PR are from test/examples

The current setup will not work if enums I added have more variants, but @noituri  is already working on that in a separate PR


wkozyra95:
Merging because some other work is blocked by this. Any comments will be addressed in follow-up.

WojciechBarczynski (commented):
LGTM, GJ

PR #507, opened 2024-04-10
Handle rtp timestamp rollover

noituri:


compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -142,21 +164,103 @@ impl AudioDepayloader {
...
+        if previous_timestamp > current_timestamp {
+            *rollover_count += 1;
+        } else {
+            // We received a packet from before the rollover, so we need to decrement the count
+            *rollover_count = rollover_count.saturating_sub(1);
+        }
+    }
+
+    (*rollover_count as u64) * (u32::MAX as u64 + 1) + current_timestamp as u64
+}
```
WojciechBarczynski, line 214: I think this won't work. If the stream starts at random PTS (as in RFC), this `rollover_count * range` is not the correct timestamp.

Assume rollover happens at 10s. If I understand correctly, if the stream starts at timestamp 9s, after rolling over, let's say after 2s, its timestamp will be set to `10s + 1s` (current timestamp is 1s since 9s -[1s]-> 10s/0s -[1s]-> 1s) but should be set to 2s.

Correct me if I missunderstod something here.

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -124,13 +138,21 @@ pub enum AudioDepayloaderNewError {
 }
 
 pub enum AudioDepayloader {
-    Opus(OpusPacket),
+    Opus {
+        depayloader: OpusPacket,
+        previous_timestamp: u32,
+        rollover_count: usize,
```
wkozyra95, line 144:
```suggestion
        rollover_state: RolloverState,
```

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -142,21 +164,103 @@ impl AudioDepayloader {
...
-                    pts: Duration::from_secs_f64(packet.header.timestamp as f64 / 48000.0),
+                    pts: Duration::from_secs_f64(timestamp as f64 / 48000.0),
                     dts: None,
                     kind,
                 }))
             }
         }
     }
 }
+
```
wkozyra95, line 197:
```suggestion
struct RolloverState {
        previous_timestamp: u32,
        rollover_count: usize,
}
```

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -142,21 +164,103 @@ impl AudioDepayloader {
...
+fn timestamp_with_rollover(
+    rollover_count: &mut usize,
+    previous_timestamp: u32,
+    current_timestamp: u32,
+) -> u64 {
```
wkozyra95, line 202:
```suggestion
impl RolloverState {

    fn timestamp_with_rollover(
        &mut self,
        current_timestamp: u32,
    ) -> u64 {
```

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -142,21 +164,103 @@ impl AudioDepayloader {
...
+    let timestamp_diff = u32::abs_diff(previous_timestamp, current_timestamp);
+    if timestamp_diff >= u32::MAX / 2 {
+        if previous_timestamp > current_timestamp {
+            *rollover_count += 1;
+        } else {
+            // We received a packet from before the rollover, so we need to decrement the count
+            *rollover_count = rollover_count.saturating_sub(1);
```
wkozyra95, line 209: If first timestamp is larger than `u32:MAX/2` then this will try to decrement 0, which will result in a zero value. This behavior is correct, but I'm not a fan how this hides the case.

Either:
- describe this case in a comment
- or use Option<u32> for previous timestamp

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -142,21 +152,110 @@ impl AudioDepayloader {
...
             }
         }
     }
 }
+
+#[derive(Default)]
+pub struct RolloverState {
+    pub previous_timestamp: Option<u32>,
+    pub rollover_count: usize,
+}
```
wkozyra95, line 182:
```suggestion
struct RolloverState {
    previous_timestamp: Option<u32>,
    rollover_count: usize,
}
```

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -142,21 +152,110 @@ impl AudioDepayloader {
...
+
+impl RolloverState {
+    fn timestamp_with_rollover(&mut self, current_timestamp: u32) -> u64 {
```
wkozyra95, line 185: maybe just `timestamp` ? but this is also fine,  no strong opinion

wkozyra95 (approved):

PR #508, opened 2024-04-10
Generate example inputs for docs

wkozyra95:
Generate some input samples that can be used to show layout component examples. We can use those files as inputs and use it to generate examples of different layouts and embed it on website as WebP

https://github.com/membraneframework/video_compositor/assets/9753141/3f592d19-4fc8-40bc-8a69-21810d3f45c4

### DIFF
--- a/generate/src/compositor_instance.rs
+++ b/generate/src/compositor_instance.rs
@@ -1,0 +1,121 @@
+use anyhow::{anyhow, Result};
+use compositor_render::use_global_wgpu_ctx;
+use crossbeam_channel::Sender;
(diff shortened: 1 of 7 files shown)


generate/src/compositor_instance.rs
```diff
@@ -0,0 +1,121 @@
+use anyhow::{anyhow, Result};
```
wkozyra95, line 1: Mostly copied from integration_tests crate

WojciechBarczynski (commented):
LGTM

PR #511, opened 2024-04-11
Add AAC RTP inputs

jerzywilczek:


src/types/from_register_input.rs
```diff
@@ -1,13 +1,94 @@
...
+    }
+
+    let mut output = BytesMut::new();
+
+    const BAD_ASC_FORMAT: &str = "Not all of the provided AudioSpecificConfig is hex digits.";
+    if !asc.chars().all(|c| char::is_ascii_hexdigit(&c)) {
+        return Err(TypeError::new(BAD_ASC_FORMAT));
+    }
+
+    for a in asc.as_bytes().chunks(2) {
```
wkozyra95, line 33: If you modify container from the loop then try to declare that mut variable as close as possible.

Also in this specific case I would consider functional approach

```
    let output = asc.as_bytes().chunks(2).map(|chunk|{
        ...
        let parsed = match u8::from_str_radix(&val, 16) {
            Ok(res) => res,
            // This can't happen, but just to be safe
            Err(_) => Err(TypeError::new(BYTE_PARSE_ERROR)),
        };
    })
    .collect()?;
```

wkozyra95:
```suggestion
    for chunk in asc.as_bytes().chunks(2) {
```

This is not Go :D

src/types/from_register_input.rs
```diff
@@ -1,13 +1,94 @@
...
+        let fallback = &[a[0], 0];
+        let val = if a.len() == 2 { a } else { fallback };
+        // the value is already checked for being ascii
+        let val = String::from_utf8_lossy(val);
```
wkozyra95, line 37:
```suggestion
        let val = match a {
           [b1, b2, ..] => &[b1, b2],
           [b1] => &[b1, 0],
           [] => &[0, 0],
        };
        // the value is already checked for being ascii
        let val = String::from_utf8_lossy(val);
```

Btw is it even correct that there would be an odd number of hex digits?

jerzywilczek: > Btw is it even correct that there would be an odd number of hex digits?
Sort of. The spec says that a non-byte aligned asc should be padded with zeroes. I'm not sure whether the receiver or the sender should add the padding, but it doesn't hurt to make sure on our side, especially since this is a well defined behavior

src/types/register_input.rs
```diff
@@ -53,6 +60,17 @@ pub enum InputRtpAudioOptions {
         /// For more information, check out [RFC](https://datatracker.ietf.org/doc/html/rfc6716#section-2.1.7).
         forward_error_correction: Option<bool>,
     },
+
+    Aac {
+        /// Specific to the AAC codec.
+        /// The config should be encoded as described in [RFC 3640](https://datatracker.ietf.org/doc/html/rfc3640#section-4.1)
+        /// The value has to be provided.
+        audio_specific_config: Option<String>,
```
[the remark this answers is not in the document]
jerzywilczek: oh yeah, thats true. I thought it was optional for a long time and then only remembered to change the docs

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -204,9 +218,273 @@ impl RolloverState {
...
+    let first = reader.get_u8();
+    let second = reader.get_u8();
+
+    let mut profile = (0b11111000 & first) >> 3;
+    let frequency: u32;
+    let channel: u8;
+    let frame_length: u32;
+
+    if profile == 31 {
+        profile = ((first & 0b00000111) << 3) + ((second & 0b11100000) >> 5) + 32;
```
wkozyra95, line 322: maybe write 32 as binary

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -204,9 +218,273 @@ impl RolloverState {
...
+            if reader.remaining() < 3 {
+                return Err(AudioSpecificConfigParseError::TooShort);
+            }
+
+            let mut rest = [0; 3];
+            reader.read_exact(&mut rest).unwrap();
+            frequency = (((second & 0b01111111) as u32) << 17)
+                | ((rest[0] as u32) << 9)
+                | ((rest[1] as u32) << 1)
+                | (((rest[2] & 0b10000000) >> 7) as u32);
```
wkozyra95, line 370: I would order those operation based on what bits they are setting e.g.
```suggestion
            frequency = (((rest[2] & 0b10000000) >> 7) as u32)
                | ((rest[1] as u32) << 1)
                | ((rest[0] as u32) << 9)
                | (((second & 0b01111111) as u32) << 17);
```

src/types/from_register_input.rs
```diff
@@ -1,13 +1,86 @@
...
+        .map(|byte| {
+            let byte = match byte {
+                &[b1, b2, ..] => [b1, b2],
+                &[b1] => [b1, 0],
+                [] => [0, 0],
+            };
+
+            let byte = String::from_utf8_lossy(&byte);
+
+            const BYTE_PARSE_ERROR: &str =
```
wkozyra95, line 33: Up to you, but I would move this const to the begining of that function

src/types/from_register_input.rs
```diff
@@ -36,15 +114,7 @@ impl TryFrom<RtpInputStream> for pipeline::RegisterInputOptions {
...
-            audio: audio.as_ref().map(|audio| input::rtp::InputAudioStream {
-                options: match audio {
-                    InputRtpAudioOptions::Opus {
-                        forward_error_correction,
-                    } => decoder::AudioDecoderOptions::Opus(decoder::OpusDecoderOptions {
-                        forward_error_correction: forward_error_correction.unwrap_or(false),
-                    }),
-                },
-            }),
+            audio,
```
wkozyra95, line 117:
```suggestion
            audio: audio.map(TryInto::try_into).transpose()?,
```
Not tested, but I think sth like that should work

src/types/register_input.rs
```diff
@@ -53,6 +60,15 @@ pub enum InputRtpAudioOptions {
         /// For more information, check out [RFC](https://datatracker.ietf.org/doc/html/rfc6716#section-2.1.7).
         forward_error_correction: Option<bool>,
     },
+
+    Aac {
+        /// The config should be encoded as described in [RFC 3640](https://datatracker.ietf.org/doc/html/rfc3640#section-4.1)
```
wkozyra95, line 65: Remember to add docs how to get that value

wkozyra95 (approved):

WojciechBarczynski (commented):
Have you tried using this to stream HLS/some other live source? How can you get ASC in such cases?

jerzywilczek:
@WojciechBarczynski same way

WojciechBarczynski (approved):

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -204,9 +218,273 @@ impl RolloverState {
...
+pub enum AudioSpecificConfigParseError {
+    #[error("ASC is not long enough")]
+    TooShort,
+
+    #[error("Illegal value in ASC")]
+    IllegalValue,
+}
+
+// MPEG-4 part 3, sections 1.6.2.1 & 4.4.1
+fn parse_asc(asc: &[u8]) -> Result<Asc, AudioSpecificConfigParseError> {
```
WojciechBarczynski, line 304: Extract this. This file is already huge

noituri (approved):

compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
jerzywilczek: this comment only loaded for me after i clicked merge, so I'll maybe do it in a subsequent PR

jerzywilczek: sorry

WojciechBarczynski: np

src/types/register_input.rs
```diff
@@ -53,6 +60,19 @@ pub enum InputRtpAudioOptions {
...
```
wkozyra95, line 65: - Missing `.` at the end
- This is rendered as one big blog of text, Maybe add a newline after it.
- I don't think this explanation is good enough
  - You are recommending passing an additional flag to ffmpeg command, but additional to what?
  - What if someone is using GStreamer to deliver RTP, even if you don't know how to access that value with gstreamer explain it. For example, instead of explaining how to get it with gstreamer you could explain how to get it from hls, rtmp, mp4/mp3 or some other container file

PR #512, opened 2024-04-11
Inherit width/height in absolute positioning

wkozyra95:
When creating absolutely positioned element inherit the width/height of a parent if it's not provided.

I'll update docs in a follow up.

https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/35


WojciechBarczynski (commented):
LGTM

PR #531, opened 2024-04-23
Do not warn on closed channel

wkozyra95:
When output receives eos based on `should_send_eos` condition the renderer is still active and producing frame. As a result this case is producing very noisy logs.

I'm changing that to debug for now to unblock next rc release, but in follow up I'll rework that code to warn if channel is closed but EOS was not sent yet.


WojciechBarczynski (commented):
LGTM

PR #536, opened 2024-04-24
Simplify handling EOS

### LINKED ISSUE #532: Rework renderer thread code to warn only if channel is closed but EOS was not yet sent

wkozyra95:
Closes #532

### DIFF
--- a/compositor_pipeline/src/pipeline.rs
+++ b/compositor_pipeline/src/pipeline.rs
@@ -16,7 +16,7 @@ use compositor_render::RendererOptions;
 use compositor_render::{error::UpdateSceneError, Renderer};
 use compositor_render::{EventLoop, InputId, OutputId, RendererId, RendererSpec};
 use crossbeam_channel::{bounded, Receiver};
-use tracing::{debug, error, info, trace, warn};
+use tracing::{error, info, trace, warn};
 
 use crate::audio_mixer::AudioMixer;
 use crate::audio_mixer::MixingStrategy;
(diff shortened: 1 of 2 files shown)


compositor_pipeline/src/pipeline/pipeline_output.rs
```diff
@@ -107,6 +112,73 @@ impl Pipeline {
...
+    pub(super) fn get_output_video_sender(
+        pipeline: &Arc<Mutex<Pipeline>>,
+        output_id: &OutputId,
+    ) -> Option<Sender<PipelineEvent<Frame>>> {
+        let mut guard = pipeline.lock().unwrap();
+        let Some(output) = guard.outputs.get_mut(output_id) else {
+            warn!(
+                ?output_id,
+                "Failed to send output frame. Output does not exists.",
+            );
```
WojciechBarczynski, line 125: If I understand correctly, those logs can happen even if there is no error. In the case of output unregistering, outputs may be unregistered in the middle of the rendering/mixing, because mixer/renderer are working on a different threads than the pipeline, causing those logs to appear. Since this is a warning, not an error, I don't have a strong opinion on that, but it seems weird to display warn logs in perfectly normal situations. I noticed this kind of log once while working on demos.

wkozyra95: Yes, I prefer to keep this warning. If we want to avoid false positives we could call this method before render on all outputs. But this would requires some refactor, because if we take that approach we can't always unregister_output when this function returns None

compositor_pipeline/src/pipeline.rs
```diff
@@ -317,6 +318,19 @@ fn run_renderer_thread(
...
+        let output_frame_senders: HashMap<_, _> = output_ids
+            .into_iter()
+            .filter_map(|id| {
+                let Some(sender) = Pipeline::get_output_video_sender(&pipeline, &id) else {
+                    renderer.unregister_output(&id);
+                    return None;
+                };
+                Some((id, sender))
+            })
+            .collect();
```
WojciechBarczynski, line 332:
```suggestion
        let output_frame_senders: HashMap<_, _> = pipeline
            .lock()
            .unwrap()
            .outputs
            .keys()
            .filter_map(|id| {
                let Some(sender) = Pipeline::get_output_video_sender(&pipeline, id) else {
                    renderer.unregister_output(id);
                    return None;
                };
                Some((id.clone(), sender))
            })
            .collect();
```

WojciechBarczynski (approved):

compositor_pipeline/src/pipeline.rs
```diff
@@ -385,40 +374,29 @@ fn run_audio_mixer_thread(
...
+        let output_samples_senders: HashMap<_, _> = output_ids
+            .into_iter()
+            .filter_map(|output_id| {
+                let Some(sender) = Pipeline::get_output_audio_sender(&pipeline, &output_id) else {
+                    audio_mixer.unregister_output(&output_id);
+                    return None;
+                };
+                Some((output_id, sender))
+            })
+            .collect();
```
WojciechBarczynski, line 388: same

compositor_pipeline/src/pipeline.rs
WojciechBarczynski: Nevermind, I haven't noticed this send in `get_output_video_sender`

WojciechBarczynski (approved):

PR #547, opened 2024-04-26
Wgpu ctx init more friendly

ludolpif:
By asking wgpu adapter.features() before adapter.request_device(..., required_features), don't let panic on missing feature but error!() and return Err(CreateWgpuCtxError::NoAdapter).
Output log contains no less nor more details with this PR. May create room to future improvements like : try another adapter or let start without some non-critical features.


wkozyra95:
Closing in favor of https://github.com/membraneframework/live_compositor/pull/548

PR #548, opened 2024-04-26
Wgpu ctx init more laxist

ludolpif:
By splitting required_features into critical_features and nominal_mode_features and adding a bit of logic, let compositor start in a degraded mode on poor's man/hw instead of not running at all if it will affect only some functionalities of the compositor.

### DIFF
--- a/compositor_render/src/wgpu/ctx.rs
+++ b/compositor_render/src/wgpu/ctx.rs
@@ -71,7 +71,7 @@ impl WgpuCtx {
         }
         let required_features =
             wgpu::Features::TEXTURE_BINDING_ARRAY | wgpu::Features::PUSH_CONSTANTS;
-        let nominal_mode_features =
+        let optional_features =
             wgpu::Features::SAMPLED_TEXTURE_AND_STORAGE_BUFFER_ARRAY_NON_UNIFORM_INDEXING
                 | wgpu::Features::UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING;
 
--- a/compositor_render/src/wgpu/ctx.rs
+++ b/compositor_render/src/wgpu/ctx.rs
@@ -82,8 +82,7 @@ impl WgpuCtx {
         }
         let missing_nominal_mode_features = nominal_mode_features.difference(adapter.features());
         if !missing_nominal_mode_features.is_empty() {
-            error!("Selected adapter or it's driver don't support nominal mode wgpu {missing_nominal_mode_features:?}.");
-            error!("Starting in degraded mode, some valid and correct user shaders may not able to run on this adapter.");
+            error!("Selected adapter or its driver does not support optional wgpu features. Missing features: {missing_optional_features:?}).");
         }
         let required_features = critical_features
             .union(nominal_mode_features.difference(missing_nominal_mode_features));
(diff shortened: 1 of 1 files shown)


compositor_render/src/wgpu/ctx.rs
```diff
@@ -69,6 +69,24 @@ impl WgpuCtx {
...
+
+        let missing_critical_features = critical_features.difference(adapter.features());
+        if !missing_critical_features.is_empty() {
+            error!("Selected adapter or it's driver don't support critical wgpu {missing_critical_features:?}. Aborting.");
+            return Err(CreateWgpuCtxError::NoAdapter);
+        }
+        let missing_nominal_mode_features = nominal_mode_features.difference(adapter.features());
+        if !missing_nominal_mode_features.is_empty() {
+            error!("Selected adapter or it's driver don't support nominal mode wgpu {missing_nominal_mode_features:?}.");
+            error!("Starting in degraded mode, some valid and correct user shaders may not able to run on this adapter.");
```
wkozyra95, line 86:
```suggestion
            error!("Selected adapter or its driver does not support optional wgpu features. Missing features: {missing_optional_features:?}).");
```

wkozyra95 (changes requested):
Configuring wgpu features will need to be reworked, but if this resolves your problem I'm ok with this approach in the meantime.

ludolpif:
All changes committed, thanks for the time taken.

compositor_render/src/wgpu/ctx.rs
```diff
@@ -69,6 +69,23 @@ impl WgpuCtx {
             error!("Selected adapter is CPU based. Aborting.");
             return Err(CreateWgpuCtxError::NoAdapter);
         }
+        let required_features =
```
wkozyra95, line 72: I did not comment specifically on each variable, but I would expect that if you change this variable name that you also will change `missing_critical_features`.
The same applies to other cases like this.

compositor_render/src/wgpu/ctx.rs
```diff
@@ -69,6 +69,23 @@ impl WgpuCtx {
...
+        let missing_critical_features = critical_features.difference(adapter.features());
+        if !missing_critical_features.is_empty() {
+            error!("Selected adapter or its driver does not support required wgpu features. Missing features: {missing_critical_features:?}).");
+            return Err(CreateWgpuCtxError::NoAdapter);
+        }
+        let missing_nominal_mode_features = nominal_mode_features.difference(adapter.features());
+        if !missing_nominal_mode_features.is_empty() {
+            error!("Selected adapter or its driver does not support optional wgpu features. Missing features: {missing_optional_features:?}).");
+        }
+        let required_features = critical_features
```
wkozyra95, line 87: There is no `critical_features` variable, this will no compile

wkozyra95 (changes requested):

compositor_render/src/wgpu/ctx.rs
ludolpif: Solved in 6ecc7e356fd1d0eab9ee220c198dd067446f7103.
Note : with critical_features renamed to required_features as asked, I need to find a third name for the result of the computation of required + optional - missing_optional. I choose requested_features. I pass requested_features var to the required_feature arg of wgpu adapter.request_device() creating of slightly misleading naming thing. It seems reusing required_features name was worse.

ludolpif:
Note that the log lines repeats feature twice with the new messages:
```
2024-04-29T18:23:47.777885Z ERROR compositor_render::wgpu::ctx:
   Selected adapter or its driver does not support optional wgpu features.
   Missing features: Features(UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING)).
```

compositor_render/src/wgpu/ctx.rs
wkozyra95: You did not have to change that variable name, rust supports shadowing and it's relatively common to use it, but requested_feature is fine too.

wkozyra95 (approved):

PR #554, opened 2024-05-06
[docs] Rewrite markdown docs generation

noituri:
- I removed `GenerationStrategy` trait in favor of `overrides` hash map field in `DocsConfig`
  - You can override markdown generation for specific types (like here)
- Rewritten markdown generation
  - Typescript definition generation is easier to read
  - Generating properties under type definition is better than the previous implementation but there are a lot of different corner cases which make it hard to understand


src/bin/generate_docs/schema_parser.rs
noituri, line 1: Nothing really changed here. Moved some functions around and removed `SchemaParser` struct (now there's `parse_schema` function)

src/bin/generate_docs/generation.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    pub fn inc_indent(&mut self) {
+        self.indent += Self::INDENT_SIZE;
+    }
+
+    pub fn dec_indent(&mut self) {
+        self.indent -= Self::INDENT_SIZE;
+    }
+
+    pub fn calculate_generation_length<F: Fn(&mut MarkdownGenerator)>(&self, f: F) -> usize {
```
wkozyra95, line 77: It's a hack. but I suspect separating serialization and keeping track of indents would be hard if writing was not handled together with serialization.

So in the end I'm fine with leaving this if there is no good alternative

wkozyra95 (commented):
Overall looks good, just few naming changes

src/bin/generate_docs/generation/properties.rs
```diff
@@ -0,0 +1,138 @@
...
+        let name = definition.name.clone();
+
+        match &definition.kind {
+            Kind::Object { fields } => self.generate_object_properties(fields.clone()),
+            Kind::Union(variants) => self.generate_union_properties(name, variants.clone()),
+            _ => {}
+        }
+    }
+
+    pub fn generate_object_properties(&mut self, fields: Vec<(String, TypeDefinition)>) {
```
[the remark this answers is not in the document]
noituri: They can be private unless they're used inside overrides

src/bin/generate_docs/definition.rs
```diff
@@ -49,6 +38,12 @@ impl TypeDefinition {
         let desc1 = self.description.clone();
         let desc2 = other.description.clone();
 
+        let name = if other.name.is_some() {
+            other.name.clone()
+        } else {
+            self.name.clone()
+        };
```
wkozyra95, line 45:
```suggestion
        let name = other.name.or(self.name).clone();
```
Wouldn't sth like this work?

Also can you add comment explaining what other.name and self.name represents, when would you see one specified and when the other?

src/bin/generate_docs/definition.rs
```diff
@@ -49,6 +38,12 @@ impl TypeDefinition {
         let desc1 = self.description.clone();
```
wkozyra95, line 38: can you add comment to the function description with an example of what could be merged (and why it makes sense)?

src/bin/generate_docs/definition.rs
noituri: Actually, `self.name` is always `None` because `self` always represents an anonymous type (the method was always used on anonymous types).

I removed the whole method. I think it's easier to understand the code when it's placed directly where it's used.

wkozyra95 (approved):

PR #556, opened 2024-05-07
Add tile transition example animation

wkozyra95:
20240508_12h45m16s_grim


WojciechBarczynski (approved):
LGTM
I would change the colors of `input_2` and `input_3` to more distinctive ones (they are pretty similar).

noituri (approved):

PR #577, opened 2024-05-27
Update website content

WojciechBarczynski:


docs/src/pages/index.tsx
```diff
@@ -91,7 +99,80 @@ function HomepageHeader() {
...
 
-function ProsCards() {
+function HowItWorks() {
+  return (
+    <div className="container margin-top--md">
+      <Heading as="h1" className="margin-bottom--md text--center text--primary">
+        How it works?
+      </Heading>
+      <p className={clsx('text--center', styles.sectionSubheading)}>
+        Send inputs as RTP streams or MP4 files. Configure mixing with HTTP requests. Get the mixed
```
wkozyra95, line 109: I'm not realy sure what, but maybe we could do sth different than putting that in a single line? Maybe bullet points?

docs/src/pages/index.tsx
```diff
@@ -183,33 +266,40 @@ function Features() {
...
       <Feature
-        text="WebGPU APIs"
-        secondaryText="Utilize existing WebGPU-based components or craft your own WGSL shader to achieve any desired effect and directly leverage GPU parallel processing capabilities."
+        text="Built-in components"
+        secondaryText="Use built-in components to process streams, render text, images, GIFs or websites, and combine them into output streams."
+        image={<FaPuzzlePiece className={styles.featureIcon} style={{ fontSize: 80 }} />}
+      />
+
+      <Feature
+        text="Highly customizable"
```
wkozyra95, line 282: WebGPU logo does not fit here with this title. Either change logo or title. I personally would prefer to keep this point specific about WebGPU

docs/src/pages/index.tsx
```diff
@@ -183,33 +266,40 @@ function Features() {
...
+
+      <Feature
+        text="Smooth transitions"
+        secondaryText="Smoothly transition layouts with animations using built-in transition mechanism."
+        image={<GiFeather className={styles.featureIcon} style={{ fontSize: 80 }} />}
+      />
+
       <Feature
         text="Real-time processing"
         secondaryText="Process live video streams in real-time with low latency. Blazingly fast processing makes video conferencing, live-streaming, and broadcasting an everyday bread and butter for Live Compositor."
```
wkozyra95, line 295: I would consider removing "Blazingly fast processing makes video conferencing, live-streaming, and broadcasting an everyday bread and butter for Live Compositor.". It sounds a bit weird.

wkozyra95 (changes requested):

docs/docusaurus.config.ts
```diff
@@ -21,6 +21,7 @@ const config: Config = {
   onBrokenMarkdownLinks: 'throw',
   onBrokenAnchors: 'throw',
   trailingSlash: false,
+  staticDirectories: ['static'],
```
wkozyra95, line 24: this is already static directory by default

docs/static/img/how_it_works.png
wkozyra95, line 1: Change the colors, blue on the red background does not look good

docs/docusaurus.config.ts
```diff
@@ -72,9 +72,18 @@ const config: Config = {
       },
       items: [
         {
-          href: 'https://github.com/membraneframework/live_compositor',
+          label: 'Docs',
```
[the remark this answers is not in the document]
WojciechBarczynski: I fixed it by creating SVGs that look like buttons. This looks way better now.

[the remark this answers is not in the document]
wkozyra95: - It looks like you are using a pure white color which contrasts too much. I would use the same color as for text or even something dimmer (or brighter on light theam)
- Spacing between docs-github is inconsistent with spacing between github-theme switch

docs/src/pages/index.tsx
```diff
@@ -183,33 +259,46 @@ function Features() {
...
         image={wgpuLogo}
       />
+
       <Feature
-        text="Real-time processing"
-        secondaryText="Process live video streams in real-time with low latency. Blazingly fast processing makes video conferencing, live-streaming, and broadcasting an everyday bread and butter for Live Compositor."
-        image={<MdLiveTv className={styles.featureIcon} style={{ fontSize: 80 }} />}
-        inverted
+        text="Animated transitions"
+        secondaryText="Dynamically change the composition with customizable animated transitions."
```
wkozyra95, line 288:
```suggestion
        secondaryText="Apply scene changes with animated transitions."
```
or maybe `Trigger`, `Run` (not really sure what fits better)
scene<->composition would also be fine, but i think scene is intuitive enough on it's own and this is the term we use in docs

docs/src/pages/index.tsx
```diff
@@ -183,33 +259,46 @@ function Features() {
...
+        image={<MdLiveTv className={styles.featureIcon} style={{ fontSize: 80 }} />}
       />
+
       <Feature
-        text="Web rendering"
-        inverted
-        secondaryText="You can render any website and combine it with video streams or other elements using the Chromium browser embedded inside the compositor."
-        image={<FaChrome className={styles.featureIcon} style={{ fontSize: 80 }} />}
+        text="Offline processing"
+        secondaryText="You can also use LiveCompositor for non-real-time use cases, like mixing MP4 files."
```
wkozyra95, line 300: Rewrite this sentence in a similar style as other features.

wkozyra95:
Also, I think it's important to mention that we have audio support

docs/docusaurus.config.ts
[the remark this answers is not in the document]
WojciechBarczynski: It's `var(--ifm-navbar-link-color)` configurable in the `custom.css` file. I think this color looks fine. It's the same color that the mode selector next to it uses

docs/src/pages/index.tsx
```diff
@@ -183,44 +264,120 @@ function Features() {
...
+            </Heading>
+          </div>
+          <div className="card__body container">
+            <p>
+              LiveCompositor can be deployed as a standalone server. Language-agnostic API allows
+              you to seamlessly integrate it into your existing solutions. Docker images and
+              deployment section in docs are meant to make the deployment process as easy as
+              possible.
+            </p>
+            <div className="row margin--sm" style={{ justifyContent: 'center' }}>
```
wkozyra95, line 336: - I don't think this looks good
  - centering those buttons on large screens looks weird (it looks fine on small screens)
  - I'm not convinced that you want to style the differently (but I also don't object if you want)
  - They fill too large, too much out of place compared to other elements next to it.
  
It's hard to say what to do without seeing how it will look, but I would suggest changing alignment to left or right, and keeping minimal width based on content instead of stretching it static width. The same as about us card

docs/src/pages/index.tsx
```diff
@@ -234,38 +391,60 @@ function IntegrationMembranePlugin() {
...
+              <Link
+                className={clsx(
+                  'button',
+                  'button--primary',
+                  'button--lg',
+                  'margin--sm',
+                  'col',
+                  styles.hoverPrimary
+                )}
+                to="https://hexdocs.pm/membrane_video_compositor_plugin/Membrane.VideoCompositor.html">
```
wkozyra95, line 408: Don't link to old plugin

docs/src/pages/index.tsx
```diff
@@ -183,44 +264,120 @@ function Features() {
...
+                  'button',
+                  'button--secondary',
+                  'button--outline',
+                  'button--lg',
+                  'margin--sm',
+                  'col',
+                  styles.hoverSecondary
+                )}
+                to="https://github.com/membraneframework/live_compositor/tree/master/build_tools/docker">
+                <IconContext.Provider value={{ style: { verticalAlign: 'middle' } }}>
```
wkozyra95, line 361: This link is to dockefiles not docker images, it's not the same thing

docs/src/pages/index.tsx
```diff
@@ -234,38 +391,60 @@ function IntegrationMembranePlugin() {
...
+                  'button--outline',
+                  'button--lg',
+                  'margin--sm',
+                  'col',
+                  styles.hoverSecondary
+                )}
+                to="https://github.com/membraneframework/membrane_live_compositor_plugin">
+                <IconContext.Provider value={{ style: { verticalAlign: 'middle' } }}>
+                  <FaGithub style={{ marginRight: 5 }} />
+                  Plugin repo
```
wkozyra95, line 427: I would avoid abbreviations like repo, at least on the main page. 

Plugin repository/Sources/Repository/View on GitHub

docs/src/css/custom.css
```diff
@@ -29,6 +29,46 @@
...
+  padding: 0px;
+  margin-right: 10px;
+}
+
+.navbar-docs-link {
+  padding: 0px;
+  margin-right: 10px;
+}
+
+.navbar-github-link::before {
```
wkozyra95, line 42: did you change the color of those icons in the navbar? I don't see the difference.

wkozyra95: ok never mind I see your comment. I still think it should be less contrasting, but I don't have a strong opinion, so it's up to you.

docs/index.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.png';
```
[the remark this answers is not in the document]
wkozyra95: This feels wrong, but apparently, this is the correct approach.

wkozyra95 (approved):

PR #579, opened 2024-05-28
[docs] Add simple transition guide

wkozyra95:


docs/pages/guides/view-transition.md
```diff
@@ -0,0 +1,622 @@
...
+    end
+    ```
+  </TabItem>
+</Tabs>
+
+In the first update request you can see that rescaler has width of 480, and in the second one it is changed
+to 1280 and `transition.duration_ms: 2000` was added.
+
+It is important that component has the same `"id"` in both initial state and the update that starts the
+transition, otherwise it will switch immediately to the new state without a transition.
```
WojciechBarczynski, line 132:
```suggestion
In the first update request, you can see that the rescaler has a width of 480, and in the second one, it is changed
to 1280 and `transition.duration_ms: 2000` was added.

The component must have the same `"id"` in both the initial state and the update that starts the
transition, otherwise it will switch immediately to the new state without a transition.
```

docs/pages/guides/view-transition.md
```diff
@@ -0,0 +1,622 @@
...
+    <img src={SimpleTransition} style={{ width: 600 }} />
+    Output stream
+</div>
+
+### Transition on one of the sibling components
+
+In the above scenario you saw how transition on single component behave, but let's see what happens with
+components that are not a part of the transition, but their size and position still depends on other components.
+
+Add second input stream wrapped with `Rescaler`, but without any transition options.
```
WojciechBarczynski, line 144:
```suggestion
### Transition on one of the sibling components

In the above scenario you saw how transition on a single component behaves, but let's see what happens with
components that are not a part of the transition, but their size and position still depend on other components.

Add a second input stream wrapped with `Rescaler`, but without any transition options.
```

docs/pages/guides/view-transition.md
```diff
@@ -0,0 +1,622 @@
...
+
+### Transition between different modes
+
+Currently, a state before the transition and after needs to use the same type of configuration. In particular:
+- It is not possible to transition a component between static and absolute positioning.
+- It is not possible to transition a component between using `top` and `bottom` fields (the same for `left`/`right`).
+- It is not possible to transition a component with known `width`/`height` to state with dynamic `width/height` based
+on parent layout.
+
+Let's try the same example as in first scenario with a single input, but instead change `Rescaler` component to be absolutely positioned in the second update.
```
WojciechBarczynski, line 283:
```suggestion
### Transition between different modes

Currently, a state before the transition and after needs to use the same type of configuration. In particular:
- It is not possible to transition a component between static and absolute positioning.
- It is not possible to transition a component between using `top` and `bottom` fields (the same for `left`/`right`).
- It is not possible to transition a component with known `width`/`height` to a state with dynamic `width/height` based
on the parent layout.

Let's try the same example as in the first scenario with a single input, but instead, change the `Rescaler` component to be absolutely positioned in the second update.
```

docs/pages/guides/view-transition.md
```diff
@@ -0,0 +1,622 @@
...
+        notify_child: {:live_compositor, request},
+        stop_timer: :start_transition
+      ], state}
+    end
+    ```
+  </TabItem>
+</Tabs>
+
+As you can see on the resulting stream transition did not happen because the `Rescaler` component
+in the initial scene was using static positioning and after the update it was positioned absolutely.
```
WojciechBarczynski, line 395:
```suggestion
As you can see the resulting stream transition did not happen because the `Rescaler` component
in the initial scene was using static positioning and after the update it was positioned absolutely.
```

docs/pages/guides/view-transition.md
```diff
@@ -0,0 +1,622 @@
...
+
+<div style={{textAlign: 'center'}}>
+    <img src={UnsupportedTransition} style={{ width: 600 }} />
+    Output stream
+</div>
+
+### Different interpolation functions
+
+All of the above examples were using default linear interpolation, but there are also few other
+modes available.
```
WojciechBarczynski, line 405:
```suggestion
### Different interpolation functions

All of the above examples use default linear interpolation, but there are also a few other
modes available.
```

WojciechBarczynski (approved):

docs/pages/guides/view-transition.md
```diff
@@ -0,0 +1,622 @@
...
+      }
+      {[
+        notify_child: {:live_compositor, request},
+        stop_timer: :start_transition
+      ], state}
+    end
+    ```
+  </TabItem>
+</Tabs>
+
```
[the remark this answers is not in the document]
wkozyra95: That would fit better in a separate guides that handles scheduling and offline processing

PR #581, opened 2024-05-29
Provide a way to configure wgpu features

wkozyra95:


WojciechBarczynski (approved):
I have no clue why anyone would want to use it

wkozyra95:
@ludolpif btw this will be a breaking change for you, now if you want to disable all optional wgpu features you will need to set `LIVE_COMPOSITOR_REQUIRED_WGPU_FEATURES` to the empty string. Otherwise compositor will fail on startup if those features are not supported.

https://compositor.live/docs/deployment/configuration#live_compositor_required_wgpu_features

ludolpif:
[resending now my reply of Jun 11 because github notification mail reply seems never process incoming emails for me]

> This is very fine, thanks to the time taken !

PR #584, opened 2024-06-03
Fix pottentially flaky integration test

wkozyra95:
It is testing scheduled updates, so it's fine to make streams required without affecting what is tested


WojciechBarczynski (commented):
LGTM

PR #589, opened 2024-06-11
Add support for inputs that produce raw video

wkozyra95:


compositor_pipeline/src/pipeline/decoder/audio.rs
```diff
@@ -82,46 +66,107 @@ pub fn start_audio_decoder_thread(
...
+            };
+        }
+    }
 
     if samples_sender.send(PipelineEvent::EOS).is_err() {
         debug!("Failed to send EOS message.")
     }
 }
 
+pub fn start_audio_decoder_thread(
```
wkozyra95, line 115: diff is a bit misleading here I added resampler funcitions above, only change here was inlining run_decoder_thread iniside this function

compositor_pipeline/src/pipeline/decoder.rs
```diff
@@ -1,65 +1,19 @@
...
+use compositor_render::Frame;
+use crossbeam_channel::Receiver;
 
 pub use audio::AacDecoderError;
 
 mod audio;
 mod video;
 
-pub fn start_decoder(
-    input_id: InputId,
```
wkozyra95, line 15: technically this code could still be here, but given that it would also need to handle a case without decoder it seems more appropriate to move it to the input

compositor_pipeline/src/pipeline/decoder/audio.rs
```diff
@@ -20,22 +20,6 @@ mod fdk_aac;
 mod opus;
 mod resampler;
 
-struct DecodedSamples {
```
wkozyra95, line 23: moved to pipeline/struct.rs becaus:
- we will need to create those samples from inputs that produce raw data
- I didn't want to make it `pub(crate)`

WojciechBarczynski (approved):

compositor_pipeline/src/pipeline/decoder/audio.rs
jerzywilczek: agreed, this needs to be changed

compositor_pipeline/src/pipeline/decoder/audio.rs
```diff
@@ -51,13 +35,13 @@ trait AudioDecoderExt {
     fn decoded_sample_rate(&self) -> u32;
 }
 
-pub fn start_audio_decoder_thread(
-    opts: AudioDecoderOptions,
+pub fn start_audio_resampler_only_thread(
```
jerzywilczek, line 38: I don't like the word `only` being included in the name of this function.

I spent a while thinking about this and I don't see a good way to make this better.

jerzywilczek (approved):
looks good to me

PR #598, opened 2024-06-25
Convert YUV422 non-planar to RGB with shader

wkozyra95:
Tested by configuring ffmpeg to produce planar yuv422 and manually converting it to non planar with
(specifying uyvy422 in ffmpeg does not work and still produces planar format)

```
        Pixel::YUV422P => {
            let y_plane = copy_plane_from_av(decoded, 0);
            let u_plane = copy_plane_from_av(decoded, 1);
            let v_plane = copy_plane_from_av(decoded, 2);
            let width = decoded.plane_width(0);
            let height = decoded.plane_height(0);
            let mut result = BytesMut::with_capacity((width * height * 2) as usize);
            for ((y, u), v) in y_plane.chunks(2).zip(u_plane.iter()).zip(v_plane.iter()) {
                result.put_u8(*u);
                result.put_u8(y[0]);
                result.put_u8(*v);
                result.put_u8(y[1]);
            }
            FrameData::PixelYUV422(result.freeze())
```


compositor_render/src/wgpu/format/pixel_yuv_to_rgba.rs
```diff
@@ -0,0 +1,99 @@
...
+        let sampler = Sampler::new(device);
+
+        let pipeline_layout = device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
+            label: Some("PixelYUV422 to RGBA color converter render pipeline layout"),
+            bind_group_layouts: &[yuv_textures_bind_group_layout, &sampler.bind_group_layout],
+            push_constant_ranges: &[],
+        });
+
+        let pipeline = device.create_render_pipeline(&wgpu::RenderPipelineDescriptor {
+            label: Some("Pixel YUV 4:2:2 to RGBA color converter render pipeline"),
```
jerzywilczek, line 30:
```suggestion
            label: Some("Interleaved YUV 4:2:2 to RGBA color converter render pipeline"),
```

jerzywilczek: I would change pixel yuv to interleaved yuv everywhere

jerzywilczek (approved):

compositor_render/src/wgpu/format/pixel_yuv_to_rgba.wgsl
```diff
@@ -0,0 +1,52 @@
...
+@group(1) @binding(0) var sampler_: sampler;
+
+@fragment
+fn fs_main(input: VertexOutput) -> @location(0) vec4<f32> {
+    var dimensions = textureDimensions(texture);
+    var eps = 0.0001;
+    var half_pixel_width = 0.5 / f32(dimensions.x);
+
+    // x_pos represents index of a column(pixel) on the output texture
+    // - dimensions.x represents half of the output width, so we need multiply * 2
```
WojciechBarczynski, line 31: Only for u and v planes. This description can be misleading.

compositor_render/src/wgpu/format/pixel_yuv_to_rgba.wgsl
```diff
@@ -0,0 +1,52 @@
...
+    var tex_coords = vec2((f32(x_pos/2) / f32(dimensions.x)) + half_pixel_width, input.tex_coords.y);
+
+    var uyvy = textureSample(texture, sampler_, tex_coords);
+
+    var u = uyvy.x;
+    var v = uyvy.z;
+    var y = uyvy.y;
+    if (x_pos % 2 != 0) {
+        y = uyvy.w;
+    }
```
WojciechBarczynski, line 45: No idea why u and v are first. Maybe add here some links to a description of the memory layout of YUV422.

Are you sure this is a standard? There are no other possible layouts?

WojciechBarczynski (approved):

compositor_render/src/wgpu/format/pixel_yuv_to_rgba.rs
wkozyra95: I think I took that from here https://www.intel.com/content/www/us/en/docs/ipp/developer-guide-reference/2021-11/pixel-and-planar-image-formats.html

My main concern is that was potentially intended to maybe refer to all single plane format. Would it be correct to call RGBA or BGRA interleaved format?

compositor_render/src/wgpu/format/pixel_yuv_to_rgba.wgsl
wkozyra95: there are no u and v planes this is not a planar format it refers to width of texture in pixels

compositor_render/src/wgpu/format/pixel_yuv_to_rgba.wgsl
WojciechBarczynski: Ok

compositor_render/src/wgpu/format/pixel_yuv_to_rgba.wgsl
WojciechBarczynski: Ok, I don't know how to formulate this, but the "Y component" has the same resolution. Only "U and V components" have reduced width.

wkozyra95: Input texture if in RGBA format (but each component is redefined).  Single "pixel" in the input texture has 4 bytes which includes two Y "value", So width of texture is half of the output one when talking about pixels.

WojciechBarczynski: Ok, at first, I didn't get that. Now it makes sense.

PR #601, opened 2024-06-28
Some DeckLink profile attributes should not be freed

wkozyra95:
It was not crashing, but according to docs some string attributes should not be freed.

### DIFF
--- a/decklink/src/api/profile.rs
+++ b/decklink/src/api/profile.rs
@@ -41,8 +41,15 @@ impl ProfileAttributes {
     }
 
     pub fn get_string(&self, id: ffi::StringAttributeId) -> Result<Option<String>, DeckLinkError> {
+        // List of attributes that should not be freed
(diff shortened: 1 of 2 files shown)


decklink/src/api/profile.rs
```diff
@@ -41,8 +41,15 @@ impl ProfileAttributes {
     }
 
     pub fn get_string(&self, id: ffi::StringAttributeId) -> Result<Option<String>, DeckLinkError> {
+        // List of attributes that should not be freed
+        const STATIC_STRING_ATTRIBUTES: [enums::ffi::StringAttributeId; 1] =
```
wkozyra95, line 45: I could just use a simple comparison, but I had already written this before realizing that it is just one value. And overall I think this shows the purpose of the logic better, so I would prefer to leave it even though it could be simplified

WojciechBarczynski (commented):
LGTM

PR #614, opened 2024-07-08
Select DeckLink by persistent ID + add override for pixel format

wkozyra95:
- Add an option to specify the persistent id of the DeckLink device
  - REST API expects hex string
  - internally it's just an integer
  - Error messages/responses will return string variant when no device is matched correctly
- Add an internal option to enforce pixel format (from REST API it is always forcing YUV422 8 bit)
  - If the device supports 10-bit and 8-bit yuv it usually defaults to 10-bit which is not supported in compositor.
- If no DeckLink device is matched correctly, the error message will include a list of all Decklink devices

Example error message in case of mismatch
```
Error stack:
 - Input initialization error while registering input for stream "input_1".
 - No DeckLink device matches specified options. Found devices: [DeckLinkInfo { display_name: Some("DeckLink Quad HDMI Recorder (1)"), persistent_id: Some("DEEB490"), subdevice_index: Some(0) }, DeckLinkInfo { display_name: Some("DeckLink Quad HDMI Recorder (2)"), persistent_id: Some("DEEB491"), subdevice_index: Some(1) }, DeckLinkInfo { display_name: Some("DeckLink Quad HDMI Recorder (3)"), persistent_id: Some("DEEB492"), subdevice_index: Some(2) }, DeckLinkInfo { display_name: Some("DeckLink Quad HDMI Recorder (4)"), persistent_id: Some("DEEB493"), subdevice_index: Some(3) }]
 ```

### DIFF
--- a/compositor_pipeline/src/pipeline/input/decklink/find_device.rs
+++ b/compositor_pipeline/src/pipeline/input/decklink/find_device.rs
@@ -2,17 +2,34 @@ use decklink::{
     get_decklinks, FlagAttributeId, IntegerAttributeId, StringAttributeId, VideoIOSupport,
 };
 
-use super::{DeckLinkError, DeckLinkOptions};
+use super::{DeckLinkError, DeckLinkInfo, DeckLinkOptions};
 
 pub(super) fn find_decklink(opts: &DeckLinkOptions) -> Result<decklink::DeckLink, DeckLinkError> {
     let decklinks = get_decklinks()?;
+
+    let decklinks_info = decklinks
+        .iter()
+        .map(|decklink| {
+            let attr = decklink.profile_attributes()?;
+            Ok(DeckLinkInfo {
+                display_name: attr.get_string(StringAttributeId::DisplayName)?,
+                persistent_id: attr
+                    .get_integer(IntegerAttributeId::PersistentID)?
+                    .map(|value| format!("{value:X}")),
+                subdevice_index: attr
+                    .get_integer(IntegerAttributeId::SubDeviceIndex)?
+                    .map(|i| i as u32),
+            })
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: decklink/src/info.rs [hunk 1]
```diff
@@ -91,6 +92,9 @@ impl ProfileAttributes {
             number_of_subdevices: self.get_integer(IntegerAttributeId::NumberOfSubDevices)?,
             subdevice_index: self.get_integer(IntegerAttributeId::SubDeviceIndex)?,
             persistent_id: self.get_integer(IntegerAttributeId::PersistentID)?,
+            persistent_id_hex_str: self
```
wkozyra95, line 95: This struct is only used for debug purposes to print params in example code, so instead overriding debug info to print all those information both hex in dec I opted just to add field.

compositor_pipeline/src/pipeline/input/decklink/find_device.rs
```diff
@@ -2,17 +2,34 @@ use decklink::{
...
+        .collect::<Result<_, DeckLinkError>>()?;
+
     for mut decklink in decklinks.into_iter() {
         if is_selected_decklink(opts, &mut decklink)? {
             return Ok(decklink);
         }
     }
-    Err(DeckLinkError::NoMatchingDeckLink)
+
+    Err(DeckLinkError::NoMatchingDeckLink(decklinks_info))
```
wkozyra95, line 32: This error message will need to include a persistent id otherwise users have no good way to find their decklink id easily.

An alternative approach is running an example from decklink crate, but this is not sth we want to recommend to users in general.

WojciechBarczynski (approved):

PR #615, opened 2024-07-09
Add output type that returns data via channel

wkozyra95:
Add:
- Output type that returns RGBA wgpu textures via channel
- Output type that returns h264 and opus encoded chunks
- Publically expose `wgpu::Device` and `wgpu::Queue` (method `wgpu_ctx` on pipeline)


src/types/register_output.rs
```diff
@@ -71,11 +71,6 @@ pub enum AudioEncoderOptions {
 
         /// (**default="voip"**) Specifies preset for audio output encoder.
         preset: Option<OpusEncoderPreset>,
-
```
wkozyra95, line 74: Options was ignored

WojciechBarczynski (approved):

PR #617, opened 2024-07-09
Examples of using channel based outputs

wkozyra95:
- Add 2 examples of using output types that return data via Rust channels
- When downloading texture padding should be to 256 bytes per row not 265 pixels. (It worked before because we added that padding consistently in chromium code, and on output pixel size is always 1 byte per YUV plane)
- Fix pad_to_256 function

### DIFF
--- a/integration_tests/examples/wgpu_textures_output.rs
+++ b/integration_tests/examples/wgpu_textures_output.rs
@@ -1,0 +1,173 @@
+use core::panic;
+use std::{fs::File, io::Write, path::PathBuf, thread, time::Duration};
+
+use compositor_pipeline::{
+    audio_mixer::{AudioChannels, AudioMixingParams, AudioSamples, InputParams, MixingStrategy},
+    pipeline::{
+        input::{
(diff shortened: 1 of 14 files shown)


integration_tests/examples/wgpu_textures_output.rs
```diff
@@ -0,0 +1,173 @@
...
+                    };
+                    audio_dump
+                        .write_all(format!("{:?} {:?}\n", packet.start_pts, samples).as_bytes())
+                        .unwrap();
+                } else {
+                    return;
+                };
+            }
+        })
+        .unwrap();
```
WojciechBarczynski, line 147: Does this thread stop? Why don't you use `send_done_audio` or sth similar here?

integration_tests/examples/wgpu_textures_output.rs
WojciechBarczynski, line 1: You "test" here both, video and audio outputs. Maybe use `raw_output`?

WojciechBarczynski (approved):

PR #621, opened 2024-07-10
Add input type that returns data via channel

wkozyra95:
- Add input type that can take wgpu::Textures and raw audio
- Add example of using that input (video only)

### DIFF
--- a/compositor_render/src/wgpu/texture.rs
+++ b/compositor_render/src/wgpu/texture.rs
@@ -40,13 +40,6 @@ enum InputTextureState {
     Rgba8UnormWgpuTexture(Arc<wgpu::Texture>),
 }
 
-#[derive(Debug, Clone, Copy, PartialEq)]
-enum InputTextureKind {
-    PlanarYuvTextures,
-    InterleavedYuv422Texture,
-    Rgba8UnormWgpuTexture,
-}
-
 impl InputTextureState {
     fn resolution(&self) -> Resolution {
         match &self {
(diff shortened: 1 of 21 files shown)


integration_tests/src/test_input.rs
```diff
@@ -0,0 +1,92 @@
+use compositor_render::{scene::RGBColor, Resolution};
+
+#[derive(Debug, Clone)]
+pub struct TestInput {
```
wkozyra95, line 4: Mostly copied form snapshot test example generation, but rgba instead of yuv

compositor_render/src/wgpu/texture/base.rs
```diff
@@ -51,13 +49,40 @@ impl Texture {
...
-        ctx.queue.submit(Some(encoder.finish()));
-        dst_texture
+        let actual = (
+            self.texture.size(),
+            self.texture.mip_level_count(),
+            self.texture.sample_count(),
+            self.texture.dimension(),
+            self.texture.format(),
+            self.texture.usage(),
+        );
```
WojciechBarczynski, line 76: One should call it on the source and the other on self

compositor_render/src/wgpu/texture.rs
```diff
@@ -77,122 +86,135 @@ impl InputTexture {
...
+                let should_recreate =
+                    state.resolution() == new_resolution && state.kind() == new_texture_kind;
+                (OptionalState::Some(state), should_recreate)
             }
+            OptionalState::None => (OptionalState::None, true),
         };
-        self.state_mut().unwrap()
+        self.0 = new_state;
+        should_recreate
+    }
```
WojciechBarczynski, line 214: Revered conditions

WojciechBarczynski (approved):

PR #625, opened 2024-07-16
Fix input queueing when some of the inputs do not produce frames/samples

wkozyra95:
Steps to reproduce:
- start more than one stream when one of them is shorter than the others
- when the first stream is finished other streams might disappear

What happens:
- normally queue is
  - checking if all inputs are ready (and pulling from channel if necessary)
  - for each input:
    - dropping old frames
    - taking first frame (first not dropped should be the one we want to use)
- if one input is done without sending EOS, but just stops sending data
  - checking if all inputs are ready is lazy, so if it checks that one specific input that ended it does not check any further
  - for each input:
    - dropping old frames will work, but because previous step did not enqueue new frames from the channel we are left with one outdated frame
    - taking first frame return old frame (latter renderer can see that frame is older by more than 500ms and drops it, replacing with a black screen)
  
  
Alternative approach could be to replace call to `any` https://github.com/membraneframework/live_compositor/blob/master/compositor_pipeline/src/queue/video_queue.rs#L103 with non-lazy alternative, but I think this is cleaner.

### DIFF
--- a/compositor_pipeline/src/queue/audio_queue.rs
+++ b/compositor_pipeline/src/queue/audio_queue.rs
@@ -153,6 +153,9 @@ impl AudioQueueInput {
         pts_range: (Duration, Duration),
         queue_start: Instant,
     ) -> PipelineEvent<Vec<InputSamples>> {
+        // ignore result, we only need to ensure samples are enqueued
+        self.check_ready_for_pts(pts_range, queue_start);
+
         // range in queue pts time frame
         let (start_pts, end_pts) = pts_range;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 ### 🐛 Bug fixes
 
+- Fix input queueing when some of the inputs do not produce frames/samples . ([#625](https://github.com/membraneframework/live_compositor/pull/625) by [@wkozyra95](https://github.com/wkozyra95))
+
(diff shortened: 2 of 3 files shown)


compositor_pipeline/src/queue/audio_queue.rs
```diff
@@ -153,6 +153,9 @@ impl AudioQueueInput {
         pts_range: (Duration, Duration),
         queue_start: Instant,
     ) -> PipelineEvent<Vec<InputSamples>> {
+        // ignore result, we only need to ensure samples are enqueued
+        self.check_ready_for_pts(pts_range, queue_start);
```
WojciechBarczynski, line 157: It's awful. I hate it. Over and over we rely on some internal behavior of this function (enqueuing frames / reading them from the channel).

I'll try try to simplify this logic, cause reading all the queue code slowly feels like reading brainfuck.

WojciechBarczynski: I would prefer having a function like `get_frames` and calling it here and in the `check_ready_for_pts` function

wkozyra95: that would not work with current architecture:
- get frames needs to return function whether it's ready or not
- currently get frames relies that returned frame will be used (it's needed for example to check whether EOS was sent)

And if you are suggesting to change the above, then your suggestion boils down to almost full rewrite of queuing code.

We can talk about it more at the office, topic is to complex for discussion via comments.

PR #626, opened 2024-07-16
Fix GStreamer receive UDP command

brzep:
Spotted missing part of command in docs while refactoring `examples` dir, so fixed it.


brzep:
made mistake while creating branch so deleted both branch and pr

PR #628, opened 2024-07-16
[docs] Add search

wkozyra95:


docs/docusaurus.config.ts
```diff
@@ -97,6 +97,12 @@ const config: Config = {
       theme: prismThemes.duotoneDark,
       additionalLanguages: ['wgsl', 'http', 'elixir', 'bash'],
     },
+    algolia: {
+      appId: 'AB30AX8OU1',
+      apiKey: '4dae5f71952b8ebd63dd7645128c3b24',
```
WojciechBarczynski, line 102: It's a [search-only API key](https://www.algolia.com/doc/guides/security/api-keys/#predefined-api-keys)? Right?

wkozyra95: yes, docs for this field specifically state that this value is safe to commit

WojciechBarczynski (commented):
LGTM

PR #629, opened 2024-07-17
Move types to `compositor_api` crate

wkazmierczak:
File `error.rs` moved, due to error connected with implementing trait `From<TypeError>` for type `ApiError`.


WojciechBarczynski:
@wkozyra95 What do you think about keeping type conversions and errors in this new crate?

There are a few problems with keeping them in `compositor_pipeline`:
- Rust's orphan rule prohibits implementing foreign traits for foreign types (otherwise implementing any new trait for any rust lib would be a breaking change for this lib). Implementing `From` in `compositor_pipeline` for `compositor_renderer` types breaks this rule. I didn't find a simple way to leave those conversions in `compositor_pipeline` and I think it's best to have them next to types definitions. This means, that `compositor_api` would have those conversions and would depend on `compositor_renderer`. I don't know if it's acceptable. WDYT?
- ApiError has 2 impls:
```rust
impl<T> From<T> for ApiError
where
    T: std::error::Error + 'static,
    PipelineErrorInfo: for<'a> From<&'a T>,
{
    fn from(err: T) -> Self {
        let stack: Vec<String> = ErrorStack::new(&err).map(ToString::to_string).collect();
        let err_info = PipelineErrorInfo::from(&err);
        ApiError {
            error_code: err_info.error_code,
            message: stack.first().unwrap().clone(),
            stack,
            http_status_code: match err_info.error_type {
                ErrorType::UserError => StatusCode::BAD_REQUEST,
                ErrorType::ServerError => StatusCode::INTERNAL_SERVER_ERROR,
                ErrorType::EntityNotFound => StatusCode::NOT_FOUND,
            },
        }
    }
}

impl From<TypeError> for ApiError {
    fn from(err: TypeError) -> Self {
        ApiError::malformed_request(&err)
    }
}
```

If TypeError is an external type, rust sees the second definition as an override of the first one (if someone adds `T: std::error::Error` for `TypeError`, the compiler wouldn't know which impl it should use).

This means, that it's tricky to keep those `impl` and `TypeError` in different crates.

WDYT? Do you have any suggestions on how to do it better?

WojciechBarczynski (approved):

WojciechBarczynski:
@wkazmierczak wait with merge for @wkozyra95 response

src/lib.rs
```diff
@@ -1,8 +1,6 @@
 pub mod config;
-pub mod error;
+// pub mod error;
```
wkozyra95, line 2: remove

src/main.rs
```diff
@@ -1,12 +1,11 @@
 use log::info;
 
 mod config;
-mod error;
+// mod error;
```
wkozyra95, line 4: remove

integration_tests/Cargo.toml
```diff
@@ -19,6 +19,7 @@ web_renderer = [
 ]
 
 [dependencies]
+compositor_api = { path = "../compositor_api" }
```
wkozyra95, line 22: add features to compositor_api and make sure it's passed to other `compositor_*` dependencies

integration_tests/Cargo.toml
```diff
@@ -7,20 +7,15 @@ license = "BUSL-1.1"
...
+decklink = ["live_compositor/decklink"]
+web_renderer = ["dep:compositor_chromium", "live_compositor/web_renderer"]
 
 [dependencies]
-live_compositor = { path = ".." }
-compositor_pipeline = { path = "../compositor_pipeline" }
+live_compositor = { workspace = true }
+compositor_api = { workspace = true }
+compositor_pipeline = { workspace = true }
+compositor_chromium = { workspace = true, optional = true }
```
wkozyra95, line 18:
```suggestion
compositor_chromium = { workspace = true, optional = true }
```

compositor_api/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[features]
+decklink = ["compositor_pipeline/decklink"]
+web_renderer = ["compositor_render/web_renderer"]
+
+[dependencies]
+compositor_pipeline = { path = "../compositor_pipeline" }
+compositor_render = { path = "../compositor_render" }
```
wkozyra95, line 12:
```suggestion
compositor_pipeline = { workspace = true }
compositor_render = { workspace = true }
```

wkozyra95 (approved):

PR #632, opened 2024-07-17
Switch to nextest

wkozyra95:


generate/src/compositor_instance.rs
```diff
@@ -103,7 +103,7 @@ impl CompositorInstance {
 
 fn get_free_port() -> u16 {
     static LAST_PORT: OnceLock<AtomicU16> = OnceLock::new();
-    let port = LAST_PORT.get_or_init(|| AtomicU16::new(10_000 + (rand::random::<u16>() % 10_000)));
+    let port = LAST_PORT.get_or_init(|| AtomicU16::new(10_000 + (rand::random::<u16>() % 50_000)));
```
wkozyra95, line 106: This should be safe enough with retry to avoid collisions. The proper way to handle it would to be have some persistent state and mutex that allow concurrent access from multiple processes, so I do no really want to go that route if I don't have to.

compositor_chromium/examples/chromium_simple.rs
wkozyra95, line 1: rename to avoid warning about name collisions

WojciechBarczynski (commented):
LGTM

PR #648, opened 2024-07-26
[website] SEO and load time optimization

WojciechBarczynski:
Change list:
- Added `robots.txt` to specify permissions for website crawlers. I allowed unrestricted access to the whole content of the website.
- Changed `.png` and `.jpeg` images to `.webp` format. `.webp` offers similar quality with significantly reduced size. It improved website loading and responsiveness times by around 30-40%.
- Added metadata tags that are used by crawlers. They appear in the HTML head. In `docosaurus` you can specify additional content of the metadata section in the `themeConfig`
- Switched to a single `H1` header. Other sections headers were changed to `H2` and subtitles to `H3`. I adjusted font sizes to match font sizes before this change and slightly increased (`2.0rem` -> `2.25rem`) font size in the hero title section. All 5 SEO checkers I used recommended doing that.

Things recommended by SEO checkers I didn't implement (I didn't include everything):
- Removing repeating header content across pages. We repeat some headers in docs, like `Static positioning`, but it makes sense and makes docs coherent/readable, so I didn't change that.
- Some security recommendations, like using HSTS. I think this is configurable on the deployment level.


compositor_api/src/types/component.rs
WojciechBarczynski, line 1: New clippy

WojciechBarczynski:
I deployed the website from the branch using GH Actions workflow dispatch (GUI). You can test it: https://compositor.live/

wkozyra95 (approved):

PR #650, opened 2024-07-31
add generating playground example input PNGs

brzep:
Simple script generating sample inputs for playground.

### DIFF
--- a/generate/src/bin/generate_playground_inputs.rs
+++ b/generate/src/bin/generate_playground_inputs.rs
@@ -1,0 +1,166 @@
+use std::{fs, path::PathBuf, process::Command, thread};
+
+use anyhow::Result;
+use compositor_render::Resolution;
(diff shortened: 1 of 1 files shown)


WojciechBarczynski:
@wkozyra95 
Overall, what inputs would you want to add to the playground? Those are ok, but they are pretty boring. I would prefer to have something like BigBuckBunny/video call/sports event camera/TV/gameplay screenshots highlighting potential use cases for LiveCompsoitor.

wkozyra95:
@WojciechBarczynski 
As long as we use static images it does not make sense to add stuff like this. When we start to support video then yes adding real videos will be needed, but even then those generic inputs will still be needed (but maybe with timer, the same as in docs)

wkozyra95 (approved):

WojciechBarczynski (approved):

PR #661, opened 2024-08-09
[test] Add offline processing integration test

WojciechBarczynski:


integration_tests/src/tests/offline_processing.rs
```diff
@@ -0,0 +1,171 @@
...
+    config.queue_options.never_drop_output_frames = true;
+    let instance = CompositorInstance::start(Some(config));
+    let (msg_sender, msg_receiver) = crossbeam_channel::unbounded();
+    start_server_msg_listener(instance.api_port, msg_sender);
+
+    instance.send_request(
+        "input/input_1/register",
+        json!({
+            "type": "mp4",
+            "url":  BUNNY_URL,
```
WojciechBarczynski, line 30:
```suggestion
            "url": BUNNY_URL,
```

integration_tests/src/tests/offline_processing.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    Command::new("ffprobe")
+        .args(["-v", "error", OUTPUT_FILE])
+        .output()
+        .map_err(|e| anyhow!("Invalid mp4 file. FFprobe error: {}", e))?;
+
+    Ok(())
+}
+
+pub fn start_server_msg_listener(port: u16, event_sender: Sender<tungstenite::Message>) {
```
wkozyra95, line 108: can you extract websocket helpers to the crate in a way that can be adapted to use in other tests too?

wkozyra95: I'm not sugesting it but just FYI, as an alternative to the websockets you could subscribe to the events directlly (you will get a channel) 

https://github.com/software-mansion/live-compositor/blob/master/compositor_render/src/event_handler.rs

integration_tests/src/tests/offline_processing.rs
```diff
@@ -0,0 +1,171 @@
...
+    for msg in msg_receiver.iter() {
+        if let tungstenite::Message::Text(msg) = msg {
+            if msg.contains("\"type\":\"OUTPUT_DONE\",\"output_id\":\"output_1\"") {
+                info!("breaking");
+                break;
+            }
+        }
+    }
+
+    Command::new("ffprobe")
```
wkozyra95, line 100: Can we get duration or bitrate with regexp from the output 

Sth like this should be easy to match, but I'm not sure if ffmpeg output is consistent enough
```
  Duration: 00:09:56.47, start: 0.000000, bitrate: 2119 kb/s
```

integration_tests/src/tests/utils.rs
```diff
@@ -0,0 +1,68 @@
+use crossbeam_channel::Sender;
```
wkozyra95, line 1: move that somwhere else, lets keep tests directory only for tests

wkozyra95 (approved):

PR #667, opened 2024-08-12
Allow passing wgpu::Device and wgpu::Queue

wkozyra95:


src/snapshot_tests/utils.rs
```diff
@@ -67,6 +84,7 @@ pub(super) fn create_renderer() -> Renderer {
         framerate: Framerate { num: 30, den: 1 },
         stream_fallback_timeout: Duration::from_secs(3),
         wgpu_features: WgpuFeatures::default(),
+        wgpu_ctx: Some(global_wgpu_ctx(false, Default::default())),
```
WojciechBarczynski, line 87: Are our required features == with `wgpu::Features` default? I was under the impression that we need more than that.

WojciechBarczynski (approved):

PR #668, opened 2024-08-12
Add separate devShell for NixOS

wkozyra95:


wkozyra95:
@noituri Can you check if that segfault still happen in ffmpeg on macOS?

PR #673, opened 2024-08-19
TS SDK

wkozyra95:
### Packages

- `live-compositor` - main package user interacts, minimal dependencies, this package would be the only that third party libs should depend
- `@live-compositor/node` - Implements node specific stuff, used to create instance of the compositor, there will be `react` and `react-native` variants in the future
- `@live-compositor/core` - Implements platform independent part that will be used by packages like  `@live-compositor/node` or react/react-native
- `examples`
  - `simple` - basic View + Text example with some state updates.
  - `dynamic-text` - Component that shows text adding letter by letter (and removing it back in the loop)
  - `news-ticker` - Component that implements moving bar of text that loops on itself (video below)

https://github.com/user-attachments/assets/c9b14144-be25-4703-9562-ca26a6da035f


ts/live-compositor/src/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  RescalerProps,
+  WebView,
+  WebViewProps,
+  Shader,
+  ShaderProps,
+  Tiles,
+  TilesProps,
+};
+
+export { ShaderParam, ShaderParamStructField, EasingFunction, Transition };
```
wkozyra95, line 39: I'm not sure about that, problem here is that types are partially defined in API and partially overiden here

ts/live-compositor/README.md
```diff
@@ -0,0 +1,45 @@
...
+    },
+  });
+  await compositor.start();
+}
+run();
+```
+
+## Documentation
+
+Components props are documented via TypeScript types definitions in this package. You can also refer to LiveCompositor (e.g. [View](https://compositor.live/docs/api/components/View)), but keep in mind to convert field names from snake case to camel case.
```
wkozyra95, line 43: I plan to add separate docs for typescript, but I did not think it through yet how it will work

ts/live-compositor/src/components/View.ts
```diff
@@ -0,0 +1,104 @@
...
+   */
+  transition?: Transition;
+  /**
+   * (**default=`"hidden"`**) Controls what happens to content that is too big to fit into an area.
+   */
+  overflow?: Api.Overflow;
+  /**
+   * (**default=`"#00000000"`**) Background color in a `"#RRGGBBAA"` format.
+   */
+  backgroundColorRgba?: Api.RGBAColor;
```
wkozyra95, line 74: I'm considering renaming that to `backgroundColor` and supporting more formats than the LiveCompositor server does.

ts/live-compositor/src/types/registerInput.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export type RegisterMp4Input = {
+  /**
+   * URL of the MP4 file.
+   */
+  url?: string | null;
+  /**
+   * Path to the MP4 file.
+   */
+  path?: string | null;
```
wkozyra95, line 45: maybe serverPath to avoid confusion if the node is deployed in a different location than the compositor.

WojciechBarczynski (approved):

ts/live-compositor/src/components/Image.ts
```diff
@@ -0,0 +1,29 @@
+import * as Api from '../api';
+import LiveCompositorComponent, { SceneBuilder, SceneComponent } from '../component';
+
+export type ImageProps = {
+  children?: undefined;
+
+  /**
+   * Id of a component.
+   */
+  id?: Api.ComponentId | null;
```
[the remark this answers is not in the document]
wkozyra95: ? means no value or undefined, null is not allowed there if it's just `?`, but on the other hand in other places I was removing that `| null` (generation from json schema is producing that)

PR #682, opened 2024-08-21
[website] Add tailwind + RTC.ON banner

wkozyra95:
When scrolling down it covers a copyright section on the website, but I don't think it's worth fixing
Also in the future I would prefer to start slowly using tailwind classes over infima/docusaurus.

20240821_18h48m29s_grim
20240821_18h48m48s_grim


docs/src/components/RtcOnBanner.tsx
```diff
@@ -0,0 +1,55 @@
+import Link from '@docusaurus/Link';
```
wkozyra95, line 1: mostly copied from membrane.stream repo

WojciechBarczynski (approved):

PR #683, opened 2024-08-22
Initial support for running `compositor_render` on web

noituri:


compositor_render/src/wgpu/ctx.rs
```diff
@@ -42,8 +42,14 @@ impl WgpuCtx {
     }
 
     fn check_wgpu_ctx(device: &wgpu::Device, features: wgpu::Features) {
+        #[cfg(not(target_arch = "wasm32"))]
```
wkozyra95, line 45: I would do that with `if` or `match`

Cargo.toml
```diff
@@ -54,10 +51,18 @@ opus = "0.3.0"
 rubato = "0.15.0"
 futures-util = "0.3.30"
 tokio = { version = "1", features = ["full"] }
-wgpu = { version = "0.19.1", features = ["naga-ir"] }
 schemars = { git = "https://github.com/membraneframework-labs/schemars", rev = "a5ad1f9", features = [
     "preserve_order",
 ] }
+shared_memory = "0.12.4"
+wgpu = { version = "0.19.1", default-features = false, features = [
```
wkozyra95, line 58: What are you actually changing here, is this disabling some default features, if so which?

compositor_api/Cargo.toml
```diff
@@ -9,9 +9,12 @@ decklink = ["compositor_pipeline/decklink"]
...
 
 [dependencies]
-compositor_pipeline = { workspace = true }
 compositor_render = { workspace = true }
 serde = { workspace = true }
 schemars = { workspace = true }
 bytes = { workspace = true }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+compositor_pipeline = { workspace = true }
```
wkozyra95, line 18: it would be cleaner to control that with a feature, but if it's fine if you want to leave it this way

compositor_render/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license = "BUSL-1.1"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [features]
-web_renderer = ["dep:compositor_chromium"]
+web_renderer = ["dep:compositor_chromium", "shared_memory"]
```
wkozyra95, line 9:
```suggestion
web_renderer = ["dep:compositor_chromium", "dep:shared_memory"]
```
is this correct?

compositor_render/src/transformations/layout/params.rs
```diff
@@ -11,6 +11,17 @@ pub(super) struct LayoutNodeParams {
     pub(super) background_color: RGBAColor,
 }
 
+impl LayoutNodeParams {
+    pub fn empty() -> Self {
```
wkozyra95, line 15: not sure if it is idiomatic, but maybe Default?

Cargo.toml
noituri: By default `wgpu` has `webgpu` feature enabled which has higher priority than `webgl` feature. Currently, compositor does not run on `webgpu` due to lack of support for push  constants on `webgpu` (they are not emulated like on `webgl`)

compositor_render/Cargo.toml
noituri: Web renderer works on my branch so I'm assuming that it works 😅. I'll take a closer look into it

Cargo.toml
noituri: Also, I added `fragile-send-sync-non-atomic-wasm` feature, which makes the wgpu types `Send` (by default they are not `Send` and `Sync` on wasm). This feature is slightly unsafe because on wasm those types should not  be shared between threads and it removes the compile time checks.

But, since we target web browsers which run the app on single thread it should not be problem for us. `LayoutProvider` needs to be  `Send` because `compositor_pipeline` uses threads. We could also solve it by making 2 `LayoutProvider`s (one Send and the other one non-Send) and enabling one based on `#[cfg(target_arch = "wasm32")]`. I decided to use the feature because it was just the easier way

compositor_render/Cargo.toml
noituri: I'm not sure why it work without `dep:`. I'll add `dep:` anyway for consistency

wkozyra95 (approved):
I would recommend rethinking file structure. e.g.
- you have utils file that only contains stuff related to wgpu
- you have app module that seems to sth top level, but it is only used in utils package

compositor_web/src/wasm.rs
```diff
@@ -0,0 +1,122 @@
...
+        _surface: surface,
+    })
+}
+
+#[wasm_bindgen]
+pub struct LiveCompositorRenderer {
+    renderer: Renderer,
+    input_uploader: InputUploader,
+    output_downloader: OutputDownloader,
+    _surface: wgpu::Surface<'static>,
```
WojciechBarczynski, line 51: What this `_surface` do?

.github/workflows/lint.yml
```diff
@@ -56,7 +56,9 @@ jobs:
               run: cargo fmt --all --check
 
             - name: 📎 Run clippy
-              run: cargo clippy --workspace --all-targets --all-features -- -D clippy::todo -D warnings
+              run: |
+                cargo clippy --workspace --all-targets --all-features -- -D clippy::todo -D warnings
+                # cargo clippy -p compositor_web --all-targets --all-features --target=wasm32-unknown-unknown -- -D clippy::todo -D warnings
```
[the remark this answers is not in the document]
noituri: I forgot to remove it. Currently it does not work well because web renderer is giving a lot of warnings and had to comment it out

compositor_render/src/transformations/layout/apply_layouts.wgsl
```diff
@@ -17,15 +17,15 @@ struct Layout {
 }
 
 @group(0) @binding(0) var texture: texture_2d<f32>;
-@group(1) @binding(0) var<uniform> layouts: array<Layout, 128>;
+@group(1) @binding(0) var<uniform> layouts: array<Layout, 100>;
```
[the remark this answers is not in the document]
noituri: `128` was too large for webGL. I had to make the array smaller

compositor_web/src/wasm.rs
noituri: It represents a surface that you can render onto. In webGL you can't render in the headless mode and surface is required. AFAIK `wgpu` uses it to retrieve WebGL context

.github/workflows/lint.yml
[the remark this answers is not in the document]
wkozyra95: If you want leave it commented out add a comment with explanation, if not just remove it

WojciechBarczynski (approved):

wkozyra95:
Do not merge everything into this branch, if this is ready then merge it

PR #684, opened 2024-08-22
Automatically rename existing mp4 files under `register_output.mp4.path` location to `[path].old.[index]`

WojciechBarczynski:
When working on the release demo I found it extremely annoying, that I couldn't automatically overwrite the output file, so I decided to add this option.

Update:
Decided to go with renaming the file instead

### DIFF
--- a/compositor_pipeline/src/pipeline/output/mp4.rs
+++ b/compositor_pipeline/src/pipeline/output/mp4.rs
@@ -1,8 +1,4 @@
-use std::{
-    fs,
-    path::PathBuf,
-    ptr,
-};
+use std::{fs, path::PathBuf, ptr};
 
 use compositor_render::{event_handler::emit_event, OutputId};
 use crossbeam_channel::Receiver;
--- a/compositor_pipeline/src/pipeline/output/mp4.rs
+++ b/compositor_pipeline/src/pipeline/output/mp4.rs
@@ -2,9 +2,9 @@ use std::{fs, path::PathBuf, ptr};
 
 use compositor_render::{event_handler::emit_event, OutputId};
 use crossbeam_channel::Receiver;
-use ffmpeg_next::{self as ffmpeg};
+use ffmpeg_next as ffmpeg;
 use log::error;
-use tracing::debug;
+use tracing::{debug, warn};
 
 use crate::{
     audio_mixer::AudioChannels,
@@ -65,6 +65,11 @@ impl Mp4FileWriter {
                 old_index += 1;
             }
 
+            warn!(
+                "Output file {} already exists. Renaming to {}.",
+                options.output_path.to_string_lossy(),
+                new_path_for_old_file.to_string_lossy()
+            );
             if let Err(err) = fs::rename(options.output_path.clone(), new_path_for_old_file) {
                 error!("Failed to rename existing output file. Error: {}", err);
             };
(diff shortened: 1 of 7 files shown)


wkozyra95 (commented):
Overwrite is not really clear in the context of the output. Yes if you read the docs you will know what it does, but I have a feeling that it does not fit here.

Option to select whether a file should be overiden or not does not seem to be necessary, and if possible I prefer to avoid adding unnecessary stuff. To solve problem from PR description I would check if file already exists and rename it e.g.
`test.mp4` -> `test.mp4.bak.0`/test.mp4.bak.1`/test.mp4.bak.2`

WojciechBarczynski:
@wkozyra95 
The name comes from `ffmpeg` `-y` option docs: https://ffmpeg.org/ffmpeg.html#Main-options

It is a simple quality of usage improvement. Sure, users can do it outside the compositor, but if it's a commonly occurring problem, I don't see the reason not to implement it.

Your solution with versioned files is less ergonomic, as users have to check the name of the generated file to play it. My "workflow" while working with offline processing examples was:
1. Write script generating output MP4
2. Play it
3. Fix stuff
4. Play it
5. Fix stuff
6. Play it ...

Without this change, I needed to delete the file before generating it (I often forgot about it / tried deleting the file even though it wasn't generated). Your solution is also far from perfect for this since I will need to check the name of the last generated file to play it.

In the end, I think that if `ffmpeg` supports this as one of the main options, it's worth adding it.

WojciechBarczynski:
I'm not putting it in the changelog, as it seems too specific/minor to be included there.

compositor_pipeline/src/pipeline/output/mp4.rs
```diff
@@ -1,8 +1,8 @@
-use std::{path::PathBuf, ptr};
+use std::{fs, path::PathBuf, ptr};
 
 use compositor_render::{event_handler::emit_event, OutputId};
 use crossbeam_channel::Receiver;
-use ffmpeg_next as ffmpeg;
+use ffmpeg_next::{self as ffmpeg};
```
wkozyra95, line 5: ?

compositor_pipeline/src/pipeline/output/mp4.rs
```diff
@@ -51,9 +51,23 @@ impl Mp4FileWriter {
...
+                ));
+                if !new_path_for_old_file.exists() {
+                    break;
+                }
+                old_index += 1;
+            }
+
+            if let Err(err) = fs::rename(options.output_path.clone(), new_path_for_old_file) {
+                error!("Failed to rename existing output file. Error: {}", err);
+            };
```
wkozyra95, line 70: Add warn log that file already exists, renaming to <new path>

wkozyra95 (approved):

wkozyra95:
> I'm not putting it in the changelog, as it seems too specific/minor to be included there.

I don't know, this is sth that might not be worth documenting, but I think CHANGELOG is a place where it might be worth mentioning it. No current section really fits that, we could add `Other` section maybe or in expo we had `Chores` (https://github.com/expo/eas-cli/blob/main/CHANGELOG.md)

Up to you, I don't have strong opinion on that.

compositor_pipeline/src/pipeline/output/mp4.rs
WojciechBarczynski: I imported and removed some stuff from `ffmpeg_next` accidentally and missed that. Reverting.

PR #686, opened 2024-08-23
Offline processing guide

WojciechBarczynski:


docs/pages/guides/basic-layouts.md
```diff
@@ -54,7 +54,7 @@ Update output to render a [`View`](../api/components/View.md) component with an
           type: :view,
           background_color_rgba: "#4d4d4dff",
           children: [
-            %{ type: :input_stream, input_id: :input_1 }
```
WojciechBarczynski, line 57: I decided to use strings instead of atoms for input ids everywhere for consistency.
It's also the typing used in docs: https://hexdocs.pm/membrane_live_compositor_plugin/Membrane.LiveCompositor.html#t:input_id/0

docs/pages/guides/offline-processing.md
```diff
@@ -0,0 +1,526 @@
...
+  </TabItem>
+</Tabs>
+
+### Add MP4 output `output_1`
+
+Here we won't focus on configuring the composition itself, as it's the same as in any other LiveCompositor use case.
+For more advanced layouts, check out the [`Basic Layouts guide`](./basic-layouts.md).
+
+We will configure the output to simply:
+- Rescale video from `input_1` to output resolution using the [`Rescaler`](../api/components/Rescaler.md) component.
```
wkozyra95, line 210: make it simpler, pass through input1 without recalling

docs/pages/guides/offline-processing.md
```diff
@@ -0,0 +1,526 @@
...
+      ]
+      {[spec: spec], %{}}
+    end
+    ```
+  </TabItem>
+</Tabs>
+
+### Schedule update output to show both inputs
+
+We are going to schedule output update 5s after the beginning and configure it to:
```
wkozyra95, line 301:
```suggestion
Let's add a second input at the 5-second mark of the video:
```

docs/pages/guides/offline-processing.md
```diff
@@ -0,0 +1,526 @@
...
+- Show both inputs side-by-side using the [`Tiles`](../api/components/Tiles.md) component.
+- Mix audio from both inputs, where `input_2` volume is reduced to 35% of the original volume.
```
wkozyra95, line 303: input_2 does not exists for membrane example

wkozyra95 (commented):
I did not review it fully, but I think this approach has a few major problems:
- This guide is very complicated and either throws a ton of information at the user or does not explain stuff at all.
- You need to go through the entire guide to test anything here.
- Parts of this guide seems more like reference than guide.

I added some comments in the PR, but I don't think the core issues with this approach can be improved. I believe that format for other guides does not work here.

In general, I think that the threshold for adding a guide here should be that at least it needs to be better than providing one big example code with some comments. I don't believe this is the case here.

PR #689, opened 2024-08-26
[ts-sdk] Add `useInputStreams` hook

wkozyra95:
- Add `useInputStreams` to interact with new streams
- Add lerna + improve build/release automation (e.g. running npm run watch for all workspaces)
- Ensure package is build when publishing

### DIFF
--- a/ts/@live-compositor/core/package.json
+++ b/ts/@live-compositor/core/package.json
@@ -6,12 +6,16 @@
   "scripts": {
     "lint": "eslint .",
     "typecheck": "tsc --noEmit",
-    "watch": "tsc --watch",
+    "watch": "tsc --watch --preserveWatchOutput",
     "build": "tsc",
-    "clean": "rimraf dist"
+    "clean": "rimraf dist",
+    "prepublishOnly": "npm run clean && npm run build"
   },
   "author": "",
   "license": "MIT",
+  "files": [
+    "/dist"
+  ],
   "devDependencies": {
     "@types/react-reconciler": "0.28.8"
   },
--- a/ts/@live-compositor/core/src/compositor.ts
+++ b/ts/@live-compositor/core/src/compositor.ts
@@ -1,4 +1,4 @@
-import { RegisterInput, RegisterOutput } from 'live-compositor';
+import { ContextStore, RegisterInput, RegisterOutput } from 'live-compositor';
 import { ApiClient, Api } from './api';
 import Output from './output';
 import { CompositorManager } from './compositorManager';
@@ -14,6 +14,7 @@ export async function createLiveCompositor(manager: CompositorManager): Promise<
 export class LiveCompositor {
   private manager: CompositorManager;
   private api: ApiClient;
+  private store: ContextStore = new ContextStore();
   private outputs: Record<string, Output> = {};
 
   public constructor(manager: CompositorManager) {
@@ -22,8 +23,8 @@ export class LiveCompositor {
   }
 
   public async registerOutput(outputId: string, request: RegisterOutput): Promise<object> {
-    const output = new Output(outputId, request, this.api);
-    const { video: initialVideo } = output.scene();
+    const output = new Output(outputId, request, this.api, this.store);
+    const initialVideo = output.scene();
 
     const apiRequest = intoRegisterOutput(request, initialVideo);
     const result = await this.api.registerOutput(outputId, apiRequest);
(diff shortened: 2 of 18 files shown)


ts/@live-compositor/core/package.json
```diff
@@ -6,12 +6,16 @@
   "scripts": {
     "lint": "eslint .",
     "typecheck": "tsc --noEmit",
-    "watch": "tsc --watch",
+    "watch": "tsc --watch --preserveWatchOutput",
```
wkozyra95, line 9: Terminal will not be cleared after file is saved and errors/warnings are printed again. This is needed to stream `npm run output` from multiple commands

ts/@live-compositor/core/package.json
```diff
@@ -6,12 +6,16 @@
...
     "build": "tsc",
-    "clean": "rimraf dist"
+    "clean": "rimraf dist",
+    "prepublishOnly": "npm run clean && npm run build"
   },
   "author": "",
   "license": "MIT",
+  "files": [
+    "/dist"
```
wkozyra95, line 17: only include dist directory to minimize size of the package

ts/@live-compositor/core/src/compositor.ts
```diff
@@ -14,6 +14,7 @@ export async function createLiveCompositor(manager: CompositorManager): Promise<
 export class LiveCompositor {
   private manager: CompositorManager;
   private api: ApiClient;
+  private store: ContextStore = new ContextStore();
```
wkozyra95, line 17: this value is shared between all outputs

noituri (approved):

ts/@live-compositor/core/src/compositor.ts
WojciechBarczynski, line 17: [NIT]
Personal opinion here, but `store` as field name sounds weird. It doesn't indicate what is "stored" there. I would use `context`.

ts/@live-compositor/core/src/output.ts
```diff
@@ -1,57 +1,91 @@
...
-      this.audioOptions = intoAudioInputsConfiguration(output.audio.initial);
+      audioOptions = intoAudioInputsConfiguration(output.audio.initial);
     }
-    this.initialized = true;
+    this.throttledUpdate = throttle(async () => {
+      await api.updateScene(this.outputId, {
+        video: this.videoRenderer && { root: this.videoRenderer.scene() },
+        audio: audioOptions,
+      });
+    }, 30);
```
WojciechBarczynski, line 35: Where `30` come from? With 60FPS per frame updates happen every 16ms.

WojciechBarczynski (approved):

ts/@live-compositor/core/src/output.ts
wkozyra95: > Where 30 come from? 

I think this is optimal value for limiting the updates

> With 60FPS per frame updates happen every 16ms.

I agree with the math, but I don't see what point you are making.

ts/@live-compositor/core/src/compositor.ts
wkozyra95: this is store, context it's what returned from the hooks

ts/@live-compositor/core/src/output.ts
WojciechBarczynski: With this throttling, it's impossible to do per-frame updates in 60FPS.

ts/@live-compositor/core/src/compositor.ts
WojciechBarczynski: What do you mean by `store`? Every data structure can be referred to as a `store`. Every struct, map, list, etc. "stores" sth. My point is that this name doesn't indicate what is stored there. 

When you see:
```
compositor.store
```
It's impossible to tell what is in `store` without checking this `LiveCompositor` class definition/field type.

ts/@live-compositor/core/src/output.ts
wkozyra95: I don't think it should be possible to do any per frame stuff, there is no way to synchronize it either way

ts/@live-compositor/core/src/compositor.ts
wkozyra95: - this internal implementation, this is not user facing
- store is well established name in react and probably even in general in web ecosystem

PR #691, opened 2024-08-27
Add output resolution settings

brzep:
added fields allowing for output resolution customisation:
- input validation with error messages
- integration with submit button
fixed submit button not actually being disabled on error


docs/src/components/PlaygroundSettingsOutput.tsx
```diff
@@ -0,0 +1,157 @@
...
+  setValidity,
+}: ResolutionInputFieldProps) {
+  enum ValidationResult {
+    Ok = 'Ok',
+    TooLargeError = 'TooLarge',
+    TooSmallError = 'TooSmall',
+    UnevenError = 'Uneven',
+    ParsingError = 'ParsingError',
+  }
+  function validationResultMessage(validationResult: ValidationResult, maxValue: number) {
```
wkozyra95, line 89: if you use pure function do not declare them inside component

docs/src/components/PlaygroundSettingsOutput.tsx
```diff
@@ -0,0 +1,157 @@
...
+  function setInputValidationResult(result: ValidationResult) {
+    if (result == ValidationResult.Ok) {
+      setValidity(true);
+    } else {
+      setValidity(false);
+    }
+    setValidationResult(result);
+  }
+
+  function parseAndValidate(input_string: string): number {
```
wkozyra95, line 113: extract validation logic to pure function and here only call that function and update state

docs/src/components/PlaygroundSettingsOutput.tsx
```diff
@@ -0,0 +1,157 @@
...
+}
+
+function ResolutionInputField({
+  id,
+  value,
+  maxValue,
+  onValueChange,
+  setValidity,
+}: ResolutionInputFieldProps) {
+  enum ValidationResult {
```
wkozyra95, line 82: move enum from component

wkozyra95 (approved):

PR #692, opened 2024-08-27
WASM checks in CI

noituri:

### DIFF
--- a/.github/workflows/wasm.yml
+++ b/.github/workflows/wasm.yml
@@ -1,0 +1,40 @@
+name: wasm build
+on:
+    push:
+        branches: [master]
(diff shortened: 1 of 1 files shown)


WojciechBarczynski (commented):
I don't like that it substantially increases build time in every CI check. I would rather have a separate job that runs on pushes to `compositor_web`

noituri:
> I don't like that it increases build time by almost 2min in every CI check. I would rather have a separate job that runs on pushes to `compositor_web`

I'll make it a separate job but I don't want to limit it to `compositor_web` since changes outside of that crate can break it.

WojciechBarczynski (approved):

PR #694, opened 2024-08-27
Fix `compositor_render` sometime ignoring output format

WojciechBarczynski:


compositor_render/src/state/render_loop.rs
```diff
@@ -91,31 +91,74 @@ pub(super) fn read_outputs(
...
+                                view_formats: &[wgpu::TextureFormat::Rgba8Unorm],
+                            });
+                    let view = wgpu_texture.create_view(&wgpu::TextureViewDescriptor::default());
+                    let compositor_texture = crate::wgpu::texture::Texture {
+                        texture: wgpu_texture,
+                        view,
+                    };
+                    ctx.wgpu_ctx
+                        .utils
+                        .fill_r8_with_value(ctx.wgpu_ctx, &compositor_texture, 0.0);
```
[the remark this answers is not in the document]
wkozyra95: transparent is fine, no strong opinion

compositor_render/src/state/render_loop.rs
```diff
@@ -91,31 +91,74 @@ pub(super) fn read_outputs(
...
+                    let pending_download = output.output_texture.start_download(ctx.wgpu_ctx);
+                    partial_textures.push(PartialOutputFrame::PendingYuvDownload {
+                        output_id: output_id.clone(),
+                        pending_download,
+                        resolution: output.output_texture.resolution().to_owned(),
+                    });
+                }
+                OutputFrameFormat::RgbaWgpuTexture => {
+                    let output_resolution = output.output_texture.resolution();
+                    let wgpu_texture =
```
wkozyra95, line 122: this too low level code to keep here I would:
- Add create_texture method in RGBATexture struct and try to reuse some code
or
- Run `RGBATexture::new()` and extract raw texture from it

compositor_render/src/state/render_loop.rs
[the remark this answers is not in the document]
WojciechBarczynski: The drawback is that YUV returns black in such cases. It would be weird to return different output content for different output types (with the same scene).

Only the fack that RGBA supports transparency and YUV doesn't make this somewhat justifiable.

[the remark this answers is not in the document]
wkozyra95: this is not public API, it's different output type, I think it's fine to skip unnecessary operations

compositor_render/src/state/render_loop.rs
```diff
@@ -91,31 +91,52 @@ pub(super) fn read_outputs(
...
+                }
+                OutputFrameFormat::RgbaWgpuTexture => {
+                    let resolution = output.output_texture.resolution();
+                    let rgba_texture = RGBATexture::new(ctx.wgpu_ctx, resolution);
+                    ctx.wgpu_ctx.utils.fill_r8_with_value(
+                        ctx.wgpu_ctx,
+                        rgba_texture.texture(),
+                        0.0,
+                    );
+                    let wgpu_texture = rgba_texture.texture().copy_wgpu_texture(ctx.wgpu_ctx);
```
wkozyra95, line 128: just extract the texture from from rgba_texture you don't need to copy it

wkozyra95 (approved):

compositor_render/src/state/render_loop.rs
WojciechBarczynski: I would need to either:
1. Make it pub
```rust
pub struct RGBATexture(pub Texture);
```
2. Add method returning texture
```rust
    pub fn texture(&self) -> &Texture {
        &self.0
    }

    pub fn texture_owned(self) -> Texture {
        self.0
    }
```

I can't do
```rust
let wgpu_texture = rgba_texutre.texture().texture
```
since `texture()` returns `&compositor_render::wgpu::texture::base::Texture` and I can't move out `wgpu::Texture` out of ref.

wkozyra95: yes In know. I would go with option 2. Not sure about the name, but don't have better idea

PR #698, opened 2024-08-27
Return `InputAlreadyRegistered` error on duplicated input register with existing `InputId`

WojciechBarczynski:


compositor_pipeline/src/pipeline/pipeline_input.rs
```diff
@@ -17,50 +15,52 @@ pub struct PipelineInput {
...
+impl Pipeline {
+    pub(super) fn register_pipeline_input<NewInputResult>(
+        &mut self,
+        input_id: InputId,
+        input_options: &dyn InputOptionsExt<NewInputResult>,
+        queue_options: QueueInputOptions,
+    ) -> Result<NewInputResult, RegisterInputError> {
+        if self.inputs.contains_key(&input_id) {
+            return Err(RegisterInputError::AlreadyRegistered(input_id));
+        }
```
WojciechBarczynski, line 27: For consistency with `register_output`

wkozyra95 (changes requested):
Do not change the method signature, it will block queue

compositor_pipeline/src/pipeline/pipeline_input.rs
```diff
@@ -17,50 +15,52 @@ pub struct PipelineInput {
     pub(super) video_eos_received: Option<bool>,
 }
 
-pub(super) fn register_pipeline_input<NewInputResult>(
-    pipeline: &Arc<Mutex<Pipeline>>,
-    input_id: InputId,
-    input_options: &dyn InputOptionsExt<NewInputResult>,
-    queue_options: QueueInputOptions,
-) -> Result<NewInputResult, RegisterInputError> {
-    let pipeline_ctx = pipeline.lock().unwrap().ctx.clone();
```
wkozyra95, line 26: maybe add comment here that we intentionally do not hold lock here so we don't block the processing. I think I saw change like that in PR a few times already

compositor_pipeline/src/pipeline/pipeline_input.rs
```diff
@@ -17,12 +17,18 @@ pub struct PipelineInput {
...
 
+/// This method doesn't take pipeline lock for the whole scope,
+/// because input registration can potentially take a relatively long time.
 pub(super) fn register_pipeline_input<NewInputResult>(
     pipeline: &Arc<Mutex<Pipeline>>,
     input_id: InputId,
     input_options: &dyn InputOptionsExt<NewInputResult>,
     queue_options: QueueInputOptions,
 ) -> Result<NewInputResult, RegisterInputError> {
+    if pipeline.lock().unwrap().inputs.contains_key(&input_id) {
```
wkozyra95, line 28: you need to check it second time before adding input to the hash map

wkozyra95 (changes requested):

compositor_pipeline/src/pipeline/pipeline_input.rs
```diff
@@ -56,7 +65,10 @@ pub(super) fn register_pipeline_input<NewInputResult>(
         }
     }
 
-    guard.inputs.insert(input_id.clone(), pipeline_input);
+    match guard.inputs.entry(input_id.clone()) {
```
wkozyra95, line 68: if you do that, you are leaving partially modified state, you need to abort it before output stream end condition is modified

wkozyra95 (changes requested):

wkozyra95 (approved):

PR #699, opened 2024-08-27
Add MP4 `loop` option

WojciechBarczynski:


compositor_api/src/types/register_input.rs
```diff
@@ -36,6 +36,8 @@ pub struct Mp4Input {
     pub url: Option<String>,
     /// Path to the MP4 file.
     pub path: Option<String>,
+    /// (**default=`false`**) If input should be played in the loop.
+    pub looped: Option<bool>,
```
wkozyra95, line 40:
```suggestion
    pub loop: Option<bool>,
```
change also rest of the code, not only the public API

WojciechBarczynski: loop is keyword in rust. I can't use it directly here.

I can use `serde(rename)` to name it `loop` in the public HTTP API (I checked - docs generate correctly).

WojciechBarczynski: Using raw identifiers is also an option, but a very bad one in my opinion.

WojciechBarczynski: https://doc.rust-lang.org/reference/keywords.html#strict-keywords

WojciechBarczynski: PS It was exactly the reason I chose `looped` - I didn't want to use different names for this option for HTTP and Rust APIs

wkozyra95: In that case I would use `shouldLoop` for code and `loop` for http api. You should never trade the quality of public HTTP API in favor of internal implementation detail.

compositor_pipeline/src/pipeline/input/mp4.rs
```diff
@@ -19,16 +19,19 @@ pub mod mp4_file_reader;
...
 pub(crate) enum Mp4ReaderOptions {
     NonFragmented {
         file: PathBuf,
+        looped: bool,
     },
     #[allow(dead_code)]
     Fragmented {
         header: Bytes,
         fragment_receiver: Receiver<PipelineEvent<Bytes>>,
+        looped: bool,
```
wkozyra95, line 34: change other usages of looped too

wkozyra95 (approved):

PR #702, opened 2024-08-28
Make text component work on wasm

noituri:
Font has to be prefetched before it can be used. I was not able to load browser's fonts so every font has to be provided by the user.

### DIFF
--- a/compositor_render/src/state.rs
+++ b/compositor_render/src/state.rs
@@ -1,6 +1,8 @@
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
+use glyphon::fontdb;
+
 use crate::error::{RegisterRendererError, UnregisterRendererError};
 
 use crate::scene::{Component, OutputScene};
@@ -47,7 +49,7 @@ pub struct Renderer(Arc<Mutex<InnerRenderer>>);
 
 struct InnerRenderer {
     wgpu_ctx: Arc<WgpuCtx>,
-    text_renderer_ctx: TextRendererCtx,
+    text_renderer_ctx: Arc<TextRendererCtx>,
     chromium_context: Arc<ChromiumContext>,
 
     render_graph: RenderGraph,
@@ -156,6 +158,11 @@ impl Renderer {
         Ok(())
     }
 
+    pub fn register_font(&self, font_source: fontdb::Source) {
+        let ctx = self.0.lock().unwrap().text_renderer_ctx.clone();
+        ctx.add_font(font_source);
+    }
+
     pub fn render(&self, input: FrameSet<InputId>) -> Result<FrameSet<OutputId>, RenderSceneError> {
         self.0.lock().unwrap().render(input)
     }
@@ -185,7 +192,7 @@ impl InnerRenderer {
 
         Ok(Self {
             wgpu_ctx: wgpu_ctx.clone(),
-            text_renderer_ctx: TextRendererCtx::new(),
+            text_renderer_ctx: Arc::new(TextRendererCtx::new()),
             chromium_context: Arc::new(ChromiumContext::new(opts.web_renderer, opts.framerate)?),
             render_graph: RenderGraph::empty(),
             renderers: Renderers::new(wgpu_ctx)?,
--- a/compositor_render/src/transformations/text_renderer.rs
+++ b/compositor_render/src/transformations/text_renderer.rs
@@ -5,8 +5,8 @@ use std::{
 };
 
 use glyphon::{
-    AttrsOwned, Buffer, Color, FontSystem, Metrics, Shaping, SwashCache, TextArea, TextAtlas,
-    TextBounds,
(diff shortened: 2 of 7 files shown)


WojciechBarczynski:
I fought about solving the slow `FontSystem` init problem and I think we can allow to specify what fonts should be registered. It seems that [this function](https://docs.rs/cosmic-text/0.10.0/cosmic_text/struct.FontSystem.html#method.new) is the problem and there should be an option to use [this one](https://docs.rs/cosmic-text/0.10.0/cosmic_text/struct.FontSystem.html#method.new_with_fonts) instead.

My initial idea for API is:
- Rust API:
Add enum `Fonts::{System, Specified(Vec<Source>), SystemWithAdditional(Vec<Source>)}` and use it in `RendererOptions` passed to Renderer in the constructor.
- Public API:
Add env var `LIVE_COMPOSITOR_FONTS` that specifies what fonts should be registered, like:
`SYSTEM;[path_to_first_font];[path_to_second_font];...` (default=`system`).

Since `TextRendererCtx` is initialized on compositor init, this configuration is similar to other config options.
I wouldn't say I like using env var for that, but it's consistent with other config options.


That being said, I don't think it's beneficial to introduce 2 separate APIs for configuring the same stuff. This PR and the solution described above allow to register custom fonts, but changes from this PR don't solve the slow init problem. I'm unsure if it's worth merging this PR.

Do you have any thoughts or ideas?
@wkozyra95

wkozyra95:
> That being said, I don't think it's beneficial to introduce 2 separate APIs for configuring the same stuff. This PR and the solution described above allow to register custom fonts, but changes from this PR don't solve the slow init problem. I'm unsure if it's worth merging this PR.

We need an API to register fonts, it is a separate issue from slow start, but it's still needed.

> I fought about solving the slow FontSystem init problem and I think we can allow to specify what fonts should be registered. It seems that [this function](https://docs.rs/cosmic-text/0.10.0/cosmic_text/struct.FontSystem.html#method.new) is the problem and there should be an option to use [this one](https://docs.rs/cosmic-text/0.10.0/cosmic_text/struct.FontSystem.html#method.new_with_fonts) instead.

- I would consider adding a boolean option as an ENV to either load all, or do not load any. (not sure about it)
- I'm ok with leaving an option from Rust to provide those paths.
- It's not a great API for regular users because they need to know where the fonts are.
- In most cases, you don't want to rely on what is installed in the filesystem, you are more likely to prefer providing your own fonts via register API than installing them in the filesystem.

compositor_web/src/wasm.rs
```diff
@@ -117,4 +120,16 @@ impl LiveCompositorRenderer {
...
+    pub async fn register_font(&mut self, font_url: String) -> Result<(), JsValue> {
+        let text_renderer_ctx = self.renderer.text_renderer_ctx();
+        let bytes = download(&font_url).await?;
+        text_renderer_ctx.add_font(Source::Binary(Arc::new(bytes.to_vec())));
+        Ok(())
+    }
+}
+
+async fn download(url: &str) -> Result<Bytes, JsValue> {
+    let resp = reqwest::get(url).await.map_err(to_js_error)?;
```
wkozyra95, line 133: what about error handling, if you get 404 you will try to parse some error message as font

compositor_render/src/state.rs
```diff
@@ -177,6 +177,11 @@ impl Renderer {
         let guard = self.0.lock().unwrap();
         (guard.wgpu_ctx.device.clone(), guard.wgpu_ctx.queue.clone())
     }
+
+    pub fn text_renderer_ctx(&self) -> Arc<TextRendererCtx> {
```
wkozyra95, line 181: add register font on renderer directlly

compositor_render/src/transformations/text_renderer.rs
```diff
@@ -215,6 +215,11 @@ impl TextRendererCtx {
             swash_cache: Mutex::new(SwashCache::new()),
         }
     }
+
+    pub fn add_font(&self, source: Source) {
+        let mut font_system = self.font_system.lock().unwrap();
+        font_system.db_mut().load_font_source(source);
```
wkozyra95, line 221: what about errors?

noituri: `load_font_source` doesn't return Result. In case of invalid font data it logs a warn to console.

compositor_render/src/state.rs
```diff
@@ -156,6 +158,11 @@ impl Renderer {
         Ok(())
     }
 
+    pub fn register_font(&self, font_source: fontdb::Source) {
+        let guard = self.0.lock().unwrap();
```
wkozyra95, line 162: make sure not to hold lock for the entire time

wkozyra95 (approved):

WojciechBarczynski:
LGTM

PR #703, opened 2024-08-29
[ts-sdk] Support WebSocket events

wkozyra95:


ts/@live-compositor/node/src/ws.ts
```diff
@@ -0,0 +1,60 @@
...
+        const event = parseEvent(data);
+        if (event) {
+          for (const listener of this.listeners) {
+            listener(event);
+          }
+        }
+      });
+
+      ws.on('close', () => {
+        this.ws = null;
```
wkozyra95, line 42: we might want to add reconnect in the future

ts/live-compositor/src/hooks/store.ts
```diff
@@ -1,9 +1,7 @@
 export type InputStreamInfo = {
   inputId: string;
-  // TODO: add input stream state
-  // state: 'registered' | 'playing' | 'finished';
-
-  // TODO: add input type (maybe other options too)
+  videoState?: 'ready' | 'playing' | 'finished';
```
wkozyra95, line 3: - not sure about naming
- current API does not allow to distinguish if we have audio and video tracks when registering, (we could sent it as part of the register request response)

WojciechBarczynski (approved):

PR #704, opened 2024-08-29
[ts-sdk] Assign ComponentId automatically with `useId` hook

wkozyra95:


WojciechBarczynski (approved):
I'm guessing that this makes transitions easier, as a user doesn't have to provide IDs, but some description of the purpose of those changes would be nice.

wkozyra95:
> I'm guessing that this makes transitions easier, as a user doesn't have to provide IDs, but some description of the purpose of those changes would be nice.

The purpose of those changes is to assign a default id for all Components, so user does not have to think about that vale, which is described here. The fact that transition will be easier is one of the effects. In the future we might use that for e.g. for partial updates or sth else that requires us to identify components

PR #708, opened 2024-08-29
Add a way to configure connecting to local compositor instance in development and to production when deployed

wkazmierczak:
Default connection set to production, when you want to connect with dev use command `npm run dev`


docs/package.json
```diff
@@ -3,17 +3,18 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "docusaurus": "docusaurus",
```
wkozyra95, line 6: don't use NODE_ENV for that, do not modify other scrips, only `dev` script should be added, and old ones shoud stay as they were

docs/src/pages/playground.tsx
```diff
@@ -97,6 +97,9 @@ function Homepage() {
     input_5: InputResolution.Resoultion1920x1080,
     input_6: InputResolution.Resoultion1920x1080,
   });
+  const { siteConfig } = useDocusaurusContext();
+  const environment = siteConfig.customFields.environment as string;
```
wkozyra95, line 101: I don't like passing that through the components, this should be a global value. Is there a way to access that field without a hook?

wkazmierczak: As far as I know and have tested, hook is needed, but I'll double-check.

docs/src/api.ts
```diff
@@ -1,4 +1,7 @@
-const BACKEND_URL: URL = new URL('http://localhost:8081');
+// The @generated/docusaurus.config file is built at runtime, not present during linting.
+// Disabling the rule prevents linting errors for this valid import.
```
wkozyra95, line 2:
```suggestion
```

wkozyra95 (approved):

PR #712, opened 2024-08-29
[docs] Add MP4 loop version docs info

WojciechBarczynski:
Screenshot 2024-08-30 at 11 55 48

### DIFF
--- a/compositor_api/src/types/register_input.rs
+++ b/compositor_api/src/types/register_input.rs
@@ -36,8 +36,7 @@ pub struct Mp4Input {
     pub url: Option<String>,
     /// Path to the MP4 file.
     pub path: Option<String>,
-    /// (**default=`false`**) If input should be played in the loop.
-    /// Added in v0.4.0.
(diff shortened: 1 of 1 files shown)


compositor_api/src/types/register_input.rs
```diff
@@ -37,6 +37,7 @@ pub struct Mp4Input {
     /// Path to the MP4 file.
     pub path: Option<String>,
     /// (**default=`false`**) If input should be played in the loop.
+    /// Added in v0.4.0.
```
wkozyra95, line 40:
```suggestion
    /// (**default=`false`**) If input should be played in the loop. <span class="badge badge--primary">Added in v0.4.0</span> 
```

wkozyra95 (commented):
btw, especially for small PRs like this it's useful to include a screenshot

compositor_api/src/types/register_input.rs
wkozyra95: 20240830_10h10m14s_grim

wkozyra95 (approved):

PR #715, opened 2024-08-30
[ts-sdk] Add audio support

wkozyra95:
Public API
- Add props to the InputStream
- Add hook `useInputAudio`
- Modify hook `useInputStreams` to return map instead of a list
- Add unregister output
- If multiple Input streams will present in scene then audio will be summed

Internal
- Refactor compositor context. Now we have one global store for entire compositor instance and per-output context that is specific to single output. (Output is not a store in the react sense)
- I need to expose some types between `@live-compositor/core` and `live-compositor` so I exported _liveCompositorInternals object that groups everything. Those types are not intended to be used by user
- I was not able to figure out a better way to shutdown an output, so I added a wrapper component that is waiting for exteranal store with just one boolean value (shutdown). If shutdown is set we change a scene to just a `<View />`


ts/live-compositor/src/components/InputStream.ts
```diff
@@ -12,9 +14,28 @@ export type InputStreamProps = {
...
    */
   inputId: Api.InputId;
+  /**
+   * Audio volume represented by a number between 0 and 1.
+   */
+  volume?: number;
+  /**
+   * Mute audio.
+   */
+  mute?: boolean;
```
WojciechBarczynski, line 24: Why do you add `mute`? Can't they just set `volume=0` if somebody wants to mute audio? Having 2 separate fields for that doesn't make sense IMO. It also introduces ambiguity when both values are specified.

wkozyra95: > Why do you add mute? 

I think it's a useful API, especially in React you can do `<InputStream mute/>`. I also considered using `muted` instead (I think mute is better, but muted is the same as in `<video/>` HTML tag), but I don't have strong opinions on that.

> Can't they just set volume=0

Yes, as you can see below it's literally what it does internally

> Having 2 separate fields for that doesn't make sense IMO.

It's often true, but I disagree with it as a general statement. This is a tradeoff, we are getting simpler API to do sth specific, but increase the complexity because user needs to know how those 2 ways interact. You can argue if you want that tradeoff is not worth it here, but not in general.

> It also introduces ambiguity when both values are specified.

Does it? Can you imagine someone seeing those 2 options and assuming that volume has higher priority than mute?

ts/@live-compositor/core/src/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+    running = true;
+    void start();
+  };
+}
+
+export async function sleep(timeout_ms: number): Promise<void> {
+  await new Promise<void>(res => {
+    setTimeout(() => {
+      res();
+    }, timeout_ms);
```
noituri, line 38:
```suggestion
export async function sleep(timeoutMs: number): Promise<void> {
  await new Promise<void>(res => {
    setTimeout(() => {
      res();
    }, timeoutMs);
```

noituri (approved):

WojciechBarczynski (commented):
LGTM

PR #723, opened 2024-09-03
[wasm] Return `FrameSet` on render 

noituri:
Currently wasm's `render` returns a map `OutputId => Bytes`. But, `render` method from `compositor_render` returns `FrameSet` which contains more additional information. In this PR I changed the wasm implimentation to also return `FrameSet`


wkozyra95 (commented):
Can you add some PR description?

wkozyra95 (approved):

PR #724, opened 2024-09-03
[wasm] NPM package for wasm compositor

noituri:
This PR adds npm package that wraps the wasm API. The install process is not completely automatic and requires some extra steps from the user:

1. npm install the package
2. Bundle wasm file from `node_modules/@live-compositor/browser-render/dist/live-compositor.wasm`  (i.e. move it to `public` directory of your webapp) 
3. Import `initCompositor` from `@live-compositor/browser-render` and call it with url to the wasm file

I've included an example web app which uses Vite + React. It served me as a playground for testing changes. If needed I will rewrite the example in the followup PR.


ts/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "lint": "lerna run lint",
     "lint:fix": "lerna run lint -- --fix",
-    "build": "lerna run build",
+    "build": "lerna run build --concurrency=1",
```
noituri, line 8: https://github.com/lerna/lerna/issues/1948#issuecomment-468354267

ts/@live-compositor/browser-render/scripts/buildWasm.ts
```diff
@@ -0,0 +1,12 @@
+import { spawn } from 'child_process';
+import * as path from 'path';
+
+function build() {
+  const compositorWebPath = path.resolve(__dirname, '../../../../compositor_web');
+  const outputPath = path.resolve(__dirname, '../src/generated');
+  const args = ['build', '--target', 'web', '--release', '-d', outputPath, compositorWebPath];
+
+  spawn('wasm-pack', args, { stdio: 'inherit' });
```
wkozyra95, line 9: either make function async and await here, or at least return the promise from spawn function

ts/@live-compositor/browser-render/scripts/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../../tsconfig.base.json",
+  "compilerOptions": {
+    "types": ["node"],
```
wkozyra95, line 4: node? this should run in browser

ts/@live-compositor/browser-render/src/api.ts
```diff
@@ -0,0 +1,11 @@
+import * as api from './api.generated';
```
wkozyra95, line 1:
```suggestion
import * as Api from './api.generated';
```

for consistency with other packages

btw, maybe import it from `live-compositor` directlly instead of copying it here

ts/@live-compositor/browser-render/src/wasm.ts
```diff
@@ -0,0 +1,11 @@
+import init, * as wasm from './generated/compositor_web';
+
+/**
+ * Loads and initializes all resources required for the live compositor to work.
+ * @param wasmFileUrl {string} - An URL for `live-compositor.wasm` file. The file is located in `dist` folder.
+ */
+export async function initCompositor(wasmFileUrl: string) {
```
wkozyra95, line 7: initRenderer or initLiveCompositorRenderer

ts/@live-compositor/browser-render/src/renderer.ts
```diff
@@ -0,0 +1,81 @@
...
+};
+
+export type Framerate = {
+  num: number;
+  den: number;
+};
+
+export type FrameSet<T> = {
+  ptsMs: number;
+  frames: Map<T, Frame>;
```
wkozyra95, line 18: don't use map, use regular object

ts/@live-compositor/browser-render/scripts/tsconfig.json
noituri: This is needed for `buildWasm.ts` script. It uses types from node

wkozyra95: So maybe rename `buildWasm.ts` -> `buildWasm.js`

wkozyra95 (approved):

WojciechBarczynski (commented):
LGTM

PR #726, opened 2024-09-04
[playground] Improve experience when the server is not accessible or under heavy load

wkazmierczak:
- [x] Attempt to send request multiple times.
- [x] Set timeout for fetch.


docs/src/api.ts
```diff
@@ -53,21 +53,64 @@ async function createError(response: Response): Promise<ApiError> {
...
+async function fetchWithTimeout(
+  path: string,
+  requestObject: RequestObject,
+  timeout: number
+): Promise<Response> {
+  const backendUrl: URL = new URL(
     config.customFields.environment === 'development'
       ? 'http://localhost:8081'
       : 'https://playground.compositor.live'
   );
```
wkozyra95, line 68:
```suggestion
  const backendUrl: string = config.customFields.environment === 'development'
      ? 'http://localhost:8081'
      : 'https://playground.compositor.live';
```

you could also move this value to the global const on top of the file

docs/src/api.ts
```diff
@@ -53,21 +53,64 @@ async function createError(response: Response): Promise<ApiError> {
...
+      })
+      .catch(error => {
+        clearTimeout(timer);
+        rej(new ApiError(error.message));
+      });
+  });
+}
+
+async function retryRequest(func, path, requestObject) {
+  const delay = 1000;
```
wkozyra95, line 93:
```suggestion
  const DELAY_MS = 1000;
```
use upper case for const, if representing time value add unit to the name
the same applies to values bellow

docs/src/api.ts
```diff
@@ -53,21 +58,59 @@ async function createError(response: Response): Promise<ApiError> {
...
+        res(response);
+      })
+      .catch(error => {
+        clearTimeout(timer);
+        rej(new ApiError(error.message));
+      });
+  });
+}
+
+async function retryRequest(path, requestObject) {
```
wkazmierczak, line 92: I wasn't sure what exactly should remain in `retryRequest` function, so for now I did it like this

docs/src/api.ts
```diff
@@ -52,22 +57,66 @@ async function createError(response: Response): Promise<ApiError> {
...
-      : 'https://playground.compositor.live'
-  );
-  const renderImageUrl = new URL('/render_image', backend_url);
-  let response;
-  try {
-    response = await fetch(renderImageUrl, requestObject);
-  } catch (error) {
-    throw new ApiError(error.message);
-  }
+  const response = await retryWithSleep(fn);
```
wkozyra95, line 64:
```suggestion
  const response = await retryWithTimeout(async () => (
    await fetchWithTimeout('/render_image', requestObject, TIMEOUT_MS)
  ), TIMEOUT_MS);
```

docs/src/api.ts
```diff
@@ -52,22 +57,66 @@ async function createError(response: Response): Promise<ApiError> {
...
+        res(response);
+      })
+      .catch(error => {
+        clearTimeout(timer);
+        rej(new ApiError(error.message));
+      });
+  });
+}
+
+async function retryWithSleep<T>(fn: () => Promise<T>): Promise<T> {
```
wkozyra95, line 99:
```suggestion
async function retryWithSleep<T>(fn: () => Promise<T>, timeoutMs: number): Promise<T> {
```

PR #729, opened 2024-09-05
Add corners rounding support to layout shader

WojciechBarczynski:
I tested it and it seems to be working ok.

Screenshot 2024-09-06 at 12 39 50


compositor_render/src/wgpu/format.rs
```diff
@@ -16,44 +16,20 @@ pub struct TextureFormat {
...
     rgba_to_yuv: RgbaToYuvConverter,
-
-    planar_yuv_layout: wgpu::BindGroupLayout,
-    interleaved_yuv_layout: wgpu::BindGroupLayout,
-    rgba_layout: wgpu::BindGroupLayout,
 }
 
 impl TextureFormat {
     pub fn new(device: &wgpu::Device) -> Self {
-        let planar_yuv_layout = PlanarYuvTextures::new_bind_group_layout(device);
```
WojciechBarczynski, line 27: I modified it for consistency

WojciechBarczynski:
It's ready for wgpu 0.19, but since @jerzywilczek updated wgpu to the newer version, that has a buffer binding size check, I'll need to change how buffers are passed to layout shader. I'll try to do it using storage buffers.

PR #742, opened 2024-09-06
[playground] Fix layout issues + add beter default scene

wkozyra95:
- Switch top level playground styles to tailwind. Technically nothing here was necessary to do with tailwind, but I would like to start slowly switching to it in the rest of the playground and new code.
- Add default scene that shows everything and has a layout like sth that look "real"
- improve some styling for light mode

20240906_19h58m49s_grim


docs/src/pages/playground.tsx
```diff
@@ -151,21 +102,19 @@ function Homepage() {
...
-    <div className={styles.page}>
-      <div className={styles.leftSide}>
-        <div className={styles.codeEditorBox}>
-          {showReactEditor ? (
-            <PlaygroundReactEditor code={code} onCodeChange={setCode} />
-          ) : (
-            <PlaygroundCodeEditor onChange={setScene} initialCodeEditorContent={INITIAL_SCENE} />
-          )}
-        </div>
+    <div className="flex flex-row flex-wrap p-8 lg:h-[calc(100vh-110px)]">
```
wkozyra95, line 105: unfortunately calculating size of a footer and nav bar is the easiest way, otherwise components deep inside the tree can easily force their size.

WojciechBarczynski (approved):

PR #743, opened 2024-09-09
[wasm] Fix bundler config

wkozyra95:
- In the dist directory index.ts is generated for just index.d.ts, the section that should generate d.ts did not work. Type files were showing up only because `declaration: true`
- Node 20 is required to build wasm
- watch command should build with rollup not tsc
- Add `build-sdk` command to avoid building wasm when it is not needed

### DIFF
--- a/build_tools/nix/flake.nix
+++ b/build_tools/nix/flake.nix
@@ -52,13 +52,14 @@
             gst_all_1.gst-plugins-ugly
             gst_all_1.gst-libav
 
-            nodejs_18
+            nodejs_20
             rustfmt
             clippy
             cargo-watch
             cargo-nextest
             rust-analyzer
             clang-tools
+            llvmPackages.bintools
           ];
         in
         {
--- a/ts/@live-compositor/browser-render/rollup.config.mjs
+++ b/ts/@live-compositor/browser-render/rollup.config.mjs
@@ -7,24 +7,27 @@ export default [
     input: 'src/index.ts',
     output: {
       file: 'dist/index.js',
-      format: 'es'
+      format: 'es',
     },
     plugins: [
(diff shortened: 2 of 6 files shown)


build_tools/nix/flake.nix
```diff
@@ -52,13 +52,14 @@
...
 
-            nodejs_18
+            nodejs_20
             rustfmt
             clippy
             cargo-watch
             cargo-nextest
             rust-analyzer
             clang-tools
+            llvmPackages.bintools
```
wkozyra95, line 62: adds lld to the PATH

ts/@live-compositor/browser-render/scripts/buildWasm.mjs
```diff
@@ -1,5 +1,5 @@
-import * as path from 'path';
-import { spawn as nodeSpawn } from 'child_process';
+import path from 'node:path';
```
wkozyra95, line 1: ESM style imports

ts/@live-compositor/browser-render/rollup.config.mjs
```diff
@@ -7,24 +7,27 @@ export default [
...
     output: {
       file: 'dist/index.js',
-      format: 'es'
+      format: 'es',
     },
     plugins: [
       typescript(),
       copy({
         targets: [
-          { src: 'src/generated', dest: 'dist' },
```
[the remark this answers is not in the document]
noituri: Previously I was importing the wasm file directly in the code and the typings for the wasm file were not bundled automatically so I had to manually copy them. Since I've changed the way wasm is imported in the package, this copy is no longer needed.

noituri (approved):

PR #748, opened 2024-09-09
Support input NV12 textures

### LINKED ISSUE #750: Add NV12 format conversion to prepare the compositor for ingesting output from the vulkan decoder

jerzywilczek:
Closes #750

### DIFF
--- a/compositor_render/src/wgpu/texture.rs
+++ b/compositor_render/src/wgpu/texture.rs
@@ -14,6 +14,7 @@ use super::WgpuCtx;
 mod base;
 mod bgra;
 mod interleaved_yuv422;
+mod nv12;
 mod planar_yuv;
 mod rgba;
 pub mod utils;
@@ -23,6 +24,7 @@ pub type RGBATexture = rgba::RGBATexture;
 pub type PlanarYuvTextures = planar_yuv::PlanarYuvTextures;
 pub type PlanarYuvVariant = planar_yuv::YuvVariant;
 pub type InterleavedYuv422Texture = interleaved_yuv422::InterleavedYuv422Texture;
+pub type NV12TextureView<'a> = nv12::NV12TextureView<'a>;
 
 pub type Texture = base::Texture;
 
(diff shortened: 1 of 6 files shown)


compositor_render/src/wgpu/texture.rs
```diff
@@ -182,6 +189,15 @@ impl InputTexture {
                             error!("Invalid texture passed as an input: {err}")
                         }
                     }
+                    InputTextureState::Nv12WgpuTexture(texture) => {
+                        let texture = NV12TextureView::from_wgpu_texture(texture.as_ref());
+                        let bind_group = texture.new_bind_group(ctx, ctx.format.nv12_layout());
```
wkozyra95, line 194: what will happen if the passed texture is not nv12?

e.g. for rgba texture with do some validation inside fill_from_wgpu_texture

wkozyra95 (approved):
LGTM, but make sure it does not panic when passed texture is invalid

compositor_render/src/wgpu/texture/nv12.rs
```diff
@@ -0,0 +1,83 @@
...
+                    binding: 0,
+                    resource: wgpu::BindingResource::TextureView(&self.view_y),
+                },
+                wgpu::BindGroupEntry {
+                    binding: 1,
+                    resource: wgpu::BindingResource::TextureView(&self.view_uv),
+                },
+            ],
+        })
+    }
```
WojciechBarczynski, line 82: This can be created just once

WojciechBarczynski (approved):

compositor_render/src/wgpu/texture/nv12.rs
jerzywilczek: and it is indeed created just once. all texture types have a similar function and all of them are created just once

compositor_render/src/wgpu/texture.rs
jerzywilczek: ok, handled that

PR #756, opened 2024-09-11
[ts-sdk] Move node-example

wkozyra95:
non node examples will have to be in a separate project, so I'm moving node examples one level down


ts/@live-compositor/browser-render/package-lock.json
```diff
@@ -1,317 +0,0 @@
-{
```
wkozyra95, line 1: In workspace setup there is only one lock file in the root that have all the dependencies

noituri (approved):

PR #767, opened 2024-09-11
[tests] Fix integration_tests docker example

wkozyra95:
regression from examples refactor


WojciechBarczynski (approved):

WojciechBarczynski:
Also, consider prefixing the PR name with sth like `[tests]`

wkozyra95:
> Also, consider prefixing the PR name with sth like `[tests]`

Fine, but this is a bit borderline. The purpose of `[]` tags is to mark PRs that have some important/distinctive grouping. Tests are usually done together with the code changes, so it's rare to include PRs that just have test changes so the usefulness of this tag is a bit questionable

PR #769, opened 2024-09-12
Add eslint rule `no-floating-promises` in docs and demos dirs

wkazmierczak:


demos/utils/run.ts
```diff
@@ -20,7 +20,7 @@ export async function runCompositorExample(
 
     await fn();
   } catch (err) {
-    logError(err);
+    void logError(err);
```
wkozyra95, line 23:
```suggestion
    await logError(err);
```

demos/utils/gst.ts
```diff
@@ -31,7 +31,7 @@ export function gstStreamWebcam(ip: string, port: number, displayOutput: boolean
 
 function checkGstPlugins(plugins: string[]) {
   plugins.forEach(plugin => {
```
wkozyra95, line 33: this function is incorrect, exception will not be thrown in checkGstPlugins
convert it to regular for loop, or `await Promise.all(plugins.map(plugin => ...))`

wkozyra95 (approved):

PR #773, opened 2024-09-17
Add logger to axum router

wkazmierczak:


src/routes.rs
```diff
@@ -111,3 +117,42 @@ where
         }
     }
 }
+
+async fn log_request_response_body(
```
wkozyra95, line 121:
```suggestion
async fn body_logger_middleware(
```
plus move everything to separate file

src/routes.rs
```diff
@@ -111,3 +117,42 @@ where
...
+    next: Next,
+) -> Result<impl IntoResponse, Response> {
+    let request = buffer_request_body(request).await?;
+    let response = next.run(request).await;
+    let response = buffer_response_body(response).await?;
+
+    Ok(response)
+}
+
+async fn buffer_request_body(request: Request) -> Result<Request, Response> {
```
wkozyra95, line 132: this should only happen if sth specific is set e.g. logger option to trace

in our logger settings we can write `live_compositor::routes=debug` to enable debugging in that one file, check if is there a way to provide a custom values that can enable specific log.

If above is not possible then lets enable that only when trace loggin is on

wkozyra95 (changes requested):

src/middleware.rs
```diff
@@ -0,0 +1,53 @@
...
+    let response = next.run(request).await;
+    let response = buffer_response_body(response).await?;
+
+    Ok(response)
+}
+
+async fn buffer_request_body(request: Request) -> Result<Request, Response> {
+    let (parts, body) = request.into_parts();
+
+    let bytes = body
```
wkozyra95, line 26: if logging is enabled then do not read/restore that body

src/middleware.rs
```diff
@@ -0,0 +1,53 @@
...
+        .collect()
+        .await
+        .map_err(|err| (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()).into_response())?
+        .to_bytes();
+
+    if log_enabled!(target: "live_compositor::routes", Level::Debug) {
```
wkozyra95, line 32: there is already file route that use sth that does not already exist e.g. `live_compositor::log_request_body`

src/middleware.rs
```diff
@@ -0,0 +1,52 @@
...
+async fn buffer_request_body(request: Request) -> Result<Request, Response> {
+    let (parts, body) = request.into_parts();
+
+    let bytes = body
+        .collect()
+        .await
+        .map_err(|err| (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()).into_response())?
+        .to_bytes();
+
+    trace!(target: "live_compositor::log_request_body", "Request body: {:?}", str::from_utf8(&bytes).unwrap());
```
wkozyra95, line 35: - print other parts of the response too, like route or method, for response status code
- do not unwrap

Also, it would be great to try parsing requests as JSON values to remove the escaped characters

wkozyra95 (approved):

PR #778, opened 2024-09-20
Add sample input videos

brzep:
Created script for generating sample videos with input name and timer in multiple resolutions and variants.
Added sample videos to static dir.


wkozyra95 (commented):
Given that this is not a blocker let's wait for fix to encoder flushing and regenerate it without tune zerolatency option

PR #782, opened 2024-09-22
Make encoder preset optional

WojciechBarczynski:
I don't have a strong option, but I noticed that the encoder field documentation was inconsistent with the implementation. We can make this field required and change docs instead. It's your call.


wkozyra95 (approved):

wkozyra95:
CI is not passing because you changed the type in handwritten types, but did not regenerate types from rust `npm run generate-types` in `ts` directory

PR #783, opened 2024-09-23
[docs] Rework getting started section + Add TypeScript

wkozyra95:
- Add React version of guides. It is not perfect there are still some places in those guides when we refer to input_id but it should be `inputId`, but reworking that would take to much time.
- Rework intro section to start promoting typescript as the primary way of using a compositor
- Add React code example to the main page

## API changes

- changed `cubic-bezier` back to `cubic_bezier`. We don't rename other values (we rename only field names) so for consistency lets stay with underscore

### DIFF
--- a/docs/pages/guides/quick-start.md
+++ b/docs/pages/guides/quick-start.md
@@ -12,6 +12,16 @@ This guide will explain basic LiveCompositor setup.
 ### Start the compositor
 
 <Tabs queryString="lang">
+  <TabItem value="react" label="React">
+    ```tsx
+    import LiveCompositor from "@live-compositor/node"
+
+    async function start() {
+        const compositor = new LiveCompositor();
+        await compositor.init();
+    }
+    ```
+  </TabItem>
   <TabItem value="http" label="HTTP">
     Start the compositor server. Check out [configuration page](../deployment/configuration.md) for available configuration options.
   </TabItem>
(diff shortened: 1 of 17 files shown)


docs/pages/guides/quick-start.md
```diff
@@ -43,6 +53,21 @@ This guide will explain basic LiveCompositor setup.
...
+    ```tsx
+    await compositor.registerInput("input_1", {
+      type: "rtp_stream",
+      transportProtocol: "tcp_server",
+      port: 9001,
+      video: {
+        decoder: "ffmpeg_h264"
+      }
+    })
+    ```
```
WojciechBarczynski, line 66: I think it would be great for react guide, if we could have full working code expendable (with all imports / previously defined code etc.), such that users could copy-paste it.

WojciechBarczynski (approved):

noituri (approved):

docs/src/components/ExampleSceneJsx.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <Shader
+        shaderId="replace_green_screen"
+        resolution={{ width: 1920, height: 1080 }}>
+        <InputStream inputId="tv" />
+        <Image imageId="background" />
+      </Shader>
+      <Rescaler top={20} left={20} width={640} height={360}>
+        <InputStream inputId="bunny" />
+      </Rescaler>
+      <View bottom={0} left={0} height={120}>
```
WojciechBarczynski, line 24: backgorund color

PR #787, opened 2024-09-24
Terminate compositor when parent process is terminated

wkozyra95:


src/server.rs
```diff
@@ -12,6 +12,7 @@ use crate::{config::read_config, logger::init_logger, routes::routes, state::Api
 pub fn run() {
     let config = read_config();
     init_logger(config.logger.clone());
+    listen_for_parent_termination();
```
WojciechBarczynski, line 15: Since `listen_for_parent_termination` depends on correctly reading `parent_id` (parent process is not killed before `let ppid = parent_id();` executes), I would put it as a first call in the `run` command.

WojciechBarczynski (approved):

PR #796, opened 2024-09-27
Add option to skip system fonts initialization

wkozyra95:
font initialization time:
- (before) for me it was 60ms but it depends on the number of fonts that are installed (also I run that over and over so it was cached in memory most likely)
- (after) sometimes 1ms sometimes few microseconds (probably depends on some system cache)
- For context, wgpu context init takes about 200-250 ms which is now almost 100% of the startup time(not tested how long it takes to start http server)


compositor_pipeline/src/pipeline.rs
```diff
@@ -158,6 +160,10 @@ impl Pipeline {
         Ok((pipeline, event_loop))
     }
 
+    pub fn renderer_mut(&mut self) -> &mut Renderer {
```
wkozyra95, line 163: exposing renderer so user can register fonts

WojciechBarczynski (approved):
Have you measured initialization speed improvement? What are the results? Let us add such info in the PR description in this type of PR.

PR #802, opened 2024-10-01
Add ESM + CJS support for `live-compositor` and `@live-compositor/core`

wkozyra95:
For node we usually want to support CJS modules, but for projects like vite ESM seems to be needed. This PR builds `live-compositor` and `@live-compositor/core` packages to both esm and cjs


ts/@live-compositor/browser-render/package.json
```diff
@@ -23,14 +23,5 @@
   },
   "peerDependencies": {
     "live-compositor": "^0.1.0"
-  },
-  "nx": {
-    "targets": {
```
wkozyra95, line 28: I run `npm run build` often in root and this was very anoying

ts/.eslintrc.base.json
```diff
@@ -16,6 +17,17 @@
       "tsconfig.json"
     ]
   },
+  "settings": {
+    "import/parsers": {
```
wkozyra95, line 21: By default `eslint-plugin-import` can't detect correctly `exports` section to detect esm, this seems to resolve the issue

noituri (approved):
🫡

WojciechBarczynski (approved):

PR #820, opened 2024-10-11
[vk-video] Process modification of short-term reference picture list

### LINKED ISSUE #804: Add support for `ref_pic_list_modification_l0` field to the hardware decoder
Reference picture list modification should be relatively simple, since it should be just copying the algorithm from The Book

jerzywilczek:
This implements sections 8.2.4.3.1 from the spec.

Closes #804

### DIFF
--- a/vk-video/src/parser/reference_manager.rs
+++ b/vk-video/src/parser/reference_manager.rs
@@ -2,7 +2,10 @@ use std::sync::Arc;
 
 use h264_reader::nal::{
     pps::PicParameterSet,
-    slice::{DecRefPicMarking, NumRefIdxActive, RefPicListModifications, SliceHeader},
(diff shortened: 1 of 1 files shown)


WojciechBarczynski (approved):

vk-video/src/parser/reference_manager.rs
```diff
@@ -36,6 +39,9 @@ pub enum ReferenceManagementError {
 
     #[error("pic_order_cnt_lsb is not present in a slice header, but is required for decoding")]
```
jerzywilczek, line 40: remove this error since you added a generic 'bad bytestrram' one dummy.

also remove the one about not supporting ref list modification.

noituri (approved):

PR #827, opened 2024-10-16
[vk-video] Add support for long-time references

### LINKED ISSUE #805: Add support for long-term reference pictures to the hardware decoder
This requires adding support for both handling the `long_term_reference` flag and adaptive memory control through the `memory_management_control_operation` field.

jerzywilczek:
This PR implements mainly section 8.2.5.4 from the spec.

Closes #805.

### DIFF
--- a/vk-video/src/parser.rs
+++ b/vk-video/src/parser.rs
@@ -32,8 +32,10 @@ pub struct DecodeInformation {
 }
 
 #[derive(Debug, Clone)]
+#[allow(non_snake_case)]
 pub(crate) struct ReferencePictureInfo {
     pub(crate) id: ReferenceId,
+    pub(crate) LongTermPicNum: Option<u64>,
     pub(crate) picture_info: PictureInfo,
 }
 
--- a/vk-video/src/parser/reference_manager.rs
+++ b/vk-video/src/parser/reference_manager.rs
@@ -3,15 +3,14 @@ use std::sync::Arc;
 use h264_reader::nal::{
     pps::PicParameterSet,
     slice::{
-        DecRefPicMarking, ModificationOfPicNums, NumRefIdxActive, RefPicListModifications,
-        SliceHeader,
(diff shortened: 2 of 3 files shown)


vk-video/src/parser.rs
```diff
@@ -32,8 +32,10 @@ pub struct DecodeInformation {
 }
 
 #[derive(Debug, Clone)]
+#[allow(non_snake_case)]
 pub(crate) struct ReferencePictureInfo {
     pub(crate) id: ReferenceId,
+    pub(crate) LongTermPicNum: Option<u64>,
```
noituri, line 38: I assume the spec uses this case but I think we should stick to rust conventions

vk-video/src/parser/reference_manager.rs
```diff
@@ -101,92 +128,297 @@ impl ReferenceContext {
...
+        for (i, picture) in self.pictures.short_term.iter().enumerate() {
+            if decode_picture_numbers_for_short_term_ref(
+                picture.header.frame_num.into(),
+                current_frame_num,
+                sps,
+            )
+            .PicNum
+                == pic_num_to_remove
+            {
+                return Ok(self.pictures.short_term.remove(i));
```
noituri, line 195: Extracting `.PicNum` to a new variable would make the condition nicer to read

vk-video/src/parser.rs
jerzywilczek: i think not. The big problem with the case is that there are some variables in the spec which are the same and only differ in the case. To be able to tell which one I'm writing about I decided to stick to using exact names from the spec. I know it's ugly, but it's the only way for us to be able to tell which spec variables we're talking about in the code

jerzywilczek: e.g. there's a variable called `frame_num`, which is a syntax element of slice headers. Later the spec defines the variable `FrameNum`, which is calculated in some way and is different from `frame_num`

jerzywilczek: the spec uses both those names, sometimes in the same section

noituri (approved):

PR #833, opened 2024-10-18
Forbid output update after EOS

### LINKED ISSUE #552: Scene update after EOS should be ignored
If a scene update is sent after the output sends EOS then the renderer is updated and starts producing frames for that output.
As a result "Received new frame from renderer after EOS" message is printed for each frame.

Expected behavior: Scene updates after EOS should be silently ignored

noituri:
Fixes #552

### DIFF
--- a/compositor_pipeline/src/pipeline.rs
+++ b/compositor_pipeline/src/pipeline.rs
@@ -378,6 +378,10 @@ impl Pipeline {
             .outputs
             .get(&output_id)
             .ok_or_else(|| UpdateSceneError::OutputNotRegistered(output_id.clone()))?;
+
+        if output.output.is_video_finished() {
+            return Err(UpdateSceneError::UpdateAfterEOS(output_id));
+        }
         let (Some(resolution), Some(frame_format)) = (
             output.output.resolution(),
             output.output.output_frame_format(),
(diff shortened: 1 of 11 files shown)


compositor_pipeline/src/pipeline.rs
```diff
@@ -378,6 +378,13 @@ impl Pipeline {
             .outputs
             .get(&output_id)
             .ok_or_else(|| UpdateSceneError::OutputNotRegistered(output_id.clone()))?;
+
+        if let Some(cond) = &output.video_end_condition {
+            if cond.did_output_end() {
+                return Err(UpdateSceneError::UpdateAfterEOS(output_id.clone()));
```
wkozyra95, line 384: What if the audio was already finished, but the video did not? I don't have a strong opinion, but I think I would not throw an error here and at most log a warning here. (the same for audio)

wkozyra95 (approved):

compositor_pipeline/src/pipeline.rs
noituri: Since I also don't have a strong opinion on this, I think I'll ignore all updates after EOS and log a warning instead.

integration_tests/src/tests/offline_processing.rs
```diff
@@ -123,7 +123,7 @@ pub fn offline_processing() -> Result<()> {
     if !(1.9..=2.1).contains(&duration) {
         return Err(anyhow!("Invalid duration: {}", duration));
     }
-    if !(930_000..=1_000_000).contains(&bit_rate) {
+    if !(930_000..=1_008_200).contains(&bit_rate) {
```
wkozyra95, line 126: - if something changed in the compositor that affected this test then adjust lower bounds too.
- if you are doing something like this make sure to communicate it clearly. You added that change after getting PR approved and I only noticed this by accident.

PR #835, opened 2024-10-22
[tests] Change expected bitrate in offline processing test

noituri:


wkozyra95:
please adjust it to sth reasonable, I assume you increased it to 1_008_200 because the resulting value was around that. Is there any reason that the result might be around `938_200`?

wkozyra95 (approved):

PR #839, opened 2024-10-25
Switch to SRGB textures + add padding to outptu texture

wkozyra95:
- Fix passing output_resolution with WASM (alignment to 16 bytes)
- Switch all textures to srgb

Snapshots: https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/46

### DIFF
--- a/compositor_render/src/wgpu/format/rgba_to_yuv.wgsl
+++ b/compositor_render/src/wgpu/format/rgba_to_yuv.wgsl
@@ -23,26 +23,35 @@ fn vs_main(input: VertexInput) -> VertexOutput {
 
 var<push_constant> plane_selector: u32;
 
+fn linear_to_srgb(linear: vec3<f32>) -> vec3<f32> {
+    let cutoff = step(linear, vec3<f32>(0.0031308));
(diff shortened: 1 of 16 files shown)


compositor_render/src/wgpu/format/rgba_to_yuv.wgsl
```diff
@@ -23,26 +23,35 @@ fn vs_main(input: VertexInput) -> VertexOutput {
 
 var<push_constant> plane_selector: u32;
 
+fn linear_to_srgb(linear: vec3<f32>) -> vec3<f32> {
```
wkozyra95, line 26: I have to do this in shader because WebGl does not support creating non srgb views on srgb textures.
We could do this better if we switch to WebGPU only

compositor_web/Cargo.toml
```diff
@@ -36,3 +36,6 @@ glyphon = { workspace = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 getrandom = { version = "0.2", features = ["js"] }
+
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = true
```
noituri, line 41: Isn't `wasm-opt` enabled by default?

noituri (approved):

PR #845, opened 2024-10-31
Remove spec-compliant reference picture list init and modification.

jerzywilczek:
Turns out these parts don't need to be done exactly according to the spec. The vulkan docs are not very clear on this unfortunately, but it seems to work.


wkozyra95 (approved):

WojciechBarczynski (commented):
LGTM

PR #848, opened 2024-11-04
[tests] Regenerate snapshot tests (no system fonts)

wkozyra95:
If we use font embedded in compositor the tests seem to work more reliably between platforms and allowed_error can be set lower


src/snapshot_tests/tests.rs
```diff
@@ -1722,7 +1720,6 @@ fn view_snapshot_tests() -> Vec<TestCase> {
         },
         TestCase {
             name: "view/border_radius_box_shadow_overflow_fit",
-            only: true,
```
wkozyra95, line 1725: left by mistake in previous PR

PR #853, opened 2024-11-08
Add file logger

wkozyra95:
When the compositor is run deep inside some other solution it is sometimes hard to get logs from it. Add file logger should solve that problem


WojciechBarczynski:
Shouldn't this be in the changelog and docs?

wkozyra95:
> Shouldn't this be documented in the changelog and docs?

Changelog yes, for docs I don't plan to update the website, but decided to add it in the markdown file, so when we move to the new site does options are not missed

WojciechBarczynski (commented):
LGTM

PR #857, opened 2024-11-12
New architecture for vk-video initialization.

### LINKED ISSUE #821: Add an example with a simple wgpu-based video player to vk-video

jerzywilczek:
This change allows users to initialize `vk-video` with wgpu surfaces for rendering to an on-screen window. This was previously overlooked and there was no possibility to guarantee a `VulkanCtx` was able to render on screen.

This patch also adds an example h264 player to vk-video.

Closes #821.

### DIFF
--- a/compositor_pipeline/src/pipeline/graphics_context.rs
+++ b/compositor_pipeline/src/pipeline/graphics_context.rs
@@ -1,14 +1,23 @@
 use crate::error::InitPipelineError;
-use compositor_render::{create_wgpu_ctx, error::InitRendererEngineError};
+use compositor_render::{create_wgpu_ctx, error::InitRendererEngineError, WgpuComponents};
 use std::sync::Arc;
+
+#[cfg(feature = "vk-video")]
+#[derive(Debug, Clone)]
+pub struct VulkanCtx {
+    pub device: Arc<vk_video::VulkanDevice>,
+    pub instance: Arc<vk_video::VulkanInstance>,
+}
 
 #[derive(Debug)]
 pub struct GraphicsContext {
     pub device: Arc<wgpu::Device>,
     pub queue: Arc<wgpu::Queue>,
+    pub adapter: Arc<wgpu::Adapter>,
+    pub instance: Arc<wgpu::Instance>,
 
     #[cfg(feature = "vk-video")]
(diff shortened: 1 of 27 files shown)


vk-video/examples/player/player/shader.wgsl
```diff
@@ -0,0 +1,46 @@
...
+    output.tex_coords = input.tex_coords;
+
+    return output;
+}
+
+@group(0) @binding(0) var y_texture: texture_2d<f32>;
+@group(0) @binding(1) var uv_texture: texture_2d<f32>;
+@group(0) @binding(2) var sampler_: sampler;
+
+fn srgb_to_linear(srgb: vec3<f32>) -> vec3<f32> {
```
wkozyra95, line 25: This is only needed if you write to SRGB texture. I added that in the shader because WASM does not support multiple TextureViews on single Texture, but a better way would be to remove this conversion and when writing to Srgb texture use non-Srgb view

compositor_pipeline/src/pipeline/graphics_context.rs
```diff
@@ -8,7 +8,9 @@ pub struct GraphicsContext {
     pub queue: Arc<wgpu::Queue>,
```
wkozyra95, line 9: remember about adding adapter and instance here too

compositor_pipeline/src/pipeline/graphics_context.rs
```diff
@@ -8,7 +8,9 @@ pub struct GraphicsContext {
...
 
     #[cfg(feature = "vk-video")]
-    pub vulkan_ctx: Option<Arc<vk_video::VulkanCtx>>,
+    pub vulkan_device: Option<Arc<vk_video::VulkanDevice>>,
```
wkozyra95, line 11: maybe group that together into a single option?
- `Option<(Device, Instance)>` 
- `Option<VulkanCtx>`

vk-video/examples/player/player/shader.wgsl
jerzywilczek: oh, ok, I'll try that

wkozyra95 (approved):

WojciechBarczynski (approved):

PR #875, opened 2024-11-20
Add hardware encoding to the compositor API.

jerzywilczek:


compositor_api/src/types/from_register_input.rs
```diff
@@ -106,10 +110,20 @@ impl TryFrom<RtpInput> for pipeline::RegisterInputOptions {
...
+
+                            #[cfg(feature = "vk-video")]
+                            VideoDecoder::VulkanVideo => decoder::VideoDecoderOptions {
+                                decoder: pipeline::VideoDecoder::VulkanVideoH264,
+                            },
+
+                            #[cfg(not(feature = "vk-video"))]
+                            VideoDecoder::VulkanVideo => {
+                                return Err(TypeError::new(NO_VULKAN_VIDEO))
+                            }
```
WojciechBarczynski, line 126: We may want to ship compiled binaries with this feature enabled (at least for Linux x86). Scripts for generating binaries should be updated to handle that (this feature is not enabled by default if I understand it correctly).

What happens when someone tries to enable this feature on a device that doesn't support the vulkan video? Does it crash on compile, on input register or in run-time? If compilation has to be done on a device with vulkan video support, we might not be able to build those binaries using GH Actions.

wkozyra95: We should ship only one version. For macos it would be disabled, and for linux always be enabled. But yes good point release script probably needs to be updated, I don't see anywhere where this feature was defined so I'm not sure if it is enabled by default

compositor_api/src/types/register_input.rs
```diff
@@ -122,8 +124,25 @@ pub enum InputRtpAudioOptions {
...
+    FfmpegH264,
+
+    /// Use hardware decoder based on Vulkan Video.
+    ///
+    /// This should be faster and more scalable than teh ffmpeg decoder, if the hardware and OS
+    /// support it.
+    ///
+    /// This requires hardware that supports Vulkan Video. Another requirement is this program has
+    /// to be compiled with the `vk-video` feature enabled.
+    #[serde(rename = "vulkan_video")]
```
wkozyra95, line 146:
```suggestion
```
I don't think this is need, you already have `rename_all = "snake_case"`

compositor_api/src/types/from_register_input.rs
wkozyra95: not sure if feature is needed, maybe just change it based on OS

jerzywilczek: If we want to remove the feature we can use something like this to define an alias for the OS support `cfg` line: https://lib.rs/crates/cfg_aliases

I would like to keep the feature in though, maybe some users want to experiment with weird configurations if moltenvk adds support for vulkan video

compositor_api/src/types/register_input.rs
```diff
@@ -122,8 +124,23 @@ pub enum InputRtpAudioOptions {
...
+    /// Use the software decoder based on ffmpeg.
+    FfmpegH264,
+
+    /// Use hardware decoder based on Vulkan Video.
+    ///
+    /// This should be faster and more scalable than teh ffmpeg decoder, if the hardware and OS
+    /// support it.
+    ///
+    /// This requires hardware that supports Vulkan Video. Another requirement is this program has
+    /// to be compiled with the `vk-video` feature enabled.
```
WojciechBarczynski, line 144: This might be confusing for someone who just wants to use precompiled binaries. Maybe we should mention what features are supported in the precompiled binaries. I would also add here mention like:
```rust
    /// This requires hardware that supports Vulkan Video. Another requirement is this program has
    /// to be compiled with the `vk-video` feature enabled (enabled by default on Linux x86, disabled by default on other platforms).
```

WojciechBarczynski (approved):

wkozyra95 (approved):

PR #880, opened 2024-11-22
[ts-sdk] Add offline processing

wkozyra95:
### Api changes

- Add `Slides`/`Slide` components
- Add `Show` component
- Add the `useAfterTimestamp` hook (internally used by `Show`)
- Add `useCurrentTimestamp` returns the current timestamp (does not trigger re-render for every time change)
  - Alternative 1:  Add a hook that returns a function that returns a timestamp
  - Alternative 2: Add a hook that returns some context object that has a method that returns a timestamp
  - Alterantive 3: Do not expose it, but instead make `useAfterTimestamp` return the current timestamp (and maybe change name of that hook)
- Add `useBlockingTask` hook
  - For live cases run the async function on mount. Hook returns result if promise resolved or undefined otherwise
    - Similiar to https://react.dev/reference/react/use
  - For offline cases it blocks rendering until promise resolves and returns the result after.
- Keep `newBlockingTask` function prrivate (internal implementation of `useBlockingTask`)
  - When it would be usfull as public API?
    - It could be useful if you want to implement something more advanced than use useBlockingTask, e.g. instead of running on mount it would rerun on specific props changes, or it should trigger after some PTS
  - Alternative 1: Make that function public and add some hook that returns `TimeContext` object. (newBlockingTask needs the context as argument)
  - Alternative 2: Add hook e.g. `useBlockingTaskBuilder` that returns a function to start a new blocking task.
- Temporarily make public `useTimeLimitedComponent`. 
  - That hook is intended to be private after we start using it in `InputStream` or `Mp4` components. It is public for now, so we can write some examples and test internalls.
  - Alternative 1: Make that hook public so users can write their own components with a duration.

### TODO in follow up

- useInputStream will return correct states for offline processing based on offset and calculated duration.

```tsx
function ExampleAppWithMultipleScreens() {
  return (
    <>
      <Show timestampMS={{ end: 3000 }}>
        <SomeComponent description="show from the beginning until 3 second" />
      </Show>
      <Show timestampMs={{ start: 3000, end: 6000 }}>
        <SomeComponent description="show between 3s and 6s />
      </Show>
      <Show delayMs={3000}>
        <SomeComponent description="3 seconds after mount" />
      </Show>
    </>
  );
}
```

```tsx
function ExampleApp() {
  return (
    <View>
      <Slides>
        <Slide>
          <SomeComponent description="duration based on components inside Input" />
        </Slide>
        <Slide>
          <SomeComponent description="duration based on components inside Input" />
        </Slide>
        <Slide durationMs={3_000}>
          <SomeComponent  description="duration 3 seconds, ignore components" />
        </Slide>
      </Slides>
    </View>
  );
}
```

```tsx
function ExampleAppWithBlockingTask() {
  const text = useBlockingTask(async function() {
     const response = await fetch("https://some/data")
     return await response.text()
  });
  return (
    <View>
      <Text>{text}</Text>
    </View>
  );
}
```

Currently, there are no components that support automatic slide change. I'm planning to add
- Support for InputStream if it is mp4 to set it based on it's duration
- New Mp4 component that is basically InputStream but does not require registration
- (maybe) some fake component that does not render to anything but if it is inserted in scene it is forcing some duration

### DIFF
--- a/ts/@live-compositor/core/src/offline/output.ts
+++ b/ts/@live-compositor/core/src/offline/output.ts
@@ -1,0 +1,206 @@
+import type { Outputs } from 'live-compositor';
+import { _liveCompositorInternals, View } from 'live-compositor';
+import type React from 'react';
+import { createElement, useSyncExternalStore } from 'react';
+import type { ApiClient, Api } from '../api.js';
+import Renderer from '../renderer.js';
+import type { RegisterOutput } from '../api/output.js';
+import { intoAudioInputsConfiguration } from '../api/output.js';
+import { sleep } from '../utils.js';
+
+type AudioContext = _liveCompositorInternals.AudioContext;
+type OfflineTimeContext = _liveCompositorInternals.OfflineTimeContext;
+type InstanceContextStore = _liveCompositorInternals.InstanceContextStore;
+
+class OfflineOutput {
+  api: ApiClient;
+  outputId: string;
+  audioContext: AudioContext;
+  timeContext: OfflineTimeContext;
+  outputShutdownStateStore: OutputShutdownStateStore;
+  durationMs: number;
+  updateTracker: UpdateTracker = new UpdateTracker();
+
+  videoRenderer?: Renderer;
+  initialAudioConfig?: Outputs.AudioInputsConfiguration;
+
+  constructor(
+    outputId: string,
+    registerRequest: RegisterOutput,
+    api: ApiClient,
+    store: InstanceContextStore,
+    durationMs: number
+  ) {
+    this.api = api;
+    this.outputId = outputId;
+    this.outputShutdownStateStore = new OutputShutdownStateStore();
+    this.durationMs = durationMs;
+
+    const supportsAudio = 'audio' in registerRequest && !!registerRequest.audio;
+    if (supportsAudio) {
+      this.initialAudioConfig = registerRequest.audio!.initial ?? { inputs: [] };
+    }
+
+    const onUpdate = () => this.updateTracker.onUpdate();
+    this.audioContext = new _liveCompositorInternals.AudioContext(onUpdate, supportsAudio);
+    this.timeContext = new _liveCompositorInternals.OfflineTimeContext(onUpdate);
+
+    if (registerRequest.video) {
+      const rootElement = createElement(OutputRootComponent, {
+        instanceStore: store,
+        audioContext: this.audioContext,
+        timeContext: this.timeContext,
+        outputRoot: registerRequest.video.root,
+        outputShutdownStateStore: this.outputShutdownStateStore,
+      });
+
+      this.videoRenderer = new Renderer({
+        rootElement,
+        onUpdate,
+        idPrefix: `${outputId}-`,
+      });
+    }
+  }
+
+  public scene(): { video?: Api.Video; audio?: Api.Audio; schedule_time_ms: number } {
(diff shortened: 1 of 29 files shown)


ts/examples/node-examples/src/combine_mp4.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { useState } from 'react';
+
+function ExampleApp() {
+  return (
+    <Tiles transition={{ durationMs: 200 }}>
+      <InputTile inputId="input_1" />
+      <Show delayMs={2000}>
+        <InputTile inputId="input_2" />
+      </Show>
+      <Show timestampMs={{ start: 5000, end: 8000 }}>
```
BrtqKr, line 23: either `<Show timeRange={{ start: 5000, end: 8000 }}>` or `<Show timeRangeMs={{ start: 5000, end: 8000 }}>`, but I prefer the first one - I'm not sure if that many libraries specify the unit in the props in this manner. You can usually safely assume that it's ms.

ts/examples/node-examples/src/combine_mp4.tsx
```diff
@@ -0,0 +1,90 @@
...
+} from 'live-compositor';
+import { downloadAllAssets } from './utils';
+import path from 'path';
+import { useState } from 'react';
+
+function ExampleApp() {
+  return (
+    <Tiles transition={{ durationMs: 200 }}>
+      <InputTile inputId="input_1" />
+      <Show delayMs={2000}>
```
BrtqKr, line 20: I'd only remove the unit to unify it with the timestamp, if necessary

ts/examples/node-examples/src/combine_mp4.tsx
```diff
@@ -0,0 +1,90 @@
...
+  useAfterTimestamp,
+  Show,
+} from 'live-compositor';
+import { downloadAllAssets } from './utils';
+import path from 'path';
+import { useState } from 'react';
+
+function ExampleApp() {
+  return (
+    <Tiles transition={{ durationMs: 200 }}>
```
BrtqKr, line 18: I'd only remove the unit to unify it with the timestamp, if necessary

ts/live-compositor/src/components/Show.ts
```diff
@@ -0,0 +1,39 @@
+import type React from 'react';
+
+import { useAfterTimestamp } from '../hooks.js';
+import { LiveCompositorContext } from '../context/index.js';
+import { useContext, useEffect, useState } from 'react';
+
+export type ShowProps = {
+  timestampMs?: { start?: number; end?: number };
+  delayMs?: number;
```
BrtqKr, line 9: same with delay

ts/@live-compositor/core/src/offline/output.ts
```diff
@@ -0,0 +1,206 @@
...
+        "Render for a specific timestamp took too long, make sure you don't have infinite update loop."
+      );
+      this.promiseRes();
+    }, MAX_RENDER_TIMEOUT_MS);
+    await this.promise;
+    clearTimeout(this.renderTimeout);
+    clearTimeout(this.updateTimeout);
+  }
+}
+
```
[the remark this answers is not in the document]
wkozyra95: Yes, this mechanism similliar to throttling. We want to send update when we know that onUpdate was not called for 20ms. The second time limit (`renderTimeout`) is to just protect against some infinite loop

noituri (approved):
LGTM

PR #881, opened 2024-11-22
WHIP-server

wkazmierczak:


compositor_pipeline/src/pipeline/input/whip.rs
```diff
@@ -0,0 +1,248 @@
...
+            debug!("Closing RTP depayloader thread.");
+            break;
+        };
+
+        let _ = sender.send(chunk);
+    }
+}
+
+fn generate_token() -> String {
+    let mut rng = rand::thread_rng();
```
wkozyra95, line 220: not like that, google how to generate random tokens in Rust

compositor_pipeline/src/pipeline/whip_whep.rs
```diff
@@ -0,0 +1,200 @@
...
+mod utils;
+
+use tokio::task;
+
+use crate::{queue::PipelineEvent, Pipeline};
+
+use super::EncodedChunk;
+
+pub(crate) const VIDEO_PAYLOAD_TYPE: u8 = 96;
+pub(crate) const AUDIO_PAYLOAD_TYPE: u8 = 111;
```
wkozyra95, line 47: why 111?
why audio? I assume this is specific to OPUS

compositor_pipeline/src/pipeline/whip_whep.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl WhipWhepState {
+    pub fn new() -> Arc<Self> {
+        Arc::new(WhipWhepState {
+            input_connections: Arc::from(Mutex::new(HashMap::new())),
+        })
+    }
+}
+
+pub async fn init_pc(
```
wkozyra95, line 113: please look through my comments on whip client PR, some of the stuff from there applies in this PR
e.g. init_peer_connection

compositor_pipeline/src/pipeline/whip_whep/config.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+#[derive(Debug, Copy, Clone)]
+pub enum LoggerFormat {
+    Pretty,
+    Json,
+    Compact,
+}
+
+impl FromStr for LoggerFormat {
```
wkozyra95, line 16: don't read envs from inside this crate, env should only be read in top level crate and passed down via options

compositor_pipeline/src/pipeline/whip_whep/handlers.rs
```diff
@@ -0,0 +1,260 @@
...
+    peer_connection.on_track(Box::new(move |track, _, _| {
+        let track_kind = track.kind();
+        let video_sender_clone = video_sender.clone();
+        let audio_sender_clone = audio_sender.clone();
+        let state_clone = state.clone();
+        let input_id_clone = input_id.clone();
+        let depayloader_clone = depayloader.clone();
+
+        tokio::spawn(async move {
+            if track_kind == RTPCodecType::Video {
```
wkozyra95, line 115: seems like match would fit better here

compositor_pipeline/src/pipeline/whip_whep/handlers.rs
```diff
@@ -0,0 +1,260 @@
...
+    let sdp = peer_connection.local_description().await.unwrap();
+    info!("Sending SDP answer: {sdp:?}");
+
+    Response::builder()
+        .status(StatusCode::CREATED)
+        .header("Content-Type", "application/sdp")
+        .header("Access-Control-Expose-Headers", "Location")
+        .header("Location", format!("/session/{}", id))
+        .body(Body::from(sdp.sdp.to_string()))
+        .unwrap()
```
wkozyra95, line 172: Is it safe?

compositor_pipeline/src/pipeline/whip_whep/handlers.rs
```diff
@@ -0,0 +1,260 @@
...
+    }));
+
+    // Set the remote SDP offer
+    peer_connection
+        .set_remote_description(RTCSessionDescription::offer(offer).unwrap())
+        .await
+        .unwrap();
+
+    // Create and set the local SDP answer
+    let answer = peer_connection.create_answer(None).await.unwrap();
```
wkozyra95, line 155: are all of those  unwraps safe?

compositor_pipeline/src/pipeline/whip_whep/handlers.rs
brzep: no, it is not, it should be randomized as in [this section of WHIP RFC](https://datatracker.ietf.org/doc/html/draft-ietf-wish-whip#section-5-4.2.1)
I would also consider implementing safety mechanisms as in the rest of _Security considerations_ section

compositor_pipeline/src/pipeline.rs
```diff
@@ -183,22 +191,27 @@ impl Pipeline {
...
-            ctx: PipelineCtx {
-                output_sample_rate: opts.output_sample_rate,
-                output_framerate: opts.queue_options.output_framerate,
-                download_dir: download_dir.into(),
-                event_emitter,
-                tokio_rt: Arc::new(Runtime::new().map_err(InitPipelineError::CreateTokioRuntime)?),
-                #[cfg(feature = "vk-video")]
-                vulkan_ctx: preinitialized_ctx.and_then(|ctx| ctx.vulkan_ctx),
-            },
+            ctx: ctx.clone(),
```
brzep, line 214: I think this clone is unnecessary

compositor_pipeline/src/pipeline/input/whip.rs
```diff
@@ -0,0 +1,248 @@
...
+pub struct OutputAudioStream {
+    pub options: encoder::EncoderOptions,
+    pub payload_type: u8,
+}
+
+#[derive(Debug, Clone)]
+pub struct WhipStream {
+    pub video: Option<InputVideoStream>,
+    pub audio: Option<InputAudioStream>,
+}
```
brzep, line 77: To me, name containing `Stream` is a bit misleading, it suggest some kind of stream of packets or sth like this, whereas it seems to be just input encoders and payload type options. Also, aren't `WhipReceiverOptions` and `WhipStream` basically the same?

compositor_pipeline/src/pipeline/input/whip.rs
```diff
@@ -0,0 +1,254 @@
...
+    mut receiver: tokio::sync::mpsc::Receiver<PipelineEvent<EncodedChunk>>,
+    sender: Sender<PipelineEvent<EncodedChunk>>,
+) {
+    loop {
+        let Some(chunk) = receiver.blocking_recv() else {
+            debug!("Closing RTP depayloader thread.");
+            break;
+        };
+
+        let _ = sender.send(chunk);
```
brzep, line 222: shouldn't be closing a channel handled in some way?

compositor_pipeline/src/pipeline/whip_whep/handlers.rs
```diff
@@ -0,0 +1,256 @@
...
+    info!("[whip] got headers: {headers:?}");
+    info!("[whip] got request: {offer}");
+
+    let input_id = InputId(Arc::from(id.clone()));
+
+    // Validate that the Content-Type is `application/sdp`
+    if let Some(content_type) = headers.get("Content-Type") {
+        if content_type.as_bytes() != b"application/sdp" {
+            error!("Invalid Content-Type, expecting application/sdp");
+            return Err(WhipServerError::InternalError(
```
brzep, line 54: Isn't it a `BAD_REQUEST`? It's users responsibility to provide required header type.

compositor_pipeline/src/pipeline/whip_whep/handlers.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    if let Ok(connections) = state_clone.input_connections.lock() {
+        if let Some(connection) = connections.get(&input_id) {
+            video_sender = connection.video_sender.clone();
+            audio_sender = connection.audio_sender.clone();
+            depayloader = connection.depayloader.clone();
+            bearer_token = connection.bearer_token.clone();
+            authorize_token(bearer_token, headers.get("Authorization"))?;
+        } else {
+            return Err(WhipServerError::InternalError(
```
brzep, line 79: Not sure if it's server error, maybe `NOT_FOUND`?

compositor_pipeline/src/pipeline/input/whip.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct WhipStream {
+    pub video: Option<InputVideoStream>,
+    pub audio: Option<InputAudioStream>,
+}
+
+pub struct WhipReceiver {
+    should_close: Arc<AtomicBool>,
```
brzep, line 80: is this should_close handled anywhere?

compositor_pipeline/src/pipeline/whip_whep.rs
```diff
@@ -0,0 +1,202 @@
...
+    if !pipeline_ctx.start_whip_whep {
+        return;
+    }
+
+    let state = pipeline_ctx.whip_whep_state;
+    let port = pipeline_ctx.whip_whep_server_port;
+
+    let app = Router::new()
+        .route("/status", get(status))
+        .route("/whip/:id", post(handle_whip))
```
wkozyra95, line 60: use more descriptive name, what does it mean to handle whip?

compositor_pipeline/src/pipeline/whip_whep.rs
```diff
@@ -0,0 +1,202 @@
...
+        .route("/session/:id", patch(whip_ice_candidates_handler))
```
wkozyra95, line 61: what does it mean to handle ice candiadates? I suspect `whip::new_ice_candidates_handle` would be better and rename handlers to whip_handlers maybe

compositor_pipeline/src/pipeline/input/whip.rs
```diff
@@ -0,0 +1,286 @@
...
+                    start_time_vid: None,
+                    start_time_aud: None,
+                    depayloader,
+                },
+            );
+        } else {
+            return Err(WhipReceiverError::WhipWhepStateAddInput(input_id.clone()));
+        }
+
+        let video = match (video_rx_crossbeam, opts.video) {
```
wkozyra95, line 118: you are checking video in bunch of places, in general you want to only check it in one place and do everything there.

When you are creating video_rx_crossbeam you know on type level that it is not None, but here you need to handle again which is not necessary.

Separation like this might be good if those places become to complicated, but in most cases there are better ways (type safe) that can solve those situation

compositor_pipeline/src/pipeline/whip_whep/whip_handlers/create_whip_session.rs
```diff
@@ -0,0 +1,217 @@
...
+        .header("Location", format!("/session/{}", encode(&id)))
+        .body(body)?;
+    Ok(response)
+}
+
+fn get_whip_input_connection_options(
+    state: Arc<WhipWhepState>,
+    input_id: InputId,
+) -> Result<WhipInputConnectionOptions, WhipServerError> {
+    if let Ok(connections) = state.input_connections.lock() {
```
wkozyra95, line 149: unwrap on mutex is fine, I don't think it is worth complexity to handle  it everywhere. Mutex lock only fails if it is poisoned (some different thread needs to panic for this to happen)

compositor_pipeline/src/pipeline/input/whip.rs
```diff
@@ -0,0 +1,286 @@
...
+                        };
+                    });
+                }
+            }
+            connections.remove(&self.input_id);
+        }
+    }
+}
+
+pub async fn handle_track(
```
wkozyra95, line 219: naming

compositor_pipeline/src/pipeline/whip_whep/whip_handlers/create_whip_session.rs
```diff
@@ -0,0 +1,217 @@
...
+                        )
+                        .await;
+                    }
+                }
+                _ => {
+                    debug!("RTPCodecType not supported!")
+                }
+            }
+        });
+        Box::pin(async {})
```
wkozyra95, line 111: why are you spawning new tokio task above, instead of doing all the work in `Box::pin(async {})`?

compositor_pipeline/src/pipeline/whip_whep/whip_handlers/terminate_whip_session.rs
```diff
@@ -0,0 +1,63 @@
...
+        ));
+    }
+    validate_token(bearer_token, headers.get("Authorization")).await?;
+    let peer_connection;
+
+    if let Ok(mut connections) = state.input_connections.lock() {
+        if let Some(connection) = connections.get_mut(&input_id) {
+            peer_connection = connection.peer_connection.clone();
+            connection.peer_connection = None;
+            drop(connection.audio_sender.clone());
```
wkozyra95, line 40: no idea what you wanted achive here, but it does not make sense. You are cloning value and then droping that cloned value.

wkozyra95 (changes requested):
I don't think I checked everything in this PR.

Some of the issues I noticed suggest to me that it is not just missing/forgetting something, but more issue with understanding what some part of the code does. So, after addressing my comments make sure to go through the PR and verify if you understand everything there. 

If you are not sure about some parts, make sure to bring them up either by marking them in PR, or we can talk about them in the office.

compositor_pipeline/src/pipeline/whip_whep.rs
```diff
@@ -0,0 +1,194 @@
...
+        .with_graceful_shutdown(shutdown_signal(shutdown_signal_receiver))
+        .await
+    {
+        error!("Cannot serve WHIP/WHEP server task: {err:?}");
+    };
+    state.clear_state();
+}
+
+async fn shutdown_signal(receiver: oneshot::Receiver<()>) {
+    let ctrl_c = async {
```
wkozyra95, line 70: remove this, you should not access os signals in this crate

compositor_pipeline/src/pipeline/whip_whep.rs
```diff
@@ -0,0 +1,194 @@
...
+    };
+
+    tokio::select! {
+        _ = ctrl_c => {},
+        _ = terminate => {},
+    }
+}
+
+pub async fn status() -> Result<(StatusCode, axum::Json<Value>), WhipServerError> {
+    Ok((StatusCode::OK, axum::Json(json!({}))))
```
wkozyra95, line 92: feels like something that could be inline, but I don't have any strong opinion on that

compositor_pipeline/src/pipeline/whip_whep/validate_bearer_token.rs
```diff
@@ -0,0 +1,42 @@
+use std::time::Duration;
+
+use axum::http::HeaderValue;
+use rand::{rngs::StdRng, Rng, SeedableRng};
+use tokio::time::sleep;
+use tracing::warn;
+
+use crate::pipeline::whip_whep::error::WhipServerError;
+
```
wkozyra95, line 9: if you have separate file for token related stuff then you could rename this file and move token generation here too

compositor_pipeline/src/pipeline/whip_whep/validate_bearer_token.rs
```diff
@@ -0,0 +1,42 @@
...
+            })?;
+
+            if let Some(token_from_header) = auth_str.strip_prefix("Bearer ") {
+                if token_from_header == bearer_token {
+                    Ok(())
+                } else {
+                    let mut rng = StdRng::from_entropy();
+                    let millis = rng.gen_range(50..1000);
+                    sleep(Duration::from_millis(millis)).await;
+                    warn!("Invalid or mismatched token provided");
```
wkozyra95, line 27: Logging warnings/errors is needed when we don't have a way to return that information to the user. In this case this Result will be send in the response,

And if we want to additionally log errors that are sent as resposnses, then why would this error be more important than any other that might be sent in response to this request. We should either log all error responses or none of them

compositor_pipeline/src/pipeline/whip_whep.rs
```diff
@@ -0,0 +1,194 @@
...
+        input_id: InputId,
+        peer_connection: Arc<RTCPeerConnection>,
+    ) -> Result<(), WhipServerError> {
+        let mut connections = self.input_connections.lock().unwrap();
+        if let Some(connection) = connections.get_mut(&input_id) {
+            connection.peer_connection = Some(peer_connection);
+            Ok(())
+        } else {
+            Err(WhipServerError::InternalError(
+                "Peer connection initialization error".to_string(),
```
wkozyra95, line 170: This error description is misleading, this can happen if there if input with that id does not exist.

compositor_pipeline/src/pipeline/whip_whep.rs
```diff
@@ -0,0 +1,194 @@
...
+        &self,
+        input_id: InputId,
+    ) -> Result<WhipInputConnectionOptions, WhipServerError> {
+        let connections = self.input_connections.lock().unwrap();
+        if let Some(connection) = connections.get(&input_id) {
+            if let Some(peer_connection) = connection.peer_connection.clone() {
+                warn!("There is another stream streaming for given input {input_id:?}");
+                if peer_connection.connection_state() == RTCPeerConnectionState::Connected {
+                    return Err(WhipServerError::InternalError(format!(
+                        "There is another stream streaming for given input {input_id:?}"
```
wkozyra95, line 147: stream streaming?

wkozyra95 (commented):
Please look through the flow where errors are returned and when logged and make sure that:
- error massages make sense and describe the actual error
- we do not have unnecessary logs and they have approriate level 
  - especially warn/error logs should not logged in general if the Result is sent as a response. If you want to log error responses send by a WHIP server then all error responses should be logged not just those few places you added manually
  - log and result is fine for debug/trace levels

compositor_pipeline/src/pipeline/whip_whep/bearer_token.rs
```diff
@@ -0,0 +1,52 @@
...
+            let auth_str = auth_str.to_str().map_err(|_| {
+                WhipServerError::Unauthorized("Invalid UTF-8 in header".to_string())
+            })?;
+
+            if let Some(token_from_header) = auth_str.strip_prefix("Bearer ") {
+                if token_from_header == bearer_token {
+                    Ok(())
+                } else {
+                    let mut rng = StdRng::from_entropy();
+                    let millis = rng.gen_range(50..1000);
```
wkozyra95, line 36: comparing strings takes a few nanoseconds, if you only sleep by increment of full milliseconds then it does not help much. There are 2 options:
- sleep for between (0, 1000) nanoseconds
- keep the original range (50, 1000) milliseconds, but with nanosecond precision

compositor_pipeline/src/pipeline/whip_whep/whip_handlers/create_whip_session.rs
```diff
@@ -0,0 +1,163 @@
...
+        ));
+    }
+    Ok(())
+}
+
+pub async fn gather_ice_candidates_for_one_second(peer_connection: Arc<RTCPeerConnection>) {
+    let (sender, mut receiver) = watch::channel(peer_connection.ice_gathering_state());
+
+    peer_connection.on_ice_gathering_state_change(Box::new(move |gatherer_state| {
+        let gathering_state = match gatherer_state {
```
wkozyra95, line 139: why are you translating from one type to the other if the only thing you do in the end is compare to RTCIceGatheringState::Complete ?

wkozyra95 (approved):

PR #887, opened 2024-11-28
Download CEF binaries to `~/.cache`

noituri:
Most of the CEF build time is spent on actually downloading CEF itself which happens every time `OUT_DIR` changes. I moved the default `cef_root` to `~/.cache` which helps with the recompilation a lot (less than 1s).


wkozyra95 (approved):
Not sure if it matters for the cache specifically, but I would use tools like https://github.com/dirs-dev/dirs-rs to find directories like that on a specific system.

PR #889, opened 2024-11-29
Move OPUS decoder buffer to heap

wkozyra95:
Not sure if it related to Rust update, but audio decoder is failing for me on master. Moving OPUS buffer to heap seems to fix it

### DIFF
--- a/build_tools/nix/flake.lock
+++ b/build_tools/nix/flake.lock
@@ -5,11 +5,11 @@
         "nixpkgs-lib": "nixpkgs-lib"
       },
       "locked": {
-        "lastModified": 1722555600,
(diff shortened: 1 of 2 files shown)


build_tools/nix/flake.lock
```diff
@@ -5,11 +5,11 @@
         "nixpkgs-lib": "nixpkgs-lib"
       },
       "locked": {
-        "lastModified": 1722555600,
```
wkozyra95, line 8: updates rust from 1.79 to 1.82

WojciechBarczynski:
Have you tested it? Tests with audio don't work

wkozyra95:
> Have you tested it? Tests with audio don't work

yes, I'll look into that

PR #890, opened 2024-12-06
Return duration of an Mp4 when possible + fix `loop` option

wkozyra95:


compositor_pipeline/src/pipeline/input/mp4.rs
```diff
@@ -89,72 +87,312 @@ impl Mp4 {
...
+                        audio_track =
+                            Some(TrackProvider::Value(audio_thread.unwrap().join().unwrap()));
+                        should_close.store(true, Ordering::Relaxed);
+                        if let Some(video_thread) = video_thread {
+                            video_track = Some(TrackProvider::Handle(video_thread));
+                        }
+                    }
+                }
+                if has_audio {
+                    offset = Duration::from_nanos(last_audio_sample_pts.load(Ordering::Relaxed));
```
wkozyra95, line 314: If an audio track is present use it specifically. Using lasdt video sample or last sample in general is leading to audio artifacts on the start of a new track

compositor_pipeline/src/pipeline/input/mp4/reader.rs
```diff
@@ -0,0 +1,245 @@
...
+
+            data.freeze()
+        };
+
+        Some(Track {
+            sample_unpacker: Box::new(sample_unpacker),
+            sample_count: track.sample_count(),
+            timescale: track.timescale(),
+            track_id,
+            duration: track.duration(),
```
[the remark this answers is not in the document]
jerzywilczek: The duration of the reader is calculated from a part of the mp4, which contains "overall information which is media-independent, and relevant to the entire presentation considered as a whole", and the duration of the track is specific to the track. Not sure which one we should use either. Maybe the reader duration should theoretically be used when looping the mp4, but I don't think it's worth changing the looping code to use it instead.

jerzywilczek (approved):
Looks good

compositor_pipeline/src/pipeline/input/mp4.rs
wkozyra95: buffers between queue<->decoder and decoder-input are in chunks, if chunks (or frame/sample batches) do not represent similar duration then timestamps for audio and video here can differ significantly. 

For mp4 (15 fps) I tested, audio processing was behind the video by over 2 seconds, because the buffer can fit a lot more video at low fps.

2 address that we would have to use channels between elements that have sized based on media duration instead.

jerzywilczek: oh, that makes a lot of sense.

WojciechBarczynski (commented):
LGTM

PR #892, opened 2024-12-10
Add initial benchmarking setup.

jerzywilczek:
Here's an initial benchmarking setup. For context, the end goal for this is for us to be able to run a set of these benchmarks on any computer and figure out what kind of smelter setups can be run on it. 

You can run it with:
```sh
cargo run --bin benchmark -- --help
```
to get the information on how to run a single test.

The full command is quite verbose, and looks like this:
```sh
cargo run --release --bin benchmark -- \
    --framerate 24 \
    --decoder-count maximize \
    --file-path examples/assets/BigBuckBunny.mp4 \
    --output-width 1280 \
    --output-height 720 \
    --encoder-preset ultrafast \
    --warm-up-time 10 \
    --measured-time 10 \
    --video-decoder vulkan_video_h264 \
    --framerate-tolerance 1.05
```
The help message will tell you all of the options for each of the arguments.

It can currently run a single test configuration. A single configuration is comprised of a couple of constant parameters (the resolution, the input video, the encoder preset etc.) and currently two variable parameters: framerate and decoders count. The variable parameters can be set to:
- a constant
- exponential iteration, meaning that the test will run a couple of iterations with the parameter being multiplied by 2 in every consecutive run, until the test stops running
- maximizing, meaning that the benchmark will binsearch the maximum number the configuration works for

Importantly, you can set multiple parameters to iterate and a single parameter to maximize, which would mean the setup would run for all iterative combinations and for each of them find the maximum number for the maximized parameter.

For example, if we set the framerate to iterate and the decoders count to maximize, the setup will run with framerate set to 1, 2, 4, 8, 16 etc. and for each framerate find the maximum number of inputs. The iteration will terminate when it is impossible to run the system due to the super high framerate even with a single decoder.

What I plan to add to this later:
- more parameters that can be variable: encoder count, encoder preset, resolution etc.
- a possibility to run without one of the elements of the pipeline, e.g. just the decoders and the renderer, without the encoder, or just the encoders, without the decoders
- a predefined set of benchmark runs that can be run with a single command, once we decide on what we want to measure

Additionally to asking whether you like this system or think it should be changed somehow, I would like to get your opinion on my handling of the arguments that are set to iterate or maximize (the `run_args` function and the other functions it calls). I feel a bit like using zig for the last couple weeks got into my head and that there is a better, more idiomatic way of doing what I'm doing here, but I just can't figure it out.

I also fixed an unrelated typo in vk-video, because it was annoying me.

### DIFF
--- a/integration_tests/src/bin/benchmark/main.rs
+++ b/integration_tests/src/bin/benchmark/main.rs
@@ -1,0 +1,328 @@
+use std::{
+    sync::{Arc, Mutex},
+    time::{Duration, Instant},
+};
+
+use clap::Parser;
+use compositor_pipeline::{
+    pipeline::{
+        encoder::VideoEncoderOptions,
+        input::{
+            mp4::{Mp4Options, Source},
+            InputOptions,
+        },
+        output::EncodedDataOutputOptions,
+        GraphicsContext, Options, OutputVideoOptions, PipelineOutputEndCondition,
+        RegisterInputOptions, RegisterOutputOptions,
+    },
+    queue::{self, QueueInputOptions, QueueOptions},
+    Pipeline,
+};
+
+use compositor_pipeline::pipeline::encoder::ffmpeg_h264::Options as H264OutputOptions;
+use compositor_render::{
+    scene::{
+        Component, HorizontalAlign, InputStreamComponent, RGBAColor, TilesComponent, VerticalAlign,
+    },
+    web_renderer::WebRendererInitOptions,
+    Framerate, InputId, OutputId, Resolution,
+};
+use live_compositor::{
+    config::{read_config, LoggerConfig},
+    logger,
+};
+use tracing::warn;
+
+mod args;
+
+use args::{Args, Argument, SingleBenchConfig};
+
+fn main() {
+    let args = Args::parse();
+    let config = read_config();
+    ffmpeg_next::format::network::init();
+    let logger_config = LoggerConfig {
+        level: "compositor_pipeline=error,vk-video=info,benchmark=info".into(),
+        ..config.logger
+    };
+    logger::init_logger(logger_config);
+
+    let ctx = GraphicsContext::new(
+        false,
+        wgpu::Features::SAMPLED_TEXTURE_AND_STORAGE_BUFFER_ARRAY_NON_UNIFORM_INDEXING
+            | wgpu::Features::UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING,
+        Default::default(),
+        None,
+    )
+    .unwrap();
+
+    if cfg!(debug_assertions) {
+        warn!("This benchmark is running in debug mode. Make sure to run in release mode for reliable results.");
+    }
+
+    let reports = run_args(ctx, &args);
+    SingleBenchConfig::log_report_header();
+    for report in reports {
+        report.log_as_report();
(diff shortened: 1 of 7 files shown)


integration_tests/src/bin/benchmark/main.rs
```diff
@@ -0,0 +1,302 @@
...
+                })),
+            },
+        },
+    )
+    .unwrap();
+
+    Pipeline::start(&pipeline);
+
+    let start_time = Instant::now();
+    while Instant::now() - start_time < bench_config.warm_up_time {
```
wkozyra95, line 286:
```suggestion
    while start_time.elapsed() < bench_config.warm_up_time {
```

wkozyra95: alternative approach would be to just do thread sleep and reading from receiver in a separate channel

wkozyra95 (approved):

integration_tests/src/bin/benchmark/main.rs
```diff
@@ -0,0 +1,302 @@
...
+                    source: Source::File(bench_config.file_path.clone()),
+                }),
+                queue_options: QueueInputOptions {
+                    offset: Some(Duration::ZERO),
+                    required: true,
+                    buffer_duration: None,
+                },
+            },
+        )
+        .unwrap();
```
brzep, line 234: I came across an error 
```
InputError(InputId("input_121"), Mp4(IoError(Os { code: 24, kind: Uncategorized, message: "Too many open files" })))
```
It should be handled somehow imo. Maybe just log it and assume that last count was max? I think it would be pretty anoying when I got such error after running test for longer time.

Maybe make `run_single_test` return `Result<bool, some-kind-of-error>` and handle those on higher level?

integration_tests/src/bin/benchmark/main.rs
```diff
@@ -0,0 +1,302 @@
...
+                    continue;
+                } else {
+                    return any_succeeded;
+                }
+            }
+        }
+    }
+
+    run_args_maximize(ctx, args, arguments, reports)
+}
```
noituri, line 115: I'm not sure about this code. On the first sight I thought there might be an infinite recursion. While now I understand what's happening here, the flow here is hard to follow. Also, I tried to run `--decoder-count` with `iterate_exp` and it didn't do anything. I'm not sure if I used it correctly tho

noituri (commented):
🤓

integration_tests/src/bin/benchmark/main.rs
jerzywilczek: fixed it quickly for now

integration_tests/src/bin/benchmark/main.rs
jerzywilczek: you must have not set any argument to `maximize`, at which point the test is kind of pointless. This should be caught and reported as a problem: there's nothing you can learn by just setting all arguments to `iterate` and none to maximize. If you set framerate to `iterate_exp` and decoders to `maximize` it will work.

I'll add code to catch this sort of a mistake.

jerzywilczek: I also added comments in some places, because I don't have ideas on how to make this easier to read. I also changed the return type of this function to `struct FurtherIterationPossible(bool)` to make what the returned `bool` means.

noituri (approved):

PR #893, opened 2024-12-12
Publish arm docker images as part of the release process

noituri:
Example multi-platform builds:
- slim: https://github.com/software-mansion/live-compositor/pkgs/container/smelter/333356027?tag=test5
- web-renderer: https://github.com/software-mansion/live-compositor/pkgs/container/smelter/333354551?tag=test5-web-renderer

### DIFF
--- a/.github/workflows/package_for_release.yml
+++ b/.github/workflows/package_for_release.yml
@@ -128,7 +128,7 @@ jobs:
                 name: smelter_with_web_renderer_darwin_aarch64.tar.gz
                 path: smelter_with_web_renderer_darwin_aarch64.tar.gz
 
-    docker:
+    docker-amd64:
       runs-on: ubuntu-latest
       steps:
           - name: 📥 Checkout repo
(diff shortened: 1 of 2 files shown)


.github/workflows/package_for_release.yml
```diff
@@ -140,12 +140,12 @@ jobs:
 
           - name: 🔨 Build Smelter image
             run: |
-              docker buildx build --platform linux/amd64 -t ghcr.io/software-mansion/smelter:${{ github.sha }} -f build_tools/docker/slim.Dockerfile .
+              docker buildx build --platform linux/amd64 -t ghcr.io/software-mansion/smelter:${{ github.sha }}-amd64 -f build_tools/docker/slim.Dockerfile .
```
wkozyra95, line 143: you can have an image with the same tag for different architectures

.github/workflows/package_for_release.yml
```diff
@@ -145,7 +145,42 @@ jobs:
...
+              echo "AMD64_IMAGE_ID=$(docker inspect --format='{{index .RepoDigests 0}}' ghcr.io/software-mansion/smelter:${{ github.sha }})" >> "$GITHUB_ENV"
+
+          - name: 🔨 Build Smelter ARM64 image
+            run: |
+              docker buildx build --platform linux/arm64 -t ghcr.io/software-mansion/smelter:${{ github.sha }} -f build_tools/docker/slim.Dockerfile .
+
+          - name: ⏩ Push ARM64 image
+            run: |
+              docker push ghcr.io/software-mansion/smelter:${{ github.sha }}
+              echo "ARM64_IMAGE_ID=$(docker inspect --format='{{index .RepoDigests 0}}' ghcr.io/software-mansion/smelter:${{ github.sha }})" >> "$GITHUB_ENV"
```
wkozyra95, line 172: Is this command show digest of a image that was build last? Seems hacky is there a way to avoid relaying on that build order?

scripts/release.sh
```diff
@@ -40,11 +40,8 @@ IMAGE_NAME="ghcr.io/software-mansion/smelter"
 docker pull "${IMAGE_NAME}:${COMMIT_HASH}"
 docker pull "${IMAGE_NAME}:${COMMIT_HASH}-web-renderer"
 
-docker tag "${IMAGE_NAME}:${COMMIT_HASH}" "${IMAGE_NAME}:${RELEASE_TAG}"
-docker tag "${IMAGE_NAME}:${COMMIT_HASH}-web-renderer" "${IMAGE_NAME}:${RELEASE_TAG}-web-renderer"
-
-docker push "${IMAGE_NAME}:${RELEASE_TAG}"
-docker push "${IMAGE_NAME}:${RELEASE_TAG}-web-renderer"
+docker buildx imagetools create -t "${IMAGE_NAME}:${RELEASE_TAG}" "${IMAGE_NAME}:${COMMIT_HASH}"
```
wkozyra95, line 43: Is the docker pull above needed? I assumed that it only fetches image for one arch, so this command must not rely on what actual is fetched locally. (just guessing here)

wkozyra95 (approved):
LGTM, just few questions to make sure I understand

scripts/release.sh
noituri: You're correct. The pull is not needed here. Fixed

.github/workflows/package_for_release.yml
noituri: Discussed outside of GitHub

PR #895, opened 2024-12-16
Support other color formats

BrtqKr:
[#896](https://github.com/software-mansion/live-compositor/issues/896)


compositor_api/src/types/from_color.rs
```diff
@@ -0,0 +1,268 @@
+use compositor_render::scene;
+
+use super::util::*;
+
+impl TryFrom<RGBAColor> for scene::RGBAColor {
+    type Error = TypeError;
+    fn try_from(value: RGBAColor) -> std::result::Result<Self, Self::Error> {
+        let s = &value.0.trim();
+        match s.chars().next() {
```
wkozyra95, line 9: Something like this would be more readable
```
if let Some(color) = parse_named_color() {
  return Ok(color)
}
if s.starts_with('#') {
  ...
  return
}
if s.starts_with('rgb(') {
  ...
  return
}
if s.starts_with('rgba(') {
  ...
  return
}
```

compositor_api/src/types/from_color.rs
```diff
@@ -0,0 +1,268 @@
...
+    ))
+}
+
+fn parse_color_channel(value: &str) -> Result<u8, TypeError> {
+    u8::from_str_radix(value, 16).map_err(|_err| {
+        TypeError::new("Invalid format. Color representation is not a valid hexadecimal number.")
+    })
+}
+
+fn parse_named_color(color_name: &str) -> Option<&'static str> {
```
wkozyra95, line 115: either return `Option<RGBAColor>` or `Option<RGBColor>`

compositor_api/src/types/from_color.rs
```diff
@@ -0,0 +1,268 @@
...
+        return Err(TypeError::new("Invalid RGB format."));
+    }
+    let (r, g, b) = (parts[0].trim(), parts[1].trim(), parts[2].trim());
+    let a = "255"; // default full opacity
+    parse_color_components_from_decimal(r, g, b, a)
+}
+
+fn parse_rgba(s: &str) -> Result<scene::RGBAColor, TypeError> {
+    let inner_s = s.trim_start_matches("rgba(").trim_end_matches(')');
+    let parts: Vec<&str> = inner_s.split(',').collect();
```
wkozyra95, line 63: You could do sth like this
```suggestion
    let parts: Vec<u32> = inner_s.split(',').map(|c|parse_color_channel(c.trim())).collect::<Result<Vec<_>,_>()?;
```
not sure if the `::<Result<Vec<_>,_>` will be needed or Rust can figure out the type

ts/live-compositor/src/components/Text.ts
```diff
@@ -93,8 +93,8 @@ function sceneBuilder(props: TextProps, children: SceneComponent[]): Api.Compone
     max_height: style?.maxHeight,
     font_size: style?.fontSize ?? DEFAULT_FONT_SIZE,
     line_height: style?.lineHeight,
-    color_rgba: style?.color && intoApiRgbaColor(style?.color),
-    background_color_rgba: style?.backgroundColor && intoApiRgbaColor(style?.backgroundColor),
+    color: style?.color && intoApiRgbaColor(style?.color),
```
wkozyra95, line 96: is the change in sdk added by mistake or intentional? I'm fine either way but keep in mind that you will have to use binaries build from source and point `LIVE_COMPOSITOR_PATH` to that binary.

If you build with `cargo build --release --no-default-features` then executable is in `./target/release/main_process`

(no default features is important to build without chrome support)

ts/live-compositor/src/api.generated.ts
```diff
@@ -552,6 +590,22 @@ export type Component =
        * effect if the previous scene already contained a `Rescaler` component with the same id.
        */
       transition?: Transition | null;
+      /**
+       * (**default=`0.0`**) Radius of a rounded corner.
+       */
+      border_radius?: number | null;
```
wkozyra95, line 596: not part of this PR, but you could add properties from this diff too new docs if we don't plan to version old compositor releases

ts/live-compositor/src/components/Text.ts
wkozyra95: also after this change you don't need intoApiRgbaColor

compositor_api/src/types/from_color.rs
BrtqKr: it couldn't be handled like that because alpha in the original format is a float between 0 and 1. I've processed it slightly differently as a whole

ts/live-compositor/src/components/Text.ts
BrtqKr: leftover, sorry for that

compositor_api/src/types/from_color.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    if !(0.0..=1.0).contains(&a) {
+        return Err(TypeError::new(
+            "Alpha value out of range. It must be between 0.0 and 1.0",
+        ));
+    }
+    let alpha = (a * 255.0).round() as u8;
+
+    Ok(scene::RGBAColor(r, g, b, alpha))
+}
```
wkozyra95, line 109:
```suggestion
fn parse_rgb(s: &str) -> Result<scene::RGBAColor, TypeError> {
    let inner_s = s.trim_start_matches("rgb(").trim_end_matches(')');
    let parts: Vec<&str> = inner_s.split(',').collect();
    match parts.as_slice() {
        [r, g, b] => Ok(scene::RGBAColor(
            parse_color_channel(r, 10)?,
            parse_color_channel(g, 10)?,
            parse_color_channel(b, 10)?,
            255,
        )),
        _ => Err(TypeError::new("Invalid RGB format.")),
    }
}

fn parse_rgba(s: &str) -> Result<scene::RGBAColor, TypeError> {
    let inner_s = s.trim_start_matches("rgba(").trim_end_matches(')');
    let parts: Vec<&str> = inner_s.split(',').collect();
    let [r, g, b, a] = parts.as_slice() else {
        return Err(TypeError::new("Expected three color components and alpha channel."));
    };

    let a = a
        .parse::<f32>()
        .map_err(|_| TypeError::new("Alpha channel parsing failed."))?;

    if !(0.0..=1.0).contains(&a) {
        return Err(TypeError::new(
            "Alpha value out of range. It must be between 0.0 and 1.0",
        ));
    }

   Ok(scene::RGBAColor(
            parse_color_channel(r, 10)?,
            parse_color_channel(g, 10)?,
            parse_color_channel(b, 10)?,
            (a * 255.0).round() as u8,
        )) 
}
```

wkozyra95 (approved):

PR #897, opened 2024-12-16
Enable faststart and fix seeking in outputted mp4s.

jerzywilczek:
It turned out that in order to be able to seek in produced mp4s, the muxer needs information about whether a packet is a keyframe or not.

The naming of the `IsKeyframe` enum and its variants is quite awkward, I'm open to suggestions on what to name them.

I also renamed the `mp4` example to `mp4_output` to better reflect what it does.


wkozyra95 (approved):

WojciechBarczynski (commented):
LGTM

PR #905, opened 2024-12-19
[ts-sdk] Update generated API types

noituri:
The API types were not regenerated after 53b726d4e00ad09e737457fd6b884a954bf19614


wkozyra95 (changes requested):
This will fix web, but it will break node in the old setup. I added docs yesterday how the process should look from now on https://github.com/software-mansion/live-compositor/blob/master/DEVELOPMENT.md#process-for-introducing-api-changes

In the new approach we will maintain SDK code to be compatibles with Rust sources and not the released version as it was up untill now. However, if you regenerating types then PR should include all the changes necessary, otherwise we will miss a lot of stuff.

There is a lot of new stuff, so I recommend just closing this PR and in the meantime, either 
- checkout to older commit to build WASM and checkout back
- temporally revert pr with color changes on the branch you are working

PR #906, opened 2024-12-20
[web-wasm] Decouple frame production from input implementation

noituri:
The current `Input` implementation assumed that every input type will have a similar frame production and frame buffering strategy. That approach does not work well with camera and share-screen input. This PR makes it possible to add different strategies for frame production.

- Each frame producer has to implement `InputFrameProducer` interface.
- `DecodingFrameProducer` accepts any `InputSource` (Encoded chunk producer). It produces frames from encoded chunks and handles the frame buffering.
- The future input types such as `Camera` & `Share screen` will require different implementation of `InputFrameProducer` which  accepts `MediaStream` and does no buffering.


wkozyra95 (approved):
I did not check too much, but I plan to take a look at that logic latter on, so if this works I'm fine with merging this as it is.

PR #907, opened 2025-01-07
[ts-sdk] Ensure compositor instance cleanup (Node.js and browser)

wkozyra95:
- Add LiveCompositor.terminate()
- Fix issues in rendering that caused LiveCompositor not to be garbage collected. (stored in global react state)
- Improve API of WASM package to make it simpler, plus add helpful component that wraps canvas


ts/@live-compositor/browser-render/src/wasm.ts
```diff
@@ -5,7 +5,7 @@ import init, * as wasm from './generated/compositor_web';
  * @param wasmModuleUrl {string} - An URL for `live-compositor.wasm` file. The file is located in `dist` folder.
  */
 export async function loadWasmModule(wasmModuleUrl: string) {
-  await init(wasmModuleUrl);
+  await init({ module_or_path: wasmModuleUrl });
```
wkozyra95, line 8: passing string works, but it triggers warning

ts/@live-compositor/core/src/live/output.ts
```diff
@@ -87,25 +90,18 @@ class Output {
...
 
-  public close(): void {
-    this.throttledUpdate = () => {};
-    // close will switch a scene to just a <View />, so we need replace `throttledUpdate`
-    // callback before it is called
-    this.outputShutdownStateStore.close();
+  public async close(): Promise<void> {
+    this.throttledUpdate.setFn(async () => {});
+    this.renderer.stop();
+    await this.throttledUpdate.waitForPendingCalls();
```
wkozyra95, line 96: I changed the throttledUpdate to class because I need to be able to wait for pending requests to finish, otherwise `LiveCompositor.terminate()` might kill the server before pending request are done

PR #916, opened 2025-01-10
[web-wasm] Move compositor to web worker + add camera (copied from #913)

wkozyra95:


ts/@live-compositor/web-wasm/src/wasmInstance.ts
```diff
@@ -0,0 +1,246 @@
...
+          stream: trackProcessor.readable,
+        },
+      },
+      [trackProcessor.readable],
+    ];
+  } else if (body.type === 'screen_capture') {
+    const mediaStream = await navigator.mediaDevices.getDisplayMedia({
+      audio: false,
+      video: {
+        width: { max: 2000 },
```
wkozyra95, line 223: without this screen capture on my monitor is too large and wgpu is not able to allocate texture (limit was somewhere around 2000, but this it not exact number

ts/@live-compositor/web-wasm/src/wasmInstance.ts
```diff
@@ -0,0 +1,246 @@
...
+        height: { max: 2000 },
+      },
+    });
+    const videoTrack = mediaStream.getVideoTracks()[0];
+    assert(videoTrack);
+    // @ts-ignore
+    const trackProcessor = new MediaStreamTrackProcessor({ track: videoTrack });
```
wkozyra95, line 230: screenshare is not closed on unregister currently. I'll address that inthe  followup

PR #918, opened 2025-01-13
Add `register_font` endpoint

wkazmierczak:


src/routes/register_request.rs
```diff
@@ -139,3 +142,46 @@ pub(super) async fn handle_image(
...
+
+    let bytes = field
+        .bytes()
+        .await
+        .map_err(|err| ApiError::malformed_request(&err))?;
+
+    tokio::task::spawn_blocking(move || {
+        Pipeline::register_font(
+            &api.pipeline.lock().unwrap(),
+            Source::Binary(Arc::new(bytes.to_vec())),
```
wkozyra95, line 181: create binary outside of the task

Currently you are 
- crearating Bytes object
- converting it to Vec
- and creating the Arc<Vec<>> from that

I think you can create Source::Binary  without that copy

src/routes/register_request.rs
```diff
@@ -139,3 +142,46 @@ pub(super) async fn handle_image(
     .await
     .unwrap()
 }
+
+pub(super) async fn handle_font(
```
wkozyra95, line 146: axum repo has an example how to do this https://github.com/tokio-rs/axum/blob/main/examples/multipart-form/src/main.rs

wkozyra95 (approved):

noituri (approved):

PR #931, opened 2025-01-27
Add padding to `View` component

### LINKED ISSUE #926: Add padding to view component

noituri:
Snapshot tests: https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/52

The padding behavior:
- Static children:
  - If the size is defined for the child, grow the child's size by its padding, and offset its position with parent padding
  - If the size is not defined, shrink the child to accommodate the parent's padding, and offset its position with the parent's padding
- Absolute children:
  -  If the size is defined for the child, grow the child's size by its padding
  - `padding_top` and `padding_left` is added to `top` and `left` parameters respectively
  - For `right`/`bottom`, `padding_right`/`padding_bottom` is subtracted, and then `padding_left`/`padding_top` is added

Closes #926


compositor_api/src/types/component.rs
```diff
@@ -92,6 +92,18 @@ pub struct View {
 
     /// List of box shadows.
     pub box_shadow: Option<Vec<BoxShadow>>,
+
+    /// (**default=`0.0`**) Padding on top side in pixels.
+    pub padding_top: Option<f32>,
```
wkozyra95, line 97: I think we talked about adding also option padding and maybe (padding_vertical and horizontal)

compositor_render/src/scene/view_component/layout.rs
```diff
@@ -72,12 +72,18 @@ impl ViewComponentParam {
...
             .iter_mut()
             .map(|child| {
-                let position = match child {
-                    StatefulComponent::Layout(layout) => layout.position(pts),
-                    ref non_layout_component => Position::Static {
-                        width: non_layout_component.width(pts),
-                        height: non_layout_component.height(pts),
-                    },
+                let (position, padding) = match child {
+                    StatefulComponent::Layout(StatefulLayoutComponent::View(view)) => {
```
wkozyra95, line 76: Why do you need to know padding of a child?

compositor_render/src/scene/layout.rs
```diff
@@ -161,18 +161,28 @@ impl StatefulLayoutComponent {
...
         parent_size: Size,
+        child_padding: Padding,
+        parent_padding: Padding,
         pts: Duration,
     ) -> NestedLayout {
-        let width = position.width.unwrap_or(parent_size.width);
-        let height = position.height.unwrap_or(parent_size.height);
+        let width = position
+            .width
+            .unwrap_or(parent_size.width + child_padding.horizontal());
```
wkozyra95, line 170: parent_size is the max size so why are you adding here more?

compositor_render/src/scene/layout.rs
```diff
@@ -161,18 +161,28 @@ impl StatefulLayoutComponent {
...
+            VerticalPosition::BottomOffset(bottom) => {
+                parent_size.height - bottom - height - parent_padding.bottom + parent_padding.top
+            }
         };
         let left = match position.position_horizontal {
-            HorizontalPosition::LeftOffset(left) => left,
-            HorizontalPosition::RightOffset(right) => parent_size.width - right - width,
+            HorizontalPosition::LeftOffset(left) => left + parent_padding.left,
+            HorizontalPosition::RightOffset(right) => {
+                parent_size.width - right - width - parent_padding.right + parent_padding.left
```
wkozyra95, line 184: why parent padding is taken into account? parent padding should not have any affect on absolutely positioned element

snapshot_tests/view/column_view_padding_static_children.scene.json
```diff
@@ -0,0 +1,49 @@
...
+        {
+          "type": "view",
+          "background_color": "#FF0000FF",
+          "direction": "column",
+          "width": 300,
+          "height": 300,
+          "children": [
+            {
+              "type": "view",
+              "padding_top": 10,
```
wkozyra95, line 16: you are not testing anything here, without any children padding does not do anything

snapshot_tests/view/column_view_padding_static_children.scene.json
```diff
@@ -0,0 +1,49 @@
+{
```
wkozyra95, line 1: All the test include children that have specified size and overflow. You have not test that validate if this works for view with undefined size

wkozyra95 (changes requested):

compositor_render/src/scene/layout.rs
noituri: The parent padding does affect the absolutely positioned child element. At least that's the case in CSS's `position: relative` and IIRC ours `absolute` is modeled after `position: relative` unless I misunderstood it

compositor_render/src/scene/view_component/layout.rs
noituri: Not necessary anymore. I removed it

compositor_api/src/types/component.rs
noituri: It was added only to react sdk but I'll add it to api as well

wkozyra95 (approved):

PR #934, opened 2025-01-28
Add register font to `ts-sdk`

wkazmierczak:


ts/@live-compositor/node/package.json
```diff
@@ -21,12 +21,15 @@
     "@types/node": "^20.14.10",
     "@types/node-fetch": "^2.6.11",
     "@types/uuid": "^10.0.0",
-    "@types/ws": "^8.5.12"
+    "@types/ws": "^8.5.12",
+    "@types/react": "^18.3.3"
```
wkozyra95, line 25: Why is this needed?

ts/@live-compositor/node/src/live/compositor.ts
```diff
@@ -0,0 +1,92 @@
...
+  public async init(): Promise<void> {
+    await this.coreCompositor!.init();
+  }
+
+  public async registerOutput(
+    outputId: string,
+    root: ReactElement,
+    request: CoreOutput.RegisterOutput
+  ): Promise<void> {
+    assert(this.coreCompositor);
```
wkozyra95, line 33: there is no need for those asserts

ts/@live-compositor/node/src/live/compositor.ts
```diff
@@ -0,0 +1,92 @@
...
+import { createLogger } from '../logger';
+import LocallySpawnedInstance from '../manager/locallySpawnedInstance';
+import assert from 'assert';
+import type { ReactElement } from 'react';
+import type { Renderers } from 'live-compositor';
+import FormData from 'form-data';
+import fetch from 'node-fetch';
+
+export default class LiveCompositor {
+  private coreCompositor?: CoreLiveCompositor;
```
wkozyra95, line 16: why is this field optional? It is always initialized in the constructor

ts/@live-compositor/node/src/offline/compositor.ts
```diff
@@ -0,0 +1,70 @@
...
+import type { ReactElement } from 'react';
+import type { Renderers } from 'live-compositor';
+import fetch from 'node-fetch';
+import FormData from 'form-data';
+
+export default class OfflineCompositor {
+  private coreCompositor?: CoreLiveCompositor;
+  private nodeCompositorManager: NodeCompositorManager;
+
+  public constructor(manager?: NodeCompositorManager) {
```
wkozyra95, line 19: the same as offline

ts/@live-compositor/core/src/compositorManager.ts
```diff
@@ -16,3 +16,7 @@ export interface CompositorManager {
   registerEventListener(cb: (event: unknown) => void): void;
   terminate(): Promise<void>;
 }
+
+export interface NodeCompositorManager extends CompositorManager {
+  sendMultipartRequest(request: MultipartRequest): Promise<object>;
```
wkozyra95, line 21: after thinking it through I think you should move that method to CompositorManager and remove that new manager

For `web-wasm` just implement empty method that throws an error.

ts/@live-compositor/node/src/fetch.ts
```diff
@@ -33,6 +34,28 @@ export async function sendRequest(baseUrl: string, request: ApiRequest): Promise
...
+  baseUrl: string,
+  request: MultipartRequest
+): Promise<object> {
+  const response = await fetch(new URL(request.route, baseUrl), {
+    method: request.method,
+    body: request.body as FormData,
+    agent: url => (url.protocol === 'http:' ? httpAgent : httpsAgent),
+  });
+  if (response.status >= 400) {
+    const err: any = new Error(`Multipart request to compositor failed.`);
```
wkozyra95, line 47:
```suggestion
    const err: any = new Error(`Request to compositor failed.`);
```
multipart is not important information for the user. If you want to add more information it should be maybe route/error code/response body, but whether it is multipart or json should not mater to someone that needs to handle this exception

ts/@live-compositor/node/src/fetch.ts
```diff
@@ -33,6 +34,28 @@ export async function sendRequest(baseUrl: string, request: ApiRequest): Promise
...
+): Promise<object> {
+  const response = await fetch(new URL(request.route, baseUrl), {
+    method: request.method,
+    body: request.body as FormData,
+    agent: url => (url.protocol === 'http:' ? httpAgent : httpsAgent),
+  });
+  if (response.status >= 400) {
+    const err: any = new Error(`Request to compositor failed.`);
+    err.response = response;
+    try {
```
wkozyra95, line 49: I think we talked about it already, if json() failes then `text()` will also fail because body was already read. This implementation and `sendRequest` is incorrect

wkozyra95 (approved):
Merging it for now, but in follow up PR
- please address error handling in sendRequest
- make sure passing assets works in vite example
- modify vite examples to not use the url directlly

PR #938, opened 2025-01-31
[web-wasm] Add minimal audio support

wkozyra95:


ts/@live-compositor/web-wasm/src/mainContext/output/stream.ts
```diff
@@ -42,16 +52,15 @@ export async function handleRegisterStreamOutput(
...
   }
 
-  // @ts-ignore
-  const audioTrack = new MediaStreamTrackGenerator({ kind: 'audio' });
-  stream.addTrack(audioTrack);
+  let mixer: AudioMixer | undefined;
+  if (request.audio) {
+    const mixer = new MediaStreamAudioMixer();
+    stream.addTrack(mixer.outputMediaStream());
+  }
```
noituri, line 61: Should `mixer` be shadowed here?

PR #951, opened 2025-02-05
Develop benchmarks

### LINKED ISSUE #937: Develop benchmarks
Develop more benchmark options:
- [x] resolution
  - [x] presets
  - [x] iteration
  - [x] maximize
- [x] outputs
  - [x] iteration
  - [x] maximize
- [x] run without encoder
- [x] run without decoder
- [x] add csv export option
- [x] readable output formatting

brzep:
Develop more benchmark options:

- [x] resolution
  - [x] presets 
  - [x] iteration
  - [x] maximize
- [x] outputs
  - [x] iteration
  - [x] maximize
- [x] run without encoder
- [x] run without decoder
- [x] add csv export option
- [x] readable output formatting
- renamed `encoder/decoder_count` to `output/input_count` to be more meaningful

closes #937

### DIFF
--- a/integration_tests/src/bin/benchmark/args.rs
+++ b/integration_tests/src/bin/benchmark/args.rs
@@ -2,7 +2,7 @@ use std::{path::PathBuf, time::Duration};
 
 use compositor_pipeline::pipeline::{self, encoder::ffmpeg_h264};
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Argument {
     IterateExp,
     Maximize,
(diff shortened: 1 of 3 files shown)


integration_tests/src/bin/benchmark/args.rs
```diff
@@ -99,6 +99,104 @@ impl From<EncoderPreset> for ffmpeg_h264::EncoderPreset {
...
 
+#[derive(Debug, Clone, Copy, clap::ValueEnum)]
+#[clap(rename_all = "snake_case")]
+pub enum ResolutionPreset {
+    Uhd,
+    Qhd,
+    Fhd,
+    Hd,
+    Sd,
+}
```
jerzywilczek, line 110: I think it would be cool if the only thing you specified on the cmd was the number ++ p: `--resolution 1080p`. I would hate if you had to specify the resolution like this: `--resolution res_480_p`.

Also, `derive(ValueEnum)` does nothing here I think, since you implement parsing manually anyway.

integration_tests/src/bin/benchmark/args.rs
```diff
@@ -99,6 +99,104 @@ impl From<EncoderPreset> for ffmpeg_h264::EncoderPreset {
...
+                "invalid resolution preset, available options: sd, hd, fhd, qhd, uhd".to_string(),
+            ),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum ResolutionConstant {
+    Preset(ResolutionPreset),
+    Value(u32, u32),
```
jerzywilczek, line 132:
```suggestion
    Value(Resolution),
```
That's how I'd do it, but it's up to you

integration_tests/src/bin/benchmark/args.rs
```diff
@@ -108,19 +206,24 @@ pub struct Args {
...
+    /// [possible values: uhd, qhd, fhd, hd, sd or `<width>x<height>`]
     #[arg(long)]
-    pub output_height: u32,
+    pub output_resolution: ResolutionConstant,
 
     #[arg(long)]
-    pub encoder_preset: EncoderPreset,
+    pub disable_encoder: bool,
+
+    #[arg(long, global = true, required_unless_present("disable_encoder"))]
```
jerzywilczek, line 225: We don't need [`global`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.global) here, as we don't have subcommands.

Also, you can use [`default_value`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_value) instead of `Option`

brzep: I don't think `default_value` changes much here, it's not used anyway if `disable_encoder` is present. Apart from that, I would highly consider adding default values to all parameters, so you don't need to specify all of them every time what is a bit inconvenient

jerzywilczek (approved):

integration_tests/src/bin/benchmark/main.rs
```diff
@@ -233,98 +294,248 @@ fn run_single_test(ctx: GraphicsContext, bench_config: SingleBenchConfig) -> boo
...
+            required: true,
+            buffer_duration: None,
+        },
+    )?;
+    Ok(x.video.unwrap())
+}
 
-    let end_time = Instant::now();
+fn raw_data_sender(bench_config: SingleBenchConfig, senders: Vec<Sender<PipelineEvent<Frame>>>) {
+    let mut file = File::open(bench_config.file_path).unwrap();
```
wkozyra95, line 513: I get requiring h264, but how do I get this file? Benchamarks should be self contained if possible.

Also, reading from the disk is not great, I get that it is only one read per input, but why it is necessary? Why not just generate a few static images (allocate Bytes once) and use them interchangeably to build FrameData?

wkozyra95:
Is my understanding correct that with this we can:
- Test variable amount of inputs
- Test variable amount of outputs

But we can't test a variable amount of instances? I'm assuming that a smelter with n outputs and n inputs where the output is using specific input would be the same as n smelters with 1 input and output. (I'm not saying we need that, just wanted o verify if we can or not)

jerzywilczek:
> Is my understanding correct that with this we can:
> 
>     * Test variable amount of inputs
> 
>     * Test variable amount of outputs
> 
> 
> But we can't test a variable amount of instances? I'm assuming that a smelter with n outputs and n inputs where the output is using specific input would be the same as n smelters with 1 input and output. (I'm not saying we need that, just wanted o verify if we can or not)

You mean iterating in a way where we always have the same number of encoders and decoders? and with each encoder using only one input?

wkozyra95:
> You mean iterating in a way where we always have the same number of encoders and decoders? and with each encoder using only one input?

The mode I was asking about would be that there is a constant number of encoders and decoders per instance. 

> and with each encoder using only one input?

not necessarily one, but the number of inputs and outputs per instance would be fixed

jerzywilczek:
But do you mean running actual separate instances of smelter, or one instance with k * x inputs and x outputs?

wkozyra95:
> But do you mean running actual separate instances of smelter, or one instance with k * x inputs and x outputs?

In my first question

> But we can't test a variable amount of instances?

I asked about  running multiple instances 

The rest of that comment compares that case with multiple instances of one instance n times input and outputs.

jerzywilczek:
can't run multiple instances, can't run with kx inputs and x outputs

PR #953, opened 2025-02-07
Fix image autoregister

wkozyra95:
- The rust server did not support scheduling image unregister which is necessary for the offline cases (or we could skip unregisters entirely)
- path field was not passed through the request
- Types for Mp4 and Image request were incorrect (as a fix I changed values to match, not types)
- Move `newBlocikngTask` just before the try-catch (if an exception would throw on asset type it would block rendering indefinitely

Tested on meetup example

### DIFF
--- a/src/routes/unregister_request.rs
+++ b/src/routes/unregister_request.rs
@@ -29,11 +29,10 @@ pub struct UnregisterOutput {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
-#[serde(tag = "type", rename_all = "snake_case")]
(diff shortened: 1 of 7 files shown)


src/routes/unregister_request.rs
```diff
@@ -29,11 +29,10 @@ pub struct UnregisterOutput {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
-#[serde(tag = "type", rename_all = "snake_case")]
-pub enum UnregisterRenderer {
```
wkozyra95, line 33: This type was not used for anything before

PR #954, opened 2025-02-07
[web-wasm] Make `SmelterRenderer` safe to use asynchronously

noituri:


compositor_web/src/wasm/renderer.rs
noituri, line 1: Moved from `wasm.rs`

compositor_web/src/wasm/renderer.rs
```diff
@@ -0,0 +1,140 @@
...
+                },
+            ),
+            api::ImageSpec::Gif { url, .. } => (url, ImageType::Gif),
+        };
+
+        let Some(url) = url else {
+            return Err(JsValue::from_str("Expected `url` field in image spec"));
+        };
+
+        let bytes = download(&url).await?;
```
wkozyra95, line 100: shouldn't stuff like this be before the mutext lock?

wkozyra95 (commented):
Can you test this with our vite examples, but change onCanvasCreated callback to onCanvasStarted and see if there are some issues there?

PR #955, opened 2025-02-11
[web-wasm] Ensure wasm module is loaded only once

noituri:
It's a potential fix for https://github.com/software-mansion/smelter/issues/874

When wasm module is loaded more than once it can override the wasm's internals. I tested it on `wgpu 24.0.1`  MacOS / Linux and it seems to be working

### DIFF
--- a/ts/smelter-browser-render/src/wasm.ts
+++ b/ts/smelter-browser-render/src/wasm.ts
@@ -4,8 +4,17 @@ import init, * as wasm from './generated/smelter/compositor_web';
  * Loads and initializes wasm module required for the smelter to work.
  * @param wasmModuleUrl {string} - An URL for `smelter.wasm` file. The file is located in `dist` folder.
  */
-export async function loadWasmModule(wasmModuleUrl: string) {
-  await init({ module_or_path: wasmModuleUrl });
-}
+export const loadWasmModule = (() => {
+  let loadResult: Promise<wasm.InitOutput> | undefined = undefined;
+  return async (wasmModuleUrl: string) => {
+    if (loadResult) {
+      await loadResult;
+      return;
+    }
+
+    loadResult = init({ module_or_path: wasmModuleUrl });
+    await loadResult;
+  };
+})();
 
 export { wasm };


ts/smelter-browser-render/src/wasm.ts
```diff
@@ -4,8 +4,15 @@ import init, * as wasm from './generated/smelter/compositor_web';
  * Loads and initializes wasm module required for the smelter to work.
  * @param wasmModuleUrl {string} - An URL for `smelter.wasm` file. The file is located in `dist` folder.
  */
-export async function loadWasmModule(wasmModuleUrl: string) {
-  await init({ module_or_path: wasmModuleUrl });
-}
+export const loadWasmModule = (() => {
```
wkozyra95, line 7: when this happens? wasm is loaded in worker code, so there should never be a case when this happens unless someone is using browser-render directlly?

noituri: Correct, it only happens in the Counter example which uses the renderer directly

wkozyra95 (approved):
By the way if init fails once then next attempt to call it will succeed, but it will not do anything

Is this intended?

noituri:
> By the way if init fails once then next attempt to call it will succeed, but it will not do anything
> 
> Is this intended?

I'm not sure when `init` could potentially fail, and if it's safe to call init again after failure (will it panic again or will it silently continue with undefined behavior?). I decided to do nothing in that case, although I'm not entirely sure if it's a good approach

ts/smelter-browser-render/src/wasm.ts
```diff
@@ -4,8 +4,15 @@ import init, * as wasm from './generated/smelter/compositor_web';
...
+  let isLoaded = false;
+  return async (wasmModuleUrl: string) => {
+    if (isLoaded) {
+      return;
+    }
+    isLoaded = true;
+    await init({ module_or_path: wasmModuleUrl });
+  };
```
wkozyra95, line 15:
```suggestion
  let loadResult = undefined;
  return async (wasmModuleUrl: string) => {
    if (loadResult) {
      return await loadResult;
    }
    loadResult = init({ module_or_path: wasmModuleUrl });
    await loadResult
  };
```

PR #956, opened 2025-02-11
Update `wgpu` to `24.0.1`

noituri:
After recent changes it looks like we can finally update `wgpu` safely. I added changes to vk-video from  reverted "Update wgpu to 23.0.0" commit.


vk-video/src/vulkan_decoder.rs
```diff
@@ -715,6 +715,11 @@ impl VulkanDecoder<'_> {
             }
         }
 
+        // this has to be done with Option and mut, because the closure we create has to be FnMut.
+        // this means we cannot consume its captures, so we have to take the option to be able to
+        // drop the resource.
```
jerzywilczek, line 720: This is no longer true, as 24.0 includes my changes added after this comment. The closure can be `FnOnce` now, so there's no need to wrap it in `Some`. Same for all `DropCallbacks`.

jerzywilczek (changes requested):

jerzywilczek (approved):
:rocket:

wkozyra95 (approved):

PR #979, opened 2025-02-25
[ts-sdk] Fix input queue on slow devices

wkozyra95:


ts/package.json
```diff
@@ -15,6 +15,7 @@
     "generate-types": "node ./scripts/generateTypes.mjs"
   },
   "devDependencies": {
+    "@eslint/js": "^9.21.0",
```
wkozyra95, line 18: on pnpm 10 this package no longer resolves without explicitlly installing it

ts/smelter-web-wasm/src/compositor/compositor.ts
```diff
@@ -36,7 +36,16 @@ export default class Smelter {
   private coreSmelter?: CoreSmelter;
   private instance?: WasmInstance;
   private options: SmelterOptions;
-  private logger: Logger = pino({ level: 'warn' });
+  private logger: Logger = pino({
+    level: 'warn',
+    browser: {
+      asObject: true,
+      write: {
+        debug: console.log,
```
wkozyra95, line 44: instead of console.trace that generates a  lot of output that makes it hard to read

ts/smelter-web-wasm/src/workerContext/input/QueuedInput.ts
```diff
@@ -113,18 +113,18 @@ export class QueuedInput implements Input {
...
-    if (!nextFrame) {
-      return;
-    } else if (nextFrame.type === 'frame') {
-      this.frames.push(this.newFrameRef(nextFrame.frame));
-    } else if (nextFrame.type === 'eos') {
-      this.receivedEos = true;
-      if (this.enqueueInterval) {
-        clearInterval(this.enqueueInterval);
+    while (this.frames.size() < MAX_BUFFER_FRAME_COUNT) {
+      const nextFrame = this.source.nextFrame();
```
wkozyra95, line 117: it only made single pass every 50ms and buffer was not filed, so dropping frames could not do much

PR #988, opened 2025-02-27
Add VP8 codec

wkazmierczak:


compositor_pipeline/src/pipeline/input/rtp/depayloader.rs
```diff
@@ -130,6 +140,35 @@ impl VideoDepayloader {
...
+            }
+            VideoDepayloader::VP8 {
+                depayloader,
+                buffer,
+                rollover_state,
+            } => {
+                let kind = EncodedChunkKind::Video(VideoCodec::VP8);
+                let vp8_chunk = depayloader.depacketize(&packet.payload)?;
+
+                if vp8_chunk.is_empty() {
```
wkozyra95, line 153: are you sure that batching packets to construct entire frame is necessary for vp8? For h264 decoder needs to get all chunks for a frame at once, so something like this is necessary.

wkozyra95 (approved):

PR #993, opened 2025-03-04
Fix muxing AAC in MP4 (missing metadata)

wkozyra95:
Mp4 that was generated before does not work in quicktime

### DIFF
--- a/compositor_pipeline/src/pipeline/output.rs
+++ b/compositor_pipeline/src/pipeline/output.rs
@@ -4,7 +4,7 @@ use compositor_render::{
     error::RequestKeyframeError, Frame, OutputFrameFormat, OutputId, Resolution,
 };
 use crossbeam_channel::{bounded, Receiver, Sender};
-use mp4::{Mp4FileWriter, Mp4OutputOptions};
+use mp4::{Mp4AudioEncoderConfig, Mp4EncoderConfig, Mp4FileWriter, Mp4OutputOptions};
 use tracing::debug;
 
 use crate::{audio_mixer::OutputSamples, error::RegisterOutputError, queue::PipelineEvent};
@@ -12,7 +12,7 @@ use crate::{audio_mixer::OutputSamples, error::RegisterOutputError, queue::Pipel
 use self::rtp::{RtpSender, RtpSenderOptions};
 
 use super::{
-    encoder::{AudioEncoderOptions, Encoder, EncoderOptions, VideoEncoderOptions},
+    encoder::{AudioEncoder, AudioEncoderOptions, Encoder, EncoderOptions, VideoEncoderOptions},
     types::EncoderOutputEvent,
     PipelineCtx, Port, RawDataReceiver,
 };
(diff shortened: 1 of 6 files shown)


compositor_pipeline/src/pipeline/output/mp4.rs
```diff
@@ -145,46 +149,56 @@ fn init_ffmpeg_output(
...
+            // This value is ignored when order is AV_CHANNEL_ORDER_UNSPEC
+            u: ffmpeg::ffi::AVChannelLayout__bindgen_ty_1 { mask: 0 },
+            // Field doc: "For some private data of the user."
+            opaque: ptr::null_mut(),
+        };
 
-            let id = stream_count;
-            stream_count += 1;
+        let id = stream_count;
+        // stream_count += 1;
```
wkozyra95, line 192: left as a comment, linter does not like the fact that value is never read after

compositor_pipeline/src/pipeline/output.rs
```diff
@@ -124,8 +124,31 @@ impl OutputOptionsExt<Option<Port>> for OutputOptions {
...
-                let writer = Mp4FileWriter::new(output_id.clone(), mp4_opt.clone(), packets, ctx)
-                    .map_err(|e| RegisterOutputError::OutputError(output_id.clone(), e))?;
+                let encoder_config = Mp4EncoderConfig {
+                    video: encoder.video.as_ref().map(|_| ()),
+                    audio: encoder
+                        .audio
+                        .as_ref()
+                        .map(|encoder| match encoder {
+                            AudioEncoder::Opus(_opus_encoder) => {
+                                Err(RegisterOutputError::UnknownError(
```
wkozyra95, line 134: I added an unknown error because this would be a bug in our code if this triggered.

compositor_pipeline/src/pipeline/output/mp4.rs
```diff
@@ -145,46 +149,56 @@ fn init_ffmpeg_output(
...
+            .add_stream(codec)
+            .map_err(OutputInitError::FfmpegMp4Error)?;
+
+        // If audio time base doesn't match sample rate, ffmpeg muxer produces incorrect timestamps.
+        stream.set_time_base(ffmpeg::Rational::new(1, sample_rate));
+
+        let codecpar = unsafe { &mut *(*stream.as_mut_ptr()).codecpar };
+        let Mp4AudioEncoderConfig::Aac(encoder_config) = encoder;
+        unsafe {
+            codecpar.extradata = ffmpeg_next::ffi::av_malloc(encoder_config.len()) as *mut u8;
```
WojciechBarczynski, line 170: @wkozyra95 Are you sure it should be `encoder_config.len() + AV_INPUT_BUFFER_PADDING_SIZE` instead of `encoder_config.len()`?

Screenshot 2025-03-06 at 14 05 32

https://ffmpeg.org/doxygen/6.1/structAVCodecParameters.html#a9befe0b86412646017afb0051d144d13

https://github.com/membraneframework/membrane_rtmp_plugin/blob/50c3ad6cc43ebc42e3cdde3053c930b61bbfec42/c_src/membrane_rtmp_plugin/sink/rtmp_sink.c#L126

Issue #701, opened 2024-08-28
Make user shaders work on WASM


noituri:
User shaders don't work due to lack of `TEXTURE_BINDING_ARRAY` feature

PR #996, opened 2025-03-05
[ts-sdk] Make shaders work on web

noituri:
Shaders on web have to be slightly different from the native version, because we can't use `binding_array` on web.
Due to those limitations, on the web, we'll only allow one texture to be passed to the user shader

Closes: #701

### DIFF
--- a/compositor_render/src/transformations/shader/validation/shader_header_web.wgsl
+++ b/compositor_render/src/transformations/shader/validation/shader_header_web.wgsl
@@ -1,0 +1,16 @@
+struct VertexInput {
+    @location(0) position: vec3<f32>,
+    @location(1) tex_coords: vec2<f32>,
+}
+
+struct BaseShaderParameters {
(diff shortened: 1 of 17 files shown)


compositor_render/src/transformations/shader/validation/shader_header_web.wgsl
```diff
@@ -0,0 +1,16 @@
+struct VertexInput {
+    @location(0) position: vec3<f32>,
+    @location(1) tex_coords: vec2<f32>,
+}
+
+struct BaseShaderParameters {
+    plane_id: i32,
+    time: f32,
+    output_resolution: vec2<u32>,
+    texture_count: u32,
```
noituri, line 10: I'm not sure if `texture_count` is needed here. Currently, we only allow one texture but in the future we might support more

wkozyra95 (approved):

PR #1000, opened 2025-03-06
Fix green tint on decklink input

wkozyra95:


wkozyra95:
It has not been tested yet. This is definitely a bug, but it might not be everything needed to fix the issue. I'm merging it, so it is included in the release.

PR #1007, opened 2025-03-10
[ts-sdk] Add `Blob` as MP4 input

### LINKED ISSUE #1001: [ts-sdk] Add support for mp4 blob

noituri:
Closes #1001

### DIFF
--- a/ts/smelter-core/src/api/input.ts
+++ b/ts/smelter-core/src/api/input.ts
@@ -12,10 +12,18 @@ import { _smelterInternals } from '@swmansion/smelter';
  * to compositor, but also additional variants that are specific to WASM like camera
  */
 export type RegisterInputRequest =
-  | Api.RegisterInput
+  | RegisterRtpStreamInputRequest
+  | RegisterMp4InputRequest
+  | RegisterWhipInputRequest
+  | RegisterDecklinkInputRequest
   | { type: 'camera' }
   | { type: 'screen_capture' }
   | { type: 'stream'; stream: any };
+
+export type RegisterRtpStreamInputRequest = Extract<Api.RegisterInput, { type: 'rtp_stream' }>;
+export type RegisterMp4InputRequest = { blob?: any } & Extract<Api.RegisterInput, { type: 'mp4' }>;
+export type RegisterWhipInputRequest = Extract<Api.RegisterInput, { type: 'whip' }>;
+export type RegisterDecklinkInputRequest = Extract<Api.RegisterInput, { type: 'decklink' }>;
 
 export type InputRef = _smelterInternals.InputRef;
 export const inputRefIntoRawId = _smelterInternals.inputRefIntoRawId;
@@ -56,6 +64,7 @@ function intoMp4RegisterInput(input: Inputs.RegisterMp4Input): RegisterInputRequ
     type: 'mp4',
     url: input.url,
     path: input.serverPath,
+    blob: input.blob,
     loop: input.loop,
     required: input.required,
     offset_ms: input.offsetMs,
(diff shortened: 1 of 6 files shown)


ts/examples/vite-example/src/components/SmelterVideo.tsx
noituri, line 1: This was not used by any example, but `MultipleOutputs` had pretty much the same implementation so I put everything here

ts/smelter-core/src/api/input.ts
```diff
@@ -13,6 +13,7 @@ import { _smelterInternals } from '@swmansion/smelter';
  */
 export type RegisterInputRequest =
   | Api.RegisterInput
+  | { type: 'mp4_blob'; blob: any }
```
noituri, line 16: `Api.RegisterInput` is a generated type and already contains definitions for `mp4`; hence, the new type `mp4_blob`

wkozyra95: I don't like adding new type just for that, it might make sense if it was more generic to handle some other cases types in the future, but not in this form

ts/smelter/src/components/Mp4.ts
```diff
@@ -32,17 +32,23 @@ function Mp4(props: Mp4Props) {
   useEffect(() => {
     const newInputId = newInternalStreamId();
     setInputId(newInputId);
-    const pathOrUrl: Pick<RegisterMp4Input, 'url' | 'serverPath'> =
-      props.source.startsWith('http://') || props.source.startsWith('https://')
-        ? { url: props.source }
-        : { serverPath: props.source };
+
+    let source: Pick<RegisterMp4Input, 'url' | 'serverPath' | 'blob'>;
```
wkozyra95, line 36: source might be confused with propsSource, maybe rename it to sourceObject or something like this

Also unrelated to your changes, but modify
```
const { muted, volume, ...otherProps } = props;
```
to 
```
const { muted, volume, source, ...otherProps } = props;
```

because we pass otherProps to InnerInputStream component, and source prop should not be passed there

wkozyra95 (commented):
What happens when somone would pass Blob object in node? Is the error clear?

ts/smelter/src/components/Mp4.ts
```diff
@@ -20,26 +20,32 @@ export type Mp4Props = Omit<ComponentBaseProps, 'children'> & {
...
+    } else if (source.startsWith('http://') || source.startsWith('https://')) {
+      sourceObject = { url: source };
     } else {
-      source = { serverPath: props.source };
+      sourceObject = { serverPath: source };
+    }
+
+    // If blob and run on Node.js
+    if (sourceObject.blob && typeof window === 'undefined') {
+      throw new Error('Blob is not supported on Node.js');
```
noituri, line 48: Instead of this, we could use different `Mp4Props` depending on whether the type is imported from wasm or node

ts/smelter/src/components/Mp4.ts
```diff
@@ -20,26 +20,32 @@ export type Mp4Props = Omit<ComponentBaseProps, 'children'> & {
...
+      sourceObject = { blob: source };
+    } else if (source.startsWith('http://') || source.startsWith('https://')) {
+      sourceObject = { url: source };
     } else {
-      source = { serverPath: props.source };
+      sourceObject = { serverPath: source };
+    }
+
+    // If blob and run on Node.js
+    if (sourceObject.blob && typeof window === 'undefined') {
```
wkozyra95, line 47: fine for now, but I don't want platform-specific code here in general

wkozyra95 (approved):

Issue #1014, opened 2025-03-17
Web renderer crashes when more than 1 instance of smelter is running


noituri:
If we try to run a new smelter instance while another instance is running, CEF will try to connect to the existing instance instead of creating a new CEF browser. Since this case is not handled correctly, it will crash smelter.

It can be fixed by either:
- Providing unique `root_cache_path` for a smelter instance
- Implementing [this method](https://bitbucket.org/chromiumembedded/cef/src/8db54e92c2e200f60f4acdcd08c2348db2e159ba/include/cef_browser_process_handler.h?at=master#cef_browser_process_handler.h-119)

PR #1016, opened 2025-03-19
Fix web renderer crashing when multiple Smelter instances are run

noituri:
When two instances of smelter share the same CEF `root_cache_path`, one instance becomes the main browser window, and the other acts as an opened "tab" in that browser. We don't handle that case at all, hence the crash. Fixed it by giving each instance a unique `root_cache_path` so now each instance works independently. 

Fixes: #1014

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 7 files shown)


wkozyra95:
add to changelog

wkozyra95 (approved):

PR #1018, opened 2025-03-19
Clip border radius

noituri:
Snapshots PR: https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/54

Closes: #1017


wkozyra95 (commented):
can you check if you add large border, set border radius too somthing  large  and set width + height to the same value the resulting shape is a circle?

wkozyra95 (approved):

PR #1020, opened 2025-03-21
Prevent decoder from resetting PTS, add flag `send_eos`

wkazmierczak:
- Prevent resetting pts to zero in audio and video decoder.
 - Add a send_eos flag, allowing control over whether eos is sent to the queue. This is needed for WHIP input to be able to reconnect to the same input without dropping the queue."

### DIFF
--- a/compositor_pipeline/src/pipeline/decoder/video/ffmpeg_h264.rs
+++ b/compositor_pipeline/src/pipeline/decoder/video/ffmpeg_h264.rs
@@ -25,6 +25,7 @@ pub fn start_ffmpeg_decoder_thread(
     chunks_receiver: Receiver<PipelineEvent<EncodedChunk>>,
     frame_sender: Sender<PipelineEvent<Frame>>,
     input_id: InputId,
+    send_eos: bool,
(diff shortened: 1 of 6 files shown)


compositor_pipeline/src/pipeline/decoder/video/ffmpeg_h264.rs
```diff
@@ -179,15 +182,9 @@ fn frame_from_av(
...
-        *pts_offset = Some(-pts)
-    }
-    let pts = original_pts
-        .map(|original_pts| original_pts + pts_offset.unwrap_or(0))
-        .ok_or_else(|| {
-            DecoderFrameConversionError::FrameConversionError("missing pts".to_owned())
-        })?;
+    let pts = decoded.pts().ok_or_else(|| {
+        DecoderFrameConversionError::FrameConversionError("missing pts".to_owned())
+    })?;
```
wkozyra95, line 187:
```suggestion
    let Some(pts) = decoded.pts() else {
        return Err(DecoderFrameConversionError::FrameConversionError("missing pts".to_owned()))
    };
```

PR #1021, opened 2025-03-21
Configurable color space support for different runtime

wkozyra95:
Changes:
- different modes for platforms/runtimes:
  - default: color blending in linear colorspace
  - cpu optimized: color blending in srgb color space with premultiplied alpha
  - webGL backend: color blending in linear colorspace, but requires additional steps to convert srgb<->rgb
- images
  - svg image is rendered in sRGB with alpha premultiplied on top of it.
    - Gpu optimized:  To fix that we need to remove premultiplication in the raw sRGB data and add it back in linear space
    - Cpu optimized: we will use it as ti is for color blending
    - WebGl the same as GPU but requires copying texture via a buffer to reinterpret it sRGB
  - Other images are returned as sRGB without premultiplication so pipeline is simpler
 - Fix wasm logger (we used both tracing_wasm and wasm_log), it should only be one
 - Add `WebView` example to ts sdk


src/snapshot_tests/yuv_tests.rs
```diff
@@ -63,9 +63,9 @@ fn yuv_test_gradient() {
     run_case(
         yuv_case,
         &[
-            91, 0, 0, 255, 106, 6, 5, 255, 161, 0, 0, 255, 169, 3, 3, 255, 204, 0, 0, 255, 210, 2, 2, 255, 238, 0, 0, 255, 242, 2, 1, 255,
-            91, 0, 0, 255, 106, 6, 5, 255, 161, 0, 0, 255, 169, 3, 3, 255, 204, 0, 0, 255, 210, 2, 2, 255, 238, 0, 0, 255, 242, 2, 1, 255,
-        ],
+            88, 0, 0, 255, 103, 7, 7, 255, 159, 0, 0, 255, 167, 4, 3, 255, 204, 0, 0, 255, 210, 2, 2, 255, 238, 0, 0, 255, 242, 2, 1, 255,
```
wkozyra95, line 66: rgb case is correct, so closer to that seems like an improvment

PR #1033, opened 2025-03-27
Fix for not working `chromium_embedding`

### LINKED ISSUE #1031: `chromium_embedding` no longer works
It's caused by V8 being sandboxed, which makes it impossible to pass a pointer as `ArrayBuffer`.

It can be fixed by:
- Copying video frame to `ArrayBuffer` (copies are expensive)
- Disabling sandbox (CEF has to be recompiled)
- Downgrading CEF version

noituri:
It's a temporary fix #1031 

The fix introduces additional frame data copy on embedding frames.


wkozyra95 (approved):

wkozyra95:
backport it to previous version too

wkozyra95 (approved):

PR #1037, opened 2025-03-28
replace schemars fork

wkozyra95:


noituri (approved):

wkozyra95:
@noituri If I remember correctly we only needed this to preserve newlines in json schema, which is not needed anymore since we don't use it for docs

noituri:
> @noituri If I remember correctly we only needed this to preserve newlines in json schema, which is not needed anymore since we don't use it for docs. Is that correct?

That's correct

PR #1038, opened 2025-03-31
[web-wasm] Reduce number of input frame copies

### LINKED ISSUE #1034: Reduce amount of frame copies made in `smelter-web-wasm`

noituri:
Tested on Chrome and Safari. This PR improved performance only on macOS. The approach used here actually makes the performance worse on most platforms (except macOS, probably due to the fact that CPU and GPU share the same memory on  macs)

Closes #1034

### DIFF
--- a/compositor_web/src/wasm/input_uploader.rs
+++ b/compositor_web/src/wasm/input_uploader.rs
@@ -7,7 +7,7 @@ use super::types;
 
 #[derive(Default)]
 pub struct InputUploader {
-    textures: HashMap<InputId, Texture>,
+    textures: HashMap<InputId, Arc<wgpu::Texture>>,
 }
 
 impl InputUploader {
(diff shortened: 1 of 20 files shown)


wkozyra95 (approved):

compositor_web/src/wasm/input_uploader.rs
```diff
@@ -1,37 +1,68 @@
...
+
+    /// On most platforms it's more performant to copy VideoFrame data to CPU
+    /// and then upload it wgpu::Texture. But on macOS using dedicated wgpu copy_external_image_to_texture function
+    /// results in better performance.
+    use_copy_external: bool,
+}
+
+impl Default for InputUploader {
+    fn default() -> Self {
+        let user_agent = match web_sys::window() {
```
wkozyra95, line 24: for simplicity I would do that in js and pass to WASM selected mode when renderer is instantiated

PR #1061, opened 2025-04-14
Add codec negotiation on WHIP output

wkazmierczak:
Add starting encoder based on  WHIP codec negotiation. Not working correctly with Twitch because Twitch requires tracks (H264, Opus) to be added before the negotiation process.


integration_tests/examples/whip_client.rs
```diff
@@ -34,17 +34,13 @@ fn client_code() -> Result<()> {
         "output/output_1/register",
         &json!({
             "type": "whip",
-            "endpoint_url": "https://g.webrtc.live-video.net:4443/v2/offer", // Twitch WHIP endpoint URL
+            "endpoint_url": "https://whip.vdo.ninja", // Video ninja (vdo.ninja/whip) WHIP endpoint URL
```
wkozyra95, line 37: revert the change or set it based on env

compositor_pipeline/src/pipeline/output/whip/payloader.rs
```diff
@@ -222,12 +231,24 @@ impl VideoPayloader {
...
             ),
+            VideoPayloader::VP8 {
+                ref mut payloader,
+                ref mut context,
+            } => payload(
+                payloader,
+                context,
+                chunk,
+                mtu,
+                VIDEO_PAYLOAD_TYPE,
```
wkozyra95, line 242: is this correct? payload type should not be a const? it should be based on SDP

compositor_api/src/types/from_register_output.rs
```diff
@@ -288,6 +304,29 @@ fn maybe_video_options(
     Ok((Some(encoder_options), Some(output_options)))
 }
 
+fn maybe_whip_video_options(
```
wkozyra95, line 307: inline if only used in one place

compositor_api/src/types/register_output.rs
```diff
@@ -74,6 +74,19 @@ pub struct OutputVideoOptions {
...
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct OutputWhipVideoOptions {
+    /// Output resolution in pixels.
+    pub resolution: Resolution,
+    /// Defines when output stream should end if some of the input streams are finished. If output includes both audio and video streams, then EOS needs to be sent on both.
+    pub send_eos_when: Option<OutputEndCondition>,
+    /// Video encoder options.
+    pub encoder: Option<VideoEncoderOptions>,
+    /// Root of a component tree/scene that should be rendered for the output. Use [`update_output` request](../routes.md#update-output) to update this value after registration. [Learn more](../../concept/component.md).
```
wkozyra95, line 86: we don't use that for generation docs sot remove those markdown relative urls

compositor_api/src/types/from_register_output.rs
```diff
@@ -161,38 +170,45 @@ impl TryFrom<WhipOutput> for pipeline::RegisterOutputOptions<output::OutputOptio
             ));
         }
 
+        if video.as_ref().filter(|v| v.encoder.is_some()).is_some() {
```
wkozyra95, line 173: filter on option is wierd, you could `video.and_then(|v| v.encoder).is_some()`

but personally I would prefer something like this:
```
if let Some(OutputWhipVideoOptions{ encoder: Some(encoder), .. })) = &video {
```

wkozyra95 (approved):

PR #1063, opened 2025-04-15
Refactor benchmark code

wkozyra95:
Prepare the codebase so we can define benchmark suites in code. Actual definitions will land in a separate PR

Regression:
- It's not possible to iterate over args, just maxmize or specific value
- Remove CSV export in favor or JSON

### DIFF
--- a/integration_tests/src/bin/benchmark/benchmark_pass.rs
+++ b/integration_tests/src/bin/benchmark/benchmark_pass.rs
@@ -1,0 +1,339 @@
+use anyhow::{anyhow, Result};
+use std::{
+    path::{Path, PathBuf},
+    sync::{Arc, Mutex},
+    thread,
+    time::{Duration, Instant},
+};
+
+use compositor_pipeline::{
+    error::{RegisterInputError, RegisterOutputError},
+    pipeline::{
+        self,
+        encoder::{
+            ffmpeg_h264::{self, EncoderPreset},
+            VideoEncoderOptions,
+        },
+        input::{
+            mp4::{Mp4Options, Source},
+            InputInitInfo, InputOptions, RawDataInputOptions,
+        },
+        output::{EncodedDataOutputOptions, RawDataOutputOptions, RawVideoOptions},
+        EncodedChunkKind, EncoderOutputEvent, GraphicsContext, OutputVideoOptions,
+        PipelineOutputEndCondition, RegisterInputOptions, RegisterOutputOptions,
+    },
+    queue::{PipelineEvent, QueueInputOptions},
+    Pipeline,
+};
+use compositor_render::{scene::Component, Frame, InputId, OutputId, YuvPlanes};
+use crossbeam_channel::{Receiver, Sender, TryRecvError};
+use tracing::debug;
+
+use crate::{
+    args::Resolution, benchmark::EncoderOptions, scenes::SceneContext,
+    utils::benchmark_pipeline_options,
+};
+
+#[derive(Debug, Clone)]
+pub enum InputFile {
+    Mp4(PathBuf),
+    Raw(RawInputFile),
+}
+
+#[derive(Debug, Clone)]
+pub struct RawInputFile {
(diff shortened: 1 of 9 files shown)


integration_tests/src/bin/benchmark/benchmark_pass.rs
```diff
@@ -0,0 +1,340 @@
...
+                thread::spawn(move || {
+                    debug!(?index, "start drain in warm up mode");
+                    while start_time.elapsed() < warm_up_time {
+                        if let Err(TryRecvError::Empty) = receiver.try_receive() {
+                            thread::sleep(Duration::from_millis(1));
+                        }
+                    }
+
+                    debug!(?index, "start drain in measure mode");
+                    let mut max_pts: Duration = Duration::ZERO;
```
wkozyra95, line 171: it needs to be based on duration, previous version counted frames which is correct for raw input, but not for encoded one

integration_tests/src/bin/benchmark/benchmark_pass.rs
```diff
@@ -0,0 +1,340 @@
...
+    ) -> Result<Sender<PipelineEvent<Frame>>, RegisterInputError> {
+        let input = Pipeline::register_raw_data_input(
+            pipeline,
+            input_id,
+            RawDataInputOptions {
+                video: true,
+                audio: false,
+            },
+            QueueInputOptions {
+                offset: None,
```
wkozyra95, line 304: offset needs to be resolved based on the first frame, we don't want to have bunch of frames already in buffer when starting

integration_tests/src/bin/benchmark/benchmark_pass.rs
jerzywilczek: I don't understand why.

integration_tests/src/bin/benchmark/benchmark_pass.rs
```diff
@@ -0,0 +1,340 @@
...
+                resolution: compositor_render::Resolution {
+                    width: input.resolution.width,
+                    height: input.resolution.height,
+                },
+                pts: Duration::from_secs_f64(counter as f64 / input.framerate),
+            };
+            counter += 1;
+
+            for sender in &senders {
+                if sender.send(PipelineEvent::Data(frame.clone())).is_err() {
```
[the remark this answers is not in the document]
jerzywilczek: I don't understand what this is about

jerzywilczek (approved):
Wow, very nice. Thanks for the cleanup sorry for the mess ;(

noituri (approved):
Looks alright

integration_tests/src/bin/benchmark/benchmark_pass.rs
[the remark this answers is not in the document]
noituri: I'm assuming it's related to https://github.com/software-mansion/smelter/blob/54d2400da7d22887b7065ea859f1a9f98e78906a/compositor_pipeline/src/pipeline/input.rs#L124-L134

If the input start is delayed too much, the channels will fill up and the send will block. Although I'm not sure why's that a problem since it's run on separate thread

integration_tests/src/bin/benchmark/benchmark_pass.rs
wkozyra95: if we are getting encoded chunks we don't know how many frames there are

integration_tests/src/bin/benchmark/benchmark_pass.rs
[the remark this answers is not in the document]
wkozyra95: the problem is that we have one sender thread that sends to all inputs in the loop, so if one of the channels gets filled it will not send data to the other channel on time

integration_tests/src/bin/benchmark/benchmark_pass.rs
jerzywilczek: oh, obviously. I've been sitting in our encoder for so long I forgot those can be split into chunks

PR #1091, opened 2025-05-09
[ts-sdk] Refactor WASM types + add HTMLVideoElement as input

wkozyra95:
- Seprate binding layer from internal implementation
- Handle a case where `frame` is `HTMLVideoElement` (there is still some issue with that, output color space seems to be incorrect, but I'll verify that in follow up PRs that are using that feature)

### DIFF
--- a/compositor_web/src/types.rs
+++ b/compositor_web/src/types.rs
@@ -1,0 +1,144 @@
+use std::time::Duration;
+
+use compositor_render::InputId;
+use tracing::error;
+use wasm_bindgen::prelude::*;
+
+use crate::wasm::{self, InputFrameKind};
+
+#[wasm_bindgen(typescript_custom_section)]
+const TS_DEFINITIONS: &'static str = r#"
+export interface InputFrame {
+  inputId: string;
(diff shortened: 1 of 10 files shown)


compositor_web/src/wasm/input.rs
wkozyra95, line 1: uploader/downloader name does not make sense anymore because in some cases it's copy CPU->GPU, GPU->GPU, GPU->CPU->GPU

compositor_web/src/types.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+pub trait ObjectExt {
+    fn set<T: Into<JsValue>>(&self, key: &str, value: T);
+}
+
+impl ObjectExt for js_sys::Object {
+    fn set<T: Into<JsValue>>(&self, key: &str, value: T) {
+        js_sys::Reflect::set(self, &JsValue::from_str(key), &value.into()).unwrap();
```
wkozyra95, line 142: unwrap because set on object should not fail normally, especially if in most cases this is an object we just created

PR #1093, opened 2025-05-09
Add VP9 encoder

wkazmierczak:


compositor_api/src/types/from_register_output.rs
```diff
@@ -498,6 +513,54 @@ fn maybe_video_options(
...
+
+    let output_options = pipeline::OutputVideoOptions {
+        initial: options.initial.try_into()?,
+        end_condition: options.send_eos_when.unwrap_or_default().try_into()?,
+    };
+
+    Ok((Some(encoder_options), Some(output_options)))
+}
+
+fn maybe_video_options_h264_only(
```
wkazmierczak, line 532: Wondering if it's a good time to split `VideoEncoderOptions` into `VideoRtmpClientEncoderOptions`, `VideoMp4EncoderOptions`, etc. However, I think doing so would require a major refactor in `from_register_output.rs` and `register_output.rs`, so for now I did it like this

PR #1098, opened 2025-05-15
Add VP8, VP9 to `ts-sdk`

wkazmierczak:


ts/smelter/src/types/output/whip.ts
```diff
@@ -38,20 +57,23 @@ export type WhipAudioOptions = {
...
+       */
+      preset?: Api.OpusEncoderPreset;
+      /**
+       * (**default=`48000`**) Sample rate. Allowed values: [8000, 16000, 24000, 48000].
+       */
+      sampleRate?: number;
+    }
+  | {
+      type: 'any';
+    };
```
wkozyra95, line 79: I'm wondering whether something like this wouldn be better.
```suggestion
  | 'any';
```

any opinion on that?

wkozyra95 (approved):

PR #1111, opened 2025-05-23
Keep gif animation state between scene updates

noituri:
Snapshot PR: https://github.com/membraneframework-labs/video_compositor_snapshot_tests/pull/59


wkozyra95 (commented):
Currentlly, we do not have any relation between layout and render nodes. State should be stored in layouts and after building render nodes we can't reverse that anymore.

I'm not against chaining our current apporach, but not as workaround for a small issue. I we want to change how relation between nodes and layouts work we can do that for everything, but not just for that one thing.

So, to solve this issue with our old patterns you would have to keep gif progress in the StatefulImageComponent and when you update scene pass starting offset of the gif

PR #1116, opened 2025-05-30
[ts-sdk] Create `smelter-web-client`

noituri:


ts/smelter-web-client/LICENSE
```diff
@@ -0,0 +1,35 @@
+This Agreement governs the use of Software Mansion S.A. software suite which encompasses two distinct processing capabilities:
```
wkozyra95: LICENSE MIT, we do not include the engine in this package

ts/smelter-web-client/README.md
```diff
@@ -0,0 +1,54 @@
...
+function ExampleApp() {
+  return (
+    <View>
+      <Text style={{ fontSize: 20 }}>Hello world</Text>
+    </View>
+  );
+}
+
+async function run() {
+  const smelter = new Smelter();
```
wkozyra95: This example does not make sense for browser

ts/smelter-web-client/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@swmansion/smelter-web-client",
+  "version": "0.2.1",
+  "author": "Software Mansion <contact@swmansion.com>",
+  "license": "SEE LICENSE IN ./LICENSE",
```
wkozyra95: MIT

ts/smelter-web-client/src/smelter/live.ts
```diff
@@ -0,0 +1,107 @@
...
+  public async init(): Promise<void> {
+    await this.coreSmelter.init();
+  }
+
+  public async registerOutput(
+    outputId: string,
+    root: ReactElement,
+    request: RegisterOutput
+  ): Promise<void> {
+    await this.coreSmelter.registerOutput(outputId, root, request);
```
wkozyra95: Add returns everywhere, return type is not typed, but we should at least return the value. e.g., it is not possible to use WHIP input because we don't know the bearer token.

Node.js has the same problem, we patched that on last relase, but not on master, so for now just add returns

ts/smelter-web-client/src/ws.ts
```diff
@@ -0,0 +1,68 @@
...
+    this.url = url;
+    this.listeners = new Set();
+  }
+
+  public async connect(logger: Logger): Promise<void> {
+    const ws = new WebSocket(this.url);
+
+    let connected = false;
+    await new Promise<void>((res, rej) => {
+      ws.addEventListener('error', (err: any) => {
```
wkozyra95, line 19: if you based it on node implementation make sure if this is all correct. WebSocket in node and browser does not have to behave the same.

wkozyra95 (approved):

PR #1118, opened 2025-05-30
Add format autodetection for images

wkazmierczak:


ts/smelter/src/types/resource.ts
```diff
@@ -2,32 +2,11 @@ import type * as Api from '../api.js';
...
-    };
-
-export type ImageAssetType = RegisterImage['assetType'];
-
-export const imageAssetTypes: ImageAssetType[] = ['png', 'jpeg', 'svg', 'gif'];
+export type RegisterImage = {
+  url?: string;
+  serverPath?: string;
+  resolution?: Api.Resolution;
+};
```
wkazmierczak, line 9: The `resolution` appears in all `image_types`, but it is ignored in all of them apart from svg. I have doubts that it is the best solution, but I am not sure how to do it better. I've tried something like this
`export type BaseImage = {
  url?: string;
  serverPath?: string;
};

export type SvgImage = BaseImage & {
  url: `${string}.svg`;
  resolution?: Api.Resolution;
};

export type NonSvgImage = BaseImage & {
  url?: Exclude<string, `${string}.svg`>;
  resolution?: never;
};

export type RegisterImage = SvgImage | NonSvgImage;`
But the inline TypeScript error message wasn’t descriptive at all

wkozyra95 (approved):

PR #1120, opened 2025-06-03
[ts-sdk] Provide address to smelter instance via URL

noituri:


ts/smelter-node/src/manager/existingInstance.ts
```diff
@@ -62,4 +69,14 @@ class ExistingInstanceManager implements SmelterManager {
   }
 }
 
+function urlToString(url: URL): string {
```
wkozyra95: why are you converting it to string if you can use URL with both fetch and websockets? It is converting back and fourth

ts/smelter-node/src/manager/existingInstance.ts
```diff
@@ -62,4 +69,14 @@ class ExistingInstanceManager implements SmelterManager {
...
+  let href = url.href;
+  // Removes trailing slash
+  if (href.slice(-1) === '/') {
```
wkozyra95: URL provides a constructor that allows building url from base url. I would use that instead of string operations

ts/smelter-node/src/manager/existingInstance.ts
```diff
@@ -10,26 +10,33 @@ import { retry, sleep } from '../utils';
...
+      url = new URL(opts.url);
+    }
+
+    if (url.protocol !== 'http:' && url.protocol !== 'https:') {
+      throw new Error('Expected url to use either http or https protocol');
+    }
+
+    this.url = url;
+
+    const wsUrl = new URL('/ws', url);
```
wkozyra95:
```suggestion
    const wsUrl = new URL('ws', url);
```
your coude would change url `https://test.com/test` to `https://test.com/ws` after my suggestion it would be `https://test.com/test/ws`

on regular deployment it does not matter, but someone can have server deployed in private network and proxy trafic on some route

PR #1121, opened 2025-06-04
Add `resolution` field to `ImageComponent`

wkazmierczak:


compositor_api/src/resource/image.rs
```diff
@@ -62,12 +62,10 @@ impl TryFrom<ImageSpec> for compositor_render::RendererSpec {
             ImageSpec::Svg {
                 url,
                 path,
-                resolution,
+                resolution: _resolution,
```
noituri:
```suggestion
                ..
```

compositor_render/src/transformations/image/svg_image.rs
```diff
@@ -77,7 +72,7 @@ impl SvgAsset {
             }
             (SvgRenderer::CpuOptimized, NodeTextureState::CpuOptimized { texture, .. }) => {
                 // input is already in sRGB with pre-multiplied alpha
-                render_to_texture(ctx, &self.tree.0, texture.texture(), resolution);
+                render_to_texture(ctx, &self.tree.0, texture.texture(), state.resolution());
             }
             (SvgRenderer::WebGl(renderer), NodeTextureState::WebGl { texture, .. }) => {
                 renderer.render(ctx, &self.tree.0, texture, resolution)
```
noituri: You're using `state.resolution()` only for cpu optimized rendering but for other rendering types you rely on `SvgAsset::resolution`. Was that intentional?

noituri (approved):
LGTM

compositor_api/src/video/component.rs
```diff
@@ -253,6 +253,14 @@ pub struct Image {
 
     /// Id of an image. It identifies an image registered using a [`register image`](../routes.md#register-image) request.
     pub image_id: RendererId,
+
+    /// Width of the texture onto which the image will be rendered. If neither width nor height is specified, the texture will default to the size of the image.
```
wkozyra95: the same as docs

compositor_api/src/video/component.rs
```diff
@@ -253,6 +253,14 @@ pub struct Image {
...
+    /// The aspect ratio of the image will be maintained if only one dimension (width or height) is provided. If both dimensions are specified, the aspect ratio will be as provided.
+    pub width: Option<usize>,
```
wkozyra95: I think we use floats in other components

compositor_render/src/state/node.rs
```diff
@@ -79,9 +79,12 @@ impl RenderNode {
...
             }
-            scene::NodeParams::Image { image, start_pts } => {
-                Self::new_image_node(ctx, image, start_pts)
-            }
+            scene::NodeParams::Image(image_params) => Self::new_image_node(
+                ctx,
+                image_params.image,
+                image_params.start_pts,
+                image_params.resolution,
+            ),
```
wkozyra95:
```suggestion
            scene::NodeParams::Image(params) => Self::new_image_node(ctx, params),
```

compositor_render/src/state/node.rs
```diff
@@ -145,8 +148,14 @@ impl RenderNode {
...
-    pub(super) fn new_image_node(ctx: &RenderCtx, image: Image, start_pts: Duration) -> Self {
-        let node = InnerRenderNode::Image(ImageNode::new(ctx.wgpu_ctx, image, start_pts));
+    pub(super) fn new_image_node(
+        ctx: &RenderCtx,
+        image: Image,
+        start_pts: Duration,
+        resolution: Resolution,
+    ) -> Self {
+        let node =
+            InnerRenderNode::Image(ImageNode::new(ctx.wgpu_ctx, image, start_pts, resolution));
```
wkozyra95: same as above

compositor_render/src/scene/scene_state.rs
```diff
@@ -185,10 +185,11 @@ impl IntermediateNode {
                     .collect::<Result<_, _>>()?,
             }),
             IntermediateNode::Image(image) => Ok(Node {
-                params: NodeParams::Image {
+                params: NodeParams::Image(ImageRenderParams {
```
wkozyra95: this is shared logic, keep component specific code to minimum

compositor_render/src/scene.rs
```diff
@@ -92,7 +91,7 @@ pub(crate) enum NodeParams {
     InputStream(InputId),
     Shader(ShaderComponentParams, Arc<Shader>),
     Web(Vec<ComponentId>, Arc<WebRenderer>),
-    Image(ImageRenderParams),
+    Image(StatefulImageComponent),
```
wkozyra95: It feels like a very bad idea, you are forcing here cloning statefull components, which can easily lead to cloning that by mistake and somehow operating on cloned state by mistake

wkozyra95: ok maybe let's leave it for, now
layout has the same problem (although it is necessary there, but not here)

PR #1126, opened 2025-06-10
[ts-sdk] Add `width`, `height` to `Image` component

wkazmierczak:


compositor_render/src/transformations/image/svg_image.rs
```diff
@@ -72,7 +72,7 @@ impl SvgAsset {
             }
             (SvgRenderer::CpuOptimized, NodeTextureState::CpuOptimized { texture, .. }) => {
                 // input is already in sRGB with pre-multiplied alpha
-                render_to_texture(ctx, &self.tree.0, texture.texture(), resolution);
+                render_to_texture(ctx, &self.tree.0, texture.texture(), state.resolution());
```
wkozyra95: any reason for that change?

wkazmierczak: Just leftover from merge or previous version, fixed in following commit

PR #1135, opened 2025-06-17
[ts-sdk] Synchronize calls to smelter

noituri:
Closes https://github.com/software-mansion/smelter/issues/1133

### DIFF
--- a/ts/smelter-core/src/utils.ts
+++ b/ts/smelter-core/src/utils.ts
@@ -61,3 +61,79 @@ export async function sleep(timeoutMs: number): Promise<void> {
     }, timeoutMs);
   });
 }
+
+export class StateGuard {
+  private state:
+    | {
+        type: 'blocked';
+        blocking_promise: Promise<void>;
+        nonblocking_promises?: Set<Promise<void>>;
+      }
+    | { type: 'running'; promises: Set<Promise<void>> }
+    | { type: 'finished' };
+
+  public constructor() {
+    this.state = { type: 'finished' };
+  }
+
+  public async runBlocking<T>(fn: () => Promise<T>): Promise<T> {
+    const [blocking_promise, release] = this.createGuardPromise();
+    while (this.state.type !== 'finished') {
+      if (this.state.type === 'blocked') {
+        await Promise.allSettled(this.state?.nonblocking_promises ?? []);
+        await this.state.blocking_promise;
+      } else if (this.state.type === 'running') {
+        this.state = {
+          type: 'blocked',
+          blocking_promise,
+          nonblocking_promises: this.state.promises,
+        };
+        await Promise.allSettled(this.state?.nonblocking_promises ?? []);
+      }
+    }
+
+    try {
+      this.state = { type: 'blocked', blocking_promise };
+      return await fn();
+    } finally {
+      this.state = { type: 'finished' };
+      release();
+    }
+  }
+
+  public async run<T>(fn: () => Promise<T>): Promise<T> {
+    while (this.state.type === 'blocked') {
+      await this.state.blocking_promise;
+    }
+
+    const [promise, release] = this.createGuardPromise();
+
+    try {
+      if (this.state.type === 'running') {
+        this.state.promises.add(promise);
+      } else if (this.state.type === 'finished') {
+        this.state = { type: 'running', promises: new Set([promise]) };
+      }
+
+      return await fn();
+    } finally {
+      if (this.state.type === 'running') {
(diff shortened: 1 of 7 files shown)


ts/smelter-core/src/utils.ts
```diff
@@ -61,3 +61,71 @@ export async function sleep(timeoutMs: number): Promise<void> {
     }, timeoutMs);
   });
 }
+
+export class PromiseScheduler {
```
wkozyra95: StateGuard?

ts/smelter-core/src/utils.ts
```diff
@@ -61,3 +61,71 @@ export async function sleep(timeoutMs: number): Promise<void> {
...
+  private state:
+    | { type: 'blocked'; by: Promise<void> }
```
wkozyra95:
```suggestion
    | { type: 'blocked'; promise: Promise<void> }
```

ts/smelter-core/src/utils.ts
```diff
@@ -61,3 +61,71 @@ export async function sleep(timeoutMs: number): Promise<void> {
...
+      }
+      if (this.state.type === 'running') {
+        await Promise.allSettled(this.state.promises);
+      }
+    }
+
+    let resFn: (() => void) | undefined;
+    const promise = new Promise<void>(res => {
+      resFn = res;
+    });
```
wkozyra95: You can write helper that would be used
```
const [promise, release] = createGuardPromise();
```

ts/smelter-core/src/utils.ts
```diff
@@ -61,3 +61,79 @@ export async function sleep(timeoutMs: number): Promise<void> {
...
+
+export class StateGuard {
+  private state:
+    | {
+        type: 'blocked';
+        blocking_promise: Promise<void>;
+        nonblocking_promises?: Set<Promise<void>>;
+      }
+    | { type: 'running'; promises: Set<Promise<void>> }
+    | { type: 'finished' };
```
wkozyra95:
```suggestion
        type: 'unique';
        blocking_promise: Promise<void>;
        nonblocking_promises?: Set<Promise<void>>;
      }
    | { type: 'shared'; promises: Set<Promise<void>> }
    | { type: 'open' };
```

ts/smelter-core/src/utils.ts
```diff
@@ -61,3 +61,79 @@ export async function sleep(timeoutMs: number): Promise<void> {
...
+
+  public constructor() {
+    this.state = { type: 'finished' };
+  }
+
+  public async runBlocking<T>(fn: () => Promise<T>): Promise<T> {
+    const [blocking_promise, release] = this.createGuardPromise();
```
wkozyra95:
```suggestion
    const [promise, promiseResolveFn] = this.createGuardPromise();
```
I suggest sticking to some pattern, e.g.  first element any name second element add `ResolveFn` to name of the first

ts/smelter-core/src/utils.ts
```diff
@@ -61,3 +61,79 @@ export async function sleep(timeoutMs: number): Promise<void> {
...
+    while (this.state.type !== 'finished') {
+      if (this.state.type === 'blocked') {
+        await Promise.allSettled(this.state?.nonblocking_promises ?? []);
+        await this.state.blocking_promise;
```
wkozyra95: use camelCase

ts/smelter-core/src/utils.ts
```diff
@@ -61,3 +61,79 @@ export async function sleep(timeoutMs: number): Promise<void> {
...
+      this.state = { type: 'blocked', blocking_promise };
+      return await fn();
+    } finally {
+      this.state = { type: 'finished' };
+      release();
+    }
+  }
+
+  public async run<T>(fn: () => Promise<T>): Promise<T> {
+    while (this.state.type === 'blocked') {
```
wkozyra95: If there is a blocking call waiting, then you should wait too

ts/smelter-core/src/utils.ts
```diff
@@ -61,3 +61,73 @@ export async function sleep(timeoutMs: number): Promise<void> {
...
+        await this.state.promise;
+      } else if (this.state.type === 'shared') {
+        const unblockingPromises = this.state.promises;
+        this.state = { type: 'unique', promise };
+        await Promise.allSettled(unblockingPromises);
+      }
+    }
+
+    try {
+      this.state = { type: 'unique', promise };
```
wkozyra95: outside of the try catch

PR #1136, opened 2025-06-17
Return instance config on `/status` endpoint

wkazmierczak:


src/routes/status.rs
```diff
@@ -0,0 +1,123 @@
...
+    output_framerate: Framerate,
+    run_late_scheduled_events: bool,
+    never_drop_output_frames: bool,
+}
+
+#[derive(Serialize, Deserialize)]
+struct Framerate {
+    num: u32,
+    den: u32,
+}
```
noituri: You don't need to derive `Deserialize` here

compositor_render/src/types.rs
```diff
@@ -17,6 +17,16 @@ pub enum RenderingMode {
...
+impl Display for RenderingMode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::GpuOptimized => write!(f, "GPU optimized"),
+            Self::CpuOptimized => write!(f, "CPU optimized"),
+            Self::WebGl => write!(f, "WebGL"),
+        }
+    }
+}
+
```
noituri: I would move the `match` to `status_handler` (similarly to how you handled inputs and outputs) instead of implementing `Display`

ts/smelter-web-client/src/manager.ts
```diff
@@ -39,10 +43,24 @@ class RemoteInstanceManager implements SmelterManager {
...
-      return await this.sendRequest({
+      let status = (await this.sendRequest({
         method: 'GET',
         route: '/status',
-      });
+      })) as StatusResponse;
+
+      const expectedAheadOfTime = opts.aheadOfTimeProcessing;
+
+      console.log(expectedAheadOfTime);
```
noituri:
```suggestion
```

ts/smelter-node/src/manager/existingInstance.ts
```diff
@@ -40,14 +44,34 @@ class ExistingInstanceManager implements SmelterManager {
...
+
+      for (const [key, expected] of Object.entries(expectedConfig)) {
+        const actual = actualConfig[key as keyof typeof actualConfig];
+        if (actual !== expected) {
+          opts.logger.warn(
+            {
+              expected: expected === undefined ? 'undefined' : expected,
+              actual: actual === undefined ? 'undefined' : actual,
+            },
+            `Mismatch in ${key}`
```
wkozyra95:
```suggestion
            `Mismatch between instance config and SDK.`
```

ts/smelter-node/src/manager/existingInstance.ts
```diff
@@ -40,14 +44,34 @@ class ExistingInstanceManager implements SmelterManager {
...
+
+      const actualConfig = {
+        ahead_of_time_processing: status.queue_options?.ahead_of_time_processing,
+      };
+
+      for (const [key, expected] of Object.entries(expectedConfig)) {
+        const actual = actualConfig[key as keyof typeof actualConfig];
+        if (actual !== expected) {
+          opts.logger.warn(
+            {
```
wkozyra95, line 60:
```suggestion
            {
              key,
```

ts/smelter-node/src/manager/locallySpawnedInstance.ts
```diff
@@ -82,10 +91,39 @@ class LocallySpawnedInstanceManager implements SmelterManager {
 
     await retry(async () => {
       await sleep(500);
-      return await this.sendRequest({
+
+      let status = (await this.sendRequest({
```
wkozyra95: wrap request in some function, convert imiedietlly to cammel case types and write all logic while operating on those values

ts/smelter-node/src/manager/existingInstance.ts
```diff
@@ -40,14 +44,35 @@ class ExistingInstanceManager implements SmelterManager {
   }
 
   public async setupInstance(opts: SetupInstanceOptions): Promise<void> {
-    // TODO: verify if options match
-    // https://github.com/software-mansion/smelter/issues/877
     await retry(async () => {
       await sleep(500);
-      return await this.sendRequest({
+      let status = (await this.sendRequest({
```
wkozyra95: the whole point for extracting the request to function is to use it where possible

ts/smelter-web-client/src/manager.ts
```diff
@@ -39,10 +43,23 @@ class RemoteInstanceManager implements SmelterManager {
   public async setupInstance(opts: SetupInstanceOptions): Promise<void> {
     await retry(async () => {
       await sleep(500);
-      return await this.sendRequest({
+      let status = (await this.sendRequest({
```
wkozyra95: it is only used once in this packaged, but I would extract it to separate function too

wkozyra95 (changes requested):

PR #1158, opened 2025-07-14
Add `HLS` input

noituri:
This PR adds hls input with discontinuity handling.

Discontinuity is detected when dts difference is larger than 10s or dts is not monotonic.

This works for most streams but sometimes I found that the pts can fall behind the queue which causes the frames to get dropped. It mostly happens after discontinuity but I've seen it happen also without any detected discontinuities. 

I handle it by adjusting the pts if I notice that there is not enough packets in a buffer (they get quickly disposed by the queue). This solution works quite well compared to other players I've tested (ffplay & mpv).

### DIFF
--- a/ts/smelter/src/types/input.ts
+++ b/ts/smelter/src/types/input.ts
@@ -33,7 +33,6 @@ export type RegisterRtpInput = {
 };
 
 export type Mp4VideoDecoder = 'ffmpeg_h264' | 'vulkan_h264';
-
 export type RegisterMp4Input = {
   /**
    * URL of the MP4 file.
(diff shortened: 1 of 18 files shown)


ts/smelter/src/types/input.ts
```diff
@@ -67,6 +66,28 @@ export type RegisterMp4Input = {
...
+  required?: boolean | null;
+  /**
+   * Offset in milliseconds relative to the pipeline start (start request). If offset is
+   * not defined then stream is synchronized based on the first frames delivery time.
+   */
+  offsetMs?: number | null;
+  /**
+   * (**default=`ffmpeg_h264`**) The decoder to use for decoding video.
+   */
+  videoDecoder?: HlsVideoDecoder | null;
```
wkozyra95: remove that, we need some way to define what decoder should be used for HLS and mp4, but it won't this way

wkozyra95 (approved):
just one comment for the API, I did not check the core part in detail, but I'm in favor of merging it as soon as possible, so I can rebase on top of that in the PR with input refactor

PR #1163, opened 2025-07-16
Change sum_scale behaviour to be smoother

### LINKED ISSUE #1155: Improve SumScale behavior
https://github.com/software-mansion/smelter/blob/master/compositor_pipeline/src/audio_mixer/mix.rs#L34

Currently SumScale when mixing single 20ms chunk of audio
- find max sample value in chunk
- if it is larger than the max value then scale all samples in the chunk by a factor that will bring the highest value into valid range

Feel free to consider other alternatives, but I suggest some variation of following algorithm:

Approach 1
- start with scaling factor 1
- for each chunk
  - if highest sample value is larger than e.g. 0.8 of max value then lower scaling factor by e.g. 0.01 (factor - 0.01 or factor * 0.99 no strong opinion)
  - if max sample value is consistently below 0.5 of max value start increasing scaling factor in small increments
  - multiply each sample by scaling factor

even better if scaling factor increases, the change is applied linearly on that chunk and not just samples multiplied by the same value

Approach 2:

The same logic as currently, but instead of taking max sample from current chunk, store max value for each chunk for the last e.g minute and use that value.

The problem here is that jumps might be larger here and I'm not sure if it would be audible.

JBRS307:
Closes #1155


compositor_pipeline/src/audio_mixer/mix.rs
```diff
@@ -9,81 +9,143 @@ use super::{
...
-    input_samples: &HashMap<InputId, Vec<(i16, i16)>>,
-    samples_count: usize,
-    inputs: I,
-) -> Vec<(i64, i64)> {
-    let mut summed_samples = vec![(0i64, 0i64); samples_count];
-
-    for input_params in inputs {
-        let Some(input_samples) = input_samples.get(&input_params.input_id) else {
-            continue;
+    fn sum_clip(&self, summed_samples: Vec<(i64, i64)>) -> Vec<(i16, i16)> {
```
wkozyra95: The name implies that it will sum samples, but it receives summed samples. Maybe clip_summed_samples/scale_summed_samples, or  clip_samples/scales_samples

compositor_pipeline/src/audio_mixer/mix.rs
```diff
@@ -9,81 +9,143 @@ use super::{
...
+            .into_iter()
+            .map(|(l, r)| (l as f64, r as f64))
+            .collect();
+
+        // Assumes that summed samples is not empty (therefore unwrap is safe)
+        let max_sample = summed_samples
+            .iter()
+            .map(|(l, r)| f64::max(l.abs(), r.abs()))
+            .reduce(f64::max)
+            .expect("Assumes that summed samples is not empty");
```
wkozyra95: If there is an easy way to provide some value I would use  `unwrap_or_else` that return the value and maybe prints an error/warning

compositor_pipeline/src/audio_mixer/mix.rs
```diff
@@ -9,81 +9,143 @@ use super::{
...
+
+        let new_scaling_factor = if max_sample * self.scaling_factor > self.scaling_threshold {
+            self.scaling_factor - self.scaling_increment
+        } else if (self.scaling_factor < 1.0f64)
+            && (max_sample * self.scaling_factor < self.scaling_threshold)
```
wkozyra95: I would extract long comparison to variable e.g. is_above_treshold and move it above the if statement

maybe for consistency it is worth to move the previous comparisong to variable too

compositor_pipeline/src/audio_mixer/mix.rs
```diff
@@ -9,81 +9,143 @@ use super::{
...
+            self.scaling_factor
         };
-        for (sum, sample) in summed_samples.iter_mut().zip(input_samples.iter()) {
-            sum.0 += (sample.0 as f64 * input_params.volume as f64) as i64;
-            sum.1 += (sample.1 as f64 * input_params.volume as f64) as i64;
-        }
+        debug!("Max abs value: {max_sample}");
+        debug!("Max abs value scaled: {}", max_sample * new_scaling_factor);
+        debug!("Old scaling factor: {}", self.scaling_factor);
+        debug!("New scaling factor: {new_scaling_factor}");
```
wkozyra95: - trace instead of debug
- move everything to one log
- use syntax like this:
```
trace!(
  max_value,
  old_scalling_factor=self.scalling_factor
  new_calling_factor,
  "Processing audio sample"
}
```
- I don't have strong opinion on that, but I would drop "Max abs value scaled" since it can be calculated from remaining values that are printed

wkozyra95 (commented):
I'm fine with addressing that in a follow-up PR(or here, up to you), but the behavior for increasing the scaling factor should not be triggered immediately.

I would suggest using a smaller increment for increase, and/or start increasing after a few seconds of volume being under that threshold

compositor_pipeline/src/audio_mixer/mix.rs
```diff
@@ -9,81 +9,162 @@ use super::{
...
-                        clip_to_i16((l as f64 * scaling_factor) as i64),
-                        clip_to_i16((r as f64 * scaling_factor) as i64),
-                    )
-                })
-                .collect()
+    /// When max audio sample is below this value scaling factor is increased
+    vol_up_threshold: f64,
+
+    /// Increment by which scaling factor is decreased each chunk while lowering volume
+    vol_down_interval: f64,
```
wkozyra95: interval->increment

compositor_pipeline/src/audio_mixer.rs
```diff
@@ -68,18 +67,29 @@ impl AudioMixer {
         self.0.lock().unwrap().update_output(output_id, audio)
     }
 }
+const VOL_DOWN_THRESHOLD: f64 = i16::MAX as f64;
+const VOL_UP_THRESHOLD: f64 = 0.7f64 * i16::MAX as f64;
```
wkozyra95: Please don't use f64 suffix in the literal definition. Initially, I was fine with that, but it is really hard to quickly parse those values when quicklly looking through code

compositor_pipeline/src/audio_mixer/mix.rs
```diff
@@ -9,81 +9,162 @@ use super::{
...
+
+    /// Interval by wich svaling factor is inreased each chunk while rising volume
```
wkozyra95, line 32:
```suggestion
    /// The increment by which the scaling factor is increased for each chunk while rising volume.
```

compositor_pipeline/src/audio_mixer/mix.rs
```diff
@@ -9,81 +9,169 @@ use super::{
...
+        let interpolation_interval =
+            (new_scaling_factor - self.scaling_factor) / summed_samples.len() as f64;
+        let mut current_scaling_factor = self.scaling_factor;
+
+        let summed_samples: Vec<(f64, f64)> = summed_samples
+            .into_iter()
+            .map(|(mut l, mut r)| {
+                l *= current_scaling_factor;
+                r *= current_scaling_factor;
+                current_scaling_factor += interpolation_interval;
```
wkozyra95, line 132: Let's leave it, but for the future, I would recommend avoiding patterns like this:
- Modifying the incoming value might be misleading, you don't know if it's a reference or not, and behavior is not obvious
- interpolation_interval will cause compounding error when adding, the difference is to small to matter and is limited to a single chunk, but nevertheless I would recommend avoiding it and instead use

```
current_scaling_factor += (new_scaling_factor - self.scaling_factor) * (sample_index  / summed_samples.len())
```

PR #1164, opened 2025-07-17
[web-wasm] Add WHEP client

wkazmierczak:
Add WHEP client to `web-wasm`. 

Now register a WHEP input by providing:
- bearer token for authentication
- WHEP endpoint URL to establish the media connection

### DIFF
--- a/ts/smelter-web-wasm/src/mainContext/input/whep.ts
+++ b/ts/smelter-web-wasm/src/mainContext/input/whep.ts
@@ -1,3 +1,4 @@
+import { sleep } from '../../utils';
 import type { Input, RegisterInputResult } from '../input';
 import type { InstanceContext } from '../instance';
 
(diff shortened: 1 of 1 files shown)


ts/smelter-web-wasm/src/mainContext/input/whep.ts
```diff
@@ -0,0 +1,225 @@
...
+  };
+}
+
+async function waitForTracks(tracks: MediaStreamTrack[]): Promise<void> {
+  const maxWait = 1000;
+  const pollInterval = 100;
+  const start = Date.now();
+
+  while (tracks.length < 1 && Date.now() - start < maxWait) {
+    await new Promise(res => setTimeout(res, pollInterval));
```
noituri: You can use `sleep` function from `utils.ts` instead

ts/smelter-web-wasm/src/mainContext/input/whep.ts
```diff
@@ -0,0 +1,226 @@
...
+      transferable,
+    ],
+  };
+}
+
+async function waitForTracks(tracks: MediaStreamTrack[]): Promise<void> {
+  const maxWait = 1000;
+  const pollInterval = 100;
+  const start = Date.now();
+
```
wkozyra95:
```suggestion
  const maxWaitMs = 1000;
  const pollIntervalMs = 100;
  const startTimestamp = Date.now();

```

PR #1174, opened 2025-07-22
Support AVCC for all h264 inputs

noituri:
This fixes problems with some playlists not working


wkozyra95 (commented):
This logic does not seem as something that should be part of HLS implementation. I would implement that in the iterator and use it in `decoder_thread_video.rs`

Additionally, for the future, even if you plan to use it in only one place try to extract some abstaraction, especially stuff like parsing or transforming raw bytes. This code is extremely coupled to hls and ffmpeg and would  require significant rewrite to reuse anything. 

In this case I would:
- make repacker should take already parsed ExtraData object
- parsing and accessing extradata could be done in calling code or separate struct that wraps the packer
- implement iterator that wraps this repacker

compositor_pipeline/src/pipeline/input/mp4/reader.rs
```diff
@@ -95,73 +94,28 @@ impl<Reader: Read + Seek + Send + 'static> Mp4FileReader<Reader> {
...
-                data.extend_from_slice(&[0, 0, 0, 1]);
-                data.extend_from_slice(&nalu);
-            }
-
-            data.freeze()
-        };
+        let mut avcc_box = BytesMut::new().writer();
+        if let Err(err) = avc.avcc.write_box(&mut avcc_box) {
+            warn!("Failed to write mp4 extra data: {err}");
+        }
```
noituri: `spawn_video_decoder_thread` accepts extra data as bytes so I have to reserialize avcc_box and then parse it again in `spawn_video_decoder_thread`. This could be avoided if we'd accept `ExtraData` struct instead, but that would mean we'd have to expose it and its parse method to the caller.

wkozyra95: I'm fine with exposing ExtraData publicly by renaming it e.g. FfmpegH264ExtraData or if it's not ffmpeg specific, then use h264 in the name

My understanding is that byte structure is not specific to ffmpeg to but the name extradata is, so use the correct name there

compositor_pipeline/src/pipeline/decoder/decoder_thread_video.rs
```diff
@@ -59,16 +64,29 @@ pub fn spawn_video_decoder_thread<Decoder: VideoDecoder, const BUFFER_SIZE: usiz
...
+            _ => {
+                warn!("Could not parse extra data: {e}");
+                None
+            }
+        });
+
     let (chunk_sender, chunk_receiver) = crossbeam_channel::bounded(BUFFER_SIZE);
-    let decoded_stream =
-        VideoDecoderStream::<Decoder, _>::new(ctx, chunk_receiver.into_iter())?.flatten();
+    // TODO: This should be only used if the stream is in h264
```
wkozyra95: You have 2 options here, either conditionally add that iterator or check the condition inside. First option would require boxing the resulting iterator, so it might be better to do it inside

compositor_pipeline/src/pipeline/input/mp4/reader.rs
```diff
@@ -95,73 +94,28 @@ impl<Reader: Read + Seek + Send + 'static> Mp4FileReader<Reader> {
...
 
-        // sps and pps have to be extracted from the container, interleaved with [0, 0, 0, 1],
-        // concatenated and prepended to the first frame.
-        let sps = avc
-            .avcc
-            .sequence_parameter_sets
-            .iter()
-            .flat_map(|s| [0, 0, 0, 1].iter().chain(s.bytes.iter()));
-
-        let pps = avc
```
wkozyra95, line 106: thanks for extracting that code too

compositor_pipeline/src/pipeline/decoder/decoder_thread_video.rs
noituri: I went with the first option because it looks nicer imo

compositor_pipeline/src/pipeline/decoder/decoder_thread_video.rs
```diff
@@ -67,8 +74,15 @@ fn init_decoder_stream<Decoder: VideoDecoder, const BUFFER_SIZE: usize>(
     DecoderInitError,
 > {
     let (chunk_sender, chunk_receiver) = crossbeam_channel::bounded(BUFFER_SIZE);
-    let decoded_stream =
-        VideoDecoderStream::<Decoder, _>::new(ctx, chunk_receiver.into_iter())?.flatten();
+    let chunk_stream: Box<dyn Iterator<Item = _>> = match Decoder::LABEL {
+        FfmpegH264Decoder::LABEL | VulkanH264Decoder::LABEL => {
```
wkozyra95, line 78: LABEL is just for description, we can't use it to distinguish between decoders

PR #1175, opened 2025-07-22
Run ffplay automatically on hls input example

noituri:


wkozyra95 (commented):
please leave it output as RTMP, just add ffmpeg code that will play rtmp

wkozyra95 (approved):

PR #1187, opened 2025-07-29
Queue implementation without input synchronization

wkozyra95:
- [WIP] - Synchronization should happen in the inputs
- All PTS should be counted relative to the sync point value (initialized on app start, not on the pipeline start)
- offset is still handled in the queue, incoming timestamps are calculated based on time of arrival, and queue is recalculating them relative to queue start/offset


compositor_pipeline/src/queue.rs
```diff
@@ -61,11 +57,16 @@ impl From<&PipelineOptions> for QueueOptions {
...
 ///
-/// - PTS of inputs streams can be in any frame of reference.
-/// - PTS of frames stored in queue are in a frame of reference where PTS=0 represents.
-///   first frame/packet.
-/// - PTS of output frames is in a frame of reference where PTS=0 represents
-///   start request.
+/// - All PTS values represent duration from sync_point stored in pipeline.
+///   (That will mean that pts of queue start will not be zero)
+///   - If input has offset defined it is applied in queue, timestamps before represents
+///     packet as if there was no offset
```
noituri:
```suggestion
///   - If input has offset defined it is applied in queue, timestamps before represent
///     packets as if there was no offset
```

compositor_pipeline/src/queue.rs
```diff
@@ -61,11 +57,16 @@ impl From<&PipelineOptions> for QueueOptions {
...
+/// - Conversion from/to this time frame should happen
+///   - on input as early as possible
+///   - on output as late as possible
+/// - All public timestamp value should refer to time after start.
```
noituri:
```suggestion
/// - All public timestamp values should refer to time after start.
```

noituri (commented):
So far looks good

compositor_pipeline/src/pipeline/rtp/rtp_input/rtcp_sync.rs
```diff
@@ -0,0 +1,209 @@
...
+use tracing::{debug, warn};
+
+use crate::pipeline::rtp::rtp_input::rollover_state::RolloverState;
+
+const POW_2_32: f64 = (1i64 << 32) as f64;
+
+#[derive(Debug)]
+/// State that should be shared between different RTP tracks to use for synchronization.
+pub(crate) struct RtpNtpSyncPoint {
+    sync_point: Instant,
```
wkozyra95, line 15: Technically, sync_point is not functionally  necessary in this structs, but I think it fits here

compositor_pipeline/src/pipeline/rtp/rtp_input/rtcp_sync.rs
```diff
@@ -0,0 +1,209 @@
...
+        clock_rate: u32,
+    ) {
+        {
+            let guard = self.ntp_time.read().unwrap();
+            if guard.is_some() {
+                return;
+            }
+        }
+
+        let mut guard = self.ntp_time.write().unwrap();
```
wkozyra95, line 58: This might break if some PTS are before rollover and some after; however, this will only be called once at the beginning of the stream, so it's extremely unlikely. I might look into that in follow-up, but don't want to focus on this right now.

PR #1191, opened 2025-07-30
Audio comparison method changed.

JBRS307:


integration_tests/Cargo.toml
```diff
@@ -37,6 +37,7 @@ crossbeam-channel = { workspace = true }
 tracing = { workspace = true }
 opus = { workspace = true }
 pitch-detection = "0.3.0"
```
JBRS307: This is not needed anymore.
```suggestion
```

integration_tests/src/tests/audio_only.rs
```diff
@@ -94,9 +94,9 @@ pub fn audio_mixing_with_offset() -> Result<()> {
         &new_output_dump,
         AudioValidationConfig {
             sampling_intervals: vec![
-                Duration::from_millis(0)..Duration::from_millis(2000),
-                Duration::from_millis(2000)..Duration::from_millis(4000),
-                Duration::from_millis(8000)..Duration::from_millis(10000),
+                Duration::from_millis(0)..Duration::from_millis(10000),
+                // Duration::from_millis(2000)..Duration::from_millis(4000),
+                // Duration::from_millis(8000)..Duration::from_millis(10000),
```
JBRS307: Remove this comments for every test and set every audio test from 0 to 10

integration_tests/src/tests/required_inputs.rs
```diff
@@ -321,12 +321,14 @@ pub fn required_audio_inputs_no_offset() -> Result<()> {
...
-                Duration::from_millis(2000)..Duration::from_millis(4000),
-                Duration::from_millis(6000)..Duration::from_millis(8000),
+                Duration::from_millis(0)..Duration::from_millis(10000),
+                // Duration::from_millis(2000)..Duration::from_millis(4000),
+                // Duration::from_millis(6000)..Duration::from_millis(8000),
             ],
-            // TODO(noituri):  This should not be necessary
-            allowed_error: 55.0,
+            // In general this test failes 3 batches (each batch is around 0.34s) - 1 per channel at the same pts,
+            // and differences are not heard in output dumps.
```
JBRS307:
```suggestion
            // and any differences cannot be heard in output dumps.
```

integration_tests/src/tests/required_inputs.rs
```diff
@@ -321,12 +321,14 @@ pub fn required_audio_inputs_no_offset() -> Result<()> {
...
-                Duration::from_millis(0)..Duration::from_millis(2000),
-                Duration::from_millis(2000)..Duration::from_millis(4000),
-                Duration::from_millis(6000)..Duration::from_millis(8000),
+                Duration::from_millis(0)..Duration::from_millis(10000),
+                // Duration::from_millis(2000)..Duration::from_millis(4000),
+                // Duration::from_millis(6000)..Duration::from_millis(8000),
             ],
-            // TODO(noituri):  This should not be necessary
-            allowed_error: 55.0,
+            // In general this test failes 3 batches (each batch is around 0.34s) - 1 per channel at the same pts,
```
JBRS307:
```suggestion
            // In general this test failes 6 batches (each batch is around 0.34s) - 1 per channel at the same pts,
```

integration_tests/src/tests/required_inputs.rs
```diff
@@ -321,12 +321,14 @@ pub fn required_audio_inputs_no_offset() -> Result<()> {
...
+            // and differences are not heard in output dumps.
+            // 8 batches to fail is a safety buffer.
```
JBRS307:
```suggestion
            // 8 batches is a safety buffer.
```

integration_tests/src/validation.rs
```diff
@@ -107,20 +93,102 @@ impl Default for VideoValidationConfig {
...
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct SamplingInterval {
+    pub first_sample: usize,
+    pub samples: usize,
+}
+
+impl SamplingInterval {
+    // Intervals returned are this function are not exact
```
JBRS307:
```suggestion
    // Intervals returned by this function are not exact
```

integration_tests/src/validation.rs
```diff
@@ -107,20 +93,102 @@ impl Default for VideoValidationConfig {
...
+    // They may be slightly longer as this function fills the time with batches of size
```
JBRS307: This  line is terrible, rethink it

integration_tests/src/validation.rs
```diff
@@ -107,20 +93,102 @@ impl Default for VideoValidationConfig {
...
+        }
+
+        let time_per_batch = Duration::from_secs_f64(samples_per_batch as f64 / sample_rate as f64);
+
+        // It finds the sample that fits pts best
+        // If it is not a multiple of samples_per_batch find the highest
+        // multiple lower than current number to be the starting sample
+        let mut first_sample =
+            f64::floor(start_pts.as_secs_f64() * sample_rate as f64 / samples_per_batch as f64)
+                as usize
```
JBRS307: No need for floor
```suggestion
            (start_pts.as_secs_f64() * sample_rate as f64 / samples_per_batch as f64)
                as usize
```

integration_tests/src/validation.rs
```diff
@@ -107,20 +93,102 @@ impl Default for VideoValidationConfig {
...
+            first_sample += samples_per_batch;
+            n += 1;
+        }
+        intervals
+    }
+}
+
+pub struct AudioAnalyzeTolerance {
+    /// Tolerance of max frequency. This value is the multiplier
+    /// by which frequency resolution shall be multiplied while calculating tolerance
```
JBRS307:
```suggestion
    /// by which frequency resolution shall be multiplied when comparing values
```

integration_tests/src/validation/audio.rs
```diff
@@ -1,21 +1,117 @@
...
-    allowed_error: f32,
-    channels: AudioChannels,
-    sample_rate: u32,
+    test_config: AudioValidationConfig,
+    // time_intervals: &[Range<Duration>],
+    // channels: AudioChannels,
+    // sample_rate: u32,
+    // samples_per_batch: usize,
+    // allowed_failed_batches: u8,
+    // tolerance: AudioAnalyzeTolerance,
```
JBRS307: Remove that

integration_tests/src/validation/audio.rs
```diff
@@ -34,75 +130,164 @@ pub fn validate(
...
+        // There should not be any NaN or Infinities and if there are the test should fail
+        let max_sample = samples.iter().map(|s| s.abs()).reduce(f32::max).unwrap() as f64;
+        let amplitude = match amplitude {
+            Some(a) => a,
+            None => max_sample,
+        };
+        if amplitude == 0.0 {
+            (0.0, max_sample)
+        } else {
+            (20.0 * f64::log10(max_sample / amplitude), max_sample)
```
JBRS307: Returning the tuple, put this calculation into a named variable

integration_tests/src/validation/audio.rs
```diff
@@ -34,75 +130,164 @@ pub fn validate(
...
+
+    let mut expected_spectrum_left = calc_fft(&expected_samples_left, sample_rate)?;
+    let mut expected_spectrum_right = calc_fft(&expected_samples_right, sample_rate)?;
+    let mut actual_spectrum_left = calc_fft(&actual_samples_left, sample_rate)?;
+    let mut actual_spectrum_right = calc_fft(&actual_samples_right, sample_rate)?;
+
+    let left_scaler = expected_spectrum_left.max().1.val();
+    let right_scaler = expected_spectrum_right.max().1.val();
+    //
+    // Channels and expected and actual sample batches should be of equal length
```
JBRS307:
```suggestion
    
    // Expected and actual sample batches should be of equal length
```

integration_tests/src/validation.rs
```diff
@@ -107,20 +93,101 @@ impl Default for VideoValidationConfig {
...
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct SamplingInterval {
+    pub first_sample: usize,
+    pub samples: usize,
+}
+
+impl SamplingInterval {
+    // Intervals returned by this function to not match time stamp exactly.
```
noituri:
```suggestion
    // Intervals returned by this function do not match time stamp exactly.
```

noituri (approved):
lgtm

integration_tests/src/audio_decoder.rs
```diff
@@ -48,15 +38,13 @@ impl AudioDecoder {
         }
 
         let samples_count = self.decoder.decode(&chunk_data, &mut self.buffer, false)?;
-        self.decoded_samples.push(AudioSampleBatch {
-            samples: self.buffer[..samples_count].to_vec(),
-            pts: Duration::from_secs_f64(packet.header.timestamp as f64 / self.sample_rate as f64),
-        });
+        self.decoded_samples
```
wkozyra95: You are assuming here that audio is continuous, in a hypothetical scenario:
- We have a test that plays audio for 2 seconds, drops packets for 2 seconds, and then plays again.
- And we have a bug that will deliver the same sample, but the timestamps are wrong.

As a result, this code would treat them the same, but when playing the original sample, it has a hole between 2 and 4 seconds, while the new behavior could be in a different place.

To be fair, the old implementation would also not handle that, but I'm not sure we should remove that information from decoder output

integration_tests/src/tests/video_audio.rs
```diff
@@ -103,11 +103,14 @@ pub fn single_input_with_video_and_audio_flaky() -> Result<()> {
         OUTPUT_DUMP_FILE,
         &new_output_dump,
         AudioValidationConfig {
-            sampling_intervals: vec![
-                Duration::from_millis(0)..Duration::from_millis(2000),
-                Duration::from_millis(2000)..Duration::from_millis(4000),
-                Duration::from_millis(8000)..Duration::from_millis(10000),
-            ],
+            sampling_intervals: vec![Duration::from_millis(0)..Duration::from_millis(10000)],
+            tolerance: crate::AudioAnalyzeTolerance {
```
wkozyra95, line 107: tested how this behaves, and this test seems to detect an error if 
- shift is larger than 150ms, which is better than the old implementation
- volume is set to 0.5, which is a lot (but still better than the old implementation). Can we do something to detect that better?

integration_tests/src/validation.rs
```diff
@@ -107,20 +93,101 @@ impl Default for VideoValidationConfig {
...
+
+        // It finds the sample that fits pts best
+        // If it is not a multiple of samples_per_batch find the highest
+        // multiple lower than current number to be the starting sample
+        let mut first_sample = (start_pts.as_secs_f64() * sample_rate as f64
+            / samples_per_batch as f64) as usize
+            * samples_per_batch;
+
+        let mut intervals = vec![];
+        let mut n = 0;
```
wkozyra95: unnecessary value, you can check `intervals.len()`

integration_tests/src/validation/audio.rs
```diff
@@ -34,75 +124,167 @@ pub fn validate(
...
+fn analyze_samples(
+    actual_samples: Vec<f32>,
+    expected_samples: Vec<f32>,
     sample_rate: u32,
-) -> Result<(f64, f64)> {
-    fn get_pitch(samples: &[f64], sample_rate: u32) -> Result<f64> {
-        if samples.is_empty() {
-            return Err(anyhow::anyhow!("No samples"));
+) -> Result<(AnalyzeResult, AnalyzeResult, AnalyzeResult, AnalyzeResult)> {
+    fn split_samples(samples: Vec<f32>) -> (Vec<f32>, Vec<f32>) {
```
wkozyra95: let's leave it as it is, but in the future, try to avoid this pattern, even though you are extracting logic to functions the fact that it is defined inside it still makes `analyze_samples` very complex.

If you want to extract those functions then just write them at top level,
If you wanted them to be private, because they are coupled stronglly witn analyze_samples you can  move everything to `validation/audio/analyze_samples.rs`
Even in this approach if there some of those functions can be described without the context of analyze druation, then maybe just those could be on top level.

wkozyra95 (approved):

integration_tests/src/tests/video_audio.rs
JBRS307: Lowering the volume tolerance would result in even very small timinng inaccuracies to fail the test and little linear change in volume (i.e. multiplying by samples by 0.9 or something similar) would not be audible.

wkozyra95 (approved):

PR #1198, opened 2025-08-05
Fixed panic in `prepare_input_samples`.

### LINKED ISSUE #1192: Audio mixer's `prepare_inputs` sometimes crashes
```
2025-07-31T09:03:28.887068Z  INFO Audio decoder thread: compositor_pipeline::pipeline::resampler::dynamic_resampler: Initializing stereo audio resampler input_sample_rate=44100 output_sample_rate=48000 input_id="input_1" decoder="FDK AAC decoder"

thread '<unnamed>' panicked at compositor_pipeline/src/audio_mixer/prepare_inputs.rs:143:54:
slice index starts at 1019 but ends at 960
```

JBRS307:
Closes #1192

### DIFF
--- a/compositor_pipeline/src/audio_mixer/prepare_inputs.rs
+++ b/compositor_pipeline/src/audio_mixer/prepare_inputs.rs
@@ -92,7 +92,7 @@ fn frame_input_samples(
         sample
     });
 
-    for (batch_index, input_samples) in samples_iter.enumerate() {
+    for input_samples in samples_iter {
         let sample_count = samples_in_frame.len();
         let expected_next_sample_start_pts =
             start_pts + Duration::from_secs_f64(sample_count as f64 / sample_rate as f64);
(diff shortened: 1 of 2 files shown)


compositor_pipeline/src/audio_mixer/prepare_inputs/single_frame_tests.rs
```diff
@@ -364,4 +364,35 @@ fn test_prepare_inputs() {
             (5.0, 5.0)
         ]
     );
+
+    // Severly missaligned timestamps (to the left)
```
wkozyra95: misaligned is a very generic statement here,  for once it's subjective it's also not really true, it does not matter for this bug how much batch is outside of the exepected ranage

if you want to describe what this test case is testing, then it should be something like this:
```
The first batch ends before the result batch starts by at least one sample duration
```

Also, if you really want to show the corner case, you can use 
```
let first_batch_start = start - (5 * sample_duration) - small_error;
```
instead of
```
let first_batch_start = start - batch_duration;
```

btw I'm assuming here that this was the corner case, don't know that for sure

compositor_pipeline/src/audio_mixer/prepare_inputs.rs
```diff
@@ -129,7 +129,7 @@ fn frame_input_samples(
                     samples_to_remove_from_start
```
wkozyra95, line 130: If first batch is dropped entirely, then this warning can give false positve because we might need to drop some samples from the second batch too

Issue #1185, opened 2025-07-28
[ts-sdk] WHIP input


wkozyra95:
A better experience for WHIP input.

Currently whip url for input example would look like this `/whip/global%3Ainput` and token is not returned properlly

PR #1199, opened 2025-08-05
Better way of getting register WHIP input info

noituri:
Closes: #1185


ts/smelter-web-client/src/smelter/offline.ts
```diff
@@ -37,24 +42,38 @@ export default class OfflineSmelter {
...
-        mappedResult.bearerToken = result['bearer_token'];
+      if (request.type === 'mp4') {
+        return {
+          videoDurationMs: result.video_duration_ms,
+          audioDurationMs: result.audio_duration_ms,
+        };
+      } else if (request.type === 'whip') {
+        return {
+          bearerToken: result.bearer_token,
+          endpointRoute: `/whip/${encodeURIComponent(inputId)}`,
```
wkozyra95: this is incorrect

ts/smelter-web-client/src/smelter/offline.ts
```diff
@@ -37,24 +42,38 @@ export default class OfflineSmelter {
     });
   }
 
+  /* eslint-disable no-dupe-class-members */
```
wkozyra95: I feel like there should be a better way to disable that option, if not, then disable that globally

ts/smelter-web-client/src/smelter/live.ts
```diff
@@ -47,24 +52,38 @@ export default class Smelter {
...
-        mappedResult.bearerToken = result['bearer_token'];
+      if (request.type === 'mp4') {
+        return {
+          videoDurationMs: result.video_duration_ms,
+          audioDurationMs: result.audio_duration_ms,
+        };
+      } else if (request.type === 'whip') {
+        return {
+          bearerToken: result.bearer_token,
+          endpointRoute: `/whip/${encodeURIComponent(inputId)}`,
```
wkozyra95: also incorrect

compositor_api/src/input/whip.rs
```diff
@@ -18,6 +18,10 @@ pub struct WhipInput {
     /// Token used for authentication in WHIP protocol. If not provided, the random value
     /// will be generated and returned in the response.
     pub bearer_token: Option<Arc<str>>,
+    /// Internal use only.
+    /// Specifies prefix that is used in `input_id`.
```
wkozyra95: This api does not make sense, why are you defining an option that tells what prefix should be removed, if you can define an option that just says what id should be used

compositor_pipeline/src/pipeline/webrtc/whip_input/state.rs
```diff
@@ -14,7 +15,62 @@ use crate::pipeline::webrtc::{
 };
 
 #[derive(Debug, Clone, Default)]
-pub(crate) struct WhipInputsState(Arc<Mutex<HashMap<InputId, WhipInputConnectionState>>>);
+pub(crate) struct WhipInputsState(Arc<Mutex<WhipInputsStateInner>>);
+
+#[derive(Debug, Clone, Default)]
+struct WhipInputsStateInner {
+    /// Maps IDs without prefix to ones with prefix
+    prefixed_input_ids: HashMap<InputId, InputId>,
```
wkozyra95: You already have a hash map based on id, there is no point in having 2 independent maps that needs to be synce together

wkozyra95 (changes requested):

ts/smelter-web-client/src/smelter/offline.ts
noituri: Why? It appears to be working correctly when I was testing it

wkozyra95: a sorry my bad, I forgot that at this point it does not have `global:` prefix

compositor_api/src/input/whip.rs
```diff
@@ -18,6 +18,9 @@ pub struct WhipInput {
     /// Token used for authentication in WHIP protocol. If not provided, the random value
     /// will be generated and returned in the response.
     pub bearer_token: Option<Arc<str>>,
+    /// Internal use only.
+    /// Overrides input id which is used when referencing the input via whip server
+    pub override_input_id: Option<InputId>,
```
wkozyra95: this is not overriding input_id, input_id is still the same as it was, change the name to describe what is actually represents

compositor_api/src/input/whip.rs
```diff
@@ -18,6 +18,10 @@ pub struct WhipInput {
     /// Token used for authentication in WHIP protocol. If not provided, the random value
     /// will be generated and returned in the response.
     pub bearer_token: Option<Arc<str>>,
+    /// Internal use only.
+    /// Overrides whip session id which is used when referencing the input via whip server.
+    /// If not provided, it defaults to input id.
+    pub override_whip_session_id: Option<InputId>,
```
wkozyra95:
```suggestion
    pub override_whip_session_id: Option<Arc<str>>,
```

I'm not sure if session ID is not already an established term(it's private, so it does not matter much)

Consider `whip_session_id_override`  (noun instead of verb)

ts/smelter-node/src/offline/compositor.ts
```diff
@@ -33,9 +38,34 @@ export default class OfflineSmelter {
...
+      let result = await this.coreSmelter.registerInput(inputId, request);
+      if (request.type === 'mp4') {
+        return {
+          videoDurationMs: result.video_duration_ms,
+          audioDurationMs: result.audio_duration_ms,
+        };
+      } else if (request.type === 'whip') {
+        return {
+          bearerToken: result.bearer_token,
+          endpointRoute: `/whip/${encodeURIComponent(inputId)}`,
```
wkozyra95: is there a way to handle that in core? If not I would still prefer to extract conversion logic like this to outside of file like this. It mixes to much high level API with simple low level type conversion

compositor_pipeline/src/pipeline/webrtc/whip_input.rs
```diff
@@ -39,6 +39,7 @@ impl WhipInput {
             return Err(InputInitError::WhipWhepServerNotRunning);
         };
 
+        let input_id = options.override_whip_session_id.unwrap_or(input_id);
```
wkozyra95: name it either url suffix or session id, but not input id, also check other places because you just swapped a value here, because some other calls might expect real input_id

wkozyra95 (approved):

PR #1211, opened 2025-08-08
Add Trickle ICE to WHEP

wkazmierczak:


compositor_pipeline/src/pipeline/webrtc/server/create_whep_session.rs
```diff
@@ -24,6 +25,7 @@ pub async fn handle_create_whep_session(
     offer: String,
 ) -> Result<Response<Body>, WhipWhepServerError> {
     let output_id = OutputId(Arc::from(id.clone()));
+    let session_id: Arc<str> = Arc::from(Uuid::new_v4().to_string()); // TODO maybe simple rand better
```
wkozyra95: The session id should be generated inside add_session

wkozyra95: uuid  v4 is basically  just a radnom number, so it's fine either way

compositor_pipeline/src/pipeline/webrtc/server.rs
```diff
@@ -83,12 +87,17 @@ impl WhipWhepServer {
         let app = Router::new()
             .route("/status", get((StatusCode::OK, axum::Json(json!({})))))
             .route("/whip/:id", post(handle_create_whip_session))
-            .route("/session/:id", patch(handle_new_whip_ice_candidates))
-            .route("/session/:id", delete(handle_terminate_whip_session))
+            .route("/whip/:id", patch(handle_new_whip_ice_candidates))
+            .route("/whip/:id", delete(handle_terminate_whip_session))
```
wkozyra95, line 91: I'm wondering whether we shouldn't also add the session concept here, especially for delete to only terminate a specific peer connection. However this would require some renames, so let's leave it for follow up

PR #1212, opened 2025-08-08
Support specifying decoders for codecs in mp4 and hls

noituri:
Closes: #1200

### DIFF
--- a/compositor_api/src/input/hls.rs
+++ b/compositor_api/src/input/hls.rs
@@ -1,7 +1,9 @@
-use std::sync::Arc;
+use std::{collections::HashMap, sync::Arc};
 
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+
+use crate::VideoDecoder;
 
 /// Parameters for an input stream from HLS source.
 #[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
@@ -16,4 +18,12 @@ pub struct HlsInput {
     /// not defined then the stream will be synchronized based on the delivery time of the initial
     /// frames.
     pub offset_ms: Option<f64>,
+    /// Assigns which decoder should be used for media encoded with a specific codec.
+    pub decoder_map: Option<HashMap<InputHlsCodec, VideoDecoder>>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, PartialEq, Eq, Hash)]
+#[serde(rename_all = "snake_case")]
+pub enum InputHlsCodec {
+    H264,
 }
--- a/compositor_api/src/input/hls_into.rs
+++ b/compositor_api/src/input/hls_into.rs
@@ -1,7 +1,6 @@
-use std::time::Duration;
-
 use crate::common_pipeline::prelude as pipeline;
 use crate::*;
+use std::time::Duration;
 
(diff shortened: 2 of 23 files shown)


ts/smelter/src/types/input/rtp.ts
noituri, line 1: This code was only moved outside of `input.ts`

ts/smelter/src/types/input/whip.ts
noituri, line 1: This code was only moved outside of `input.ts`

compositor_pipeline/src/pipeline/hls/hls_input.rs
noituri, line 1: This code is getting a little bit convoluted. I don't want to introduce a lot of unrelated changes in this PR so I'll try to rewrite it in some follow up PR

compositor_api/src/input/hls.rs
```diff
@@ -16,4 +18,12 @@ pub struct HlsInput {
     /// not defined then the stream will be synchronized based on the delivery time of the initial
     /// frames.
     pub offset_ms: Option<f64>,
+    /// Configures decoders for the provided codecs.
+    pub decoder_mappings: Option<HashMap<InputHlsCodec, VideoDecoder>>,
```
wkozyra95: I think decoder_mapping is more correct; however, I would suggest `decoder_map`

compositor_api/src/input/mp4.rs
```diff
@@ -23,6 +24,12 @@ pub struct Mp4Input {
     /// Offset in milliseconds relative to the pipeline start (start request). If offset is
     /// not defined then stream is synchronized based on the first frames delivery time.
     pub offset_ms: Option<f64>,
-    /// (**default=`ffmpeg_h264`**) The decoder to use for decoding video.
-    pub video_decoder: Option<VideoDecoder>,
+    /// Configures decoders for the provided codecs.
```
wkozyra95:
```suggestion
    /// Assigns which decoder should be used for media encoded with a specific codec.
```
The same above

compositor_pipeline/src/pipeline/hls/hls_input.rs
```diff
@@ -159,15 +159,40 @@ impl HlsInput {
...
+                    VideoDecoderOptions::VulkanH264 => {
+                        VideoDecoderThread::<vulkan_h264::VulkanH264Decoder, _>::spawn(
+                            input_id,
+                            decoder_thread_options,
+                        )
+                    }
+                    _ => {
+                        // This should never happend
+                        warn!("Non h264 decoder provided to HLS. Defaulting to ffmpeg h264");
+                        VideoDecoderThread::<ffmpeg_h264::FfmpegH264Decoder, _>::spawn(
```
wkozyra95: return error here

compositor_api/src/input/hls_into.rs
```diff
@@ -19,9 +21,45 @@ impl TryFrom<HlsInput> for pipeline::RegisterInputOptions {
             buffer_duration: None,
         };
 
+        let video_decoders = match decoder_mappings {
```
wkozyra95: If you are already handling codecs here and separating video from audio, then I don't see the point of returning HashMap here.

PR #1222, opened 2025-08-13
Add interactive example.

### LINKED ISSUE #1206: Add Demo server and demo client
Basic functionalities for demo server and client

JBRS307:
Closes #1206, #1219 

- Basic functionalities
- Automatic streaming with gstreamer


integration_tests/examples/demo/main.rs
```diff
@@ -0,0 +1,118 @@
...
+    #[strum(to_string = "Start")]
+    Start,
+}
+
+fn run_demo() -> Result<()> {
+    let mut state = SmelterState::new();
+
+    let options = Action::iter().collect::<Vec<_>>();
+
+    loop {
```
wkozyra95, line 36: It's not that bad now, but this is easy to handle without repeating, e.g. by updating options when handing Start action

integration_tests/examples/demo/main.rs
```diff
@@ -0,0 +1,118 @@
...
+            Err(e) => {
+                error!("{e}");
+                break;
+            }
+        }
+    }
+
+    // Inquire handles Ctrl+c, after it causes to break
+    // out of the loop all players are killed
+    Command::new("pkill")
```
wkozyra95: store handle to processes you are spawning in state and kill them explicitly e.g. on drop

integration_tests/examples/demo/utils.rs
```diff
@@ -0,0 +1,212 @@
+use std::{
```
wkozyra95: utils/tools/misc name should be avoided where possible and in general limited to very generic functions that are not coupled with a business logic

integration_tests/examples/demo/utils.rs
```diff
@@ -0,0 +1,212 @@
...
+        };
+
+        output_handler.set_initial_scene(&self.inputs);
+
+        let output_json = output_handler.serialize_register();
+        let output_route = format!("output/{}/register", output_handler.name());
+
+        if output_handler.transport_protocol() == TransportProtocol::TcpServer {
+            examples::post(&output_route, &output_json)?;
+        }
```
wkozyra95: If you already have add workarounds like this with just RTP it suggests that OutputHandler/InputHandler might not be a good level of abstractions. Note that you will need that same logic on input for e.g. HLS input

For this specific case you could add add function that returns boolean that specifies whether init should be run before or after. However I would rethink the abstraction itself, For example you could:
- Instead of function like `start_ffmpeg_transmitter` add `on_before_register` and `on_after_register` and select gst or ffmpeg based on internal value instead of argument

integration_tests/examples/demo/utils/inputs.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+#[derive(Debug, Display, EnumIter)]
+pub enum AudioDecoder {
+    #[strum(to_string = "opus")]
+    Opus,
+    // TODO: AAC
+}
+
+pub fn input_name() -> String {
```
wkozyra95: I don't think this function helps much, it could be inlined. Plus I think it would be better to have more info in the name to simplify reading logs eg. `input_rtp_tcp_{port}_{random_number}`

integration_tests/examples/demo/utils/outputs.rs
```diff
@@ -0,0 +1,124 @@
...
+    fn name(&self) -> &str;
+    fn port(&self) -> u16;
+    fn transport_protocol(&self) -> TransportProtocol;
+    fn serialize_register(&self) -> serde_json::Value;
+    fn serialize_update(&self) -> serde_json::Value;
+    fn inputs(&mut self) -> &mut Vec<String>;
+    fn start_ffmpeg_receiver(&self) -> Result<()>;
+    fn start_gstreamer_receiver(&self) -> Result<()>;
+
+    fn set_initial_scene(&mut self, inputs: &[Box<dyn InputHandler>]) {
```
wkozyra95: I don't see the point of that function, this is something known during the construction so why is this in trait?

integration_tests/examples/demo/utils/outputs.rs
```diff
@@ -0,0 +1,124 @@
...
+use strum::{Display, EnumIter};
+
+use crate::utils::{inputs::InputHandler, TransportProtocol};
+
+pub mod rtp;
+
+pub trait OutputHandler: Debug {
+    fn name(&self) -> &str;
+    fn port(&self) -> u16;
+    fn transport_protocol(&self) -> TransportProtocol;
```
wkozyra95: very RTP specific, if this is intended to be generic for all outputs it should not have anything related to specific protocol/format

integration_tests/examples/demo/utils/outputs.rs
```diff
@@ -0,0 +1,124 @@
...
+    fn serialize_register(&self) -> serde_json::Value;
+    fn serialize_update(&self) -> serde_json::Value;
+    fn inputs(&mut self) -> &mut Vec<String>;
```
wkozyra95: I don't think there is a need to store inputs in outputs, unless maybe we will have some special scenes that only use specific inputs. If you need to inputs to create scene updates, the should be passed to update function as argument, we don't want to duplicate state

integration_tests/examples/demo/utils.rs
```diff
@@ -0,0 +1,212 @@
...
+        if output_handler.transport_protocol() == TransportProtocol::Udp {
+            examples::post(&output_route, &output_json)?;
+        }
+
+        self.outputs.push(output_handler);
+
+        Ok(())
+    }
+
+    pub fn unregister_input(&mut self) -> Result<()> {
```
wkozyra95, line 115: unregister should kill ffmpeg/gst processes if there are any, The idea I mentioned earlier with doing that in Drop would solve that

integration_tests/examples/demo/utils/inputs/rtp.rs
```diff
@@ -0,0 +1,235 @@
...
+                        break;
+                    }
+                }
+            }
+        }
+
+        Ok(rtp_input)
+    }
+
+    fn setup_video(&mut self) -> Result<()> {
```
wkozyra95: I'm not a fan off this "builder like" pattern, mixinig input object that represents state and process of providing information will not end well in my opinion.

integration_tests/examples/demo/utils/inputs/rtp.rs
```diff
@@ -0,0 +1,235 @@
...
+            audio: None,
+            transport_protocol: TransportProtocol::Udp,
+        };
+
+        let options = RtpRegisterOptions::iter().collect::<Vec<_>>();
+        loop {
+            let action = Select::new("What to do?", options.clone()).prompt()?;
+
+            match action {
+                RtpRegisterOptions::AddVideoStream => rtp_input.setup_video()?,
```
wkozyra95: why are those things action that user can select? When registering input/output just ask everything in order with option to skip if optional.

integration_tests/examples/demo/utils.rs
JBRS307: Will be changed in the next PR as for now renaming it causes problems with git

integration_tests/examples/demo/utils/outputs.rs
JBRS307: ~I don't understand what "is known" during construction.~
Nvm, I get it.

wkozyra95: This is something you set only once when creating output, so you know what type output is this at this point, there is no reason that this should be a part of a trait

integration_tests/examples/demo/utils/inputs/rtp.rs
```diff
@@ -0,0 +1,366 @@
...
+        Ok(())
+    }
+}
+
+impl InputHandler for RtpInput {
+    fn name(&self) -> &str {
+        &self.name
+    }
+
+    fn serialize(&self) -> serde_json::Value {
```
wkozyra95: If you have a separate type for building input and for storing input state it makes more sense if the builder returns both InputHandler and register request. If you think about the name "serialize", you don't serialize input, you serialize register request.

Also, this is another thing you are doing once, just after input creation, so it's does not need to be a part of thet trait

integration_tests/examples/demo/utils/inputs/rtp.rs
```diff
@@ -0,0 +1,366 @@
...
+                break;
+            }
+        }
+
+        let transport_options = options
+            .iter()
+            .filter(|o| {
+                matches!(
+                    o,
+                    RtpRegisterOptions::SetTransportProtocol | RtpRegisterOptions::Skip
```
wkozyra95: Skip here is unnecessary, just prompt for protocol, audio, and video can be not defined, but this value has a default

integration_tests/examples/demo/utils/inputs/rtp.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    fn with_prompted_transport_protocol(self) -> Result<Self> {
+        let options = TransportProtocol::iter().collect();
+
+        let selection = Select::new("Transport protocol:", options).prompt()?;
+        Ok(self.with_transport_protocol(selection))
+    }
+
+    pub fn with_video(mut self, video: RtpInputVideoOptions) -> Self {
+        self.video = Some(video);
```
wkozyra95, line 241: it's fine if you prefer this pattern, but I would just make it `&mut self` update the self.video and do not return anything in all those functions

integration_tests/examples/demo/utils/outputs/rtp.rs
```diff
@@ -0,0 +1,460 @@
...
+                break;
+            }
+        }
+
+        let transport_options = options
+            .iter()
+            .filter(|o| {
+                matches!(
+                    o,
+                    RtpRegisterOptions::SetTransportProtocol | RtpRegisterOptions::Skip
```
wkozyra95: same as in input

integration_tests/examples/demo/utils/outputs/rtp.rs
```diff
@@ -0,0 +1,460 @@
...
+        }
+    }
+}
+
+impl OutputHandler for RtpOutput {
+    fn name(&self) -> &str {
+        &self.name
+    }
+
+    fn serialize_register(&self, inputs: &[&str]) -> serde_json::Value {
```
wkozyra95: this same as for input, Builder code would be better place for it

integration_tests/examples/demo/utils/inputs/rtp.rs
```diff
@@ -0,0 +1,366 @@
...
+
+            builder = match audio_selection {
+                Some(RtpRegisterOptions::AddAudioStream) => {
+                    builder.with_audio(RtpInputAudioOptions::default())
+                }
+                Some(RtpRegisterOptions::Skip) | None => builder,
+                _ => unreachable!(),
+            };
+
+            if builder.video.is_none() && builder.audio.is_none() {
```
wkozyra95, line 221: Up to you, but for an internal example it's fine if you just remove the loop panic in cases like this. You don't have to handle clear user errors like this.

integration_tests/examples/demo/utils/inputs/rtp.rs
```diff
@@ -0,0 +1,366 @@
...
+
+        let video_options = options
+            .iter()
+            .filter(|o| {
+                matches!(
+                    o,
+                    RtpRegisterOptions::AddVideoStream | RtpRegisterOptions::Skip
+                )
+            })
+            .collect::<Vec<_>>();
```
wkozyra95: Why? Shouldn't this be just
```suggestion
        let video_options = vec![RtpRegisterOptions::AddVideoStream, RtpRegisterOptions::Skip]
```

or am I missing something?

integration_tests/examples/demo/utils/inputs/rtp.rs
JBRS307: I'll do it as you say, but leave it as `prompt_skippable` in case someone really wants to press ESC

integration_tests/examples/demo/utils/inputs/rtp.rs
JBRS307: I'll leave it like that in case of missclicks.

integration_tests/examples/demo/utils/inputs/rtp.rs
JBRS307: I'll leave it as is.

integration_tests/examples/demo/utils/outputs/rtp.rs
```diff
@@ -0,0 +1,422 @@
...
+                        self.stream_handles.push(self.start_gst_recv_udp()?);
+                        Ok(())
+                    }
+                    OutputPlayerOptions::Manual => Ok(()),
+                };
+
+                match player_result {
+                    Ok(_) => break,
+                    Err(e) => error!("{e}"),
+                }
```
wkazmierczak: This first `match` will never return `Err()`, so the second one is unnecessary

integration_tests/examples/demo/utils/outputs/rtp.rs
```diff
@@ -0,0 +1,422 @@
...
+                        Ok(())
+                    }
+                    OutputPlayerOptions::Manual => Ok(()),
+                    _ => unreachable!(),
+                };
+
+                match player_result {
+                    Ok(_) => break,
+                    Err(e) => error!("{e}"),
+                }
```
wkazmierczak: Same with `match` as above

integration_tests/examples/demo/utils/outputs/rtp.rs
JBRS307: It may return `Err()` if something fails inside `self.start_ffmpeg_receiver()?` etc. because of `?` error propagation.

integration_tests/examples/demo/utils/outputs/rtp.rs
JBRS307: Same as above

integration_tests/examples/demo/utils/outputs/rtp.rs
wkazmierczak: Whole function will return `Err()` in that case. This `error!("{e}")` will never be printed.

integration_tests/examples/demo/utils.rs
```diff
@@ -0,0 +1,182 @@
...
+            let update_json = output.serialize_update(&inputs);
+            examples::post(&update_route, &update_json)?;
+        }
+
+        let unregister_route = format!("input/{}/unregister", to_delete.name());
+        examples::post(&unregister_route, &json!({}))?;
+
+        // Input to delete is chosen from existing inputs
+        // so it is guaranteed that it exists in vec.
+        let delete_index = self
```
wkozyra95: fine for example, but in general if you have to do something like this:
- it suggest that vector might not be a best data structure to hold this data
- if you need to keep vector then just 
   - `self.inputs =  self.inputs.iter().cloned().filter(...).collect()`
   - or if you realy want to avoid cloningin `self.inputs = std::mem::take(self.inputs).into_iter().filter(...).collect()`

integration_tests/examples/demo/utils/outputs/rtp.rs
wkozyra95: @wkazmierczak is right, I think you misunderstood how `?` works

JBRS307: Yeah, and it didn't work as I wanted it to work, fixed now.

wkazmierczak (approved):

PR #1224, opened 2025-08-13
[vk-video] Handle decoder restarts

noituri:
Closes: #1202

### DIFF
--- a/vk-video/src/vulkan_decoder.rs
+++ b/vk-video/src/vulkan_decoder.rs
@@ -73,12 +73,6 @@ pub enum VulkanDecoderError {
     #[error("Invalid input data for the decoder: {0}.")]
     InvalidInputData(String),
 
-    #[error("Profile changed during the stream")]
-    ProfileChangeUnsupported,
-
-    #[error("Level changed during the stream")]
-    LevelChangeUnsupported,
-
     #[error("Monochrome video is not supported")]
     MonochromeChromaFormatUnsupported,
 
@@ -194,7 +188,7 @@ impl VulkanDecoder<'_> {
             } => {
                 return self
                     .process_reference_frame(decode_info, *reference_id)
-                    .map(Option::Some)
+                    .map(Option::Some);
             }
 
             DecoderInstruction::Idr {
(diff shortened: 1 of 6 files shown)


vk-video/src/vulkan_decoder/session_resources.rs
```diff
@@ -27,6 +27,7 @@ pub(super) struct VideoSessionResources<'a> {
     pub(crate) decoding_images: DecodingImages<'a>,
     pub(crate) sps: HashMap<u8, SeqParameterSet>,
     pub(crate) pps: HashMap<(u8, u8), PicParameterSet>,
+    pub(crate) sps_needing_reset: HashMap<u8, SeqParameterSet>,
```
noituri: I'm not sure if I should handle `pps` in similar manner. I tried implementing it both ways and it didn't really make any difference.

vk-video/src/vulkan_decoder.rs
```diff
@@ -292,6 +281,20 @@ impl VulkanDecoder<'_> {
...
+                .sps_needing_reset
+                .remove(&decode_information.sps_id)
+            {
+                video_session_resources.reset_session(
+                    &self.vulkan_device,
+                    &self.command_buffers.decode_buffer,
+                    sps,
+                    &self.sync_structures.fence_memory_barrier_completed,
+                )?;
+            }
```
jerzywilczek: I think we can do better. How about when new sps comes in, we calculate something like `max(current_session_parameters, parameters_scheduled_for_reset, new_sps)` and set that as `parameters_scheduled_for_reset`? so on IDR we would create a session that meets the requirements for all sps, not just the one the IDR happens to be using? I think that could me more correct, both because we can have a sequence like: IDR(sps 1) P(sps 2) and to avoid downgrading the session, which I think is possible right now (think a sequence like: sps1, sps2, idr1, idr2), where sps1.profile > sps2.profile. we would end up with a session with sps2.profile if I'm understanding this correctly

jerzywilczek (approved):
:rocket:

jerzywilczek:
Actually, I think there are workarounds in smelter for this. I think this PR should remove them.

PR #1225, opened 2025-08-14
Add RTMP output to interactive example.

### LINKED ISSUE #1221: Add RTMP output to interactive example.

JBRS307:
Closes #1221

Merge only after #1222


integration_tests/examples/demo/main.rs
```diff
@@ -6,9 +6,9 @@ use smelter::{config::read_config, logger::init_logger};
 use strum::{Display, EnumIter, IntoEnumIterator};
 use tracing::{debug, error};
 
-mod utils;
+mod smelter_state;
```
wkozyra95, line 12: Dividing into dirs should serve some purpose if you throw everything inside one dir, it is the same as if you just had everything on the same level.

I would rename it to state
move input/output dirs up one level

integration_tests/examples/demo/smelter_state/outputs/rtmp.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    fn serialize_update(&self, inputs: &[&str]) -> serde_json::Value {
+        json!({
+            "video": self.video.as_ref().map(|v| v.serialize_update(inputs)),
+            "audio": self.audio.as_ref().map(|a| a.serialize_update(inputs)),
+        })
+    }
+
+    fn on_before_registration(&mut self) -> Result<()> {
+        let player_options = vec![
```
wkozyra95: I would suggest prompting at the beginning for everything in the first phase in Builder, and here only run appropriate code based on selection

integration_tests/examples/demo/smelter_state/outputs/rtmp.rs
```diff
@@ -0,0 +1,315 @@
...
+
+            match player_result {
+                Ok(_) => break,
+                Err(e) => error!("{e}"),
+            }
+        }
+        Ok(())
+    }
+
+    fn on_after_registration(&mut self) -> Result<()> {
```
wkozyra95: You can add a default implementation to trait to avoid implementing empty functions

integration_tests/examples/demo/smelter_state/outputs/rtmp.rs
```diff
@@ -0,0 +1,315 @@
...
+            video: None,
+            audio: None,
+        }
+    }
+
+    pub fn prompt(self) -> Result<Self> {
+        let mut builder = self;
+
+        let video_options = vec![
+            RtmpRegisterOptions::AddVideoStream,
```
wkozyra95: add sounds wrong in this context (this would also apply to the previous PR), Configure or even Set would be more appropriate

integration_tests/examples/demo/smelter_state/outputs/rtmp.rs
JBRS307: - [x] Done

integration_tests/examples/demo/main.rs
JBRS307: - [x] Done

integration_tests/examples/demo/smelter_state/outputs/rtmp.rs
JBRS307: - [x] Done

integration_tests/examples/demo/smelter_state/outputs/rtmp.rs
JBRS307: - [x] Done

integration_tests/src/ffmpeg.rs
```diff
@@ -44,8 +44,12 @@ pub fn start_ffmpeg_rtmp_receive(port: u16) -> Result<Child> {
         .stdout(Stdio::null())
         .stderr(Stdio::null())
         .spawn()?;
+    info!("Started RTMP FFmpeg listener on port {port}.");
+    thread::sleep(Duration::from_secs(2));
+
+    rtmp_listener.stdout = None;
```
JBRS307: This solution for breaking terminal works, but is dumb. I am open to any better ideas.

integration_tests/examples/demo/inputs/whip.rs
```diff
@@ -21,7 +21,7 @@ impl InputHandler for WhipInput {
         &self.name
     }
 
-    fn on_after_registration(&mut self) -> Result<()> {
+    fn on_after_registration(&mut self, _player: InputPlayer) -> Result<()> {
```
wkozyra95: this can be deleted

integration_tests/examples/demo/inputs/mp4.rs
```diff
@@ -21,7 +21,7 @@ impl InputHandler for Mp4Input {
         &self.name
     }
 
-    fn on_after_registration(&mut self) -> Result<()> {
+    fn on_after_registration(&mut self, _player: InputPlayer) -> Result<()> {
```
wkozyra95: delete

integration_tests/examples/demo/inputs.rs
```diff
@@ -2,13 +2,19 @@ use anyhow::Result;
...
+
 pub mod mp4;
 pub mod rtp;
 pub mod whip;
 
 pub trait InputHandler: Debug {
     fn name(&self) -> &str;
-    fn on_after_registration(&mut self) -> Result<()>;
+
+    #[allow(unused_variables)]
```
wkozyra95:
```suggestion
```

integration_tests/examples/demo/generators.rs
wkozyra95, line 1: generators has a very specific meaning, so I wouldn't use that here, if you don't have a better idea I guess this could be utils.
If it's hard to name something, then likelly it's in the wrong place. I would keep that function with the code responsible for other HTTP requests and call file/dir e.g. `api.rs`

integration_tests/src/ffmpeg.rs
wkozyra95:
```
    let mut handle = Command::new("bash")
        .arg("-c")
        .arg(format!("ffmpeg -f flv -listen 1 -i {output_address} -vcodec copy -f flv - | ffplay -f flv -i - "))
        .stdout(Stdio::null())
        .stderr(Stdio::null())
        .stdin(Stdio::null())
        .spawn()?;
 ```
 works for me
 
 I merged my PR, so when you rebase you only need to add `   .stdin(Stdio::null())`

integration_tests/examples/demo/generators.rs
JBRS307: Changed to utils for now as it is only one function.

integration_tests/src/ffmpeg.rs
JBRS307: Well, on mac this solution leaves ffplay as an orphaned process which has to be killed manually in the end.

integration_tests/examples/demo/inputs/whip.rs
JBRS307: - [x] Done

integration_tests/examples/demo/inputs/mp4.rs
JBRS307: - [x] Done

integration_tests/examples/demo/inputs.rs
JBRS307: - [x] Done

integration_tests/src/ffmpeg.rs
JBRS307: - [x] Done with ffplay autoexit

PR #1240, opened 2025-08-21
Proper termination of `peer_connections` on WHEP output

wkazmierczak:


compositor_pipeline/src/pipeline/webrtc/whep_output/peer_connection.rs
```diff
@@ -41,12 +48,16 @@ pub(crate) struct PeerConnection {
 impl PeerConnection {
     pub async fn new(
         ctx: &Arc<PipelineCtx>,
-        video_encoder: Option<VideoEncoderOptions>,
-        audio_encoder: Option<AudioEncoderOptions>,
+        video_encoder: &Option<VideoEncoderOptions>,
+        audio_encoder: &Option<AudioEncoderOptions>,
```
noituri, line 53: NIT:
I think it's more common to use `Option<&T>` than `&Option<T>` but I'm fine with either

compositor_pipeline/src/pipeline/webrtc/whep_output/state.rs
```diff
@@ -60,6 +62,40 @@ impl WhepOutputsState {
...
+                            "Session {session_id:?} not found for {output_id:?}"
+                        )));
+                    }
+                }
+                None => {
+                    return Err(WhipWhepServerError::NotFound(format!(
+                        "{output_id:?} not found"
+                    )));
+                }
+            }
```
wkozyra95:
```suggestion
            let Some(output) = guard.get_mut(output_id) else {
                    return Err(WhipWhepServerError::NotFound(format!(
                        "{output_id:?} not found"
                    )));
            }
            let Some(pc) = output.sessions.remove(session_id)  else {
                   return Err(WhipWhepServerError::NotFound(format!(
                        "Session {session_id:?} not found for {output_id:?}"
                   )));
            }
            pc
```

wkozyra95: or you could chain `map_err` and `?`

compositor_pipeline/src/pipeline/webrtc/whep_output/peer_connection.rs
wkozyra95: with Option<&T> you will need to use `as_ref`` in calling code, with this approach it depends, so I would leave this version

compositor_pipeline/src/pipeline/webrtc/whep_output/peer_connection.rs
```diff
@@ -231,6 +235,65 @@ impl PeerConnection {
...
+    pub fn attach_cleanup_when_pc_failed(
+        &self,
+        outputs: WhepOutputsState,
+        output_id: &OutputId,
+        session_id: &Arc<str>,
+    ) {
+        let last_connected_time = Arc::new(Mutex::new(None::<Instant>));
+
+        self.pc.on_peer_connection_state_change(Box::new({
+            let pc_clone = self.pc.clone();
```
wkozyra95: are you sure all of those clones are necessary? maybe some diffrent codes structure could simplify it?

compositor_pipeline/src/pipeline/webrtc/whep_output/state.rs
```diff
@@ -45,13 +45,15 @@ impl WhepOutputsState {
...
+        peer_connection: PeerConnection,
     ) -> Result<Arc<str>, WhipWhepServerError> {
         let mut guard = self.0.lock().unwrap();
         match guard.get_mut(output_id) {
             Some(output) => {
                 let session_id: Arc<str> = Arc::from(Uuid::new_v4().to_string());
-                output.sessions.insert(session_id.clone(), peer_connection);
+                output
+                    .sessions
+                    .insert(session_id.clone(), Arc::from(peer_connection));
```
wkozyra95: You can drop arc here, PeerConnection already has an Arc inside

compositor_pipeline/src/pipeline/webrtc/server/create_whep_session.rs
```diff
@@ -89,7 +87,12 @@ pub async fn handle_create_whep_session(
     let sdp_answer = peer_connection.negotiate_connection(offer).await?;
     trace!("SDP answer: {}", sdp_answer.sdp);
 
-    let session_id = outputs.add_session(&output_id, Arc::new(peer_connection))?;
+    let session_id = outputs.add_session(&output_id, peer_connection.clone())?;
+
+    let cleanup_session_handler =
+        create_cleanup_session_handler(outputs, output_id.clone(), session_id.clone());
+
```
wkozyra95:
```suggestion
```

compositor_pipeline/src/pipeline/webrtc/whep_output/cleanup_session_handler.rs
```diff
@@ -0,0 +1,30 @@
...
+use compositor_render::OutputId;
+use tracing::warn;
+
+use crate::pipeline::webrtc::whep_output::state::WhepOutputsState;
+
+pub(crate) fn create_cleanup_session_handler(
+    outputs: WhepOutputsState,
+    output_id: OutputId,
+    session_id: Arc<str>,
+) -> impl Fn() -> std::pin::Pin<Box<dyn std::future::Future<Output = ()> + Send>>
```
wkozyra95:
```suggestion
) -> impl Fn() -> Pin<Box<dyn Future<Output = ()> + Send>>
```
If it still does not fit in one line, then maybe extract it somehow.

compositor_pipeline/src/pipeline/webrtc/whep_output/peer_connection.rs
```diff
@@ -289,3 +319,57 @@ fn register_codecs(
     }
     Ok(())
 }
+
+fn handle_peer_connection_cleanup<F>(
```
wkozyra95: handle_cleanup_on_disconnect or something like that, current name implies that this function should be called when cleanup is needed, while in fact it is only detecting need for cleanup

compositor_pipeline/src/pipeline/webrtc/whep_output/peer_connection.rs
```diff
@@ -289,3 +319,57 @@ fn register_codecs(
...
+        }
+        RTCPeerConnectionState::Failed | RTCPeerConnectionState::Disconnected => {
+            if let Ok(mut handle) = cleanup_task_handle.lock() {
+                if handle.is_none() {
+                    // schedule task only if none is pending, crucial in transitions failed <-> disconnected
+                    let task =
+                        tokio::spawn(maybe_cleanup_peer_connection(pc, remove_session_handler));
+                    *handle = Some(task);
+                }
+            }
```
wkozyra95:
```suggestion
            if let Ok(None @ mut handle) = cleanup_task_handle.lock() {
                    // schedule task only if none is pending, crucial in transitions failed <-> disconnected
                    let task =
                        tokio::spawn(maybe_cleanup_peer_connection(pc, remove_session_handler));
                    *handle = Some(task);
            }
```
I think something like this would work

PR #1260, opened 2025-08-27
[vk-video] Add keyframe info to encoded chunks

noituri:


vk-video/src/lib.rs
```diff
@@ -169,17 +169,24 @@ impl H264Profile {
...
     pub pts: Option<u64>,
 }
 
+/// Represents a chunk of encoded video data returned by the encoder.
+pub struct EncodedOutputChunk<T> {
+    pub data: T,
+    pub pts: Option<u64>,
+    pub is_keyframe: bool,
+}
+
```
noituri, line 189: I can't really think of better names. I'm open to suggestions

jerzywilczek: its fine

PR #1269, opened 2025-08-29
[vk-video] Move resource transitions outside of `WgpuTexturesEncoder::encode`

noituri:


vk-video/src/lib.rs
```diff
@@ -305,6 +305,7 @@ impl WgpuTexturesEncoder {
     ///
     /// # Safety
     /// - The texture cannot be a surface texture
+    /// - The texture has to be transitioned to [`wgpu::TextureUses::RESOURCE`] usage
```
jerzywilczek: I would make the docs more helpful here, maybe show an example transition that would work for this?

jerzywilczek (changes requested):

vk-video/src/lib.rs
```diff
@@ -305,6 +305,24 @@ impl WgpuTexturesEncoder {
...
+    ///           state: wgpu::TextureUses::RESOURCE,
+    ///           selector: None,
+    ///       }]
+    ///       .into_iter(),
+    ///   );
+    ///   queue.submit([encoder.finish()]);
+    ///
+    ///   // Now you can use `WgpuTexturesEncoder::encode` on `texture`
+    ///   # }
+    ///   ```
```
jerzywilczek: use the empty wgpu backend instead

jerzywilczek (approved):

PR #1275, opened 2025-09-01
Added `MP4` output to interactive example.

### LINKED ISSUE #1218: Add MP4 output to interactive example

JBRS307:
Closes #1218


integration_tests/examples/demo/outputs/mp4.rs
```diff
@@ -0,0 +1,292 @@
...
+
+            builder = match audio_selection {
+                Some(Mp4RegisterOptions::SetAudioStream) => {
+                    builder.with_audio(Mp4OutputAudioOptions::default())
+                }
+                Some(Mp4RegisterOptions::Skip) | None => builder,
+                _ => unreachable!(),
+            };
+
+            if builder.video.is_none() || builder.audio.is_none() {
```
wkozyra95: wrong condition

wkozyra95:
I renamed the PR, The main purpose of those tags is to visually filter out stuf that it does not interest you. e.g. if you see `[ts-sdk]` while looking for changes in rust you know that you can skip that commit. Those tags should be one simple world that reasonably will be used in other PRs. It should not be a reference to a specific larger task. It should inform the reader what parts of code this change touches

In this case it would be fine to `[examples]` or `[integration_tests]`

integration_tests/examples/demo/outputs/mp4.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub fn prompt(self) -> Result<Self> {
+        let mut builder = self;
+        let env_path = env::var(MP4_OUTPUT_PATH).unwrap_or_default();
+
+        let default_path = examples_root_dir().join("example_output.mp4");
+        let path_output =
+            Text::new("Output path (absolute or relative to 'smelter/integration_tests'):")
+                .with_initial_value(&env_path)
+                .with_default(default_path.to_str().unwrap())
+                .prompt()?;
```
JBRS307:
```suggestion
                .prompt_skippable()?;
```
And handle it

integration_tests/src/assets.rs
JBRS307, line 1: Check if this works by downloading all

integration_tests/examples/demo/inputs/mp4.rs
```diff
@@ -35,41 +39,30 @@ impl Mp4InputBuilder {
...
-                        continue;
-                    }
-                },
-            };
-            break;
-        }
+        let env_path = env::var(MP4_INPUT_PATH).unwrap_or_default();
+
+        let path_input =
+            Text::new("Input path (absolute or relative to 'smelter/integration_tests'):")
```
wkozyra95: Not tested, but if I run `cargo run -p integration_tests --example demo` (and the same for demo_server) from root of the repo, the relative path will be resolved as you described

wkozyra95 (approved):

integration_tests/examples/demo/inputs/mp4.rs
JBRS307: Handling of relative paths done, however take a quick look please in case I overlooked something.

JBRS307:
> Providing invalid path shuts down entire CLI

- [x] Fixed

integration_tests/examples/demo/inputs/mp4.rs
```diff
@@ -35,41 +39,38 @@ impl Mp4InputBuilder {
...
+                .with_initial_value(&env_path)
+                .prompt_skippable()?;
+
+            match path_input {
+                Some(path) if !path.trim().is_empty() => {
+                    let path = resolve_path(path.into())?;
+                    if path.exists() {
+                        break builder.with_path(path);
+                    } else {
+                        error!("Path is not valid");
```
wkozyra95, line 55: This fixes if the path does not exist, but cli still exists if register is failing for other reasons, e.g. I provided non mp4 file to trigger that.

In general it is fine if this examples does not have extensive error handling,  and exits in case of some errors, but handing when register request is failing seems like the most basic case we should handle.

Btw this applies to all requests, not just mp4 output

PR #1281, opened 2025-09-02
Update declink implementation for new queue sync mechanism

wkozyra95:
After the recent queue sync rework DeckLink implementation was not updated. This PR fixes the synchronization mechanism.

### DIFF
--- a/compositor_pipeline/src/pipeline/decklink/capture.rs
+++ b/compositor_pipeline/src/pipeline/decklink/capture.rs
@@ -1,5 +1,5 @@
 use std::{
-    sync::{Mutex, OnceLock, Weak},
+    sync::{Arc, Mutex, Weak},
     time::{Duration, Instant},
 };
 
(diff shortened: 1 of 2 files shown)


compositor_pipeline/src/pipeline/decklink/capture.rs
```diff
@@ -64,10 +65,12 @@ impl ChannelCallbackAdapter {
...
                 audio_sender,
                 span,
-                audio_resampler: Mutex::new(DynamicResampler::new(output_sample_rate)),
+                audio_resampler: Mutex::new(DynamicResampler::new(ctx.mixing_sample_rate)),
                 input,
-                start_time: OnceLock::new(),
-                offset: Mutex::new(Duration::ZERO),
+                // 15 ms is a buffer that should be enough for frame to be delivered to queue
+                sync_point: ctx.queue_sync_point + Duration::from_millis(15),
+                audio_offset: Mutex::new(None),
```
wkozyra95, line 72: Not sure what guarantees we have about timestamps here, I'm assuming here that audio and video timestamps are not related to each other, and they don't start from zero

compositor_pipeline/src/pipeline/decklink/capture.rs
```diff
@@ -263,7 +270,9 @@ impl ChannelCallbackAdapter {
         input.flush_streams()?;
         input.start_streams()?;
 
-        *self.offset.lock().unwrap() = self.start_time().elapsed();
+        // it will reset on the next packet
+        *self.video_offset.lock().unwrap() = None;
+        *self.audio_offset.lock().unwrap() = None;
```
wkozyra95, line 275: not sure if format change will affect audio timestamps, so reseting just in case

wkazmierczak (approved):
LGTM, but I am not that familiar with DeckLink codebase

PR #1304, opened 2025-09-05
[examples] Added error handling for register and unregister requests.

JBRS307:
Simple error handling for client requests.

Additionally handled situation when server is not running (inside of `examples.rs` file).

### DIFF
--- a/integration_tests/examples/demo/smelter_state.rs
+++ b/integration_tests/examples/demo/smelter_state.rs
@@ -1,6 +1,6 @@
 use std::ops::Deref;
 
-use anyhow::Result;
+use anyhow::{Context, Result};
 use inquire::Select;
 use integration_tests::examples;
 use serde_json::json;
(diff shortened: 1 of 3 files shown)


integration_tests/examples/demo/smelter_state.rs
```diff
@@ -166,7 +175,12 @@ impl SmelterState {
         }
 
         let unregister_route = format!("input/{}/unregister", to_delete);
-        examples::post(&unregister_route, &json!({}))?;
+
+        let unregister_result = examples::post(&unregister_route, &json!({}));
```
wkozyra95: I would recommend keep using `?` and print the error in the last step.
Also, you never want to silently ignore errors, even in places where returning `Err` is expected, I would normally use a debug! at minimunm.

JBRS307: - [x] Done

wkozyra95 (approved):

PR #1305, opened 2025-09-05
Init chromium context outside of the pipeline (remove EventLoop)

wkozyra95:
required to implement reset endpoint. I plan to implement a reset by dropping the Pipeline object and creating a new one.

Additionally, it makes sense to extract it because cef initialization is not tied to a specific Pipeline. I'm not even sure if it would work in a Rust application that starts more than one Pipeline


noituri (approved):
LGTM

PR #1310, opened 2025-09-08
Add `/reset` endpoint

wkozyra95:
`/reset` would sound better, but after this call, you still need to send  start request, so it would be misleading

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -82,7 +82,7 @@ impl ApiState {
         }
     }
 
-    pub fn restart(&self) -> Result<(), ApiError> {
+    pub fn reset(&self) -> Result<(), ApiError> {
         let mut guard = self.pipeline.lock().unwrap();
         guard.take();
 
--- a/src/routes.rs
+++ b/src/routes.rs
@@ -68,8 +68,8 @@ pub fn routes(state: Arc<ApiState>) -> Router {
         Ok(Response::Ok {})
     }
 
-    async fn handle_restart(State(state): State<Arc<ApiState>>) -> Result<Response, ApiError> {
-        tokio::task::spawn_blocking(move || state.restart())
+    async fn handle_reset(State(state): State<Arc<ApiState>>) -> Result<Response, ApiError> {
+        tokio::task::spawn_blocking(move || state.reset())
             .await
             .unwrap()?;
         Ok(Response::Ok {})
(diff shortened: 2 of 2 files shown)


src/state.rs
```diff
@@ -62,16 +61,29 @@ impl ApiState {
...
-        })
+        }))
     }
 
-    pub(crate) fn pipeline(&self) -> MutexGuard<'_, Pipeline> {
-        self.pipeline.lock().unwrap()
+    pub fn pipeline(&self) -> Result<Arc<Mutex<Pipeline>>, ApiError> {
+        match self.pipeline.lock().unwrap().clone() {
+            Some(pipeline) => Ok(pipeline),
+            None => self.restart(),
```
wkozyra95: This can happen if the restart failed and left smelter in dead state. Restart here might fix the issue, but the main reason for it is to return specific error why restart failed.

compositor_pipeline/src/pipeline/webrtc/server.rs
```diff
@@ -73,10 +73,20 @@ impl WhipWhepServer {
 
     async fn new(port: u16) -> Result<Self, InitPipelineError> {
         let addr = SocketAddr::from(([0, 0, 0, 0], port));
-        match tokio::net::TcpListener::bind(addr).await {
-            Ok(listener) => Ok(Self { listener }),
-            Err(err) => Err(InitPipelineError::WhipWhepServerInitError(err)),
+        let mut last_error: Option<std::io::Error> = None;
```
wkozyra95, line 76: I did not hit that issue during testing, but I suspect it can happen if there are some pending requests, so I added a retry on startup

src/state.rs
```diff
@@ -62,16 +62,35 @@ impl ApiState {
...
-        })
+        }))
     }
 
-    pub(crate) fn pipeline(&self) -> MutexGuard<'_, Pipeline> {
-        self.pipeline.lock().unwrap()
+    pub fn pipeline(&self) -> Result<Arc<Mutex<Pipeline>>, ApiError> {
+        match self.pipeline.lock().unwrap().clone() {
+            Some(pipeline) => Ok(pipeline),
+            None => Err(ApiError {
```
wkozyra95, line 76: I initially planned to attempt reset in this case, but reset has to be called from `spawn_blockin` if it's triggered from async context, and this would require `pipeline` to be async. I think this would be unnecessarily complex with low added value.

src/state.rs
```diff
@@ -72,18 +73,24 @@ impl ApiState {
...
         let mut guard = self.pipeline.lock().unwrap();
         guard.take();
+
         let options =
             pipeline_options_from_config(&self.config, &self.runtime, &self.chromium_context);
         let pipeline = Arc::new(Mutex::new(Pipeline::new(options)?));
-        *guard = Some(pipeline.clone());
-        Ok(pipeline)
+        *guard = Some(pipeline);
+        Ok(())
```
noituri, line 93: This won't clean up running `cef::Browser`s. (Probably todo for me)

noituri: nvm, it should work

PR #1321, opened 2025-09-10
Fix RTCP sync with immediate rollover

wkozyra95:
- Add tests
- Fix case where first packet and first sender report are on different "sides" of RTP rollover
- Switch intermediate calculations to seconds. Duration is causing too many issues with negative numbers

### DIFF
--- a/compositor_pipeline/src/pipeline/rtp/rtp_input/rtcp_sync/sync_test.rs
+++ b/compositor_pipeline/src/pipeline/rtp/rtp_input/rtcp_sync/sync_test.rs
@@ -269,7 +269,7 @@ fn test_rtcp_sync_pts_with_rollover_before_sender_report_second_stream() {
     assert_duration_eq(stream_2_first_pts, Duration::from_millis(100), PREC_RUNTIME);
 
     // after 10 seconds
-    let stream_1_second_pts = stream_1.pts_from_timestamp(stream_1_first_rtp_timestamp + 10_000); // rolled over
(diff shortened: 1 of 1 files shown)


compositor_pipeline/src/pipeline/rtp/rtp_input/rtcp_sync/sync_test.rs
```diff
@@ -0,0 +1,453 @@
...
+    let stream_1_first_pts = stream_1.pts_from_timestamp(stream_1_first_rtp_timestamp);
+    thread::sleep(Duration::from_millis(100));
+    let stream_2_first_pts = stream_2.pts_from_timestamp(stream_2_first_rtp_timestamp);
+
+    assert_duration_eq(stream_1_first_pts, Duration::ZERO, PREC_RUNTIME);
+    assert_duration_eq(stream_2_first_pts, Duration::from_millis(100), PREC_RUNTIME);
+
+    // after 10 seconds
+    let stream_1_second_pts = stream_1.pts_from_timestamp(stream_1_first_rtp_timestamp + 10_000); // rolled over
+    let stream_2_second_pts = stream_2.pts_from_timestamp(5_000);
```
noituri: Isn't the `stream_2` that is rolled over? (I'm talking about the comment at the end of 272 line in case it does not show up on your screen)

wkozyra95: Yes, I was writing this test by copying the previous one and switching timestamps.

PR #1336, opened 2025-09-11
Split HLS input logic (fix context panic on move)

wkozyra95:
In the past, we had trouble with ffmpeg context segfaulting on move. It looks like closure has to be defined with `move`, otherwise referenced variable leak.

This PR:
- Splits initialization from the demuxing loop. 
- We don't need a channel anymore that returns int result from thread

### DIFF
--- a/compositor_pipeline/src/pipeline/hls/hls_input.rs
+++ b/compositor_pipeline/src/pipeline/hls/hls_input.rs
@@ -500,7 +500,7 @@ fn input_with_dictionary_and_interrupt<F>(
     interrupt_fn: F,
 ) -> Result<context::Input, ffmpeg_next::Error>
 where
-    F: FnMut() -> bool,
+    F: FnMut() -> bool + 'static,
 {
     unsafe {
         let mut ps = avformat_alloc_context();


compositor_pipeline/src/pipeline/hls/hls_input.rs
```diff
@@ -454,6 +446,51 @@ fn timestamp_to_secs(timestamp: f64, time_base: ffmpeg_next::Rational) -> f64 {
...
+    }
+}
+
+struct FfmpegInputContext {
+    ctx: context::Input,
+}
+
+impl FfmpegInputContext {
+    fn new(url: &Arc<str>, should_close: Arc<AtomicBool>) -> Result<Self, ffmpeg_next::Error> {
+        let ctx = input_with_dictionary_and_interrupt(
```
wkozyra95, line 470: Internally, this calls `Box::into_raw`, so on every input registration, we are leaking closure object

compositor_pipeline/src/pipeline/hls/hls_input.rs
```diff
@@ -454,6 +446,51 @@ fn timestamp_to_secs(timestamp: f64, time_base: ffmpeg_next::Rational) -> f64 {
...
+            url,
+            Dictionary::from_iter([("protocol_whitelist", "tcp,hls,http,https,file,tls")]),
+            // move is required even though types do not require it
+            move || should_close.load(Ordering::Relaxed),
```
noituri, line 474: Add `'static` to this parameter, then it won't compile without `move`

PR #1347, opened 2025-09-15
[examples] Created instruction for interactive example.

JBRS307:
- [ ] Merge after #1345

### DIFF
--- a/integration_tests/examples/demo/INSTRUCTION.md
+++ b/integration_tests/examples/demo/INSTRUCTION.md
@@ -31,7 +31,7 @@ Sets the bearer token to be used when connecting to the `WHIP` server provided b
 
 ### `WHIP_OUTPUT_BEARER_TOKEN`
 
-Bearer token required to connect to the `WHIP` server that Smelter streams to.
+Bearer token used to connect to the `WHIP` server that Smelter streams to.
 
 ### `WHIP_OUTPUT_URL`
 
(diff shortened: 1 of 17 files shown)


integration_tests/examples/demo/INSTRUCTION.md
```diff
@@ -0,0 +1,46 @@
...
+
+Bearer token required to connect to the `WHIP` server that Smelter streams to.
+
+### `WHIP_OUTPUT_URL`
+
+URL of the `WHIP` server that Smelter should connect to.
+
+### `WHEP_OUTPUT_BEARER_TOKEN`
+
+Bearer token required that shall be used to connect to `WHEP` server provided by Smelter.
```
wkozyra95: token for whep is not required

integration_tests/examples/demo/INSTRUCTION.md
```diff
@@ -0,0 +1,46 @@
...
+
+Path at which `mp4` output file should be saved.
+
+### `RTP_INPUT_PATH`
+
+Path to file that will be used as source of `RTP` stream.
+
+### `WHIP_INPUT_BEARER_TOKEN`
+
+Bearer token required that shall be used to connect to `WHIP` server provided by Smelter.
```
wkozyra95: - This sentence could be understand in multiple ways, wording it implies that this env is required which is not true.
- I don't think the required here was used correctly, (Required bearer token might be correct, but still sound wierd)
- shall is correct grammatically, but my understanding is that this is very archaic/oldfashioned form.

I would say e.g "Default value for bearer token used... . Token value is required when registering input."

wkozyra95 (approved):

PR #1352, opened 2025-09-16
[integration-tests] Added new audio tests.

### LINKED ISSUE #1243: Create new audio mixing tests.
Create audio mixing tests that add new audio input after some time has passed.

JBRS307:
Closes #1243 

- [ ] Snapshots membraneframework-labs/video_compositor_snapshot_tests#65

- [x] Add new snapshot that changes frequency after some time to correctly test audio offsets.

### DIFF
--- a/integration-tests/src/pipeline_tests/audio_only.rs
+++ b/integration-tests/src/pipeline_tests/audio_only.rs
@@ -542,8 +542,9 @@ pub fn single_input_aac_mp4() -> Result<()> {
 }
 
 /// Single frequency input that changes after 5 seconds. Input starts streaming 2 seconds before
-/// start request with offset set to 2 seconds. This should result in 2 seconds of silence, 5
-/// seconds of lower frequency and higher frequency after that time.
+/// start request with offset set to 2 seconds. 
+///
+/// Play  2 seconds of silence, 5 seconds of lower frequency and higher frequency after that time.
 #[test]
 fn audio_offset_variable_frequency() -> Result<()> {
     const OUTPUT_DUMP_FILE: &str = "audio_offset_variable_frequency_output.rtp";
@@ -617,9 +618,10 @@ fn audio_offset_variable_frequency() -> Result<()> {
     Ok(())
 }
 
-/// Single frequency input that changes after 5 seconds. Input starts streaming 2 seconds before
-/// start request with no offset. This should result in approx. 3 seconds of lower frequency and
-/// higher frequency after that.
+/// Use input that changes frequency after 5 seconds. Input starts streaming 2 seconds before
+/// start request with no offset. 
+///
+/// Play approx. 3 seconds of lower frequency and higher frequency after that.
 #[test]
 fn audio_no_offset_variable_frequency() -> Result<()> {
     const OUTPUT_DUMP_FILE: &str = "audio_no_offset_variable_frequency_output.rtp";
--- a/integration-tests/src/pipeline_tests/audio_only.rs
+++ b/integration-tests/src/pipeline_tests/audio_only.rs
@@ -546,8 +546,8 @@ pub fn single_input_aac_mp4() -> Result<()> {
 ///
 /// Play  2 seconds of silence, 5 seconds of lower frequency and higher frequency after that time.
 #[test]
-fn audio_offset_variable_frequency() -> Result<()> {
-    const OUTPUT_DUMP_FILE: &str = "audio_offset_variable_frequency_output.rtp";
+fn audio_early_streaming_with_offset() -> Result<()> {
+    const OUTPUT_DUMP_FILE: &str = "audio_early_streaming_with_offset_output.rtp";
     let instance = CompositorInstance::start(None);
     let input_1_port = instance.get_port();
     let output_port = instance.get_port();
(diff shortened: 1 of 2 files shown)


integration_tests/src/tests/audio_only.rs
```diff
@@ -206,6 +206,241 @@ pub fn audio_mixing_no_offset() -> Result<()> {
     Ok(())
 }
 
+/// Two audio input streams mixed together with the same volume.
+/// Second one joins after 10 seconds of `thread::sleep`.
+///
+/// Play audio for 20 seconds.
+#[test]
+pub fn audio_mixing_addition_with_offset() -> Result<()> {
```
JBRS307: I am not exactly sure if this name is appropriate but didn't have any better ideas, same with the other test.

wkozyra95: What is the goal of this test?

JBRS307: On second thought, I removed version the "no offset" version because it didn't exactly check what i wanted it to check. This now tests 2 things.
1. How does smelter behave on `update` request. If everything works as it should without any issues.
2. As we've talked about it before, it checks if the testing method can handle smoothly handle adding new track to the mix. In case of test, it also checks if default tolerance is tolerant enough.

integration_tests/src/pipeline_tests/audio_only.rs
```diff
@@ -206,6 +207,118 @@ pub fn audio_mixing_no_offset() -> Result<()> {
     Ok(())
 }
 
+/// Two audio input streams mixed together with the same volume after update request.
+/// Second one joins after 10 seconds of `thread::sleep`.
+///
+/// Play audio for 20 seconds.
+#[test]
+pub fn audio_mixing_track_insertion_with_offset() -> Result<()> {
+    const OUTPUT_DUMP_FILE: &str = "audio_mixing_addition_with_offset_output.rtp";
```
wkozyra95: keep file and test name consistent

JBRS307: - [x] Done

JBRS307:
Tests with `variable_frequency` in their name are the ones that start streaming before `start` request.

integration-tests/src/pipeline_tests/audio_only.rs
```diff
@@ -427,3 +540,154 @@ pub fn single_input_aac_mp4() -> Result<()> {
 
     Ok(())
 }
+
+/// Single frequency input that changes after 5 seconds. Input starts streaming 2 seconds before
+/// start request with offset set to 2 seconds. This should result in 2 seconds of silence, 5
+/// seconds of lower frequency and higher frequency after that time.
```
wkozyra95:
```suggestion
/// Single frequency input that changes after 5 seconds. Input starts streaming 2 seconds before
/// start request with offset set to 2 seconds. 
///
/// Play  2 seconds of silence, 5 seconds of lower frequency and higher frequency after that time.
```

integration-tests/src/pipeline_tests/audio_only.rs
```diff
@@ -427,3 +540,154 @@ pub fn single_input_aac_mp4() -> Result<()> {
...
+        audio::ValidationMode::Artificial,
+        AudioValidationConfig::default(),
+    )?;
+
+    Ok(())
+}
+
+/// Single frequency input that changes after 5 seconds. Input starts streaming 2 seconds before
+/// start request with no offset. This should result in approx. 3 seconds of lower frequency and
+/// higher frequency after that.
```
wkozyra95:
```suggestion
/// Use input that changes frequency after 5 seconds. Input starts streaming 2 seconds before
/// start request with no offset. 
///
/// Play approx. 3 seconds of lower frequency and higher frequency after that.
```

integration-tests/src/pipeline_tests/audio_only.rs
```diff
@@ -427,3 +540,154 @@ pub fn single_input_aac_mp4() -> Result<()> {
 
     Ok(())
 }
+
+/// Single frequency input that changes after 5 seconds. Input starts streaming 2 seconds before
+/// start request with offset set to 2 seconds. This should result in 2 seconds of silence, 5
+/// seconds of lower frequency and higher frequency after that time.
+#[test]
+fn audio_offset_variable_frequency() -> Result<()> {
```
wkozyra95: Instead of variable frequency, I would use `two_frequencies`, `two_frequencies_input` or `frequency_change`

However, focusing on that in the name is already a mistake. You are using this specific input to test something. Test name should reflect the goal, and using this specific input does not change what you are testing, it just makes it easier to detect errors.

integration-tests/src/pipeline_tests/audio_only.rs
```diff
@@ -427,3 +540,154 @@ pub fn single_input_aac_mp4() -> Result<()> {
...
+            "type": "rtp_stream",
+            "transport_protocol": "tcp_server",
+            "port": input_1_port,
+            "audio": {
+                "decoder": "opus"
+            },
+        }),
+    )?;
+
+    let audio_input_1 = input_dump_from_disk("variable_frequency_opus_audio.rtp")?;
```
wkozyra95: consider moving this higher(before register input), it might affect test timing

PR #1353, opened 2025-09-16
[examples] Added input fields in `WHEP` player.

### LINKED ISSUE #1350: Fix `WHEP` output name in interactive example.
It is called `input_whep` and should be `output_whep`.

### LINKED ISSUE #1351: Add input fields in `WHEP` player to change output name and bearer token in interactive example.

JBRS307:
Closes #1351,
Closes #1350 

Additional changes:
- [x] Fixed wrong `WHEP` output name.


integration_tests/examples/demo/whep.html
```diff
@@ -40,13 +91,32 @@
       const endpointUrl = urlParams.get('url');
       const bearerToken = urlParams.get('token');
 
+      const urlInput = document.getElementById("url-input");
```
wkozyra95, line 94: check `chrome://webrtc-internals` page when running this, I think this will leave old connections.

Also in terms of approach. The most common pattern would be to refresh that page when testing something, so it might be better to update url based on input change, so it survives page refersh

JBRS307: - [x] Seems done

PR #1356, opened 2025-09-17
New demo v012: Handling Multiple Mp4

piotrsnow:


ts/examples/smelter-app/src/server/roomState.ts
```diff
@@ -34,6 +34,10 @@ export type RegisterInputOptions =
   | {
       type: 'kick-channel';
       kickChannelId: string;
+    }
+  | {
+      type: 'local-mp4';
+      mp4Url: string;
```
wkozyra95: is this url or file path?

ts/examples/smelter-app/src/server/roomState.ts
```diff
@@ -110,6 +114,51 @@ export class RoomState {
...
       throw new Error('Add kick support');
+    } else if (opts.type === 'local-mp4') {
+      console.log('Adding local mp4');
+      let mp4Path = path.join(process.cwd(), 'mp4s', opts.mp4Url);
+      let mp4Name = opts.mp4Url;
+      if (opts.mp4Url === 'random') {
+        const mp4sDir = path.join(process.cwd(), 'mp4s');
+        let files: string[] = [];
+        try {
+          files = fs.readdirSync(mp4sDir);
```
wkozyra95: you can read that dir at the start once, at keep that in some global list. It's fine to use sync on startup, but I wouldn't use it in request handler

ts/examples/smelter-app/src/server/roomState.ts
```diff
@@ -110,6 +114,51 @@ export class RoomState {
...
+      }
+      const inputId = `${this.idPrefix}::local::sample_streamer::${Date.now()}`;
+
+      const fileNameWithoutExt = mp4Name.replace(/\.mp4$/i, '');
+      const formattedName = fileNameWithoutExt
+        .split(/[_\- ]+/)
+        .map(word => word.charAt(0).toUpperCase() + word.slice(1))
+        .join(' ');
+
+      if (fs.existsSync(mp4Path)) {
```
wkozyra95:
```suggestion
      if (await fs.pathExist(mp4Path)) {
```
import fs from fs-extra instead of standard library

ts/examples/smelter-app/src/server/roomState.ts
wkozyra95: If you want to support random, I would suggest
```
source: { sourceType: 'path', path: string } | { sourceType: 'random' }
```
of 
```
source: { path: string } | 'random'
```
I would avoid having special string values

piotrsnow: ok final type:
`{
type:'local-mp4';
source:{
fileName: string;
url: string;
}}
`

ts/examples/smelter-app/src/server/roomState.ts
piotrsnow: as we will not use random anymore, sync read stays in init method

PR #1375, opened 2025-09-19
Update to rust `1.90`

noituri:


integration_tests/examples/demo/outputs.rs
```diff
@@ -58,6 +58,7 @@ pub enum OutputProtocol {
     Hls,
 }
 
+#[allow(dead_code)]
 #[derive(Debug, EnumIter, Display, Clone)]
```
wkozyra95, line 62: @JBRS307 are those types need or are they left by mistake?

JBRS307: This is dead code, I am going to remove it after merge

PR #1376, opened 2025-09-19
Add WHEP input

wkazmierczak:


integration-tests/examples/whep_client.rs
```diff
@@ -0,0 +1,93 @@
...
+const OUTPUT_PORT: u16 = 9002;
+
+fn main() {
+    run_example(client_code);
+}
+
+fn client_code() -> Result<()> {
+    start_ffmpeg_rtmp_receive(OUTPUT_PORT)?;
+    std::thread::sleep(Duration::from_millis(2000));
+
```
wkozyra95: don't add new examples, add it to demo instead

smelter-api/src/input/whep.rs
```diff
@@ -0,0 +1,51 @@
...
+    FfmpegVp9,
+
+    /// Use hardware decoder based on Vulkan Video.
+    ///
+    /// This should be faster and more scalable than the ffmpeg decoder, if the hardware and OS
+    /// support it.
+    ///
+    /// This requires hardware that supports Vulkan Video. Another requirement is this program has
+    /// to be compiled with the `vk-video` feature enabled (enabled by default on platforms which
+    /// support Vulkan, i.e. non-Apple operating systems and not the web).
```
wkozyra95, line 47: I think we talked about that description before. I think we changed that in website documentation, but not all in code comments were updated, so this still gets copy pasted everywhere

smelter-core/src/pipeline/webrtc/whep_input.rs
```diff
@@ -0,0 +1,188 @@
...
+        let (input_samples_sender, input_samples_receiver) = bounded(5);
+
+        let span = span!(
+            Level::INFO,
+            "WHEP client task",
+            input_id = input_id.to_string()
+        );
+        let rt = ctx.tokio_rt.clone();
+        let ctx_clone = ctx.clone();
+        rt.spawn(
```
wkozyra95:
```suggestion
        ctx.tokio_rt.spawn(
```
You would need to clone if you did `ctx_clone.tokio_rt`

smelter-core/src/pipeline/webrtc/whep_input.rs
```diff
@@ -0,0 +1,188 @@
...
+                    }
+                    Err(err) => init_confirmation_sender.send(Err(err)).unwrap(),
+                }
+            }
+            .instrument(span),
+        );
+
+        let (session_url, client) =
+            wait_with_deadline(init_confirmation_receiver, WHEP_INIT_TIMEOUT)?;
+        Ok((
```
wkozyra95: Either:
- This function should have absolute minimum of logic, `init_whep_client` should already return the same type as new_input.
- Logic involving returning result by channels and waiting for deadline should not be in top level code like that

smelter-core/src/pipeline/webrtc/whep_input.rs
```diff
@@ -0,0 +1,188 @@
...
+) -> Result<(Url, Arc<WhipWhepHttpClient>), WebrtcClientError> {
+    let client = WhipWhepHttpClient::new(&options.endpoint_url, &options.bearer_token)?;
+    let (video_preferences, video_codecs_params) =
+        resolve_video_preferences(&ctx, options.video_preferences)?;
+    let pc = RecvonlyPeerConnection::new(&ctx, &video_codecs_params).await?;
+
+    let _video_transceiver = pc.new_video_track(video_codecs_params).await?;
+    let _audio_transceiver = pc.new_audio_track().await?;
+
+    let (session_url, answer) = exchange_sdp_offers(&pc, &client).await?;
```
wkozyra95: I would consider inlining that function and maybe extracting on_track instead

smelter-api/src/input/whep.rs
wkazmierczak: I'll do that in following PR

PR #1384, opened 2025-09-24
`FFmpeg` version installation detection and libraries download.

### LINKED ISSUE #1357: More portable releases (FFmpeg compatibility)
Currently, we have a problem with distributing releases via binaries. Specific binary depends on concrete versions of libav* libraries e.g.

`readelf -d target/release/main_process`

```
  0x0000000000000001 (NEEDED)       Shared library: [libavutil.so.59]
  0x0000000000000001 (NEEDED)       Shared library: [libavformat.so.61]
  0x0000000000000001 (NEEDED)       Shared library: [libavfilter.so.10]
  0x0000000000000001 (NEEDED)       Shared library: [libavdevice.so.61]
  0x0000000000000001 (NEEDED)       Shared library: [libswscale.so.8]
  0x0000000000000001 (NEEDED)       Shared library: [libswresample.so.5]
  0x0000000000000001 (NEEDED)       Shared library: [libavcodec.so.61]
  0x0000000000000001 (NEEDED)       Shared library: [libopus.so.0]
  0x0000000000000001 (NEEDED)       Shared library: [libssl.so.3]
  0x0000000000000001 (NEEDED)       Shared library: [libcrypto.so.3]
  0x0000000000000001 (NEEDED)       Shared library: [libgcc_s.so.1]
  0x0000000000000001 (NEEDED)       Shared library: [libm.so.6]
  0x0000000000000001 (NEEDED)       Shared library: [libc.so.6]
  0x0000000000000001 (NEEDED)       Shared library: [ld-linux-x86-64.so.2]
```

The primary goal is to remove the need to have a specific version of ffmpeg installed. FFmpeg brings a lot of dependencies that we also need, so it is fine if we still require users to install some FFmpeg version.

Potential approach to consider:
Write a wrapper script on smelter binaries that detects if the current FFmpeg version is compatible. If it is not it downloads appropriate binaries and use them instead of system libraries

JBRS307:
Closes #1357 

Changes:
- `libav` libraries are now downloaded when required `FFmpeg` version is not found on the system.
    They are downloaded from [smelter-labs/smelter-dep-prebuilds](https://github.com/smelter-labs/smelter-rc)
- New `dependency_check` binary is responsible for checking and downloading libraries.
- `LD_LIBRARY_PATH` was removed from wrapper script. `runpath` is now set at compile time.
---
This detects if appropriate `FFmpeg` version is installed and downloads appropriate libraries if not.

- [x] MacOS
- [x] Linux
---
Tests:
- [x] Bundle on MacOS with MP4 offline processing
- Bundle on Linux (Ubuntu docker) with MP4 offline processing
  - [x] arm64
  - [x] x86_64
- ts-sdk
  - [x] MacOS
  - [x] Linux - ~first example run (the one that downloads smelter) fails~ SOLVED
- Membrane plugin
  - [x] MacOS
  - [x] Linux
  ---
- [x] Remove setting of `LD_LIBRARY_PATH` (use `runpath` to point to dependencies)

### DIFF
--- a/smelter-core/src/pipeline/encoder/ffmpeg_h264.rs
+++ b/smelter-core/src/pipeline/encoder/ffmpeg_h264.rs
@@ -57,8 +57,6 @@ impl VideoEncoder for FfmpegH264Encoder {
             ("threads", "0"),
         ]);
 
-        // I am not sure at the moment if default FFmpeg settings are the best for videotoolbox,
-        // but they are definitely better than these
         if codec_name != "libopenh264" && codec_name != "h264_videotoolbox" {
             ffmpeg_options.append(&[
                 // Quality-based VBR (0-51)
--- a/tools/src/bin/dependency_check.rs
+++ b/tools/src/bin/dependency_check.rs
@@ -125,14 +125,11 @@ fn unpack_ffmpeg(executable_dir: &Path) -> Result<()> {
 
     let mut decoder = flate2::bufread::GzDecoder::new(gz_input);
 
-    let decompressed_bytes = std::io::copy(&mut decoder, &mut tar_bytes)
+    let decompressed_size = std::io::copy(&mut decoder, &mut tar_bytes)
         .with_context(|| "Failed to decompress the archive")?;
-    trace!(decompressed_bytes);
+    trace!(decompressed_size);
 
-    let tar_bytes = tar_bytes;
-
-    let mut tar_archive = Archive::new(&tar_bytes[..]);
-    tar_archive.unpack(executable_dir)?;
+    Archive::new(&tar_bytes[..]).unpack(executable_dir)?;
 
     Ok(())
 }


.cargo/config.toml
```diff
@@ -1,2 +1,6 @@
 [build]
 rustflags = "--cfg web_sys_unstable_apis"
+
+[env]
+FFMPEG_VERSION = "1.0"
+FFMPEG_URL = "https://swmansion.com"
```
JBRS307: I have decided to use `env!()` instead ot `option_env!()` in `dependency_check.rs` to not complicate the CI and code itself. This are just dummies that are overwritten in code needed for rust analyzer to not recognize usage of env!() as an error.

src/bin/dependency_check.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+    Ok(command_result)
+}
+
+fn check_ffmpeg_command(required_ffmpeg_version: &str) -> Result<bool> {
+    let ffmpeg_result = Command::new("ffmpeg").arg("-version").output();
+    match ffmpeg_result {
+        Ok(ffmpeg_output) => {
+            let ffmpeg_output = String::from_utf8(ffmpeg_output.stdout)?.trim().to_string();
+            let re = Regex::new(r"(?m)^ffmpeg version \D*(\d+\.\d+)")?;
```
JBRS307: Is this regex ok or should it be more/less specific?

tools/src/bin/package_for_release/bundle_linux.rs
```diff
@@ -46,15 +46,42 @@ fn bundle_app(
         info!("Bundling smelter without web rendering");
     }
 
+    let ffmpeg_version = ffmpeg_version()?;
+    let ffmpeg_url = ffmpeg_url(&ffmpeg_version)?;
+
+    let rustc_envs = vec![
+        ("FFMPEG_VERSION", ffmpeg_version),
+        ("FFMPEG_URL", ffmpeg_url),
+    ];
```
JBRS307: The envs set in `.cargo/config.toml` are overwritten here.

src/bin/dependency_check.rs
wkozyra95: it's fine, but copy example output from terminal and write tests for it. It's more for documentation than for testing

.cargo/config.toml
wkozyra95: Can this section be defined in Cargo. toml?
If yes I would prefer to move that bin `tools` crate
If no  I would prefer option_env!() instead

.github/workflows/shellcheck.yml
```diff
@@ -5,13 +5,17 @@ on:
         branches: [master]
         paths:
             - tools/release.sh
-            - tools/src/bin/package_for_release/linux_runtime_wrapper.sh
+            - tools/src/bin/package_for_release/linux_runtime_wrapper_cef.sh
+            - tools/src/bin/package_for_release/linux_runtime_wrapper_no_cef.sh
+            - tools/src/bin/package_for_release/macos_runtime_wrapper.sh
```
wkozyra95: I think `tools/src/bin/package_for_release/*.sh` would work

.github/workflows/shellcheck.yml
```diff
@@ -28,4 +32,4 @@ jobs:
 
             - name: shellcheck
               run: |
-                shellcheck ./tools/release.sh ./tools/src/bin/package_for_release/linux_runtime_wrapper.sh
+                shellcheck ./tools/release.sh ./tools/src/bin/package_for_release/linux_runtime_wrapper_cef.sh ./tools/src/bin/package_for_release/linux_runtime_wrapper_no_cef.sh ./tools/src/bin/package_for_release/macos_runtime_wrapper.sh
```
wkozyra95:
```suggestion
                shellcheck \
                  ./tools/release.sh \
                  ./tools/src/bin/package_for_release/linux_runtime_wrapper_cef.sh \
                  ./tools/src/bin/package_for_release/linux_runtime_wrapper_no_cef.sh \
                  ./tools/src/bin/package_for_release/macos_runtime_wrapper.sh
```

src/bin/dependency_check.rs
wkozyra95: I would do the same for other non trivial regexes

.cargo/config.toml
JBRS307: It cannot. I will change that to `option_env!()` and run the binary on CI to check if it does not panic.

src/bin/ffmpeg_output_macos.txt
```diff
@@ -0,0 +1,12 @@
+ffmpeg version 8.0 Copyright (c) 2000-2025 the FFmpeg developers
```
wkozyra95: I would just inline this text as a string in the test directly

src/bin/dependency_check.rs
```diff
@@ -0,0 +1,223 @@
...
+mod dependency_check_test {
+    use super::*;
+    use std::{fs, path::PathBuf};
+
+    // TODO: (@jbrs) test for linux output
+    #[test]
+    fn ffmpeg_macos_regex_test() {
+        let crate_root = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        let ffmpeg_output =
+            fs::read_to_string(crate_root.join("src/bin/ffmpeg_output_macos.txt")).unwrap();
```
wkozyra95: as mentioned below I suggest inlining text here. However, for similar cases in tests in the future you can usually use `include_str!` macro to read file content during compilation

tools/src/bin/package_for_release/linux_runtime_wrapper_cef.sh
```diff
@@ -4,11 +4,10 @@
 #   Runtime wrapper which provides paths to native libs used by the web renderer
 #  
 
-set -eo pipefail
-
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-export LD_LIBRARY_PATH="$SCRIPT_DIR/lib:$LD_LIBRARY_PATH"
+export LD_LIBRARY_PATH="$SCRIPT_DIR/lib:$LD_LIBRARY_PATH:$SCRIPT_DIR/ffmpeg_lib"
```
wkozyra95: is this syntax correct if `LD_LIBRARY_PATH` is not defined

tools/src/bin/dependency_check.rs
wkozyra95, line 1: move to tools

wkozyra95 (commented):
Add more context to errors, (test on empty ubuntu container)

tools/src/bin/package_for_release/linux_runtime_wrapper_cef.sh
JBRS307: - [x] Done

tools/src/bin/dependency_check.rs
JBRS307: - [x] Done

JBRS307:
In the end I've decided to remove `set -eo pipefail` to scripts as `-e` makes Smelter not start in case of `dependency_check` failure and `-o pipefail` does completely nothing.

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,290 @@
...
+    match ffmpeg_result {
+        Ok(ffmpeg_output) => {
+            let ffmpeg_output = String::from_utf8(ffmpeg_output.stdout)?.trim().to_string();
+            match match_ffmpeg_version(&ffmpeg_output) {
+                Some(version) if version == required_ffmpeg_version => Ok(true),
+                Some(version) => {
+                    warn!(
+                        installed_ffmpeg_version = version,
+                        required_ffmpeg_version,
+                        "Inatelled version doesn't match the required version."
```
wkozyra95:
```suggestion
                        "Installed version doesn't match the required version."
```

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,290 @@
...
+#[cfg(target_os = "macos")]
+fn check_ffmpeg_homebrew(required_ffmpeg_version: &str) -> Result<bool> {
+    let required_ffmpeg_version_brew =
+        &required_ffmpeg_version[..required_ffmpeg_version.find(".").unwrap_or(1)];
+    let brew_output = Command::new("brew").arg("list").output()?;
+    let brew_output = String::from_utf8(brew_output.stdout)?.trim().to_string();
+
+    let ffmpeg_string = format!("ffmpeg@{required_ffmpeg_version_brew}");
+
+    let re = Regex::new(&format!("(?m){ffmpeg_string}"))?;
```
wkozyra95: I don't think this separation helps with readability 
```suggestion
    let re = Regex::new(&format!("(?m)ffmpeg@{required_ffmpeg_version_brew}"))?;
```

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,290 @@
...
+#[cfg(target_os = "linux")]
+fn check_ffmpeg(required_ffmpeg_version: &str) -> Result<bool> {
+    check_ffmpeg_command(required_ffmpeg_version)
+}
+
+#[cfg(target_os = "macos")]
+fn check_ffmpeg(required_ffmpeg_version: &str) -> Result<bool> {
+    let command_result = check_ffmpeg_command(required_ffmpeg_version)?;
+    if !command_result {
+        info!("Checking if ffmpeg is installed as homebrew keg-only");
```
wkozyra95:
```suggestion
        info!("Checking if FFmpeg is installed as homebrew keg-only");
```

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,290 @@
...
+    Ok(libraries)
+}
+
+fn download_ffmpeg(executable_dir: &Path, ffmpeg_url: &str) -> Result<()> {
+    let response = get(ffmpeg_url)
+        .with_context(|| format!("Failed to download FFmpeg libraries. URL: {ffmpeg_url}"))?;
+    let content = response.bytes()?;
+
+    fs::write(executable_dir.join(FFMPEG_ARCHIVE_NAME), &content)
+        .with_context(|| "Failed to save downloaded FFmpeg libraries to file")?;
```
wkozyra95: It's a nitpick, but it is more common to refer to libraries like libavcodec as libav, not "FFmpeg libraries"

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,290 @@
...
+        None => {
+            warn!("Failed to parse FFmpeg version.");
+            None
+        }
+    }
+}
+
+#[cfg(target_os = "macos")]
+fn check_ffmpeg_homebrew(required_ffmpeg_version: &str) -> Result<bool> {
+    let required_ffmpeg_version_brew =
```
wkozyra95, line 198: When looking at this code, I don't know what `required_ffmpeg_version`  is what it will be after removing before the dot

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,290 @@
...
+
+#[cfg(target_os = "linux")]
+const FFMPEG_ARCHIVE_NAME: &str = "ffmpeg.tar.xz";
+
+fn main() -> Result<()> {
+    tracing_subscriber::fmt().init();
+
+    #[allow(clippy::option_env_unwrap)]
+    let (required_ffmpeg_version, ffmpeg_url) = (
+        option_env!("FFMPEG_VERSION").unwrap(),
```
wkozyra95: add
- Comment describing where those values are set,
- Describe what format they have
- maybe validate them

wkozyra95: I would also consider extracting accessing and validating those envs to separate functions, and instead passing non descriptive value like `required_ffmpeg_version` calling that function when necessary

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,290 @@
...
+        Some(0) => {}
+        Some(code) => bail!("`tar` command failed with code: {code}."),
+        None => bail!("`tar` command failed."),
+    }
+    if tar_code != Some(0) {
+        bail!("`tar` command failed with code: {tar_code:?}.");
+    }
+
+    let re = if cfg!(target_os = "linux") {
+        Regex::new(r"^lib[a-zA-Z]+\.so\.\d+$")?
```
wkozyra95: for simpler regexes you don't have to write tests, but at least add a comment what you are expecting to match
e.g. are you matching file name or entire path

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,290 @@
...
+                continue;
+            }
+        };
+        fs::rename(lib, executable_dir.join(FFMPEG_LIB_DIR).join(libname))?;
+    }
+
+    Ok(())
+}
+
+fn find_ffmpeg_libs(dirpath: PathBuf, re: &Regex) -> Result<Vec<PathBuf>> {
```
wkozyra95: either better name or better comments, I have no idea how to use this function without reading and understanding its entire implementation

tools/src/bin/package_for_release/bundle_macos.rs
```diff
@@ -24,28 +24,144 @@ pub fn bundle_macos_app() -> Result<()> {
...
+        let basename = path_buf.file_name().unwrap().to_str().unwrap();
+        let exit_code = Command::new("install_name_tool")
+            .args([
+                "-change",
+                path,
+                &format!("@rpath/{basename}"),
+                smelter_bin_path.to_str().unwrap(),
+            ])
+            .spawn()?
+            .wait()?
```
noituri: You can replace that with just `.status()`

JBRS307: - [x] Done

tools/src/bin/dependency_check.rs
JBRS307: - [x] Done

tools/src/bin/dependency_check.rs
JBRS307: - [x] Done

tools/src/bin/dependency_check.rs
JBRS307: - [x] Done

tools/src/bin/dependency_check.rs
JBRS307: - [x] Done

tools/src/bin/dependency_check.rs
JBRS307: - [x] Done

JBRS307: > * maybe validate them

~Well, the easiest way of validating them would be to use `env!()` instead of `option_env!()` because if these envs were not set at compile time then it simply would not compile. However that would cause terrible issues with `clippy` and `rust-analyzer` and would require us to set dummy values somewhere, so it doesn't throw errors.~ As for runtime validation these envs are nicely validated already in `tools::package_for_release::utils`.

EDIT: I forgot we've resigned from doing it the `env!()` way. Another option of validating it is to simply run it on CI, if it doesn't fail it means it was nicely compiled.

tools/src/bin/package_for_release/bundle_macos.rs
```diff
@@ -24,28 +24,143 @@ pub fn bundle_macos_app() -> Result<()> {
...
+            bail!("Command \"install_name_tool\" failed with exit code: {exit_code:?}");
+        }
+    }
+
+    info!("Create tar.gz archive.");
+    let exit_code = Command::new("tar")
+        .args(["-czvf", output_name, "smelter"])
+        .current_dir(workdir)
+        .spawn()?
+        .wait()?
```
noituri: Same here, you can use `status()`

.github/workflows/shellcheck.yml
```diff
@@ -5,13 +5,15 @@ on:
...
             - tools/release.sh
-            - tools/src/bin/package_for_release/linux_runtime_wrapper.sh
+            - tools/src/bin/package_for_release/*.sh
             - .github/workflows/shellcheck.yml
     pull_request:
         types: [opened, synchronize]
         paths:
             - tools/release.sh
-            - tools/src/bin/package_for_release/linux_runtime_wrapper.sh
+            - tools/src/bin/package_for_release/linux_runtime_wrapper_cef.sh
```
wkozyra95: define paths consistently

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,310 @@
...
+
+fn main() -> Result<()> {
+    tracing_subscriber::fmt().init();
+    // Matches if version is in correct format i.e. `x.y`
+    let re = Regex::new(r"^\d+\.\d+$")?;
+    if !re.is_match(required_ffmpeg_version()) {
+        // NOTE: This should NEVER happen, especially that analogical check is performed at compile
+        // time
+        bail!("Version in invalid format: {}", required_ffmpeg_version());
+    }
```
wkozyra95: move that to required_ffmpeg_version

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,310 @@
...
+        Regex::new(r"^lib[a-zA-Z]+\.so\.\d+$")?
+    } else if cfg!(target_os = "macos") {
+        // Matches dynamic libav library with major version only on macos
+        // E.g. libavcodec.62.dylib
+        Regex::new(r"^lib[a-zA-Z]+\.\d+\.dylib$")?
+    } else {
+        panic!("Unknown platform");
+    };
+
+    let ffmpeg_libs_paths = extract_libav_libraries(executable_dir.join(FFMPEG_DOWNLOAD_DIR), &re)
```
wkozyra95: This is unnecessarily splitting the processing. I would preferably:
- rename extract_libav_libraries to find_libav_libraries
- do all the processing on each file separately

or at least

- to all the renaming inside extract_libav_libraries

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,310 @@
...
+const FFMPEG_LIB_DIR: &str = "ffmpeg_lib";
+const FFMPEG_DOWNLOAD_DIR: &str = "ffmpeg_download";
+
+#[cfg(target_os = "macos")]
+const FFMPEG_ARCHIVE_NAME: &str = "ffmpeg.tar.gz";
+
+#[cfg(target_os = "linux")]
+const FFMPEG_ARCHIVE_NAME: &str = "ffmpeg.tar.xz";
+
+/// FFMPEG_VERSION is set at compile time and contains FFmpeg version in `x.y` format which was used to compile
```
wkozyra95:
```suggestion
/// FFMPEG_VERSION is set at compile time (in package_for_release bin) and contains FFmpeg version in `x.y` format which was used to compile
```

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,310 @@
...
+/// Smelter. FFmpeg version is found by matching `ffmpeg -version` output during compilation.
+fn required_ffmpeg_version() -> &'static str {
+    #[allow(clippy::option_env_unwrap)]
+    option_env!("FFMPEG_VERSION").unwrap()
+}
+
+/// FFMPEG_URL is set at compile time and provides the URL of the FFmpeg prebuilt release with
```
wkozyra95:
```suggestion
/// FFMPEG_URL is set at compile time (in package_for_release bin) and provides the URL of the FFmpeg prebuilt release with
```

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    let executable_path =
+        env::current_exe().with_context(|| "Failed to get current executable directory.")?;
+    let executable_dir = executable_path.parent();
+    let executable_dir = match executable_dir {
+        Some(path) => path.to_path_buf(),
+        None => bail!("Failed to get current executable directory."),
+    };
+
+    let lib_exists = fs::exists(executable_dir.join(FFMPEG_LIB_DIR))
```
wkozyra95: What if someone hit Control C in previous run. The easiest pattern to implement is to:
- At the end write empty `.ready` file
- Check here for existence of `.ready` file
- If not make sure to remove FFMPEG_LIB_DIR and recreate it

JBRS307:
@wkozyra95 When you start review please take a look at smelter-labs/smelter-dep-prebuilds#4 first.

ts/smelter-node/src/manager/locallySpawnedInstance.ts
```diff
@@ -24,25 +24,33 @@ const VERSION = `v0.4.1`;
 type ManagedInstanceOptions = {
   port: number;
   workingdir?: string;
-  executablePath?: string;
+  mainExecutablePath?: string;
+  dependencyCheckPath?: string;
   enableWebRenderer?: boolean;
 };
 
+type ExecutablePath = {
```
wkozyra95:
```suggestion
type ExecutablePaths = {
```

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+fn prepare_dependencies(executable_dir: &Path) -> Result<()> {
+    download_ffmpeg(executable_dir).with_context(|| "libav download failed.")?;
+
+    let ffmpeg_archive_path = executable_dir.join(FFMPEG_ARCHIVE_NAME);
+
+    // Archive contains only `libav` directory in the root. This directory contains all smelter
+    // libav dependencies.
+    let tar_status = Command::new("tar")
```
wkozyra95: use lib instead https://docs.rs/tar/latest/tar/

If we don't have to add another dependecy we should do that

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,220 @@
...
+/// FFMPEG_URL is set at compile time (in package_for_release bin) and provides the URL of the FFmpeg prebuilt release with
+/// libav dynamic libraries that should be downloaded if the required FFmpeg version is not installed
+fn ffmpeg_url() -> &'static str {
+    #[allow(clippy::option_env_unwrap)]
+    option_env!("FFMPEG_URL").unwrap()
+}
+
+fn main() -> Result<()> {
+    tracing_subscriber::fmt().init();
+
```
wkozyra95: This should respect smelter logger envs

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,220 @@
...
+            None
+        }
+    }
+}
+
+#[cfg(target_os = "macos")]
+fn check_ffmpeg_homebrew() -> Result<bool> {
+    let required_ffmpeg_version_brew =
+        &required_ffmpeg_version()[..required_ffmpeg_version().find(".").unwrap_or(1)];
+    let brew_output = Command::new("brew").arg("list").output()?;
```
wkozyra95: What if there is no brew?
- Is the error clear?
- Can we provide better behavior? e.g. always download dependencies in that case? Would our binaries even work without homebrew?

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,220 @@
...
+        Err(error) => return Err(anyhow::Error::from(error).context("`tar` command failed")),
+    }
+
+    Ok(())
+}
+
+fn download_ffmpeg(executable_dir: &Path) -> Result<()> {
+    let response = get(ffmpeg_url())
+        .with_context(|| format!("Failed to download libav libraries. URL: {}", ffmpeg_url()))?;
+    let content = response.bytes()?;
```
wkozyra95, line 111: It's probably fine in this case, but normally, when you're downloading anything to a disc, you should get some writer or stream interface and pipe it to disk write. This solution requires that at some point entire file is in memory

tools/src/bin/package_for_release/linux_runtime_wrapper_no_cef.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
wkozyra95: aren't all those scripts the same now?

tools/src/bin/dependency_check.rs
JBRS307: I am aware, but as you said, it is fine in this case as it won't take more RAM than a chrome tab.

tools/src/bin/dependency_check.rs
JBRS307: Done for stdio only

ts/smelter-node/src/manager/locallySpawnedInstance.ts
JBRS307: Done. And I also commented out some code so it still works with current smelter version.

tools/src/bin/dependency_check.rs
JBRS307: We have to add it as a dependency, but only for tools crate. I think it is still worth because we won't have to run command on someone else's machine.

tools/src/bin/dependency_check.rs
JBRS307: Changed:
- If brew is not installed it is handled inside the function, so it returns `Ok(false)` in this case.
- Error message says downloaded libraries require the same dependencies as `FFmpeg` and will not work without them present.
  - It is not specified how they should be installed. It will work if they're installed by `homebrew` or inside of `/usr/local/lib`
  - Using it without `homebrew` is possible in theory, however it requires a lot of work because EACH FFmpeg dependency would have to be built from source

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,307 @@
...
+    pub(super) fn init_logger() {
+        let logger_level = match env::var("SMELTER_LOGGER_LEVEL") {
+            Ok(level) => level,
+            Err(_) => "info".to_string(),
+        };
+        let stdio_logger_level = match env::var("SMELTER_STDIO_LOGGER_LEVEL") {
+            Ok(level) => level,
+            Err(_) => logger_level.clone(),
+        };
+        let default_logger_format = LoggerFormat::Compact;
```
wkozyra95: default is json

ts/smelter-node/src/manager/locallySpawnedInstance.ts
```diff
@@ -71,17 +82,33 @@ class LocallySpawnedInstanceManager implements SmelterManager {
...
       if (err.stderr) {
         console.error(err.stderr);
       }
       if (err.stdout) {
         console.error(err.stdout);
       }
-    });
+    };
+
+    // This code will be uncomented with the next smelter release which is going to
```
wkozyra95, line 97: Maybe uncomment it, make a release in rc repo and change url to binary to rc repo, or better write this code with the assumption that dependency path does not exists

tools/src/bin/dependency_check.rs
```diff
@@ -118,11 +119,34 @@ fn download_ffmpeg(executable_dir: &Path) -> Result<()> {
...
 
+fn unpack_ffmpeg(executable_dir: &Path) -> Result<()> {
+    let gz_archive_path = executable_dir.join(FFMPEG_GZIP_ARCHIVE_NAME);
+    let tar_archive_path = executable_dir.join(FFMPEG_TAR_ARCHIVE_NAME);
+
+    let gz_input = BufReader::new(
+        File::open(&gz_archive_path).with_context(|| "Failed to open downloaded archive")?,
+    );
+    let mut tar_output = File::create(&tar_archive_path)?;
+
```
wkozyra95, line 125: try to implement that without intermediate

smelter-core/src/pipeline/encoder/ffmpeg_h264.rs
```diff
@@ -56,7 +56,10 @@ impl VideoEncoder for FfmpegH264Encoder {
             // Auto number of threads
             ("threads", "0"),
         ]);
-        if codec_name != "libopenh264" {
+
+        // I am not sure at the moment if default FFmpeg settings are the best for videotoolbox,
```
wkozyra95: remove the comment

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,297 @@
...
+    );
+    let mut tar_bytes: Vec<u8> = vec![];
+
+    let mut decoder = flate2::bufread::GzDecoder::new(gz_input);
+
+    let decompressed_bytes = std::io::copy(&mut decoder, &mut tar_bytes)
+        .with_context(|| "Failed to decompress the archive")?;
+    trace!(decompressed_bytes);
+
+    let tar_bytes = tar_bytes;
```
wkozyra95:
```suggestion
```

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    let mut tar_archive = Archive::new(&tar_bytes[..]);
+    tar_archive.unpack(executable_dir)?;
```
wkozyra95:
```suggestion
    Archive::new(&tar_bytes[..]).unpack(executable_dir)?;
```

tools/src/bin/dependency_check.rs
```diff
@@ -0,0 +1,297 @@
...
+    let gz_archive_path = executable_dir.join(FFMPEG_GZIP_ARCHIVE_NAME);
+
+    let gz_input = BufReader::new(
+        File::open(&gz_archive_path).with_context(|| "Failed to open downloaded archive")?,
+    );
+    let mut tar_bytes: Vec<u8> = vec![];
+
+    let mut decoder = flate2::bufread::GzDecoder::new(gz_input);
+
+    let decompressed_bytes = std::io::copy(&mut decoder, &mut tar_bytes)
```
wkozyra95:
```suggestion
    let mut tar_bytes: Vec<u8> = vec![];
    let decompressed_bytes = std::io::copy(&mut decoder, &mut tar_bytes)
```
also you are naming `_bytes` variables with actual data and variables that store the number of bytes, maybe use `_size` or `_size_bytes`

wkozyra95 (approved):

PR #1392, opened 2025-09-30
Add GPU selection env option

noituri:
This PR makes it possible to select specific GPU instead of relying on automatic selection.
On startup, all available device adapters are printed:
```sh
2025-09-30T12:14:48.113571Z  INFO smelter_core::graphics_context: Available adapters:
 - AdapterInfo { name: "AMD Radeon RX 7800 XT (RADV NAVI32)", device_type: DISCRETE_GPU, api_version: "1.4.318" , driver: "radv", driver_info: "Mesa 25.2.4-arch1.1",  vendor: 4098, device: 29822, supports_decoding: true, supports_encoding: true }
 - AdapterInfo { name: "AMD Radeon Graphics (RADV RAPHAEL_MENDOCINO)", device_type: INTEGRATED_GPU, api_version: "1.4.318" , driver: "radv", driver_info: "Mesa 25.2.4-arch1.1", vendor: 4098, device: 5710, supports_decoding: true, supports_encoding: true }
```

I tried to make the logs look similar to what we already had for WGPU (without vk-video)

Users can set `SMELTER_GPU_DEVICE_ID` to the device provided in the logs (e.g 5710),
or/and set  `SMELTER_GPU_DRIVER_NAME` to e.g "radv" or "Mesa 25" to filter out unwanted drivers

On vulkan, the device ID corresponds to PCI ID, so for the same device model the ID should be the same across machines.

This works both for `GraphicsContext` created with `vk-video` and `wgpu`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -127,7 +127,7 @@ fn try_read_config() -> Result<Config, String> {
         ),
         Err(_) => None,
     };
-    let gpu_driver_name = env::var("SMELTER_GPU_DRIVER_NAME").ok();
+    let gpu_driver_name = env::var("SMELTER_GPU_DEVICE_DRIVER").ok();
 
     const DEFAULT_STREAM_FALLBACK_TIMEOUT: Duration = Duration::from_millis(500);
     let stream_fallback_timeout = match env::var("SMELTER_STREAM_FALLBACK_TIMEOUT_MS") {


wkozyra95 (changes requested):
PR should be reviewable without running the code. Please if output is not obvious, include example stdout, or example value that should be provided, where somone should get that value.

It is especially true for stuff that not everyone can run on their devices.

wkozyra95:
> On vulkan, the device ID corresponds to PCI ID, so for the same device model the ID should be the same across machines.

I can test that when I'll be home, but based on description if I have 2 entries for the same device with different mesa version this env is not enough

wkozyra95 (commented):
If the output look similiar then it's solves the main issue, but I would prefer if logic for printing was not spred over the codebase. I would unify initialization to core crate and put logic for printing in the same place.

src/config.rs
```diff
@@ -118,6 +120,15 @@ fn try_read_config() -> Result<Config, String> {
...
     };
 
+    let gpu_device_id = match env::var("SMELTER_GPU_DEVICE_ID") {
+        Ok(id) => Some(
+            id.parse::<u32>()
+                .map_err(|_| "SMELTER_GPU_DEVICE_ID has to be a valid number")?,
+        ),
+        Err(_) => None,
+    };
+    let gpu_driver_name = env::var("SMELTER_GPU_DRIVER_NAME").ok();
```
wkozyra95:
```suggestion
    let gpu_driver_name = env::var("SMELTER_GPU_DEVICE_DRIVER").ok();
```
I would also prefer using consistent match if clipy allows it.

When adding this to docs, show example output and examples of what values could be provided here

noituri: Clippy prefers `.ok()`

PR #1395, opened 2025-09-30
Improve DEVELOPMENT.md instructions

wkozyra95:


DEVELOPMENT.md
```diff
@@ -174,7 +280,7 @@ cargo build -r --no-default-features
 export SMELTER_PATH=$(pwd)/target/release/main_process
```
JBRS307, line 300: If you are already doing this, I think it would be good idea to add info that without it smelter binary release is downloaded with info where it is placed.

JBRS307 (approved):

PR #1399, opened 2025-09-30
Rename compositor-chromium to libcef

wkozyra95:


wkozyra95:
@noituri I'm also considering renaming chromium_sys to libcef_sys or cef_sys. What do you think?

noituri:
> @noituri I'm also considering renaming chromium_sys to libcef_sys or cef_sys. What do you think?

I like libcef_sys

PR #1408, opened 2025-10-02
Extract common `webrtc input` modules

wkazmierczak:


smelter-core/src/pipeline/webrtc/video_input_processing_loop.rs
```diff
@@ -4,56 +4,46 @@ use crossbeam_channel::Sender;
...
+    } = video_track_ctx;
     let handle = VideoTrackThread::spawn(
-        &endpoint_id,
-        (
-            ctx.clone(),
-            video_decoder_mapping,
-            video_payload_type_mapping,
-            frame_sender,
-        ),
+        thread_instance_id,
```
wkozyra95: extract outside

smelter-core/src/pipeline/webrtc/audio_input_processing_loop.rs
```diff
@@ -3,47 +3,43 @@ use std::sync::Arc;
...
-        return Err(WhipWhepServerError::InternalError(
-            "No audio codecs negotiated".to_string(),
-        ));
-    };
-
-    let WhipWhepServerState { inputs, ctx, .. } = state;
-    let samples_sender =
-        inputs.get_with(&endpoint_id, |input| Ok(input.input_samples_sender.clone()))?;
-    let handle = AudioTrackThread::spawn(&endpoint_id, (ctx.clone(), samples_sender))?;
+pub(super) struct AudioInputTrackCtx {
```
wkozyra95:
```suggestion
pub(super) struct AudioInputLoop {
```
and make a function below a method

smelter-core/src/pipeline/webrtc/whip_input/process_tracks.rs
```diff
@@ -0,0 +1,102 @@
...
+            "No audio codecs negotiated".to_string(),
+        ));
+    };
+
+    let WhipWhepServerState { inputs, ctx, .. } = state;
+    let samples_sender =
+        inputs.get_with(&endpoint_id, |input| Ok(input.input_samples_sender.clone()))?;
+
+    let handle = AudioTrackThread::spawn(
+        format!("WHIP input audio, endpoint_id: {}", endpoint_id),
```
wkozyra95, line 44: I think this will lead to log "instance = "Input WHIP input audio, endpoint_id: some_id"". But already the way those thread utils log is not great, so I'm fine with leaving it this way and I'll address that separatlly

PR #1410, opened 2025-10-03
[WIP] Dynamic buffer (configurable per input type)

wkozyra95:


smelter-core/src/pipeline/utils/input_buffer.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    state: LatencyOptimizedBufferState,
+}
+
+impl LatencyOptimizedBuffer {
+    fn new(ctx: &PipelineCtx) -> Self {
+        Self {
+            sync_point: ctx.queue_sync_point,
+            desired_buffer: ctx.default_buffer_duration,
+            min_buffer: Duration::min(Duration::from_millis(20), ctx.default_buffer_duration),
```
wkozyra95, line 85: Those values might need to be adjusted, currentlly:
- target 80ms
- if above 130ms over 10 second then lower buffer
- if above 20ms but below 80ms increase buffer

smelter-core/src/pipeline/utils/input_buffer.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    fn pts_with_buffer(&mut self, pts: Duration) -> Duration {
+        const INCREMENT_DURATION: Duration = Duration::from_micros(100);
+
+        let next_pts = pts + self.dynamic_buffer;
+        if next_pts > self.sync_point.elapsed() + self.desired_buffer {
+            // ok
+        } else if next_pts > self.sync_point.elapsed() + self.min_buffer {
+            debug!(
```
wkozyra95, line 180: this will hit only if current PTS is between 20ms and 80ms ahead of a queue. For protocols like HLS it will most likely never hit

smelter-core/src/pipeline/utils/input_buffer.rs
```diff
@@ -0,0 +1,198 @@
...
+            InputBuffer::Const { buffer } => pts + *buffer,
+            InputBuffer::LatencyOptimized(buffer) => buffer.lock().unwrap().pts_with_buffer(pts),
+            InputBuffer::Adaptive(buffer) => buffer.lock().unwrap().pts_with_buffer(pts),
+        }
+    }
+}
+
+/// Buffer intended for low latency inputs, if input stream is not delivers on time
+/// it quickly increasing. However when buffer is stable for some time it starts to shrink to
+/// minimize the latency.
```
wkazmierczak:
```suggestion
/// Buffer intended for low latency inputs, if input stream is not delivered on time,
/// it quickly increases. However, when buffer is stable for some time it starts to shrink to
/// minimize the latency.
```

wkazmierczak (approved):

PR #1420, opened 2025-10-07
Extract common WebRTC Client utils

wkazmierczak:


smelter-core/src/pipeline.rs
```diff
@@ -32,6 +32,7 @@ mod input;
 mod instance;
 mod output;
 
+pub use crate::pipeline::webrtc::WebrtcClientError;
```
wkozyra95: nothing from this module shoud be public

PR #1427, opened 2025-10-09
Add `ffmpeg_options` (for muxer) for mp4 output

wkozyra95:


smelter-core/src/pipeline/hls/hls_output.rs
```diff
@@ -175,24 +176,16 @@ impl HlsOutput {
             .map_err(OutputInitError::FfmpegError)?;
 
         stream.set_time_base(VIDEO_TIME_BASE);
+        stream.update_codecpar(|codecpar| {
```
wkozyra95, line 179: unrelated to the change, but it is useful to reuse, plus it's safer this way because codecpar accessibility is limited

ts/smelter/src/types/output/hls.ts
```diff
@@ -30,7 +30,7 @@ export type HlsVideoEncoderOptions =
       /**
        * Raw FFmpeg encoder options. See [docs](https://ffmpeg.org/ffmpeg-codecs.html) for more.
        */
-      ffmpegOptions?: Extract<Api.VideoEncoderOptions, { type: 'ffmpeg_h264' }>['ffmpeg_options'];
+      ffmpegOptions?: Record<string, string>;
```
wkozyra95, line 33: Reusing generated type was not useful here and hard to read

PR #1437, opened 2025-10-14
Extract `WhipWhepHttpClient`

wkazmierczak:
Prepare webrtc client to reuse it in WHEP input and WHIP output


smelter-core/src/pipeline/webrtc/http_client.rs
wkozyra95, line 1: http_client.rs

PR #1439, opened 2025-10-15
Various fixes found during debugging crashes

jerzywilczek:


integration-tests/src/bin/benchmark/main.rs
```diff
@@ -65,6 +65,8 @@ fn main() {
         })
         .collect();
 
+    std::thread::sleep(std::time::Duration::from_millis(250));
```
wkozyra95, line 68: is this leftover after debuging?

noituri: It's a hack that makes the segfaults less prevalent. We figured that some of the segfaults were caused by threads not being awaited for on drop

jerzywilczek: combined with the other changes in this patch, it causes the benchmark to never segfault

src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![recursion_limit = "256"]
+
```
jerzywilczek, line 2: @wkozyra95 is this ok?
I do not understand why this is necessary, but it is.

PR #1443, opened 2025-10-16
[examples] Improve whip.html/whep.html

wkozyra95:


JBRS307 (approved):
Video's aspect ratio is not kept when shrinking viewport's width. If that is not a problem then everything is ok.

wkozyra95:
> Video's aspect ratio is not kept when shrinking viewport's width. If that is not a problem then everything is ok.

Good point, I assumed that `video[poster]` refered to poster styling while in fact it refers to any video element with poster

PR #1444, opened 2025-10-16
Fix WHIP input

wkazmierczak:


smelter-core/src/pipeline/webrtc/server/create_whip_session.rs
```diff
@@ -41,7 +42,12 @@ pub async fn handle_create_whip_session(
 
     let peer_connection = RecvonlyPeerConnection::new(&state.ctx, &video_codecs).await?;
 
-    let _video_transceiver = peer_connection.new_video_track(video_codecs).await?;
+    // When setting codec preferences, payload types should be compatible with those in the offer. Simplest way to achieve that is by setting defaults
+    let video_codecs_with_default_payload_types = reset_payload_types_to_default(video_codecs);
```
wkozyra95: What about whep? does new_video_track shouldn't also received zeroed version?
I would prefer to make that conversion as close to usage as possible

wkazmierczak: For WHEP, it is not necessary because we are the ones that send the offer. But i have tested it, and even with this payload_types reset, WHEP works fine so I can move that inside `new_video_track()`

smelter-core/src/pipeline/webrtc/supported_codec_parameters.rs
```diff
@@ -168,3 +118,15 @@ pub fn opus_codec_params() -> Vec<RTCRtpCodecParameters> {
         })
         .collect()
 }
+
+pub fn reset_payload_types_to_default(
```
wkozyra95: I would inline this:
- it is short
- used in one place
- function name is not not descriptive

PR #1449, opened 2025-10-20
Update comments in api video decoder options

wkazmierczak:


smelter-api/src/input/hls.rs
```diff
@@ -29,16 +29,10 @@ pub enum InputHlsCodec {
...
 
-    /// Use hardware decoder based on Vulkan Video.
-    ///
-    /// This should be faster and more scalable than the ffmpeg decoder, if the hardware and OS
-    /// support it.
-    ///
-    /// This requires hardware that supports Vulkan Video. Another requirement is this program has
-    /// to be compiled with the `vk-video` feature enabled (enabled by default on platforms which
-    /// support Vulkan, i.e. non-Apple operating systems and not the web).
+    /// (Required feature: vk-video)
```
wkazmierczak: Not sure if communicating this that way is fine

wkozyra95: No strong opinion, but if we have some format it's important to keep it consistent maybe it's better to just add Another sentecnce "Requires vk-video feature"

Either way event in this structure I would invert the order. Main description first, requirements latter

smelter-api/src/input/hls.rs
```diff
@@ -29,16 +29,10 @@ pub enum InputHlsCodec {
 #[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
 #[serde(rename_all = "snake_case", deny_unknown_fields)]
 pub enum HlsVideoDecoderOptions {
-    /// Use the software h264 decoder based on ffmpeg.
+    /// Software H264 decoder based on FFmpeg.
```
wkozyra95, line 32: that based on FFmpeg sounds wrong to me, but don't have better ides plus it's not related to the change so lets leave it

Also "Software", no sure if it's not possbile to use hardware accelerated one this way.

wkazmierczak: Currently we do not support providing any options to decoder, so we can use only "software" one. But ffmpeg recently added way to use hardware accelerated H264 decoder.

PR #1468, opened 2025-10-27
Fixed blinking when static tile is changed with new id.

### LINKED ISSUE #1425: Better behavior for `Tails` animation when some component is replaced with the new one ("different one")
Currently single tail blinks for the duration of the transition

JBRS307:
Closes #1425 

- Merge after #1467 
- ~Snapshots smelter-labs/smelter-snapshot-tests#68~
- Snapshots smelter-labs/smelter-snapshot-tests#69
---
Additional changes:
- Changed timestamps in already existing `tiles_transitions` tests (as mentioned [here](https://github.com/software-mansion/smelter/pull/1468#discussion_r2469985442))
---
Intended behaviour:
- When tile is changed whith a new id in the spot that was previously occupied and no animation occurs it should be visible from the beginning of transition
- Otherwise it should be visible only after transition is finished
---
TODO:
- [x] Transition fix
- [x] Tests
---
Before merging:
- [x] Delete `tiles_blink` temporary example.


integration-tests/examples/tiles_blink.rs
JBRS307, line 1: This file will be deleted before merging, I keep it for now just in case.

smelter-render/src/scene/tiles_component/interpolation.rs
```diff
@@ -20,6 +20,10 @@ impl ContinuousValue for Vec<Option<Tile>> {
             .enumerate()
             .filter_map(|(index, tile)| tile.as_ref().map(|tile| (&tile.id, index)))
             .collect();
+        let end_id_set: HashSet<&TileId> = end
+            .iter()
+            .flat_map(|tile| tile.as_ref().map(|tile| &tile.id))
```
JBRS307: This should be `filter_map()`

smelter-render/src/scene/tiles_component/interpolation.rs
JBRS307, line 1: Add some comments to explain this logic.

smelter-render/src/scene/tiles_component/interpolation.rs
```diff
@@ -36,6 +40,23 @@ impl ContinuousValue for Vec<Option<Tile>> {
...
+                        //   so tile disappears whenever any animation is triggered, for example margin change.
+                        // - Replaced tile does not exist in state represented by `end` (so no
+                        //   animation is triggered).
+                        start
+                            .iter()
+                            .flatten()
+                            .find(|start_tile| are_positions_equal(start_tile, tile))
+                            .and_then(|start_tile| match end_id_set.contains(&start_tile.id) {
+                                true => None,
+                                false => Some(ContinuousValue::interpolate(tile, tile, state)),
```
noituri: I think the interpolation from `tile` to `tile` does nothing. Returning just a `tile` here should work

smelter-render/src/scene/tiles_component/interpolation.rs
```diff
@@ -36,6 +40,23 @@ impl ContinuousValue for Vec<Option<Tile>> {
                             .as_ref()
                             .map(|old_tile| ContinuousValue::interpolate(old_tile, tile, state))
                     })
+                    .or_else(|| {
+                        // This handles case when a tile is swapped "in place" with entirely new tile
+                        // that has different ID.
```
wkozyra95: What is "This" here? This sentence could be understood in 2 ways:
- This or_else is called when description is true
- This closure is changing the behavior when description is true
Neither of those interpretation is true.

smelter-render/src/scene/tiles_component/interpolation.rs
```diff
@@ -36,6 +40,23 @@ impl ContinuousValue for Vec<Option<Tile>> {
...
+                        // It only takes effect when:
```
wkozyra95: What is it? You technically are describing the conditions, but it only make sense if someone already knows what this code is for. I had to read the code to understand this comment.

smelter-render/src/scene/tiles_component/interpolation.rs
```diff
@@ -36,6 +40,23 @@ impl ContinuousValue for Vec<Option<Tile>> {
...
+                        // - Exact position exists in the previous state represented by `start`
```
wkozyra95: "Exact position exists" What does it mean that position exists?

smelter-render/src/scene/tiles_component/interpolation.rs
```diff
@@ -52,3 +73,14 @@ impl ContinuousValue for Tile {
         }
     }
 }
+
+fn are_positions_equal(lhs: &Tile, rhs: &Tile) -> bool {
```
wkozyra95, line 78: Up to you, but we already have TilePosition struct, so you could change Tail to have 2 fields (id and position) and implement Eq trait on TilePosition.

integration-tests/src/render_tests/tiles_transitions_tests.rs
```diff
@@ -160,6 +160,124 @@ fn tiles_transitions_tests() {
...
+                "./tiles_transitions/start_with_3_inputs_all_id.scene.json"
+            )),
+            Step::RenderWithSnapshot(Duration::from_millis(0)),
+            Step::UpdateSceneJson(include_str!(
+                "./tiles_transitions/end_with_3_inputs_3_id_different_component.scene.json"
+            )),
+            Step::RenderWithSnapshot(Duration::from_millis(1)),
+            Step::RenderWithSnapshot(Duration::from_millis(150)),
+            Step::RenderWithSnapshot(Duration::from_millis(350)),
+            Step::RenderWithSnapshot(Duration::from_millis(500)),
```
wkozyra95, line 178:
```suggestion
            Step::RenderWithSnapshot(Duration::from_millis(100)),
            Step::RenderWithSnapshot(Duration::from_millis(400)),
            Step::RenderWithSnapshot(Duration::from_millis(500)),
```
Does not really matter, but I would use timestamp like this. Eaiser to see by file name, will be sorted in PR

smelter-render/src/scene/tiles_component/interpolation.rs
JBRS307: I have thought about that, but implementing `PartialEq` here would not be correct as the tolerance checking makes the relation intransitive.

integration-tests/src/render_tests/tiles_transitions_tests.rs
JBRS307: If that is the case I think it would be a good idea to change that for every test (I just copied these values from existing tests).

EDIT: Did that

smelter-render/src/scene/tiles_component/interpolation.rs
```diff
@@ -32,10 +36,25 @@ impl ContinuousValue for Vec<Option<Tile>> {
                     .get(&tile.id)
                     .and_then(|index| start.get(*index))
                     .and_then(|old_tile| {
+                        // Animation interpolation. Returns `Some(_)` only if for the tile from the `end` state
+                        // exists a tile with the same ID in the `start` state.
```
wkozyra95:
```suggestion
                        // For each tile that existed before the last update (exists in both `start` and `end`)
                        // interpolate between those 2 states.
```

smelter-render/src/scene/tiles_component/interpolation.rs
```diff
@@ -32,10 +36,25 @@ impl ContinuousValue for Vec<Option<Tile>> {
...
                         old_tile
                             .as_ref()
                             .map(|old_tile| ContinuousValue::interpolate(old_tile, tile, state))
                     })
+                    .or_else(|| {
+                        // This closure prevents tile from displaying empty square for the duration
+                        // of transition if the tile is swapped "in place" (i.e. no animation
+                        // occurs).
```
wkozyra95:
```suggestion
                        // Handle a new tile (`tile` did not exist in the `start` state). Tiles that existed before
                        // last update should be handled in the previous `and_then` clousure.
                        //
                        // - If any tile existed in the same position as the new tile before the transition and still exists
                        //   somewhere in the `end` state then do not show the new tile until the end of the transition.
                        // - Otherwise, show the new tile from the start of the transition.
```

wkozyra95 (approved):

PR #1482, opened 2025-10-29
[vk-video] Detect missing reference pictures

noituri:


smelter-core/src/pipeline/decoder/vulkan_h264.rs
```diff
@@ -36,6 +40,12 @@ impl VideoDecoderInstance for VulkanH264Decoder {
 
         let frames = match self.decoder.decode(chunk) {
             Ok(res) => res,
+            Err(DecoderError::ParserError(ParserError::ReferenceManagementError(
+                ReferenceManagementError::MissingFrame,
+            ))) => {
+                debug!("Vulkan H264 decoder detected missing frame.");
```
jerzywilczek:
```suggestion
                debug!("Vulkan H264 decoder detected a missing frame.");
```

vk-video/src/parser/reference_manager.rs
```diff
@@ -34,12 +37,14 @@ pub struct ReferenceId(usize);
...
     next_reference_id: ReferenceId,
-    previous_frame_num: usize,
+    previous_frame_num: u16,
     prev_pic_order_cnt_msb: i32,
     prev_pic_order_cnt_lsb: i32,
     MaxLongTermFrameIdx: MaxLongTermFrameIdx,
     prevFrameNumOffset: i64,
     previous_picture_included_mmco_equal_5: bool,
+    is_missing_frames: bool,
+    allow_gaps_in_frames: bool,
```
jerzywilczek: I'm not a fan of the naming here. `allow_gaps_in_frames` has the two problems that first, it sounds very similar to `gaps_in_frame_num_allowed`, which is an h264 flag describing a completely separate behavior, which we don't support, and second, is not very precise and could mean either that we allow some frames to be missing or that we allow chunks in the middle of a frame to be missing.

How about changing the first field to `detected_missed_frames`, and the second to `ignore_missed_frames`? Those names are still not as good as I would want them so I'm open to a discussion.

Also, I think we should change these names in the public API and every other place in the code

vk-video/src/parser/reference_manager.rs
```diff
@@ -667,6 +691,23 @@ impl ReferenceContext {
...
         Ok(reference_list)
     }
+
+    fn verify_frame_num(
+        &mut self,
+        sps: &SeqParameterSet,
+        header: &SliceHeader,
+    ) -> Result<(), ReferenceManagementError> {
+        let is_expected_frame_num = !sps.gaps_in_frame_num_value_allowed_flag
+            && header.frame_num != self.previous_frame_num + 1
```
jerzywilczek:
```suggestion
            && header.frame_num != self.previous_frame_num
```

It can be equal to the previous one.

> When gaps_in_frame_num_value_allowed_flag is equal to 0 and frame_num is not equal to PrevRefFrameNum and is not equal to ( PrevRefFrameNum + 1 ) % MaxFrameNum, the decoding process should infer an unintentional loss of pictures.

vk-video/src/device.rs
```diff
@@ -42,6 +42,16 @@ pub struct Rational {
     pub denominator: NonZeroU32,
 }
 
+/// Parameters for decoder creation
+#[derive(Debug, Clone, Copy, Default)]
+pub struct DecoderParameters {
+    /// If set to `false` and decoder detects missing frames, it will enter corrupted state
+    /// and won't decode any reference frames until IDR frame is provided.
```
jerzywilczek:
```suggestion
    /// If set to `false` and decoder detects missing frames, it will enter a corrupted state
    /// and won't decode any reference frames until an IDR frame is provided.
```

also, write what happens otherwise

vk-video/src/parser/reference_manager.rs
```diff
@@ -171,6 +194,7 @@ impl ReferenceContext {
         };
 
         self.previous_picture_included_mmco_equal_5 = header.includes_mmco_equal_5();
+        self.previous_frame_num = header.frame_num;
```
jerzywilczek: This should only happen if the current frame is a reference frame (that's why the variable is called `PrevRefFrameNum` in the spec). Does the assignment here take that into accout? I don't think it does, but I haven't looked at this chunk of code in a while, so I may be wrong

jerzywilczek (changes requested):

vk-video/src/parser/reference_manager.rs
jerzywilczek: also, is `previous_frame_num` the thing that is supposed to represent `PrevRefFrameNum` in our code? if yes, then we should rename it to `PrevRefFrameNum`, if not then we should add a separate variable for `PrevRefFrameNum`.

vk-video/src/parser/reference_manager.rs
noituri: Maybe it's just me but I think `ignore_missed_frames`  could be interpreted as "decoder won't throw an error, it will ignore those frames (we won't depend on those missing frames)".

What about `enable_missing_frame_detection` or something along those lines?

vk-video/src/parser/reference_manager.rs
```diff
@@ -72,14 +73,15 @@ impl ReferenceContext {
         *self = Self {
             pictures: ReferencePictures::default(),
             next_reference_id: ReferenceId::default(),
-            previous_frame_num: 0,
+            prevFrameNum: 0,
```
noituri, line 78: I'm not sure if we should preserve the `prevFrameNum` here or not. Couldn't find anything specific about it in the spec.

> Let prevFrameNum be equal to the frame_num of the previous picture in decoding order.

This sounds like we shouldn't reset it?

jerzywilczek: I'm not sure, leave it like this if it works

vk-video/src/parser/reference_manager.rs
jerzywilczek: How about something like this?

```rs
#[derive(Default)
enum MissedFrameHandling {
    /// When missed frames are detected, error on every subsequent frame that depends on them
    /// (i. e. fail on every frame until an IDR frame arrives)
    #[default]
    Strict,
    
    /// When missed frames are detected, try to decode later frames that depend on them anyway.
    /// This can produce decoded frames with very visible artifacts.
    Tolerant,
}
```

PR #1486, opened 2025-10-31
Bump to Rust `1.91`

JBRS307:


.github/workflows/rust_check.yml
```diff
@@ -29,7 +29,7 @@ jobs:
             - name: 🔧 Install the rust toolchain
               uses: dtolnay/rust-toolchain@stable
               with:
-                  toolchain: 1.90.0
+                  toolchain: 1.91.0
```
wkozyra95, line 32: search repo for `1.90.0`, I think there are other places that need to be updated

PR #1487, opened 2025-10-31
Make codec negotiation work on WHIP output

wkazmierczak:


smelter-core/src/pipeline/webrtc/whip_output.rs
```diff
@@ -110,6 +112,10 @@ impl WhipClientTask {
 
         let (session_url, answer) = exchange_sdp_offers(&pc, &client).await?;
 
+        // replace tracks with negotiated codecs
+        // webrtc-rs does not support codec negotiation properly https://github.com/webrtc-rs/webrtc/issues/737
```
wkozyra95:
```suggestion
        // webrtc-rs assigns a codec to the transceiver on creation, so we need to ensure that 
        // supported codec is set before set_remote_description https://github.com/webrtc-rs/webrtc/issues/737
        //
        // Final codec resolution is based on RTCRtpSendParameters and happens after set_remote_description call.
```

Try to avoid descriptions where you are saying something is "done properly" or "not properly" or other generic stuff like that if you can describe (even partially) what the problem is.

PR #1489, opened 2025-10-31
Implement RTP jitter buffer

wkozyra95:


smelter-core/src/pipeline/rtp/rtp_input.rs
```diff
@@ -75,11 +72,13 @@ impl RtpInput {
         let (audio_handle, audio_samples_receiver) =
             Self::start_audio_thread(&ctx, &input_id, opts.audio)?;
 
+        let jitter_buffer_init = RtpJitterBufferInitOptions::new(&ctx, opts.jitter_buffer);
```
wkozyra95, line 75: I'm open to suggestion for better naming or comment. `jitter_buffer_init` is a one struct that holds options and shared state, but it will be used to create 2 actual jitter buffers (one for audio and one for video)

smelter-core/src/pipeline/rtp/rtp_input/jitter_buffer.rs
```diff
@@ -0,0 +1,170 @@
...
+struct JitterBufferPacket {
+    packet: webrtc::rtp::packet::Packet,
+    pts: Duration,
+    received_at: Instant,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct RtpJitterBufferInitOptions {
+    mode: RtpJitterBufferMode,
+    buffer: InputBuffer,
```
wkozyra95, line 28: not really a buffer, it holds information about timestamps/size of the buffer, but it does not do any actuall buffering.

InputBuffer element should be created once for all the tracks(and shared)

smelter-core/src/pipeline/rtp/rtp_input/jitter_buffer.rs
```diff
@@ -0,0 +1,170 @@
...
+    ntp_sync_point: Arc<RtpNtpSyncPoint>,
```
wkozyra95, line 29: It's convient, but it is a bit of a hack to include that here. We only need it to create `RtpTimestampSync` in constructor. 

The same as for buffer, this is something we create once per multiple RTP tracks.

smelter-core/src/protocols/rtp.rs
```diff
@@ -22,21 +31,27 @@ pub enum RtpAudioOptions {
...
-}
-
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum RtpInputTransportProtocol {
     Udp,
     TcpServer,
 }
 
+#[derive(Debug, Clone)]
+pub enum RtpJitterBufferMode {
```
wkozyra95, line 41: I did not add any option for reliable RTP connection e.g. for TCP, because if packets are sequential it should not add any latency and return packets imiedietlly

PR #1500, opened 2025-11-04
Fixed broken `pixel_format` option for `FFmpeg` `h264` and `vp9` encoders.

JBRS307:
`pixel_format` option was set only for renderer, but not for encoder which resulted in bad colours in the output.

---
- [ ] ~Handle case when encoder doesn't support specified `pixel_format`~

### DIFF
--- a/smelter-core/src/pipeline/encoder/ffmpeg_h264.rs
+++ b/smelter-core/src/pipeline/encoder/ffmpeg_h264.rs
@@ -6,7 +6,7 @@ use smelter_render::{Frame, OutputFrameFormat};
 use tracing::{error, info, trace, warn};
 
 use crate::pipeline::encoder::ffmpeg_utils::{
-    create_av_frame, encoded_chunk_from_av_packet, read_extradata,
+    create_av_frame, encoded_chunk_from_av_packet, into_ffmpeg_pixel_format, read_extradata,
 };
 use crate::pipeline::ffmpeg_utils::FfmpegOptions;
 use crate::prelude::*;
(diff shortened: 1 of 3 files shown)


smelter-core/src/codecs.rs
```diff
@@ -62,6 +62,16 @@ pub enum OutputPixelFormat {
     YUV444P,
 }
 
+impl From<OutputPixelFormat> for ffmpeg_next::format::Pixel {
```
wkozyra95: this mapping is internal implementation, I would prefer if it was not a part of public API. Maybe add this as regular function inside `encoder/ffmpeg_utils.rs`

smelter-core/src/pipeline/encoder/ffmpeg_h264.rs
```diff
@@ -169,6 +164,30 @@ impl FfmpegH264Encoder {
     }
 }
 
+fn init_h264_encoder(
```
wkozyra95: I think extracting this function only hurts readability. This is important top level code, I wouldn't hide that

smelter-core/src/pipeline/encoder/ffmpeg_utils.rs
```diff
@@ -121,3 +121,11 @@ pub(super) fn encoded_chunk_from_av_packet(
         kind,
     })
 }
+
+pub(super) fn ffmpeg_pix_fmt(pixel_format: OutputPixelFormat) -> ffmpeg_next::format::Pixel {
```
wkozyra95:
```suggestion
pub(super) fn into_ffmpeg_pixel_format(pixel_format: OutputPixelFormat) -> ffmpeg_next::format::Pixel {
```

wkozyra95 (approved):

PR #1512, opened 2025-11-07
Avoid sending RTP packets with `payload_type=0` on WHEP output

wkazmierczak:


smelter-core/src/pipeline/webrtc/whip_output.rs
```diff
@@ -226,37 +226,48 @@ impl WhipClientTask {
...
                     if audio.timestamp > video.timestamp {
-                        if let (Some(packet), Some(track)) =
-                            (next_video_packet.take(), &video_track)
-                            && let Err(err) = track.write_rtp(&packet.packet).await
-                        {
-                            warn!("RTP write error {}", err);
-                            break;
+                        if let (Some(p), Some(track)) = (next_video_packet.take(), &video_track) {
+                            match track.write_rtp(&p.packet).await {
+                                Ok(_) => trace!(packet=?p, "Wrote video RTP packet to track"),
```
wkozyra95:
```suggestion
                                Ok(_) => trace!(packet=?p, "Video RTP packet written to track"),
```
That seems more consistent with how we write logs in other places

smelter-core/src/pipeline/webrtc/whip_output.rs
```diff
@@ -226,37 +226,48 @@ impl WhipClientTask {
                 // try to wait for both audio and video packet to be ready
                 (Some(video), Some(audio)) => {
                     if audio.timestamp > video.timestamp {
-                        if let (Some(packet), Some(track)) =
-                            (next_video_packet.take(), &video_track)
-                            && let Err(err) = track.write_rtp(&packet.packet).await
-                        {
-                            warn!("RTP write error {}", err);
-                            break;
+                        if let (Some(p), Some(track)) = (next_video_packet.take(), &video_track) {
```
wkozyra95, line 229: you can leave it, but I would try to avoid one letter variables like this. I understant you want to avoid `packet.packet`, but maybe there is some other way

PR #1520, opened 2025-11-12
Use `Ref<InputId>` instead of `InputId`

wkozyra95:


smelter-core/src/pipeline/hls/hls_input.rs
```diff
@@ -202,16 +201,16 @@ impl HlsInput {
...
     fn spawn_demuxer_thread(
-        input_id: InputId,
+        input_ref: InputRef,
         input_ctx: FfmpegInputContext,
         audio: Option<Track>,
         video: Option<Track>,
     ) {
         std::thread::Builder::new()
-            .name(format!("HLS thread for input {}", input_id.clone()))
+            .name(format!("HLS thread for input {}", input_ref.id()))
```
wkazmierczak: In some logs there is only `id` displayed, while in others, whole input_ref is shown. I''ll make it consistent.

PR #1525, opened 2025-11-13
Add RTMP server input

wkazmierczak:
Closes: #1538

### DIFF
--- a/smelter-api/src/input/rtmp.rs
+++ b/smelter-api/src/input/rtmp.rs
@@ -7,7 +7,7 @@ use serde::{Deserialize, Serialize};
 #[serde(deny_unknown_fields)]
 pub struct RtmpInput {
     /// URL where the RTMP server will listen for incoming stream.
-    /// Format: `rtmp://<ip_address>:<port>/<app>/<stream_key>`
+    /// Format: `rtmp://<ip_address>:<port>`
     pub url: Arc<str>,
     /// (**default=`false`**) If input is required and the stream is not delivered
     /// on time, then Smelter will delay producing output frames.
--- a/smelter-api/src/input/rtmp_into.rs
+++ b/smelter-api/src/input/rtmp_into.rs
@@ -22,7 +22,7 @@ impl TryFrom<RtmpInput> for core::RegisterInputOptions {
             core::QueueInputOptions {
                 required: false,
                 offset: None,
-            } => core::InputBufferOptions::Adaptive,
+            } => core::InputBufferOptions::Const(None),
             _ => core::InputBufferOptions::None,
         };
 
--- a/smelter-core/src/pipeline/rtmp/rtmp_input.rs
+++ b/smelter-core/src/pipeline/rtmp/rtmp_input.rs
@@ -37,8 +37,8 @@ impl RtmpServerInput {
         let should_close = Arc::new(AtomicBool::new(false));
         let buffer = InputBuffer::new(&ctx, opts.buffer);
 
-        let (frame_sender, frame_receiver) = bounded(5);
-        let (samples_sender, samples_receiver) = bounded(5);
+        let (frame_sender, frame_receiver) = bounded(10);
+        let (samples_sender, samples_receiver) = bounded(10);
 
         let receivers = QueueDataReceiver {
             video: Some(frame_receiver),
@@ -89,9 +89,6 @@ impl Track {
 
         let sender = &self.handle.chunk_sender;
         trace!(?chunk, buffer = sender.len(), "Sending chunk");
-        if sender.is_empty() {
-            debug!("RTMP input channel was drained");
-        }
         if sender.send(PipelineEvent::Data(chunk)).is_err() {
             debug!("Channel closed")
         }
--- a/smelter-core/src/pipeline/rtmp/rtmp_input/demux.rs
+++ b/smelter-core/src/pipeline/rtmp/rtmp_input/demux.rs
@@ -11,8 +11,8 @@ const RTMP_READ_RETRY_DELAY: Duration = Duration::from_millis(10);
 
 pub(super) fn run_demuxer_loop(
     mut input_ctx: FfmpegInputContext,
-    mut audio: Option<&mut Track>,
-    mut video: Option<&mut Track>,
+    mut audio: Option<Track>,
+    mut video: Option<Track>,
 ) {
     loop {
         let packet = match input_ctx.read_packet() {
--- a/smelter-core/src/pipeline/rtmp/rtmp_input/input_loop.rs
+++ b/smelter-core/src/pipeline/rtmp/rtmp_input/input_loop.rs
@@ -3,7 +3,7 @@ use std::{
     time::Duration,
 };
 
-use crossbeam_channel::Sender;
+use crossbeam_channel::{Receiver, Sender, bounded};
 use smelter_render::InputId;
 use tracing::{Level, debug, error, span, warn};
 
@@ -40,9 +40,6 @@ pub(super) fn spawn_input_loop(
             let _span =
                 span!(Level::INFO, "RTMP thread", input_id = input_ref.to_string()).entered();
 
-            let mut audio_track: Option<Track> = None;
-            let mut video_track: Option<Track> = None;
-
             loop {
                 if should_close.load(std::sync::atomic::Ordering::Relaxed) {
                     break;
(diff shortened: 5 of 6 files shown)


smelter-core/src/pipeline/rtmp/rtmp_input/demux.rs
```diff
@@ -0,0 +1,90 @@
...
+                std::thread::sleep(RTMP_READ_RETRY_DELAY);
+                continue;
+            }
+            Err(ffmpeg_next::Error::Other { errno: EIO }) => {
+                warn!("Input session disconnected!");
+                break;
+            }
+            Err(err) => {
+                trace!("RTMP read error {err:?}");
+                continue; // TODO maybe break here
```
wkazmierczak: consider break here

smelter-core/src/input.rs
```diff
@@ -49,6 +50,7 @@ pub struct InputInfo {
 #[derive(Debug, Clone, Copy)]
 pub enum InputProtocolKind {
     Rtp,
+    RtmpServer,
```
wkozyra95: Rtmp

smelter-api/src/input/rtmp.rs
```diff
@@ -0,0 +1,37 @@
+use std::{collections::HashMap, sync::Arc};
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct RtmpInput {
+    /// RTMP endpoint URL that Smelter listens on for incoming stream.
```
wkozyra95: No idea what url should be here based on this description

smelter-api/src/input/rtmp_into.rs
```diff
@@ -0,0 +1,51 @@
...
+        let queue_options = smelter_core::QueueInputOptions {
+            required: required.unwrap_or(false),
+            offset: offset_ms.map(|offset_ms| Duration::from_secs_f64(offset_ms / 1000.0)),
+        };
+
+        let buffer = match &queue_options {
+            core::QueueInputOptions {
+                required: false,
+                offset: None,
+            } => core::InputBufferOptions::Adaptive,
```
wkozyra95: Not sure if adaptive is a good choice. I think maybe fixed buffer would be better. Not sure, I would have to think about, we can change it later.

smelter-core/src/pipeline/rtmp/rtmp_input/track_audio.rs
```diff
@@ -0,0 +1,63 @@
...
+        },
+        rtmp::rtmp_input::{Track, stream_state::StreamState},
+        utils::input_buffer::InputBuffer,
+    },
+    thread_utils::InitializableThread,
+};
+
+use crate::prelude::*;
+
+pub(super) fn handle_audio_track(
```
wkozyra95: there more params than logic inside, and also spread between multiple files. I don't thinks this helps

smelter-core/src/pipeline/rtmp/rtmp_input/track_audio.rs
```diff
@@ -0,0 +1,63 @@
...
+    )?;
+
+    Ok(Track {
+        index: stream.index(),
+        handle,
+        state,
+    })
+}
+
+fn read_extra_data(stream: &Stream<'_>) -> Option<Bytes> {
```
wkozyra95: stuff like this (exectlly the same as in hls input) will need to be unified. If you want you can put it for now in e.g. ffmpeg_utils.rs and we can latter extract it to some common place

smelter-core/src/pipeline/rtmp/rtmp_input/track_video.rs
```diff
@@ -0,0 +1,107 @@
...
+                input_ref,
+                decoder_thread_options,
+            )?
+        }
+        VideoDecoderOptions::VulkanH264 => {
+            if !vulkan_supported {
+                return Err(InputInitError::DecoderError(
+                    DecoderInitError::VulkanContextRequiredForVulkanDecoder,
+                ));
+            }
```
wkozyra95: Didn't we already discuss that in other inputs? I don't think this does anything because spawn bellow will return the same exact error

smelter-core/src/pipeline/rtmp/rtmp_input/stream_state.rs
```diff
@@ -0,0 +1,117 @@
...
+
+struct DiscontinuityState {
+    check_timestamp_monotonicity: bool,
+    time_base: ffmpeg_next::Rational,
+    prev_timestamp: Option<f64>,
+    next_predicted_timestamp: Option<f64>,
+    offset: f64,
+}
+
+impl DiscontinuityState {
```
wkozyra95: I don't think there is something like discontinuity in RTMP. You will need some way to sync to queue somehow on reconnect, but behavior from HLS is wrong for that.

Not really sure what should we do when offset, it might be complex. Also we can consider removing offset from this input at all.

smelter-core/src/pipeline/rtmp/rtmp_input/demux.rs
```diff
@@ -0,0 +1,57 @@
...
+                warn!("Input session disconnected!");
+                break;
+            }
+            Err(err) => {
+                trace!("RTMP read error {err:?}");
+                continue;
+            }
+        };
+
+        if packet.is_corrupt() {
```
wkozyra95: remove

smelter-api/src/input/rtmp_into.rs
wkazmierczak: I'll leave adaptive for now

smelter-core/src/pipeline/rtmp/rtmp_input/stream_state.rs
```diff
@@ -0,0 +1,53 @@
...
+            .reference_pts_and_timestamp
+            .get_or_insert_with(|| (self.queue_start_time.elapsed(), pts_timestamp));
+
+        let pts_diff_secs = timestamp_to_secs(pts_timestamp - reference_timestamp, self.time_base);
+        let pts =
+            Duration::from_secs_f64(reference_pts.as_secs_f64() + f64::max(pts_diff_secs, 0.0));
+
+        let dts = dts_timestamp.map(|dts| {
+            Duration::from_secs_f64(f64::max(timestamp_to_secs(dts, self.time_base), 0.0))
+        });
```
wkozyra95: This is unnecesrlly complicated, you just need to convert 2 numbers to Duration and add fixed offset based on the first packet

smelter-core/src/pipeline/rtmp/rtmp_input/init.rs
```diff
@@ -0,0 +1,181 @@
...
+            ffmpeg_utils::read_extra_data,
+        },
+        utils::input_buffer::InputBuffer,
+    },
+    thread_utils::InitializableThread,
+};
+
+use crate::prelude::*;
+
+pub(super) fn spawn_initialization_thread(
```
wkozyra95: This is not initialization thread

smelter-core/src/pipeline/rtmp/rtmp_input/demux.rs
```diff
@@ -0,0 +1,48 @@
...
+use ffmpeg_next::ffi::{EAGAIN, EIO};
+use tracing::{trace, warn};
+
+use crate::pipeline::rtmp::rtmp_input::{Track, ffmpeg_context::FfmpegInputContext};
+
+use crate::prelude::*;
+
+const RTMP_READ_RETRY_DELAY: Duration = Duration::from_millis(10);
+
+pub(super) fn run_demuxer_thread(
```
wkozyra95: This is not a thread, this is a function. This might be fine if you have a code that spawns some thread and run this function inside somwhere close, but in this state it unclear

smelter-api/src/input/rtmp.rs
```diff
@@ -0,0 +1,38 @@
+use std::{collections::HashMap, sync::Arc};
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct RtmpInput {
+    /// URL where the RTMP server will listen for incoming stream.
+    /// Format: `rtmp://<ip_address>:<port>/<app>/<stream_key>`
```
wkozyra95:
```suggestion
    /// Format: `rtmp://<ip_address>:<port>`
```

smelter-core/src/pipeline/rtmp/rtmp_input.rs
```diff
@@ -0,0 +1,99 @@
...
+            pts,
+            dts,
+            kind,
+        };
+
+        let sender = &self.handle.chunk_sender;
+        trace!(?chunk, buffer = sender.len(), "Sending chunk");
+        if sender.is_empty() {
+            debug!("RTMP input channel was drained");
+        }
```
wkozyra95: remove, this was needed in hls for debuging specific issue

smelter-core/src/pipeline/rtmp/rtmp_input/input_loop.rs
```diff
@@ -0,0 +1,222 @@
...
+                    }
+                };
+
+                let mut audio_track =
+                    setup_audio_track(&ctx, &input_ctx, &input_ref, &buffer, &samples_sender);
+
+                let mut video_track =
+                    setup_video_track(&ctx, &input_ctx, &input_ref, &opts, &buffer, &frame_sender);
+
+                run_demuxer_loop(input_ctx, audio_track.as_mut(), video_track.as_mut());
```
wkozyra95: move instead of borrow, drop will not be necessary

drop context

smelter-core/src/pipeline/rtmp/rtmp_input/input_loop.rs
```diff
@@ -0,0 +1,222 @@
...
+        .name(format!("{media_kind} forwarder for input {input_ref}"))
+        .spawn(move || {
+            for event in receiver {
+                if let PipelineEvent::EOS = event {
+                    break;
+                }
+                if sender.send(event).is_err() {
+                    break;
+                }
+            }
```
wkozyra95, line 216: send eos at the end

smelter-api/src/input/rtmp_into.rs
wkazmierczak: changed to Const(None)

PR #1529, opened 2025-11-14
[integration-tests] Left padded timestamps in  `render_tests` files

JBRS307:
Timestamps in filenames are left padded so files are correctly ordered.

---

- Snapshots smelter-labs/smelter-snapshot-tests#70

### DIFF
--- a/integration-tests/src/render_tests.rs
+++ b/integration-tests/src/render_tests.rs
@@ -48,8 +48,7 @@ impl TestRunner {
         self.cases.push(case)
     }
 
-    fn run(mut self) {
-        calculate_timestamps_lengths(&mut self.cases);
+    fn run(self) {
         check_test_names_uniqueness(&self.cases);
         check_unused_snapshots(&self.cases, &self.snapshot_dir);
         let has_only = self.cases.iter().any(|test| test.only);
@@ -100,11 +99,11 @@ fn check_test_names_uniqueness(tests: &[TestCase]) {
     }
 }
 
-fn snapshot_save_path(test_name: &str, pts: &Duration, longest_pts_length: usize) -> PathBuf {
-    let mut pts = pts.as_millis().to_string();
-    let left_padding = longest_pts_length - pts.len();
-    pts.insert_str(0, "0".repeat(left_padding).as_str());
-    let out_file_name = format!("{test_name}_{pts}_{OUTPUT_ID}.png");
+fn snapshot_save_path(test_name: &str, pts: &Duration) -> PathBuf {
+    let pts = pts.as_millis();
+
+    // Pad timestamp with 0s on the left to the summaric length of 5.
+    let out_file_name = format!("{test_name}_{pts:05}_{OUTPUT_ID}.png");
     render_snapshots_dir_path().join(out_file_name)
 }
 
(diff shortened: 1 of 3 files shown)


integration-tests/src/render_tests.rs
```diff
@@ -48,13 +48,14 @@ impl TestRunner {
...
-    fn run(self) {
+    fn run(mut self) {
+        calculate_timestamps_lengths(&mut self.cases);
         check_test_names_uniqueness(&self.cases);
         check_unused_snapshots(&self.cases, &self.snapshot_dir);
         let has_only = self.cases.iter().any(|test| test.only);
 
         let mut failed = false;
-        for test in self.cases.iter() {
+        for test in self.cases.iter_mut() {
```
JBRS307: This doesn't have to be `mut`

integration-tests/src/render_tests/test_case.rs
```diff
@@ -194,4 +199,16 @@ impl TestCase {
...
+    pub fn calculate_and_set_timestamp_length(&mut self) {
+        self.timestamp_length = self
+            .steps
+            .iter()
+            .filter_map(|step| match step {
+                Step::RenderWithSnapshot(pts) => Some(pts.as_millis().to_string().len()),
+                _ => None,
+            })
+            .max()
+            .expect("There are no snapshots."); // TODO: (@jbrs) this message is terrible
```
JBRS307: Resolve it before merging

integration-tests/src/render_tests.rs
```diff
@@ -99,8 +100,11 @@ fn check_test_names_uniqueness(tests: &[TestCase]) {
...
 }
 
-fn snapshot_save_path(test_name: &str, pts: &Duration) -> PathBuf {
-    let out_file_name = format!("{}_{}_{}.png", test_name, pts.as_millis(), OUTPUT_ID);
+fn snapshot_save_path(test_name: &str, pts: &Duration, longest_pts_length: usize) -> PathBuf {
+    let mut pts = pts.as_millis().to_string();
+    let left_padding = longest_pts_length - pts.len();
+    pts.insert_str(0, "0".repeat(left_padding).as_str());
+    let out_file_name = format!("{test_name}_{pts}_{OUTPUT_ID}.png");
     render_snapshots_dir_path().join(out_file_name)
```
noituri: `format!` has builtin support for adding padding: https://stackoverflow.com/questions/41820818/how-can-i-0-pad-a-number-by-a-variable-amount-when-formatting-with-stdfmt

integration-tests/src/render_tests/snapshot.rs
```diff
@@ -16,11 +16,12 @@ pub(super) struct Snapshot {
     pub pts: Duration,
     pub resolution: Resolution,
     pub data: Vec<u8>,
+    pub timestamp_length: usize,
```
noituri: I don't like this field name but I also don't have better idea

integration-tests/src/render_tests/test_case.rs
```diff
@@ -194,4 +199,18 @@ impl TestCase {
             .expect("No scene update provided before render");
         Ok(output_frame)
     }
+
+    pub fn calculate_and_set_timestamp_length(&mut self) {
```
wkozyra95: 1. Avoid words like calaculate and set because they are not explaining anythying (ok maybe if writing a setter set make sense). In case of calaculate name should convey what/how are you calculating. In case of set what you are changing or if it is relevent what the change means.
2. Why are calculating anything, pick some value that is larger than timestamp and use it everywhere as const.

integration-tests/src/render_tests/snapshot.rs
wkozyra95: display_timestamp_lenght would be a better name, but I recommend removing it at all based on the other comment

noituri (approved):

PR #1534, opened 2025-11-17
[vk-video] Expose parser

noituri:


vk-video/src/parser/au_splitter.rs
```diff
@@ -152,3 +153,6 @@ fn idrs_where_idr_pic_id_differs(last: &Slice, curr: &Slice) -> bool {
         _ => false,
     }
 }
+
+/// Group of [`Nalu`]s representing one frame
+pub struct AccessUnit(pub Vec<Nalu>);
```
jerzywilczek: If it's a finished AU, the storage won't grow. Maybe make it into a boxed slice to communicate that?

vk-video/src/parser/nalu_parser.rs
```diff
@@ -144,6 +145,14 @@ pub enum ParsedNalu {
     Other(String),
 }
 
+/// H264 Network Abstraction Layer Unit
+pub struct Nalu {
+    /// Parsed nalu from [`Nalu::raw_bytes`]
+    pub parsed: ParsedNalu,
+    pub raw_bytes: Bytes,
```
jerzywilczek: I'm not a huge fan of putting `Bytes` in the public API, and I would like to stay with a `Vec<u8>`. What do you think about it? Can you measure the performance impact of one vs the other (e.g. how big chunk of a profile NAL-splitting is with a `Vec` here vs with `Bytes`)? If the performance benefit is significant I think it's OK to keep `Bytes` here, but intuitively I don't think there is much difference.

vk-video/src/parser.rs
```diff
@@ -1,29 +1,114 @@
...
+
+    #[error("Error while parsing SPS: {0:?}")]
+    SpsParseError(h264_reader::nal::sps::SpsError),
+
+    #[error("Error while parsing PPS: {0:?}")]
+    PpsParseError(h264_reader::nal::pps::PpsError),
+
+    #[error("Error while parsing a slice: {0:?}")]
+    SliceParseError(h264_reader::nal::slice::SliceHeaderError),
+}
```
jerzywilczek: All of those are h264-specific, so it should be in the h264 module together with the parser

vk-video/src/parser.rs
```diff
@@ -1,29 +1,114 @@
...
+                    self.reader.push(&nalu_bytes);
+
+                    let parsed_nalu = self.receiver.try_recv().unwrap();
+                    parsed_nalu.map(|parsed_nalu| Nalu {
+                        parsed: parsed_nalu,
+                        raw_bytes: nalu_bytes,
+                        pts,
+                    })
+                })
+                .collect::<Vec<_>>();
```
jerzywilczek: Do we need the `collect` here? I don't see why it's necessary

vk-video/src/parser.rs
```diff
@@ -77,99 +162,33 @@ pub enum DecoderInstruction {
...
-                    }
-                    ParsedNalu::Pps(pic_parameter_set) => {
-                        instructions.push(DecoderInstruction::Pps(pic_parameter_set))
-                    }
-                    ParsedNalu::Slice(slice) => {
-                        slices.push((slice, pts));
-                    }
-
-                    ParsedNalu::Other(_) => {}
+pub(crate) fn extract_decoder_instructions(
```
jerzywilczek: How about a fun name like `compile_to_decoder_instructions`? I always imagined decoding h264 as writing an emulator for an obscure game console, where videos are the game ROMs and this step is compilation of the rom data into a VM bytecode. I think it's quite fun and that the analogy works surprisingly well.

Then in the future we can think of various codecs as compiler front-ends, which they essentially will be.

vk-video/src/lib.rs
```diff
@@ -133,16 +133,17 @@ pub mod parameters {
...
 
 use ash::vk;
-use parser::Parser;
 use vulkan_decoder::{FrameSorter, VulkanDecoder};
 
 pub use adapter::VulkanAdapter;
 pub use device::VulkanDevice;
 pub use instance::VulkanInstance;
-pub use parser::{ParserError, ReferenceManagementError};
+pub use parser::{ParserError, ReferenceManagementError, h264};
```
jerzywilczek: I'd rather we exposed a `parsers` module, which would have the `h264` module inside it, instead of just exposing an `h264` module, but this might interfere with the rest of your planned PRs in this series

vk-video/src/parser.rs
noituri: I really like this analogy

vk-video/src/parser/nalu_parser.rs
noituri: The performance difference is negligible. I decided to use boxed slice in the public api instead of vec

PR #1535, opened 2025-11-17
[vk-video] Make vk-video compile on MacOS

noituri:
Depends on: 
- https://github.com/software-mansion/smelter/pull/1534

The parser will also be used on macOS so we need to make vk-video compilable on macOS.
I made sure these changes won't affect how the docs look visually.

### DIFF
--- a/smelter-core/Cargo.toml
+++ b/smelter-core/Cargo.toml
@@ -30,7 +30,7 @@ tracing-subscriber = { workspace = true }
 fdk-aac-sys = "0.5.0"
 rubato = "0.15.0"
 wgpu = { workspace = true }
-vk-video = { path = "../vk-video" }
+vk-video = { workspace = true }
 glyphon = { workspace = true }
 tokio = { workspace = true }
 serde_json = { workspace = true }
--- a/vk-video/src/parser.rs
+++ b/vk-video/src/parser.rs
@@ -17,6 +17,7 @@ pub mod h264 {
 
     pub use super::au_splitter::AccessUnit;
     pub use super::nalu_parser::{Nalu, ParsedNalu};
+    #[cfg(feature = "expose_parsers")]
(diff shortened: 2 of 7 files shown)


smelter-core/Cargo.toml
```diff
@@ -9,7 +9,7 @@ license-file = "../LICENSE"
 default = []
 decklink = ["dep:decklink"]
 web-renderer = ["smelter-render/web-renderer"]
-vk-video = ["dep:vk-video"]
```
noituri, line 12: I think `vulkan-video` would be a better feature name now but that would require changing docs and would be technically a breaking change so I'm leaving it as it is

vk-video/src/parser.rs
```diff
@@ -1,15 +1,11 @@
...
-
 mod au_splitter;
 mod nalu_parser;
 mod nalu_splitter;
-mod reference_manager;
+
+#[cfg(vulkan)]
+pub(crate) mod decoder_instructions;
+#[cfg(vulkan)]
+pub(crate) mod reference_manager;
```
noituri, line 8: I would prefer having all `#[cfg(vulkan)]` in `lib.rs` but that would require moving `decoder_instructions` and `reference_manager` modules out of `parser`. Maybe we should do that?

smelter-api/Cargo.toml
```diff
@@ -8,7 +8,6 @@ license-file = "../LICENSE"
 default = []
 decklink = ["smelter-core/decklink"]
 web-renderer = ["smelter-core/web-renderer"]
-vk-video = ["smelter-core/vk-video"]
```
noituri, line 11: This wasn't used?

vk-video/src/parser.rs
jerzywilczek: I don't think it's worth it just to avoid the cfgs

PR #1536, opened 2025-11-17
Detect missing frames in `ffmpeg_h264`

noituri:
Depends on: 
- https://github.com/software-mansion/smelter/pull/1534
- https://github.com/software-mansion/smelter/pull/1535

In this PR I added access unit splitter for `ffmpeg_h264` decoder.

The access units are checked for missing frames. It only checks the frame nums of nalus and will not work if multiple nalus have the same frame num.

The buffering in the depayloader was disabled because its access unit detection will not work if marker packet is lost

### DIFF
--- a/smelter-core/src/pipeline/decoder/ffmpeg_h264.rs
+++ b/smelter-core/src/pipeline/decoder/ffmpeg_h264.rs
@@ -2,6 +2,7 @@ use std::{iter, sync::Arc};
 
 use crate::pipeline::decoder::{
     VideoDecoder, VideoDecoderInstance,
+    au_splitter::AUSplitter,
     ffmpeg_utils::{create_av_packet, from_av_frame},
 };
 use crate::prelude::*;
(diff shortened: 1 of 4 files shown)


smelter-core/src/pipeline/decoder/ffmpeg_h264.rs
```diff
@@ -44,28 +46,40 @@ impl VideoDecoder for FfmpegH264Decoder {
...
         trace!(?chunk, "H264 decoder received a chunk.");
-        let av_packet = match create_av_packet(chunk, VideoCodec::H264, TIME_BASE) {
-            Ok(packet) => packet,
+        let chunks = match self.au_splitter.put_chunk(chunk) {
+            Ok(chunks) => chunks,
             Err(err) => {
-                warn!("Dropping frame: {}", err);
+                debug!("H264 AU splitter could not process the chunks: {err}");
                 return Vec::new();
             }
```
noituri, line 62: if no frames  are returned `DynamicVideoDecoderStream` will request a keyframe. I don't like that but this will probably be refactored later anyway so I'll leave it as it is

PR #1539, opened 2025-11-18
Pass packet loss info to decoder

wkozyra95:
- Pass information about packet loss from rtp to decoder
  - Wrap RtpPacket with RtpEvent
  - Wrap EncodedInputChunk with EncodedInputEvent (will be useful to pass decoder flush too)
- Change how the jitter buffer works:
  - Now first packet is not immediately returned when we wait as long as possible (the same logic as when there is a discontinuity in the sequence numbers)
- Jitter buffer changes were causing some RTP tests to fail, so
  - RTP input now tries to pop from the jitter buffer even if no new packets were received
  - RTP input does not have a jitter buffer when on TCP. JitterBuffer struct is still used, but we are always pop-ing regardless of the conditions


smelter-core/src/pipeline/rtp/depayloader.rs
```diff
@@ -39,7 +39,7 @@ pub fn new_depayloader(options: DepayloaderOptions) -> Box<dyn Depayloader> {
     info!(?options, "Initialize RTP depayloader");
     match options {
         DepayloaderOptions::H264 => {
-            BufferedDepayloader::<H264Packet>::new_boxed(MediaKind::Video(VideoCodec::H264))
+            SimpleDepayloader::<H264Packet>::new_boxed(MediaKind::Video(VideoCodec::H264))
```
wkozyra95: This will break ffmepeg h264 decoder. This needs to be reverted before the merge, or parser implementation needs to land first

smelter-core/src/pipeline/rtp/rtp_input.rs
```diff
@@ -270,9 +273,33 @@ fn run_rtp_demuxer_thread(
         }
     };
     loop {
-        let Ok(mut buffer) = receiver.recv() else {
-            debug!("Closing RTP demuxer thread.");
-            break;
+        let mut buffer = match receiver.recv_timeout(Duration::from_millis(10)) {
+            Ok(buffer) => buffer,
+            Err(RecvTimeoutError::Timeout) => {
+                if let Some(video) = &mut video {
```
wkozyra95, line 279: Normally, this would not need to be triggered, but our tests are sending a lot of data at the start, so recv is never called latter and data is not pop-ed from jitter buffer

smelter-core/src/pipeline/rtp/rtp_input/jitter_buffer.rs
```diff
@@ -114,42 +114,51 @@ impl RtpJitterBuffer {
...
+
+        match self.next_seq_num {
+            Some(next) => {
+                (self.on_stats_event)(RtpJitterBufferStatsEvent::RtpPacketLost);
+                self.next_seq_num = Some(next + 1);
+                Some(RtpInputEvent::LostPacket)
+            }
+            None => {
+                // first packet
+                self.pop()
```
wkozyra95, line 156: The first packet will only be popped now when we are running out of time, previouslly first packet was popped immediately after receiving it. However, that risked that if the initial packet was reorded and was the keyframe we would never request new one

PR #1547, opened 2025-11-21
Flush parser based on marker bit

wkozyra95:
Dirty workaround to flush parser based on marker bit.

### DIFF
--- a/vk-video/src/parser/nalu_splitter.rs
+++ b/vk-video/src/parser/nalu_splitter.rs
@@ -17,7 +17,7 @@ fn find_start_of_next_nalu(buf: &[u8]) -> Option<usize> {
         return None;
     };
 
-    // If the start code is at the beginning of nalu, we need to skip it, because this means we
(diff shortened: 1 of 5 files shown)


vk-video/src/parser/nalu_splitter.rs
```diff
@@ -61,4 +61,26 @@ impl NALUSplitter {
...
+
+        let mut result = Vec::new();
+        while let Some(i) = find_start_of_next_nalu(&self.buffer[self.previous_search_end..]) {
+            let nalu = self.buffer.split_to(self.previous_search_end + i);
+            self.previous_search_end = 0;
+            result.push((nalu.to_vec(), self.pts));
+        }
+
+        result.push((
+            [self.buffer.to_vec(), [0, 0, 1].to_vec()].concat(),
```
wkozyra95, line 78: Because of this, between access units the prefix will need to be pushed twice

PR #1549, opened 2025-11-24
Added `/stats` endpoint for `hls` input

### LINKED ISSUE #1546: Add `/stats` for `HLS` input.

JBRS307:
Closes #1546 

---

I have split the `input_state` into separate modules, as this file would grow significantly when adding stats for next inputs.

### DIFF
--- a/smelter-core/src/pipeline/hls/hls_input.rs
+++ b/smelter-core/src/pipeline/hls/hls_input.rs
@@ -106,18 +106,11 @@ impl HlsInput {
         stream: &Stream<'_>,
         buffer: InputBuffer,
     ) -> Result<(Track, Receiver<PipelineEvent<InputAudioSamples>>), InputInitError> {
-        let stats_sender =
-            HlsInputTrackStatsSender::new(input_ref, &ctx.stats_sender, HlsTrack::Audio);
         // not tested it was always null, but audio is in ADTS, so config is not
         // necessary
         let asc = read_extra_data(stream);
         let (samples_sender, samples_receiver) = bounded(5);
-        let state = StreamState::new(
-            ctx.queue_sync_point,
-            stream.time_base(),
-            buffer,
-            stats_sender,
-        );
+        let state = StreamState::new(ctx, input_ref, stream.time_base(), buffer, TrackKind::Audio);
         let handle = AudioDecoderThread::<fdk_aac::FdkAacDecoder>::spawn(
             input_ref.clone(),
             AudioDecoderThreadOptions {
@@ -145,15 +138,8 @@ impl HlsInput {
         video_decoders: HlsInputVideoDecoders,
         buffer: InputBuffer,
     ) -> Result<(Track, Receiver<PipelineEvent<Frame>>), InputInitError> {
-        let stats_sender =
-            HlsInputTrackStatsSender::new(input_ref, &ctx.stats_sender, HlsTrack::Video);
         let (frame_sender, frame_receiver) = bounded(5);
-        let state = StreamState::new(
-            ctx.queue_sync_point,
-            stream.time_base(),
-            buffer,
-            stats_sender,
-        );
+        let state = StreamState::new(ctx, input_ref, stream.time_base(), buffer, TrackKind::Video);
 
         let extra_data = read_extra_data(stream);
         let h264_config = extra_data
@@ -354,13 +340,15 @@ struct StreamState {
 
 impl StreamState {
     fn new(
-        queue_sync_point: Instant,
+        ctx: &Arc<PipelineCtx>,
+        input_ref: &Ref<InputId>,
         time_base: ffmpeg_next::Rational,
         buffer: InputBuffer,
-        stats_sender: HlsInputTrackStatsSender,
+        track_kind: TrackKind,
     ) -> Self {
+        let stats_sender = HlsInputTrackStatsSender::new(input_ref, &ctx.stats_sender, track_kind);
         Self {
-            queue_sync_point,
+            queue_sync_point: ctx.queue_sync_point,
             time_base,
             buffer,
 
@@ -548,7 +536,7 @@ where
 }
 
 #[derive(Clone, Copy)]
-enum HlsTrack {
+enum TrackKind {
     Audio,
     Video,
 }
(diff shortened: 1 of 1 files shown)


smelter-core/src/pipeline/hls/hls_input.rs
```diff
@@ -405,10 +439,12 @@ impl DiscontinuityState {
         if is_discontinuity {
             debug!("Discontinuity detected: {prev_timestamp} -> {timestamp}");
             self.offset += next_timestamp - timestamp;
+            // TODO: (stats) detected discontinuity
```
JBRS307: Delete this

smelter-core/src/pipeline/hls/hls_input.rs
```diff
@@ -264,7 +285,19 @@ impl HlsInput {
...
+                stats_sender.send_audio(HlsInputTrackStatsEvent::ChunkSize(packet.size() as u64));
+                stats_sender.send_audio(HlsInputTrackStatsEvent::InputBufferSize(
+                    track.state.buffer.size(),
+                ));
+                stats_sender.send_audio(HlsInputTrackStatsEvent::EffectiveBuffer(
+                    pts.saturating_sub(track.state.queue_sync_point.elapsed()),
+                ));
+                if pts_discontinuity {
+                    stats_sender.send_audio(HlsInputTrackStatsEvent::DiscontinuityDetected);
+                }
```
JBRS307: This is too long to be inlined.

smelter-core/src/pipeline/hls/hls_input.rs
```diff
@@ -236,13 +244,26 @@ impl HlsInput {
...
+                stats_sender.send_video(HlsInputTrackStatsEvent::ChunkSize(packet.size() as u64));
+                stats_sender.send_video(HlsInputTrackStatsEvent::InputBufferSize(
+                    track.state.buffer.size(),
+                ));
+                stats_sender.send_video(HlsInputTrackStatsEvent::EffectiveBuffer(
+                    pts.saturating_sub(track.state.queue_sync_point.elapsed()),
+                ));
+                if pts_discontinuity {
+                    stats_sender.send_video(HlsInputTrackStatsEvent::DiscontinuityDetected);
+                }
```
JBRS307: This is too long to be inlined

smelter-core/src/pipeline/hls/hls_input.rs
```diff
@@ -236,13 +244,21 @@ impl HlsInput {
...
                 continue;
             }
 
             if let Some(track) = &mut video
                 && packet.stream() == track.index
             {
-                let (pts, dts) = track.state.pts_dts_from_packet(&packet);
+                let (pts, dts, pts_discontinuity) = track.state.pts_dts_from_packet(&packet);
+
+                stats_sender.send_video(
```
wkozyra95: try to minimize logi like this in this context (discussed offline)

smelter-core/src/stats/mod.rs
```diff
@@ -60,18 +60,24 @@ impl StatsMonitor {
 }
 
 impl StatsSender {
-    pub fn send(&self, event: StatsEvent) {
-        if let Err(TrySendError::Full(event)) = self.0.try_send(event) {
-            warn!(?event, "Stats channel full")
-        };
+    pub fn send(&self, events: Vec<StatsEvent>) {
```
wkozyra95: I don't see the logic. Why method that sends a vector of events is named send, and single event send_evetnt. It is inconsitent.

Also consider writing generic version that is just one method that can take either
```
pub fn send<T: IntoIter<StatsEvent>>(&self, events: T) {

}

impl IntoIter for StatsEvent {
...
}
```

smelter-core/src/pipeline/hls/hls_input.rs
```diff
@@ -496,3 +546,77 @@ where
...
     }
 }
+
+struct HlsInputStatsSender {
+    input_ref: Ref<InputId>,
+    stats_sender: StatsSender,
+}
+
+impl HlsInputStatsSender {
+    fn new(input_ref: Ref<InputId>, stats_sender: StatsSender) -> Self {
```
wkozyra95:
```suggestion
    fn new(input_ref: &Ref<InputId>, stats_sender: &StatsSender) -> Self {
```
If you take borrowed values here and clone them inside then the calling side will be more concise. It won't need those additional clone calls

smelter-core/src/pipeline/hls/hls_input.rs
```diff
@@ -47,6 +47,18 @@ struct Track {
...
+
+impl Track {
+    fn pts_dts_from_packet(&mut self, packet: &Packet) -> (Duration, Option<Duration>) {
+        self.state.pts_dts_from_packet(packet, &self.stats_sender)
+    }
+
+    fn send_stats_on_packet_received(&self, packet: &Packet, packet_pts: Duration) {
+        self.stats_sender
+            .send_on_packet_received(packet, packet_pts, &self.state);
+    }
```
wkozyra95: I think those helpers are unnecessary, they are doing to little to justify. You can avoid part of this logic by passing stats_sender to in the state creation

smelter-core/src/pipeline/hls/hls_input.rs
```diff
@@ -86,7 +103,8 @@ impl HlsInput {
             audio: samples_receiver,
         };
 
-        Self::spawn_demuxer_thread(input_ref, input_ctx, audio, video);
+        let stats_sender = HlsInputStatsSender::new(input_ref.clone(), ctx.stats_sender.clone());
```
wkozyra95: This type has only 2 uses:
- sending event in one place
- send stats_sender and input_ref as one argument

Unless you have strong preditction that there will be more uses of this struct I would:
- Pass input_ref and ctx to demuxer thread as standlone arguments
- use ctx.stats_sender directlly in that one place

smelter-core/src/pipeline/hls/hls_input.rs
```diff
@@ -496,3 +546,77 @@ where
...
+}
+
+struct HlsInputTrackStatsSender {
+    input_ref: Ref<InputId>,
+    stats_sender: StatsSender,
+    track: HlsTrack,
+}
+
+impl HlsInputTrackStatsSender {
+    fn new(input_ref: Ref<InputId>, stats_sender: StatsSender, track: HlsTrack) -> Self {
```
wkozyra95: The same as above (pass refs and clone inside new)

smelter-core/src/pipeline/hls/hls_input.rs
```diff
@@ -133,8 +145,15 @@ impl HlsInput {
...
+        let stats_sender =
+            HlsInputTrackStatsSender::new(input_ref, &ctx.stats_sender, HlsTrack::Video);
         let (frame_sender, frame_receiver) = bounded(5);
-        let state = StreamState::new(ctx.queue_sync_point, stream.time_base(), buffer);
+        let state = StreamState::new(
+            ctx.queue_sync_point,
+            stream.time_base(),
+            buffer,
+            stats_sender,
+        );
```
wkozyra95:
```suggestion
        let state = StreamState::new(ctx, stream.time_base(), buffer, HlsTrack::Video);
```
and create sender inside new

I would also recommend renaming HlsTrack to e.g. TrackKind. Hls is not necessary in private type and without Kind is unclear whethre it represents Track state or Track type

wkozyra95 (approved):

PR #1587, opened 2025-12-03
Refine the V4L2 API.

jerzywilczek:
This patch does the following things:
- Add a function that lists all capturable video devices in a directory, along with their formats, capabilities etc.
- Add an example focused on the Rust API for V4L2 in Smelter.
- Make some parameters for the input optional.

### DIFF
--- a/integration-tests/examples/v4l2_rust_api.rs
+++ b/integration-tests/examples/v4l2_rust_api.rs
@@ -49,9 +49,7 @@ mod main_module {
         let input_id = InputId("input_1".into());
 
         Pipeline::register_input(&pipeline, input_id.clone(), v4l2_input_options()).unwrap();
-
         Pipeline::register_output(&pipeline, output_id, output_options(input_id.clone())).unwrap();
-
         Pipeline::start(&pipeline);
 
         std::thread::sleep(Duration::from_secs(30));


ts/smelter/src/api.generated.ts
```diff
@@ -164,20 +164,20 @@ export type RegisterInput =
        * While the numbers assigned in `/dev/video<N>` paths can differ depending on device detection order, the `by-id` paths are always the same for a given device, and the `by-path` paths should be the same for specific ports.
```
wkozyra95, line 164: This is not the only change, check out my original PR that added v4l2 types to see where you need to change types

smelter-api/src/input/v4l2.rs
```diff
@@ -19,15 +19,18 @@ pub struct V4l2Input {
     /// detection order, the `by-id` paths are always the same for a given device, and the
     /// `by-path` paths should be the same for specific ports.
     pub path: Arc<Path>,
-    /// The resolution that will be negotiated with the device.
-    pub resolution: Resolution,
     /// The format that will be negotiated with the device.
     pub format: V4l2InputFormat,
+    /// The resolution that will be negotiated with the device.
```
wkozyra95:
```suggestion
    /// The requested resolution that will be negotiated with the device.
```
or maybe desired 

normally I would think about changing the field name too, but since we released it I don't think it is worth it

smelter-api/src/input/v4l2_into.rs
```diff
@@ -14,9 +14,12 @@ impl TryFrom<V4l2Input> for core::RegisterInputOptions {
         Ok(core::RegisterInputOptions {
             input_options: core::ProtocolInputOptions::V4l2(core::V4l2InputOptions {
                 path: value.path,
-                resolution: value.resolution.into(),
                 format: value.format.into(),
-                framerate: value.framerate.try_into()?,
+                resolution: value.resolution.map(smelter_render::Resolution::from),
```
wkozyra95:
```suggestion
                resolution: value.resolution.map(Into::into),
```

smelter-core/src/pipeline/v4l2/v4l2_input.rs
```diff
@@ -133,10 +135,13 @@ impl V4l2DeviceConfig {
         }
 
         let requested_fourcc = opts.format.into();
-        let format = device.format()?;
+        let mut format = device.format()?;
```
wkozyra95: Code like this is problematic because you are modyfing mut variable from external scope. Usually if I have to write something like this I prefer if it is isolated from rest of the code, preferablly in a function, sometimes in just a code block.

Also it is usually best to keep logic separate for clarity. First resolve optional values if possible, and then write v4l2 specific code without any extra if statements. It is not always possible, e.g. you can't do that with framerate, but you can do this with resolution

For consistent approach you could extract it as `try_set_resolution` and `try_set_framerate` on `V4l2DeviceConfig`. This way you could avoid this large chains in condirions for framerate

smelter-core/src/pipeline/v4l2/v4l2_input.rs
```diff
@@ -274,3 +288,149 @@ impl InputState<'_> {
...
+    pub resolutions: Vec<V4l2ResolutionInfo>,
+}
+
+#[derive(Debug, Clone)]
+pub struct V4l2ResolutionInfo {
+    pub resolution: Resolution,
+    pub framerates: Vec<Framerate>,
+}
+
+pub fn available_v4l2_devices(
```
wkozyra95:
```suggestion
pub fn get_v4l2_devices(
```
or list

wkozyra95: plus add a comment

integration-tests/examples/v4l2_rust_api.rs
```diff
@@ -0,0 +1,125 @@
...
+
+        let pipeline = pipeline(&config);
+        let output_id = OutputId("output_1".into());
+        let input_id = InputId("input_1".into());
+
+        Pipeline::register_input(&pipeline, input_id.clone(), v4l2_input_options()).unwrap();
+
+        Pipeline::register_output(&pipeline, output_id, output_options(input_id.clone())).unwrap();
+
+        Pipeline::start(&pipeline);
```
wkozyra95, line 53:
```suggestion
        Pipeline::register_input(&pipeline, input_id.clone(), v4l2_input_options()).unwrap();
        Pipeline::register_output(&pipeline, output_id, output_options(input_id.clone())).unwrap();
        Pipeline::start(&pipeline);
```

PR #1588, opened 2025-12-05
[rtmp] Base server implementation

wkazmierczak:


rtmp/Cargo.toml
JBRS307, line 1: Enable workspace lints make clippy run with the same set of rules as in the rest of the repo.

rtmp-server/src/server.rs
```diff
@@ -0,0 +1,75 @@
...
+                    info!("New connection from: {}", peer_addr);
+
+                    let client_timeout = self.config.client_timeout_secs;
+
+                    tokio::spawn(async move {
+                        // use stream or tls_stream based on use_ssl field
+                        let result = handle_client(stream, client_timeout).await;
+
+                        if let Err(e) = result {
+                            error!("Client handler error: {}", e);
```
JBRS307: I think it would be better to do
```suggestion
                            error!(error = %e, "Client handler error");
```

rtmp-server/src/server.rs
```diff
@@ -0,0 +1,75 @@
...
+                        }
+                    });
+                }
+                Err(e) => {
+                    error!("Accept error: {}", e);
```
JBRS307: Same as above

rtmp-server/src/handshake.rs
```diff
@@ -0,0 +1,53 @@
...
+    where
+        S: AsyncRead + AsyncWrite + Unpin,
+    {
+        // C0 version
+        let mut c0 = [0u8; 1];
+        stream.read_exact(&mut c0).await?;
+
+        if c0[0] != RTMP_VERSION {
+            return Err(RtmpError::InvalidVersion(c0[0]));
+        }
```
JBRS307: According to specs, if server does not recognize the version sent by the client, it SHOULD (by the RFC definition) respond with `3` anyway. (info on that is in RTMP spec, page 7)

rtmp-server/src/handshake.rs
```diff
@@ -0,0 +1,53 @@
...
+        s2[4..8].copy_from_slice(&timestamp.to_be_bytes());
+        stream.write_all(&s2).await?;
+        stream.flush().await?;
+
+        // C2 client echoes S1
+        let mut c2 = [0u8; HANDSHAKE_SIZE];
+        stream.read_exact(&mut c2).await?;
+
+        if c2 != s1 {
+            return Err(RtmpError::HandshakeFailed("C2 does not match S1".into()));
```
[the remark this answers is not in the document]
JBRS307: This is only half true. Everything but bytes in range [4..8] should match. [4..8] in C2 contains the timestamp at which client read the S1, S1 contains only zeroes there.

rtmp-server/src/handshake.rs
```diff
@@ -0,0 +1,53 @@
...
+        rand::rng().fill_bytes(&mut s1[8..]);
+        stream.write_all(&s1).await?;
+
+        // C1 timestamp(4 bytes), zero(4 bytes), random(1528 bytes)
+        let mut c1 = [0u8; HANDSHAKE_SIZE];
+        stream.read_exact(&mut c1).await?;
+
+        // S2 echo C1 with our timestamp
+        let mut s2 = c1;
+        s2[4..8].copy_from_slice(&timestamp.to_be_bytes());
```
JBRS307: This MUST (rfc def) contain the timestamp at which we read the C1

rtmp/src/handshake.rs
```diff
@@ -0,0 +1,57 @@
...
+        let c0_read_time = Instant::now();
+
+        // S0 version
+        stream.write_all(&[RTMP_VERSION])?;
+
+        // S1 timestamp(4 bytes), zero(4 bytes), random(1528 bytes)
+        let mut s1 = [0u8; HANDSHAKE_SIZE];
+        let timestamp: u32 = 0;
+        s1[0..4].copy_from_slice(&timestamp.to_be_bytes());
+        s1[4..8].copy_from_slice(&[0u8; 4]); // zeros
```
JBRS307:
```suggestion
        s1[4..8].copy_from_slice(&[0; 4]);
```

I am not 100% sure, but it should work without explicitly typing the literal here

rtmp-server/src/server.rs
```diff
@@ -0,0 +1,212 @@
...
+    sync::{
+        Arc, RwLock,
+        mpsc::{Receiver, channel},
+    },
+    thread,
+};
+use tracing::{error, info};
+
+pub type OnConnectionCallback =
+    Arc<dyn Fn(&RtmpConnection, Receiver<Vec<u8>>, Receiver<Vec<u8>>) -> bool + Send + Sync>;
```
JBRS307: Isn't bool `Send` and `Sync` by itself already?

rtmp-server/src/server.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    pub fn on_connection<F>(&mut self, callback: F)
+    where
+        F: Fn(&RtmpConnection, Receiver<Vec<u8>>, Receiver<Vec<u8>>) -> bool
+            + Send
+            + Sync
+            + 'static,
+    {
+        self.on_connection = Some(Arc::new(callback));
+    }
```
JBRS307: In Rust, if a function is named the same as a struct field it is considered to be a getter, not a setter.

rtmp-server/src/server.rs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+impl RtmpServer {
+    pub fn new(config: ServerConfig) -> Self {
+        Self {
+            config,
+            state: Arc::new(ServerState::new()),
+            on_connection: None,
+        }
+    }
```
JBRS307: Up to you, but I would allow people to send `Option<F>` for `on_connection()` here already.

rtmp-server/src/server.rs
wkazmierczak: however, Sync + Send applies to the entire closure, not just the bool

rtmp-server/src/server.rs
wkazmierczak: I thought it was common to name it that way, but I’ll chec

Cargo.toml
```diff
@@ -16,6 +16,7 @@ members = [
     "decklink",
     "vk-video",
     "libcef",
+    "rtmp-server",
```
wkozyra95: Do not name it specifically server, we might want to add client in the future

rtmp-server/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
```
wkozyra95: I think this can be removed, both target  and lock is in parent directory

rtmp-server/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "rtmp_server"
```
wkozyra95:
```suggestion
name = "rtmp-server"
```

rtmp-server/src/server.rs
```diff
@@ -0,0 +1,212 @@
...
+    sync::{
+        Arc, RwLock,
+        mpsc::{Receiver, channel},
+    },
+    thread,
+};
+use tracing::{error, info};
+
+pub type OnConnectionCallback =
+    Arc<dyn (Fn(&RtmpConnection, Receiver<Vec<u8>>, Receiver<Vec<u8>>) -> bool) + Send + Sync>;
```
wkozyra95:
```suggestion
    Box<dyn (FnMut(&RtmpConnection, Receiver<Bytes>, Receiver<Bytes>) -> bool) + Send>;
```
- I think often callback will need to mutable
- I don't see a reason why the callback needs to be sync. If you need sync you can use mutex

rtmp-server/examples/server.rs
```diff
@@ -0,0 +1,38 @@
...
+        use_ssl: false,
+        cert_file: None,
+        key_file: None,
+        ca_cert_file: None,
+        client_timeout_secs: 30,
+    };
+
+    let mut server = RtmpServer::new(config);
+
+    server.on_connection(|conn, video_rx, _audio_rx| {
```
wkozyra95: Package all arguments into single struct. droping that struct should close the connection. 

This implementation leads to some wierd behavoir that you can close only audio or only video track. Unless there is a good reason for that I would not do that.

rtmp-server/src/server.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub port: u16,
+    pub use_ssl: bool,
+    pub cert_file: Option<String>,
+    pub key_file: Option<String>,
+    pub ca_cert_file: Option<String>,
+    pub client_timeout_secs: u64,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct RtmpConnection {
```
wkozyra95: is app / stream_key pattern a standart? What if someone has url `rtmp://127.0.0.1/abc/efg/hij`? What part is app and what is stream key?

If it is not standart, just store it as single string

rtmp-server/src/server.rs
```diff
@@ -0,0 +1,212 @@
...
+use tracing::{error, info};
+
+pub type OnConnectionCallback =
+    Arc<dyn (Fn(&RtmpConnection, Receiver<Vec<u8>>, Receiver<Vec<u8>>) -> bool) + Send + Sync>;
+
+#[allow(dead_code)] // TODO add SSL/TLS
+pub struct ServerConfig {
+    pub port: u16,
+    pub use_ssl: bool,
+    pub cert_file: Option<String>,
```
wkozyra95: use Arc<str> instead of String

rtmp-server/src/server.rs
```diff
@@ -0,0 +1,212 @@
...
+    // hardcoded for now
+    let conn_info = RtmpConnection::new("app".into(), "stream_key".into());
+
+    // check if another stream is not actively streaming on that route
+
+    let (video_tx, video_rx) = channel();
+    let (audio_tx, audio_rx) = channel();
+
+    // client
+    let authorized = if let Some(ref cb) = on_connection {
```
wkozyra95: authorized? why?

rtmp-server/src/server.rs
wkozyra95: Not even option, RtmpServer does not make sense without on_connection, so callback should be passed inside constructior and it should be optional

rtmp-server/src/server.rs
wkozyra95: What are those bytes? Are they RTMP messages/FLV chunks/h264? What about any timestamps/metadata?

rtmp-server/src/server.rs
wkazmierczak: image

rtmp/src/server.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    Ok(())
+}
+
+struct RtmpMessage {
+    type_id: u8,
+    payload: Bytes,
+}
+
+// jsut schema of structs based of my previous implementation
```
```suggestion
// jsut schema of structs
```

rtmp/src/handshake.rs
wkazmierczak: finally, I have used `.fill(0)` here

PR #1567, opened 2025-11-27
Added `bitrate` option to video encoders.

JBRS307:
Closes #1565 

---

- [x] Add options to `smelter-api` and structs in `smelter-core`
- Implement options for each encoder
  - [x] h264
  - [x] vp8
  - [x] vp9
 - Implement defaults
   - [x] h264
   - [x] vp8 - Resolution based default bitrate, `crf` not working as expected
   - [x] vp9 - set constant quality mode with resolution based `crf` as default
- [x] regenerate `JSON` schema
- [x] ts-sdk
- [x] changelog

PR #1591, opened 2025-12-08
Add `keyframe_interval_ms` in encoder options.

### LINKED ISSUE #1566: Add `keyframe_interval` option to video encoders.

JBRS307:
Closes #1566 

---

- Merge after #1567

---

- Add option to encoders:
  - [x] VP8
  - [x] VP9
  - [x] H264
- [x] ts-sdk

### DIFF
--- a/smelter-core/src/pipeline/encoder/utils.rs
+++ b/smelter-core/src/pipeline/encoder/utils.rs
@@ -26,5 +26,5 @@ pub(super) fn bitrate_from_resolution_framerate(
 
 pub(super) fn gop_size_from_ms_framerate(keyframe_interval: Duration, framerate: Framerate) -> u64 {
     let framerate = framerate.num as f64 / framerate.den as f64;
-    (framerate * keyframe_interval.as_secs_f64().round()).round() as u64
+    (framerate * keyframe_interval.as_secs_f64()).round() as u64
 }


smelter-api/src/output/hls.rs
```diff
@@ -43,6 +43,9 @@ pub enum HlsVideoEncoderOptions {
         /// Encoding bitrate. Default value depends on chosen encoder.
         bitrate: Option<VideoEncoderBitrate>,
 
+        /// (**default=`5000`**) Interval between keyframes, in milliseconds.
+        keyframe_interval_ms: Option<u64>,
```
wkozyra95: What happens if you set this value to e.g. 5000.234 ? Is this 5000 or validation error?

I don't think we will need sub millisecond precision, so if it is rounding automatically then it is fine, but if not it needs to be a float

smelter-core/src/pipeline/encoder/vulkan_h264.rs
```diff
@@ -56,7 +57,19 @@ impl VideoEncoder for VulkanH264Encoder {
...
+        let mut encoder_params =
+            device.encoder_parameters_high_quality(video_params, rate_control)?;
+
+        let gop_size_raw =
+            gop_size_from_ms_framerate(options.keyframe_interval_ms as u64, framerate) as u32;
+        let gop_size = if gop_size_raw == 0 {
+            NonZero::new(1u32).unwrap()
+        } else {
+            NonZero::new(gop_size_raw).unwrap()
+        };
```
wkozyra95: If you need NonZero then I think serde already provdes support for that type.

Also 
- if you need to do something like this then this would be better (unwrap event if you know it is safe should be a last resort)
```suggestion
        let gop_size = NonZero::new(gop_size_raw).unwrap_or(1)
```
- don't use type name in literal if it is not necessary(I did not test it, but I think 1u32 could be just 1 there)

smelter-core/src/pipeline/encoder/utils.rs
```diff
@@ -21,3 +21,9 @@ pub(super) fn bitrate_from_resolution_framerate(
         max_bitrate: max_bitrate as u64,
     }
 }
+
+pub(super) fn gop_size_from_ms_framerate(keyframe_interval_ms: u64, framerate: Framerate) -> u64 {
```
wkozyra95: do not operate on numbers when it comes to time, it should be Duration

smelter-core/src/codecs/vp8.rs
```diff
@@ -7,6 +7,7 @@ use crate::codecs::VideoEncoderBitrate;
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct FfmpegVp8EncoderOptions {
     pub bitrate: Option<VideoEncoderBitrate>,
+    pub keyframe_interval_ms: u64,
```
wkozyra95:
```suggestion
    pub keyframe_interval: Duration,
```
or if it make sense
```suggestion
    pub keyframe_interval: NonZero<Duration>,
```

smelter-api/src/output/hls.rs
```diff
@@ -43,6 +43,9 @@ pub enum HlsVideoEncoderOptions {
         /// Encoding bitrate. Default value depends on chosen encoder.
         bitrate: Option<VideoEncoderBitrate>,
 
+        /// (**default=`5000`**) Interval between keyframes, in milliseconds.
```
wkozyra95: Shouldn't this be max interval? I'm fine with this too

JBRS307: Depends on the encoder, I will check for which one it means what and change accordingly.

smelter-core/src/pipeline/encoder/vulkan_h264.rs
JBRS307: 1. `NonZero<Duration>` is not possible, so I cannot put it in api like that.
2. This suggestion does not work because there is a type mismatch, the similar thing closest to this is
`NonZero::new(gop_size_raw).unwrap_or(NonZero::new(1).unwrap())` It still required raw `.unwrap()`, but no better idea at the moment.

smelter-core/src/codecs/vp8.rs
JBRS307: `NonZero<Duration>` is not possible.

smelter-api/src/output/hls.rs
JBRS307: Changed that to `f64`. In implementation now there is a conversion which rounds that with `.round()` function.

smelter-core/src/pipeline/encoder/utils.rs
```diff
@@ -21,3 +23,8 @@ pub(super) fn bitrate_from_resolution_framerate(
         max_bitrate: max_bitrate as u64,
     }
 }
+
+pub(super) fn gop_size_from_ms_framerate(keyframe_interval: Duration, framerate: Framerate) -> u64 {
+    let framerate = framerate.num as f64 / framerate.den as f64;
+    (framerate * keyframe_interval.as_secs_f64()).round() as u64
```
JBRS307, line 29: Round before multiplication

ts/smelter/src/types/output/hls.ts
JBRS307, line 1: Update the comments in `ts-sdk` too (applies to all files)

smelter-api/src/output/hls.rs
```diff
@@ -54,6 +57,9 @@ pub enum HlsVideoEncoderOptions {
         /// Encoding bitrate. If not provided, bitrate is calculated based on resolution and framerate.
         /// For example at 1080p 30 FPS the average bitrate is 5000 kbit/s and max bitrate is 6250 kbit/s.
         bitrate: Option<VideoEncoderBitrate>,
+
+        /// (**default=`5000`**) Interval between keyframes, in milliseconds.
+        keyframe_interval: Option<f64>,
```
wkozyra95: I suggested removing `_ms` in core because it is already have a type. Types in API need to have `_ms` in the name

smelter-api/src/output/hls_into.rs
```diff
@@ -1,4 +1,5 @@
 use crate::common_core::prelude as core;
+use crate::output::common_into::duration_from_keyframe_interval;
```
wkozyra95:
```suggestion
```
make it pub(crate)

JBRS307: It requried adding one additional reexport in `output.rs` in `smelter-api`.
```rs
pub(crate) use common_into::*;
```

smelter-api/src/output/common_into.rs
```diff
@@ -111,3 +113,18 @@ impl TryFrom<VideoEncoderBitrate> for core::VideoEncoderBitrate {
         }
     }
 }
+
+pub(crate) fn duration_from_keyframe_interval(
+    keyframe_interval: &Option<f64>,
```
wkozyra95, line 118: I think in this case it is fine, because it would be really anoying to handle defaults in all encoders. However, in general when you writing utility funcitons that convert something it is not their responsibility to handle the Option, they should return value, and options should be handled by the caller

smelter-core/src/pipeline/encoder/utils.rs
```diff
@@ -21,3 +23,8 @@ pub(super) fn bitrate_from_resolution_framerate(
         max_bitrate: max_bitrate as u64,
     }
 }
+
+pub(super) fn gop_size_from_ms_framerate(keyframe_interval: Duration, framerate: Framerate) -> u64 {
+    let framerate = framerate.num as f64 / framerate.den as f64;
+    (framerate * keyframe_interval.as_secs_f64().round()).round() as u64
```
wkozyra95:
```suggestion
    (framerate * keyframe_interval.as_secs_f64()).round() as u64
```
This round lowers the precision to 1 second. I assume it is a mistake?

PR #1592, opened 2025-12-08
[rtmp] Add amf0 parser and encoder

wkazmierczak:
Crucial amf0 changes adb9b31 cherry picked to #1625

### DIFF
--- a/rtmp-server/src/amf0/parser.rs
+++ b/rtmp-server/src/amf0/parser.rs
@@ -1,0 +1,145 @@
+use bytes::{Buf, Bytes};
+use std::collections::HashMap;
+use thiserror::Error;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum AmfValue {
+    Number(f64),
+    Boolean(bool),
+    String(String),
+    Object(HashMap<String, AmfValue>),
+    Null,
+    Array(Vec<AmfValue>),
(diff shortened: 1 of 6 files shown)


rtmp-server/src/amf0/parser.rs
```diff
@@ -0,0 +1,145 @@
...
+
+        let marker = buf.get_u8();
+
+        match marker {
+            // number
+            0x00 => {
+                if buf.remaining() < 8 {
+                    return Err(ParseError::InsufficientData);
+                }
+                let number = buf.get_f64();
```
JBRS307, line 57: After reading the docs, this function treats bytes as if they are in network byte order, so it should work. Same with other functions, for example `.get_u32()`

rtmp-server/src/amf0/parser.rs
```diff
@@ -0,0 +1,145 @@
...
+            // strict array
+            0x0A => {
+                if buf.remaining() < 4 {
+                    return Err(ParseError::InsufficientData);
+                }
+                let size = buf.get_u32() as usize;
+                let mut array = Vec::with_capacity(size);
+                let mut current_buf = buf;
+
+                for _ in 0..size {
```
wkozyra95: You don't know that this message is in the buffer, you only checked if you have at least 4 bytes

PR #1595, opened 2025-12-08
Fix the V4L2 input detecting resolution change every frame

jerzywilczek:
This issue used to happen when using the OBS virtual camera as an input device. Obs would sometimes add extra 2048 bytes at the end of the buffer it sends us, which would cause us to detect a resolution change.

Now, resolution change is detected when the difference of the received buffer length and the expected length is greater than 1% of the received length.

### DIFF
--- a/smelter-core/src/pipeline/v4l2/v4l2_input.rs
+++ b/smelter-core/src/pipeline/v4l2/v4l2_input.rs
@@ -272,9 +272,13 @@ impl InputState<'_> {
                 resolution, format, ..
             } = &self.config;
 
+            // Some devices, most notably the OBS virtual camera, stuck extra bytes at the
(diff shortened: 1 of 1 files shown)


smelter-core/src/pipeline/v4l2/v4l2_input.rs
```diff
@@ -274,7 +274,8 @@ impl InputState<'_> {
 
             let data = match format {
                 V4l2Format::Yuyv => {
-                    if frame.len() != resolution.width * resolution.height * 2 {
+                    let expected_length = (resolution.width * resolution.height * 2) as f64;
+                    if (frame.len() as f64 - expected_length).abs() > expected_length * 0.01 {
```
wkozyra95, line 281: Add comment explaining that  0.01 and why we need it

wkozyra95 (approved):

PR #1596, opened 2025-12-09
[vk-video] Flush parser in decoder flush

noituri:


vk-video/src/vulkan_video.rs
```diff
@@ -248,8 +260,20 @@ impl BytesDecoder {
...
+
+        let mut result = Vec::new();
+        for unsorted_frame in unsorted_frames {
+            let mut sorted_frames = self.frame_sorter.put(unsorted_frame);
+            result.append(&mut sorted_frames);
+        }
+
+        result.append(&mut self.frame_sorter.flush());
+
+        Ok(result)
```
jerzywilczek: this duplication is starting to worry me. I think we should get rid of it. If you want, do it in this PR, if not we'll have to refactor how this works very soon.

PR #1605, opened 2025-12-11
Upgrade DeckLink SDK from 14.0 to 15.3

michael-dm:
Fixes symbol mismatch (`CreateVideoFrameAncillaryPacketsInstance_0001` vs `_0002`) causing silent DeckLink input failures.

**Changes:**
- Replace vendored SDK 14.0 headers with 15.3
- Update `video_input_frame_bytes()` to use `IDeckLinkVideoBuffer` with `StartAccess`/`EndAccess` per [migration guide](https://documents.blackmagicdesign.com/UserManuals/DeckLinkSDKMigrationGuide.pdf)
- Rename `MemoryAllocator` → `VideoBufferAllocator` to match SDK rename

**Testing:** Requires DeckLink hardware.


decklink/cpp/api.cpp
```diff
@@ -274,10 +274,17 @@ long video_input_frame_row_bytes(IDeckLinkVideoInputFrame *frame) {
...
 uint8_t *video_input_frame_bytes(IDeckLinkVideoInputFrame *frame) {
+  IDeckLinkVideoBuffer *videoBuffer = nullptr;
+  if (frame->QueryInterface(IID_IDeckLinkVideoBuffer, (void **)&videoBuffer) != S_OK) {
+    throw std::runtime_error("IDeckLinkVideoInputFrame::QueryInterface(IID_IDeckLinkVideoBuffer) failed.");
+  }
+
   void *buffer = nullptr;
-  auto result = frame->GetBytes(&buffer);
+  auto result = videoBuffer->GetBytes(&buffer);
+  videoBuffer->Release();
```
wkozyra95, line 291: https://documents.blackmagicdesign.com/UserManuals/DeckLinkSDKMigrationGuide.pdf?_v=1732780810000

This suggest that calling StartAccess before GetBytes and EndAccess after might be necessary. I'm assuming it works for you, but it might be specific to your capture card. I'm guessing here, but maybe on some cards the captured frame is not stored in memory and that Start Access might be necessary.

michael-dm:
Good catch, thanks for the link. Added `StartAccess(bmdBufferAccessRead)` before `GetBytes()` and `EndAccess` after, per the migration guide.

PR #1611, opened 2025-12-12
Update to Rust `1.92`

noituri:


wkozyra95 (approved):

.github/workflows/ts_check.yml
```diff
@@ -22,7 +22,7 @@ jobs:
             - name: 🔧 Install the rust toolchain
               uses: dtolnay/rust-toolchain@stable
               with:
-                  toolchain: 1.91.0
+                  toolchain: 1.92.0
```
JBRS307:
```suggestion
                  toolchain: 1.92
```
This way it would use the newest patch version, and not `1.92.0` specifically

PR #1613, opened 2025-12-12
[examples] Use RTMP output where possible

wkozyra95:
Originally examples were using RTP for everything, but now RTMP is a lot easier, so unless we are testing different protocol specifically it's best to RTMP


integration-tests/examples/view_border_transition.rs
```diff
@@ -1,166 +0,0 @@
-use anyhow::Result;
```
wkozyra95, line 1: rescaler example did almost the same

PR #1614, opened 2025-12-12
Decklink add support for 8bit ARGB and BGRA textures

wkozyra95:


smelter-core/src/pipeline/decklink/capture.rs
```diff
@@ -136,7 +137,7 @@ impl ChannelCallbackAdapter {
         data: bytes::Bytes,
         pts: Duration,
     ) -> Frame {
-        let data = if width != bytes_per_row * 2 {
+        let data = if width * 2 != bytes_per_row {
```
wkozyra95, line 140: This was a bug that disabled performance optimization for a case where rows of bytes do not have any padding, but did not break anything

PR #1615, opened 2025-12-12
Format autodection in DeckLink

wkozyra95:


smelter-core/src/pipeline/decklink/capture.rs
```diff
@@ -264,50 +264,38 @@ impl ChannelCallbackAdapter {
...
+                PixelFormat::Format8BitYUV
+            }
+            Colorspace::RGB444 => {
+                if new_format.bit_depth != BitDepth::Depth8Bit {
+                    warn!(
+                        "Format changed to {:?}. Forcing 8-bit.",
+                        new_format.bit_depth
+                    )
+                }
+                PixelFormat::Format8BitBGRA
```
wkozyra95, line 293: There is not enough information whether this should be BGRA or ARGB, I suspect that this is a conversion card my do on the fly.

PR #1617, opened 2025-12-15
[ts-sdk] Fix relative URL

wkozyra95:
Old version worked for deploying on `/`, but if you want to hos smelter under `https://example.com/<path prefix>` and prefix all routes then `new URL`  would override that path prefix and send request to wrong url.

Tested on node version, not tested the browser version

### DIFF
--- a/ts/smelter-node/src/fetch.ts
+++ b/ts/smelter-node/src/fetch.ts
@@ -3,6 +3,7 @@ import http from 'http';
 import https from 'https';
 import { Stream } from 'stream';
 import { promisify } from 'util';
+import path from 'path';
 
 import fetch from 'node-fetch';
 import type FormData from 'form-data';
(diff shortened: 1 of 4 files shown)


ts/smelter-node/src/manager/existingInstance.ts
```diff
@@ -69,16 +69,12 @@ class ExistingInstanceManager implements SmelterManager {
...
-      await this.sendRequest({
-        method: 'POST',
-        route: '/api/reset',
-        body: {},
-      });
-      opts.logger.info('Sent reset request to existing Smelter instance.');
-    } catch (err) {
-      opts.logger.warn({ err }, 'Failed to reset existing Smelter instance.');
-    }
+    await this.sendRequest({
```
wkozyra95, line 72: failed reset should abort

ts/smelter-node/src/fetch.ts
```diff
@@ -38,19 +35,15 @@ export async function sendMultipartRequest(
...
   });
   if (response.status >= 400) {
     const err: any = new Error(`Request to Smelter server failed.`);
     err.response = response;
-    try {
-      err.body = await response.json();
-    } catch {
-      err.body = await response.text();
-    }
+    err.body = await readErrorBody(response);
```
wkozyra95, line 46: if json failes, then text will also faile because we read body second time

PR #1623, opened 2025-12-18
[rtmp] Read RTMP chunks

wkazmierczak:


rtmp/src/chunk.rs
```diff
@@ -0,0 +1,32 @@
...
+#[derive(Debug, Clone)]
+pub(crate) struct ChunkHeader {
+    pub fmt: ChunkType,
+    pub cs_id: u32,
+    pub timestamp: u32,
+    pub timestamp_delta: u32,
+    pub msg_len: u32,
+    pub msg_type_id: u8,
+    pub msg_stream_id: u32,
+    pub has_extended_timestamp: bool,
```
JBRS307: Unnecessary field, it may be usable only when sending, however it is not planned for now.

rtmp/src/message_reader.rs
```diff
@@ -0,0 +1,210 @@
...
+            partial_payloads: HashMap::new(),
+            chunk_size: 128, // Default RTMP chunk size
+        }
+    }
+
+    // https://rtmp.veriskope.com/docs/spec/#5311-chunk-basic-header
+    fn read_basic_header(&mut self) -> Result<(ChunkType, u32), RtmpError> {
+        let byte = self.read_u8()?;
+        let fmt = ChunkType::from(byte >> 6);
+        let cs_id_initial = byte & 0b111111;
```
JBRS307:
```suggestion
        let cs_id_initial = byte & 0x3F;
```
EDIT: I misread it, now it is correct

rtmp/src/message_reader.rs
```diff
@@ -0,0 +1,210 @@
...
+                msg_type_id = self.read_u8()?;
+                msg_stream_id = self.read_u32_le()?;
+                timestamp_delta = 0;
+            }
+            ChunkType::NoMessageStreamId => {
+                // type 1 (7 bytes)
+                timestamp_delta = self.read_u24()?;
+                msg_len = self.read_u24()?;
+                msg_type_id = self.read_u8()?;
+                // reuse msg_stream_id from prev
```
JBRS307:
```suggestion
                // reuse msg_stream_id from previous chunk header
```

rtmp/src/message_reader.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    fn read_u8(&mut self) -> Result<u8, RtmpError> {
+        let mut buf = [0u8; 1];
+        self.stream.read_exact(&mut buf)?;
+        Ok(buf[0])
+    }
+    fn read_u24(&mut self) -> Result<u32, RtmpError> {
+        let mut buf = [0u8; 3];
+        self.stream.read_exact(&mut buf)?;
+        Ok(((buf[0] as u32) << 16) | ((buf[1] as u32) << 8) | (buf[2] as u32))
```
JBRS307:
```suggestion
        Ok(u32::from_be_bytes([0, buf[0], buf[1], buf[2]]))
```
Easier to read this way

rtmp/src/message_reader.rs
```diff
@@ -0,0 +1,210 @@
...
+            };
+
+            self.prev_headers.insert(cs_id, header.clone());
+
+            let accumulator =
+                self.partial_payloads
+                    .entry(cs_id)
+                    .or_insert_with(|| PayloadAccumulator {
+                        length: header.msg_len as usize,
+                        buffer: BytesMut::with_capacity(header.msg_len as usize),
```
JBRS307: Same vulnerability as the one Copilot mentioned in previous PR, properly forged header can cause memory exhaustion.

rtmp/src/server.rs
```diff
@@ -136,8 +137,17 @@ fn handle_client(
...
+    for msg_result in message_reader {
+        let msg = match msg_result {
+            Ok(msg) => msg,
+            Err(error) => {
+                error!(?error, "Error reading RTMP message");
+                break;
+            }
+        };
+
+        debug!(msg_type=?msg.type_id,  "RTMP message received");
```
```suggestion
        trace!(msg_type=?msg.type_id,  "RTMP message received");
```
Consider making it `trace!`, as it may turn out too spamy for `debug!`

rtmp/src/message_reader.rs
wkazmierczak: yeah I know, for me personally it is easier to write like this, but I'll change it

wkazmierczak: and i think `0xFF` is not correct. it should be `0x3F`

rtmp/src/message_reader.rs
wkazmierczak: I'll add checks

rtmp/src/message_reader.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        let cs_id = match cs_id_initial {
+            0 => {
+                let next_byte = self.read_u8()?;
+                64 + (next_byte as u32)
+            }
+            1 => {
+                let second_byte = self.read_u8()?;
+                let third_byte = self.read_u8()?;
+                64 + (second_byte as u32) + ((third_byte as u32) << 8) // not sure, inconsistency in spec
```
JBRS307:
```suggestion
                64 + (second_byte as u32) + ((third_byte as u32) * 256)
```
The spec explicitly specifies this formula for cs_id calculation so no inconsistency here.

Changed the code to more readable version, the same as is in spec.

wkazmierczak: text is explicit, but this illustration, made me confused
image

wkozyra95 (changes requested):
discussed offline, current implementation might lose data

rtmp/src/buffered_stream_reader.rs
```diff
@@ -0,0 +1,98 @@
...
+                        _ => {
+                            return Err(err.into());
+                        }
+                    }
+                }
+            };
+        }
+    }
+
+    pub(crate) fn peek(&self, index: usize) -> Result<u8, RtmpError> {
```
wkozyra95: I don't think any of those methods should be here (this one and below). You can expose something that returns entire slice, and something that reads X bytes from buffer.

In that scenario read_bytes would happen only once per chunk

rtmp/src/message_reader.rs
```diff
@@ -0,0 +1,104 @@
+use crate::chunk::{ChunkType, RtmpChunk, RtmpChunkReader};
```
wkozyra95: cleanup imports

rtmp/src/message_reader.rs
```diff
@@ -0,0 +1,104 @@
...
+            payload: acc.buffer.freeze(),
+        })
+    }
+}
+
+impl Iterator for RtmpMessageReader {
+    type Item = Result<RtmpMessage, RtmpError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
```
wkozyra95, line 75: This iterator should end if tcp connection is broken

rtmp/src/message_reader.rs
```diff
@@ -0,0 +1,104 @@
...
+use crate::message::RtmpMessage;
+use bytes::BytesMut;
+use std::cmp::min;
+use std::collections::HashMap;
+use std::net::TcpStream;
+use std::sync::{Arc, atomic::AtomicBool};
+
+pub struct RtmpMessageReader {
+    chunk_reader: RtmpChunkReader,
+    accumulators: HashMap<u32, PayloadAccumulator>,
```
wkozyra95, line 16: I would look for some better way of storing this data, It would be easy to prepare payload that will use all the memory in the system.

It's fine if you want to leave that for follow up PR, but if that is the case create an issue for that

rtmp/src/chunk.rs
```diff
@@ -0,0 +1,296 @@
...
+            payload_vec.push(data[cursor + i]);
+        }
+        let payload = Bytes::from(payload_vec);
+        cursor += chunk_payload_size;
+
+        Ok((RtmpChunk { header, payload }, cursor))
+    }
+}
+
+fn read_u24(data: &VecDeque<u8>, start: usize) -> u32 {
```
wkazmierczak, line 275: Maybe some wrapper on data and cursor

rtmp/src/message_reader.rs
wkazmierczak: I'll do it in follow up

PR #1630, opened 2025-12-23
Better setting of FFmpeg logger level

### LINKED ISSUE #1582: Better FFmpeg logger handling.
- Check if it is possible without manually filtering out messages sent by FFmpeg.
- Set ffmpeg logger level according to given smelter env.

JBRS307:
Closes #1582

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -24,18 +24,15 @@ pub enum FfmpegLogLevel {
     Trace,
 }
 
-impl From<FfmpegLogLevel> for i32 {
-    // Source of numeric logger level values:
-    // https://www.ffmpeg.org/doxygen/3.0/group__lavu__log__constants.html
-    fn from(value: FfmpegLogLevel) -> Self {
-        use self::FfmpegLogLevel::*;
-        match value {
-            Error => 16,
-            Warn => 24,
-            Info => 32,
-            Verbose => 40,
-            Debug => 48,
-            Trace => 56,
+impl FfmpegLogLevel {
+    fn into_i32(self) -> i32 {
+        match self {
+            FfmpegLogLevel::Error => 16,
+            FfmpegLogLevel::Warn => 24,
+            FfmpegLogLevel::Info => 32,
+            FfmpegLogLevel::Verbose => 40,
+            FfmpegLogLevel::Debug => 48,
+            FfmpegLogLevel::Trace => 56,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/logger.rs
```diff
@@ -19,7 +19,25 @@ pub enum FfmpegLogLevel {
     Error,
     Warn,
     Info,
+    Verbose,
     Debug,
+    Trace,
+}
+
+impl From<FfmpegLogLevel> for i32 {
+    // Source for numeric logger level values:
```
JBRS307:
```suggestion
    // Source of numeric logger level values:
```

src/logger.rs
```diff
@@ -19,7 +19,25 @@ pub enum FfmpegLogLevel {
...
+    Verbose,
     Debug,
+    Trace,
+}
+
+impl From<FfmpegLogLevel> for i32 {
+    // Source of numeric logger level values:
+    // https://www.ffmpeg.org/doxygen/3.0/group__lavu__log__constants.html
+    fn from(value: FfmpegLogLevel) -> Self {
+        use self::FfmpegLogLevel::*;
```
wkozyra95: I don't think there is anything wrong with pattern like this (importing variant names directlly), but it is inconsitent with the rest of the codebase, so I would preffer 
```
FfmpegLogLevel::Error => 16,
```

src/logger.rs
```diff
@@ -19,7 +19,25 @@ pub enum FfmpegLogLevel {
     Error,
     Warn,
     Info,
+    Verbose,
     Debug,
+    Trace,
+}
+
+impl From<FfmpegLogLevel> for i32 {
```
wkozyra95: From trait implementation is automatically public, but this is very local implementation detail that is used in only one place. I would  write it as regular function or method on FfmpegLogLevel e.g. into_i32 or as_i32

src/logger.rs
```diff
@@ -104,6 +98,6 @@ pub fn init_logger(opts: LoggerConfig) {
     }
 
     unsafe {
-        ffmpeg_next::sys::av_log_set_callback(Some(ffmpeg_log_callback));
+        ffmpeg_next::sys::av_log_set_level(ffmpeg_logger_level().into());
```
wkozyra95: are you sure that it works with higher error thresholds? Specifically if you set Error that Warn logs do not show up?

JBRS307: It works, double checked

PR #1631, opened 2025-12-23
[ts-sdk] Fix for `Image` component not accepting `.jpg` files

### LINKED ISSUE #1628: Jpg extension not working in `Image` component
Passing a url to an image with .jpg extension as a `source` prop on `Image` component results in an error ('Unsupported image type'). Replacing the .jpg with .jpeg in the url resolved the issue, but it's ugly, unintuitive and may not always be possible. Could you handle the .jpg extension the same way you handle the .jpeg?

JBRS307:
Closes #1628

### DIFF
--- a/ts/smelter/src/components/Image.ts
+++ b/ts/smelter/src/components/Image.ts
@@ -61,7 +61,6 @@ function Image(props: ImageProps) {
       props.source?.startsWith('http://') || props.source?.startsWith('https://')
         ? { url: props.source }
         : { serverPath: props.source };
-    const assetType = 'auto';
 
     let registerPromise: Promise<any>;
     const task = newBlockingTask(ctx);
@@ -70,7 +69,7 @@ function Image(props: ImageProps) {
       try {
         registerPromise = ctx.registerImage(newImageId, {
           ...pathOrUrl,
-          assetType,
+          assetType: 'auto',
         });
         await registerPromise;
         setIsImageRegistered(true);


ts/smelter/src/components/Image.ts
```diff
@@ -63,7 +63,12 @@ function Image(props: ImageProps) {
         ? { url: props.source }
         : { serverPath: props.source };
     const extension = props.source?.split('.').pop();
-    const assetType = extension && isValidImageType(extension) ? extension : undefined;
+    const assetType =
+      extension === 'jpg'
+        ? 'jpeg'
+        : extension && isValidImageType(extension)
+          ? extension
+          : undefined;
```
noituri: Nested ternary ifs are not really readable. I would rewrite it using normal ifs

wkozyra95: agree, and cases like this are usually a good part to extract

but in this case I would replace this logic with using `auto` field instead

ts/smelter/src/components/Image.ts
```diff
@@ -62,15 +61,9 @@ function Image(props: ImageProps) {
       props.source?.startsWith('http://') || props.source?.startsWith('https://')
         ? { url: props.source }
         : { serverPath: props.source };
-    const extension = props.source?.split('.').pop();
-    const assetType = extension && isValidImageType(extension) ? extension : undefined;
+    const assetType = 'auto';
```
wkozyra95: inline that in registerImage call below

PR #1639, opened 2026-01-12
Audio stretch/squash on input buffer change

wkozyra95:
Snapshots: https://github.com/smelter-labs/smelter-snapshot-tests/pull/73
Not sure 100% why one of the snapshots is still mismatched, but the difference is on the first sample, so I suspect it might be caused by sligh shift caused by different resampler delay

### DIFF
--- a/smelter-core/src/audio_mixer/input/mod.rs
+++ b/smelter-core/src/audio_mixer/input/mod.rs
@@ -12,7 +12,7 @@ mod resampler;
 pub(super) struct AudioMixerInput {
     input_sender: Sender<AudioMixerInputEvent>,
     result_receiver: Receiver<AudioMixerInputResult>,
-    next: Option<AudioMixerInputResult>,
+    next_chunk: Option<AudioMixerInputResult>,
 }
 
 #[derive(Debug)]
@@ -35,7 +35,7 @@ impl AudioMixerInput {
         Self {
             input_sender,
             result_receiver,
-            next: None,
+            next_chunk: None,
         }
     }
 
@@ -50,7 +50,7 @@ impl AudioMixerInput {
 
     pub fn get_samples(&mut self, pts_range: (Duration, Duration)) -> Option<Vec<(f64, f64)>> {
         loop {
-            if self.next.is_none() {
+            if self.next_chunk.is_none() {
                 let Ok(result) = self
                     .result_receiver
                     .recv_timeout(Duration::from_millis(100))
@@ -59,11 +59,11 @@ impl AudioMixerInput {
                     error!("Failed to read samples.");
                     return None;
                 };
-                self.next = Some(result)
+                self.next_chunk = Some(result)
             }
-            let next = self.next.as_ref()?;
+            let next = self.next_chunk.as_ref()?;
             if next.pts_range == pts_range {
-                return Some(self.next.take()?.samples);
+                return Some(self.next_chunk.take()?.samples);
             }
             error!("Found batch for different range. This should not happen");
             if next.pts_range.0 > pts_range.0 || next.pts_range.1 > pts_range.1 {
@@ -72,7 +72,7 @@ impl AudioMixerInput {
                 return None;
             }
             // drop the old batch and wait for next
-            self.next = None
+            self.next_chunk = None
         }
     }
 }


smelter-core/src/pipeline/utils/input_buffer.rs
```diff
@@ -114,14 +114,14 @@ impl LatencyOptimizedBuffer {
...
-        trace!(effective_buffer=?next_pts.saturating_sub(self.sync_point.elapsed()));
+        trace!(effective_buffer=?next_pts.saturating_sub(self.sync_point.elapsed()), dynamic_buffer=?self.dynamic_buffer);
 
         if next_pts > self.sync_point.elapsed() + self.max_hard_threshold {
             let first_pts = self.state.set_too_large(next_pts);
             if next_pts.saturating_sub(first_pts) > STABLE_STATE_DURATION {
                 self.dynamic_buffer = self
                     .dynamic_buffer
-                    .saturating_sub(self.dynamic_buffer / 100);
+                    .saturating_sub(self.dynamic_buffer / 200);
```
wkozyra95, line 124: This will need to be reworked, but for now I just increased the step because it squashed the audio too much

smelter-core/src/queue/audio_queue.rs
```diff
@@ -180,22 +181,13 @@ impl AudioQueueInput {
...
-            .collect::<Vec<InputAudioSamples>>();
-
-        // Drop all batches older than `end_pts`. Entire batch (all samples inside) has to be older.
-        while self
-            .queue
-            .front()
-            .is_some_and(|batch| batch.end_pts < end_pts)
+        let mut popped_samples = vec![];
+        while let Some(batch) = self.queue.front()
+            && batch.start_pts <= end_pts + Duration::from_millis(40)
```
wkozyra95, line 188: The resampler has some latency, so we need to send a bit more data. I think 10ms would be enough, howeverr this way we have a bit more space for squashing/stretching without artifacts

smelter-core/src/pipeline/decoder.rs
```diff
@@ -77,8 +67,9 @@ pub(crate) trait AudioDecoder: Sized {
     type Options: Send + 'static;
 
     fn new(ctx: &Arc<PipelineCtx>, options: Self::Options) -> Result<Self, DecoderInitError>;
-    fn decode(&mut self, chunk: EncodedInputChunk) -> Result<Vec<DecodedSamples>, DecodingError>;
-    fn flush(&mut self) -> Vec<DecodedSamples>;
+    fn decode(&mut self, chunk: EncodedInputChunk)
+    -> Result<Vec<InputAudioSamples>, DecodingError>;
```
noituri, line 71: This formatting looks wrong

tools/nix/flake.nix
```diff
@@ -63,6 +63,7 @@
             yarn
 
             # Rust
+            rustup
```
wkozyra95: I will revert this at some point. I need that temporarlly because rust 1.92 is not yet avaialble in nixpkgs-unstable channel

smelter-core/src/audio_mixer/input/resampler.rs
```diff
@@ -0,0 +1,613 @@
...
+            let samples = match self.channels {
+                AudioChannels::Mono => AudioSamples::Mono(vec![0.0; zero_samples]),
+                AudioChannels::Stereo => AudioSamples::Stereo(vec![(0.0, 0.0); zero_samples]),
+            };
+            self.resampler_input_buffer.push_back(samples)
+        }
+        if start_pts + CONTINUITY_THRESHOLD < self.input_buffer_end_pts {
+            trace!("Detected overlapping batches, dropping.");
+            return;
+        }
```
noituri, line 199: I'm not sure why we allow any overlap here (within `CONTINUITY_THRESHOLD` bounds), having any overlap seems incorrect

smelter-core/src/audio_mixer/input/resampler.rs
```diff
@@ -0,0 +1,613 @@
...
+        if !self.before_first_resample {
+            return None;
+        }
+
+        let input_buffer_start_pts = self.input_buffer_start_pts();
+
+        // if entire input buffer is in the future
+        if pts_range.1 < self.input_buffer_start_pts() {
+            let zero_samples = (0.02 * self.output_sample_rate as f64) as usize;
+            let samples = match self.channels {
```
noituri, line 296: Why `0.2`? Shouldn't `zero_samples` be calculated from `pts_range`

integration-tests/src/validation/audio/artificial.rs
```diff
@@ -96,7 +96,10 @@ fn compare(
 
         if !(frequency_match && value_match) {
             return Err(anyhow::anyhow!(
-                "Audio mismatch at sample {first_sample} on channel {channel}: actual = {actual_bins:#?}, expected = {expected_bins:#?}",
+                "Audio mismatch at sample {first_sample} on channel {channel}. (value_diff: {}, tolerance: {}, frequency_diff: {}",
```
wkazmierczak, line 99:
```suggestion
                "Audio mismatch at sample {first_sample} on channel {channel}. (value_diff: {}, tolerance: {}, frequency_diff: {})",
```

smelter-core/src/audio_mixer/input/resampler.rs
wkazmierczak: If i get it correctly, in that case pts_range is entirely before first sample, so relaying on pts_range does not make much sense. 20 ms of silence seems reasonable., but correct me if i am wrong

smelter-core/src/pipeline/decoder.rs
wkozyra95: agree, but I don't think it's bad enough to disable formatting here

smelter-core/src/audio_mixer/input/resampler.rs
wkozyra95: not sure I understand. that is way we are adding audio stretching, if there was no overlap, there wouldn't be anything to stretch/squash

smelter-core/src/audio_mixer/input/resampler.rs
wkozyra95: yes @noituri it should be based on range (range will be always 20ms, but still it shouldn't be hardcoded)

smelter-core/src/audio_mixer/input/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+use crate::{audio_mixer::input::input_thread::start_input_thread, prelude::*};
+
+mod input_thread;
+mod resampler;
+
+#[derive(Debug)]
+pub(super) struct AudioMixerInput {
+    input_sender: Sender<AudioMixerInputEvent>,
+    result_receiver: Receiver<AudioMixerInputResult>,
+    next: Option<AudioMixerInputResult>,
```
JBRS307:
```suggestion
    next_chunk: Option<AudioMixerInputResult>,
```

Issue #1494, opened 2025-11-03
After updating wgpu to 28, change the API to utilize nv12 textures being renderable


jerzywilczek:
since #1382 is merged upstream, once wgpu has a new release we should use our contributions

PR #1651, opened 2026-01-15
[vk-video] Change API to utilize NV12 textures

noituri:
This PR changes the texture encode API. Now users will have to create NV12 texture with `COPY_SRC` usage, populate it with a video frame (e.g. render onto it) and then pass it to `vk-video`. Due to synchronization and queue ownership problems between `wgpu` and `vk-video`, we copy that texture to our own internal encode texture instead of using the user provided texture directly.

Closes: #1494

### DIFF
--- a/vk-video/src/vulkan_encoder.rs
+++ b/vk-video/src/vulkan_encoder.rs
@@ -662,7 +662,7 @@ impl VulkanEncoder<'_> {
         Ok((image, buffer))
     }
 
-    fn transfer_image_to_image(
+    fn copy_wgpu_texture_to_image(
         &mut self,
         frame: &Frame<wgpu::Texture>,
     ) -> Result<wgpu::hal::vulkan::CommandEncoder, VulkanEncoderError> {
@@ -697,6 +697,7 @@ impl VulkanEncoder<'_> {
         unsafe { encoder.begin_encoding(None)? }
         let buffer = unsafe { encoder.raw_handle() };
 
+        // TODO: This should be abstracted away to some helper function
         let mut layout = self
             .tracker
             .image_layout_tracker
@@ -1196,7 +1197,7 @@ impl VulkanEncoder<'_> {
         frame: Frame<wgpu::Texture>,
         force_idr: bool,
     ) -> Result<EncodedOutputChunk<Vec<u8>>, VulkanEncoderError> {
-        let _cmd_encoder = self.transfer_image_to_image(&frame)?;
+        let _cmd_encoder = self.copy_wgpu_texture_to_image(&frame)?;
 
         let is_keyframe = force_idr || self.idr_period_counter == 0;
         let result = self.encode(self.input_image.clone(), force_idr)?;


smelter-render/src/wgpu/utils/rg8_fill_value.wgsl
```diff
@@ -0,0 +1,26 @@
...
+
+    return output;
+}
+
+var<immediate> value: vec2<f32>;
+
+@fragment
+fn fs_main(input: VertexOutput) -> @location(0) vec2<f32> {
+    return value;
+}
```
jerzywilczek, line 26: why is this necessary? cant we just have an empty render pass with `load_op: Clear(color)` or something like that?

vk-video/src/device.rs
```diff
@@ -188,14 +189,17 @@ impl VulkanDevice {
 
         let mut vk_synch_2_feature =
             vk::PhysicalDeviceSynchronization2Features::default().synchronization2(true);
+        let mut vk_video_maintenance1_feature =
+            vk::PhysicalDeviceVideoMaintenance1FeaturesKHR::default().video_maintenance1(true);
```
jerzywilczek, line 197: maybe we should add a task to our board to start using this everywhere?

vk-video/src/vulkan_encoder.rs
```diff
@@ -639,6 +658,145 @@ impl VulkanEncoder<'_> {
...
+
+        let mut layout = self
+            .tracker
+            .image_layout_tracker
+            .lock()
+            .unwrap()
+            .map
+            .get(&self.input_image.key())
+            .unwrap()
+            .clone();
```
jerzywilczek, line 709: I sort of feel like the image layout tracker should maybe have a helper for this, but I leave it up to you if you want to do it in this PR. You can just leave a TODO here

vk-video/src/vulkan_encoder.rs
```diff
@@ -639,6 +658,145 @@ impl VulkanEncoder<'_> {
         Ok((image, buffer))
     }
 
+    fn transfer_image_to_image(
```
jerzywilczek: maybe rename this to say that  this is a wgpu -> vk image transfer?

smelter-render/src/wgpu/utils/rg8_fill_value.wgsl
noituri: This is how smelter does it in other cases and I wanted to be consistent with the rest of the code. I think it would make sense to change it to just `load_op` (assuming it would work) but I would change it in a separate PR for the rest of smelter.

vk-video/src/device.rs
noituri: yup, I'll make one

noituri: https://github.com/software-mansion/smelter/issues/1676

vk-video/src/vulkan_encoder.rs
noituri: Since we want to release new version of `vk-video` soon, let's just do it in a separate PR

PR #1652, opened 2026-01-15
[rtmp] Return RTMP media data instead of just bytes

wkazmierczak:
- Instead of just Bytes return structs RtmpVideoData/RtmpAudioData with more media details.
- Fix adding `timestamp_delta` multiple time per message.

### DIFF
--- a/rtmp/src/error.rs
+++ b/rtmp/src/error.rs
@@ -42,5 +42,5 @@ pub enum RtmpError {
     InternalBufferError(&'static str),
 
     #[error("FLV tag parsing failed: {0}")]
-    FLVParsingFailed(#[from] ParseError),
+    FlvParsingFailed(#[from] ParseError),
 }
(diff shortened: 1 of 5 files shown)


rtmp/src/error.rs
```diff
@@ -39,4 +40,7 @@ pub enum RtmpError {
 
     #[error("Internal buffer error: {0}")]
     InternalBufferError(&'static str),
+
+    #[error("FLV tag parsing failed: {0}")]
+    FLVParsingFailed(#[from] ParseError),
```
wkozyra95: case

rtmp/src/server.rs
```diff
@@ -11,10 +13,32 @@ use tracing::{error, info};
 pub type OnConnectionCallback = Box<dyn FnMut(RtmpConnection) + Send + 'static>;
 pub type RtmpUrlPath = Arc<str>;
 
+#[derive(Debug, Clone)]
+pub struct RtmpAudioData {
```
wkozyra95: enum

wkozyra95 (approved):

PR #1667, opened 2026-01-27
[rtmp/flv] Added handling of `SCRIPTDATA` in `flv`

JBRS307:
I am willing to restructure that in a followup to prepare it for `amf3` handling.

### DIFF
--- a/rtmp/flv/src/amf0/decoding.rs
+++ b/rtmp/flv/src/amf0/decoding.rs
@@ -59,8 +59,11 @@ impl Decoder {
             }
             LONG_STRING => AmfValue::LongString(self.decode_long_string(buf)?),
             TYPED_OBJECT => {
-                let (class_name, pairs) = self.decode_typed_object(buf)?;
-                AmfValue::TypedObject(class_name, pairs)
+                let (class_name, properties) = self.decode_typed_object(buf)?;
+                AmfValue::TypedObject {
+                    class_name,
+                    properties,
+                }
             }
 
             // TODO add switch to AMF3 (0x11)
@@ -190,8 +193,10 @@ impl Decoder {
         let class_name = self.decode_string(buf)?;
         let pairs = self.decode_object_pairs(buf)?;
 
-        self.complexes
-            .push(AmfValue::TypedObject(class_name.clone(), pairs.clone()));
+        self.complexes.push(AmfValue::TypedObject {
+            class_name: class_name.clone(),
+            properties: pairs.clone(),
+        });
         Ok((class_name, pairs))
     }
 
--- a/rtmp/flv/src/tag/scriptdata.rs
+++ b/rtmp/flv/src/tag/scriptdata.rs
@@ -28,7 +28,10 @@ pub enum ScriptDataValue {
         timezone_offset: i16,
     },
     LongString(String),
-    TypedObject(String, HashMap<String, ScriptDataValue>),
+    TypedObject {
+        class_name: String,
+        properties: HashMap<String, ScriptDataValue>,
+    },
 }
 
 impl ScriptData {
@@ -75,12 +78,18 @@ impl From<amf0::AmfValue> for ScriptDataValue {
                 timezone_offset,
             },
             amf0::AmfValue::LongString(s) => Self::LongString(s),
-            amf0::AmfValue::TypedObject(class_name, obj) => {
-                let tag_obj = obj
+            amf0::AmfValue::TypedObject {
+                class_name,
+                properties,
+            } => {
+                let tag_properties = properties
                     .into_iter()
                     .map(|(key, value)| (key, Self::from(value)))
                     .collect();
-                Self::TypedObject(class_name, tag_obj)
+                Self::TypedObject {
+                    class_name,
+                    properties: tag_properties,
+                }
             }
         }
     }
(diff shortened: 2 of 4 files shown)


rtmp/src/handle_client.rs
```diff
@@ -62,6 +65,12 @@ pub(crate) fn handle_client(
                     .send(video_data)
                     .map_err(|_| RtmpError::ChannelClosed)?;
             }
+            MessageType::DataMessageAmf0 => {
+                let metadata = parse_data_message(msg)?;
+                sender
+                    .send(metadata)
+                    .map_err(|_| RtmpError::ChannelClosed)?;
+            }
             _ => {} // possible metadata
```
JBRS307:
```suggestion
            _ => {}
```

rtmp/src/server.rs
```diff
@@ -23,6 +23,8 @@ pub enum RtmpMediaData {
     VideoConfig(VideoConfig),
     Audio(AudioData),
     AudioConfig(AudioConfig),
+
+    Metadata(ScriptData),
 }
```
```suggestion
pub enum RtmpStreamData {
Video(VideoData),
VideoConfig(VideoConfig),
Audio(AudioData),
AudioConfig(AudioConfig),

    Metadata(ScriptData),
}
```

rtmp/flv/src/tag/scriptdata.rs
```diff
@@ -0,0 +1,86 @@
...
+    Undefined,
+    EcmaArray(HashMap<String, ScriptDataValue>),
+    StrictArray(Vec<ScriptDataValue>),
+    Date {
+        unix_time: f64,
+        timezone_offset: i16,
+    },
+    LongString(String),
+    TypedObject(String, HashMap<String, ScriptDataValue>),
+}
```
JBRS307, line 35: It is 1:1 with `AmfValue` struct (for `amf0`), however it is a different abstraction as amf has broader use than just metadata

rtmp/flv/src/tag/scriptdata.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use bytes::Bytes;
+
+use crate::{
+    amf0::{self, decoding::decode_amf_values},
+    error::ParseError,
+};
+
+#[derive(Debug, Clone)]
+pub struct ScriptData {
```
JBRS307, line 12: re-export this

rtmp/flv/src/tag/scriptdata.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub values: Vec<ScriptDataValue>,
+}
+
+#[derive(Debug, Clone)]
+pub enum ScriptDataValue {
```
JBRS307, line 17: and re-export this

rtmp/src/server.rs
```diff
@@ -11,18 +11,22 @@ use std::{
...
 
+pub use flv::tag::scriptdata::{ScriptData, ScriptDataValue};
+use flv::{AudioChannels, AudioCodec, FrameType, VideoCodec};
+
 use crate::{error::RtmpError, handle_client::handle_client};
 
 pub type OnConnectionCallback = Box<dyn FnMut(RtmpConnection) + Send + 'static>;
 
-pub enum RtmpMediaData {
+pub enum RtmpStreamData {
```
wkozyra95: RtmpEvent, RtmpMessage

If the name collides with some internal types, rename the internal ones, the public API takes precendense

rtmp/src/server.rs
```diff
@@ -11,18 +11,22 @@ use std::{
 };
 
 use bytes::Bytes;
-use flv::{AudioChannels, AudioCodec, FrameType, VideoCodec};
 use tracing::{error, info};
 
+pub use flv::tag::scriptdata::{ScriptData, ScriptDataValue};
```
wkozyra95: if you really need to re-export, move this below regular `use` statements, but I don't see any reason to do it. This crate is so small that creating intermediate APIs seems unnecessary

rtmp/flv/src/amf0/decoding.rs
```diff
@@ -22,8 +22,8 @@ pub enum DecodingError {
 
 const OBJECT_END_MARKER: [u8; 3] = [0x00, 0x00, 0x09];
 
-pub fn decode_amf_values(rtmp_msg_payload: &[u8]) -> Result<Vec<AmfValue>, DecodingError> {
-    let mut buf = Bytes::copy_from_slice(rtmp_msg_payload);
+pub fn decode_amf_values(amf_encoded_payload: &[u8]) -> Result<Vec<AmfValue>, DecodingError> {
```
wkozyra95: payload does not make sense here

```suggestion
pub fn decode_amf_values(amf_bytes: &[u8]) -> Result<Vec<AmfValue>, DecodingError> {
```

also it might be worth describing in the comment what the bytes should represent

rtmp/src/server.rs
JBRS307: There is no collision, I renamed it because after adding metadata I think it makes more sense.

rtmp/src/server.rs
JBRS307: I did reexports so all 3 types (Audio, Video and Metadata) are imported from the same module.

PR #1668, opened 2026-01-27
Improve low latency input buffer

wkozyra95:


smelter-core/src/pipeline/utils/input_buffer.rs
```diff
@@ -79,58 +79,78 @@ pub(crate) struct LatencyOptimizedBuffer {
 
     /// If effective_buffer is above this threshold for a period of time, aggressively shrink
     /// the buffer.
-    max_hard_threshold: Duration,
+    shrink_threshold_1: Duration,
+    /// If effective_buffer is above this threshold for a period of time, shrink the buffer
+    /// quickly
+    shrink_threshold_2: Duration,
```
wkozyra95, line 85: Mostly renames, the only difference is that I added one more level where shrinking happens at different speeds

smelter-core/src/pipeline/utils/input_buffer.rs
```diff
@@ -79,58 +79,78 @@ pub(crate) struct LatencyOptimizedBuffer {
...
+        );
 
-        if next_pts > self.sync_point.elapsed() + self.max_hard_threshold {
+        if next_pts > self.sync_point.elapsed() + self.shrink_threshold_1 {
             let first_pts = self.state.set_too_large(next_pts);
             if next_pts.saturating_sub(first_pts) > STABLE_STATE_DURATION {
                 self.dynamic_buffer = self
                     .dynamic_buffer
-                    .saturating_sub(self.dynamic_buffer / 200);
+                    .saturating_sub(self.dynamic_buffer / 1000);
```
wkozyra95, line 139: In this case, the buffer will be at least 1 second, so the shift will be around 1 millisecond per packet, which might be too much because audio stretching is limited to 5%

It would be better to apply this adjustment based on packet time not packet number, but it's to complex for a to little gain for now.

smelter-core/src/audio_mixer/input/resampler.rs
```diff
@@ -9,6 +9,8 @@ use tracing::{debug, error, trace, warn};
 
 use crate::{AudioChannels, AudioSamples, prelude::InputAudioSamples};
 
+const MAX_STRETCH_FACTOR: f64 = 1.05;
```
wkozyra95, line 12: 10% was affecting pitch to much

smelter-core/src/pipeline/utils/input_buffer.rs
```diff
@@ -79,58 +79,78 @@ pub(crate) struct LatencyOptimizedBuffer {
...
+            max_desired_buffer,
+            shrink_threshold_1,
+            shrink_threshold_2,
         }
     }
 
     fn recalculate_buffer(&mut self, pts: Duration) {
-        const INCREMENT_DURATION: Duration = Duration::from_micros(200);
-        const DECREMENT_DURATION: Duration = Duration::from_micros(200);
+        // Increment duration is larger than decrement, because when buffer is to small
```
noituri:
```suggestion
        // Increment duration is larger than decrement, because when buffer is too small
```

smelter-core/src/pipeline/utils/input_buffer.rs
noituri, line 85: I'm not a fan of  `shrink_threshold_1` and `shrink_threshold_2` names. I think `hard` and `soft` words, or something along those lines, would be more self-explonatory

wkozyra95: hard and soft would be misleading, that's why I renamed it. I prefer just number because it does not convey extra information and forces someone to check the code.

Also if you just need threshold order, it is the same as order of those fields

smelter-core/src/pipeline/utils/input_buffer.rs
```diff
@@ -79,58 +79,78 @@ pub(crate) struct LatencyOptimizedBuffer {
 
     /// If effective_buffer is above this threshold for a period of time, aggressively shrink
     /// the buffer.
-    max_hard_threshold: Duration,
+    shrink_threshold_1: Duration,
+    /// If effective_buffer is above this threshold for a period of time, shrink the buffer
+    /// quickly
```
JBRS307, line 84: What is a "period of time"?

JBRS307: Ok, the answer is later in the code (`STABLE_STATE_DURATION` if I am not wrong). I would consider mentioning here that the default is 10 seconds

wkozyra95: It's not a default, default implies it can be overriden

Comments should explain why, code should explain what. I think this comment describes why we need this value, by specifying what it represents. The fact that we wait for a period of time before shrinking buffer, but not when growing, is important for the algorithm, actual duration of that period is not relevant.

I would include a specific value in the comment if the const was defined somwhere else, or if the value was dynamically calculated(e.g. like sample batch usually being 20ms), but in this case, this value is set just below.

PR #1682, opened 2026-02-02
[rtmp/flv] Added decoder for `amf3` serialized messages.

JBRS307:
Please note that it is just a decoder. It will need to be adjusted to work with the rtmp, because rtmp `amf3` messages are serialized as `amf0` messages with special tag that triggers switching to `amf3`. This tag has a scope only for ONE VALUE and not the whole message.

PR #1687, opened 2026-02-03
[rtmp] Deleted the `flv` crate and made it into a module within `rtmp` crate.

JBRS307:
- Merge after #1682

---

Two main reasons:
1. Naming issues - all reasonable names are taken
2. It is useful only in use with the rtmp, as it does not contain demuxing and parsing files.

### DIFF
--- a/rtmp/src/amf0/decoding.rs
+++ b/rtmp/src/amf0/decoding.rs
@@ -9,7 +9,7 @@ const OBJECT_END_MARKER: [u8; 3] = [0x00, 0x00, 0x09];
 /// Decode AMF0 encoded messages.
 ///
 /// `amf_bytes` must include whole AMF0 values. It can be a payload of `rtmp` Data or Command message.
-pub fn decode_amf0_values(amf_bytes: Bytes) -> Result<Vec<AmfValue>, DecodingError> {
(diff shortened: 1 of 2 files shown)


rtmp/src/error.rs
JBRS307, line 1: I deleted the `flv/error.rs` file and put all errors from it here. They were re-exported anyway, there are no naming conflicts.

rtmp/src/error.rs
```diff
@@ -44,3 +43,87 @@ pub enum RtmpError {
...
+
+    #[error("Invalid sound type header value: {0}")]
+    InvalidSoundType(u8),
+
+    #[error("Invalid AacPacketType header value: {0}")]
+    InvalidAacPacketType(u8),
+}
+
+#[derive(Error, Debug, Clone, PartialEq)]
+pub enum DecodingError {
```
JBRS307: Add info, that these errors regard amf

rtmp/src/amf3/mod.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(unused)]
```
JBRS307, line 1: Currently added this to the entire `amf3` module, as now no public re-export of any amf functionality is needed and this code is not yet integrated with the RTMP server.

Issue #1686, opened 2026-02-03
Investigate openAPI types autogeneration from rust types


JBRS307:
- [x] OpenAPI spec generation - #1692 
- [x] OpenAPI spec commit and added to CI check - #1703

PR #1692, opened 2026-02-05
Add generation of `JSON` specification of OpenAPI.

JBRS307:
Part of #1686 

---

Endpoints done:
- [x] Register endpoints
- [x] Unregister endpoints
- [x] `/status`
- [x] Update endpoints
- [x] WebSocket
- [x] Request keyframe (output)
- [x] Start end reset

Will not be done:
- Stats (this is for internal use, no need to document that) 

---

OpenAPI spec json will be commited in a followup

### DIFF
--- a/tools/src/bin/generate_json_schema_openapi.rs
+++ b/tools/src/bin/generate_json_schema_openapi.rs
@@ -117,7 +117,7 @@ enum SchemaAction {
     Nothing,
 }
 
-// OpenAPI generation
+// OpenAPI specification generation
 
 #[derive(OpenApi)]
 #[openapi(
@@ -153,7 +153,7 @@ enum SchemaAction {
 struct ApiDoc;
 
 fn generate_openapi() {
-    let schema_json = ApiDoc::openapi().to_pretty_json().unwrap();
+    let openapi_json = ApiDoc::openapi().to_pretty_json().unwrap();
     let gen_path = PathBuf::from(ROOT_DIR).join("schemas/openapi_specification.json");
-    fs::write(&gen_path, schema_json).unwrap();
+    fs::write(&gen_path, openapi_json).unwrap();
 }


tools/src/bin/generate_json_schema_openapi.rs
```diff
@@ -114,3 +116,44 @@ enum SchemaAction {
     CheckIfChanged,
     Nothing,
 }
+
+// OpenAPI generation
```
JBRS307:
```suggestion
// OpenAPI specification generation
```

tools/src/bin/generate_json_schema_openapi.rs
```diff
@@ -114,3 +116,44 @@ enum SchemaAction {
...
+
+#[derive(OpenApi)]
+#[openapi(
+    info(
+        title = "Smelter",
+        description = "Real-time video compositing software",
+        version = "0.5.0",
+        license(
+            name = "",
```
JBRS307, line 129: I have no idea what to put here, but I think it should not be empty.

tools/src/bin/generate_json_schema_openapi.rs
```diff
@@ -114,3 +116,44 @@ enum SchemaAction {
...
+        smelter::routes::update_output::handle_output_update,
+        smelter::routes::update_output::handle_keyframe_request,
+        smelter::routes::status::status_handler,
+        smelter::routes::ws::ws_handler,
+    )
+)]
+struct ApiDoc;
+
+fn generate_openapi() {
+    let schema_json = ApiDoc::openapi().to_pretty_json().unwrap();
```
JBRS307:
```suggestion
    let openapi_json = ApiDoc::openapi().to_pretty_json().unwrap();
```

tools/src/bin/generate_json_schema_openapi.rs
wkozyra95: use MIT, this is license for the API not for the code

tools/src/bin/generate_json_schema_openapi.rs
```diff
@@ -7,13 +7,15 @@ use schemars::{
...
 use smelter::routes;
+use utoipa::OpenApi;
 
 const ROOT_DIR: &str = env!("CARGO_MANIFEST_DIR");
 
 fn main() {
     tracing_subscriber::fmt().init();
     let check_flag = std::env::args().any(|arg| &arg == "--check");
     generate_json_schema(check_flag);
+    generate_openapi();
```
wkozyra95, line 18: If check_flag is set generate schema in memory and compare it with file. Throw an error if they don't match

tools/.gitignore
```diff
@@ -1,5 +1,6 @@
 # Schemas generated for TypeScript demos
 /schemas/api_types.schema.json
+/schemas/openapi_specification.json
```
wkozyra95, line 3: commit it

wkozyra95: also you can use `openapi.json` it is already in schemas directory

PR #1696, opened 2026-02-06
Refactor output resampler + enforce continuity of the output stream

wkozyra95:
- The output resampler was based on a generic implementation that was used for input and output at the same time.
  - In previous PR new input resampler was implemented, so there is no point in keeping a generic implementation
  - We don't need any stretching on output so fft resampler is more efficient
- Some of the behavior was inconsistent
  - before:
    - aac encoder assumed continuous input, but opus not
    - opus encoder needs to receive a specific chunk size, but nothing was enforcing that
  - now:
    - audio mixer ensures that its output is continuous, if some batch was dropped on queue, then it will generate zeros
    - The output resampler assumes a continuous stream on its input and guarantees that its output is also continuous
    - all audio encoders assume continuous inputs


wkozyra95:
It looks like there is a problem with YouTube, but the same issue happesn on master

smelter-core/src/audio_mixer/mixer.rs
```diff
@@ -147,20 +186,31 @@ impl InternalAudioMixer {
...
 
+        let mixed_samples = self.mix_samples(input_samples, samples_count, samples_set.start_pts);
+
+        self.last_processed_batch_end = Some(samples_set.end_pts);
+        if let Some(mut samples) = maybe_zero_samples {
+            samples.merge(mixed_samples);
+            samples
+        } else {
+            mixed_samples
+        }
```
JBRS307: No strong opinion, but `match` would look nicer here

smelter-core/src/codecs/aac.rs
```diff
@@ -14,6 +14,10 @@ impl AudioEncoderOptionsExt for FdkAacEncoderOptions {
     fn sample_rate(&self) -> u32 {
         self.sample_rate
     }
+
+    fn channel(&self) -> AudioChannels {
```
JBRS307:
```suggestion
    fn channels(&self) -> AudioChannels {
```

smelter-core/src/codecs/opus.rs
```diff
@@ -22,4 +22,8 @@ impl AudioEncoderOptionsExt for OpusEncoderOptions {
     fn sample_rate(&self) -> u32 {
         self.sample_rate
     }
+
+    fn channel(&self) -> AudioChannels {
```
JBRS307:
```suggestion
    fn channels(&self) -> AudioChannels {
```

smelter-core/src/pipeline/encoder/libopus.rs
```diff
@@ -48,34 +63,57 @@ impl AudioEncoder for OpusEncoder {
...
+        trace!("Flushing libopus encoder");
+        self.inner_encode(true)
+    }
+}
+
+impl OpusEncoder {
+    fn inner_encode(&mut self, force: bool) -> Vec<EncodedOutputChunk> {
+        let mut result = vec![];
+        while self.input_buffer.frames() >= 960 || (force && self.input_buffer.frames() > 0) {
+            let samples = self.input_buffer.read_samples(960);
```
JBRS307: It would be better if it was put in a named const, something like `OPUS_ENCODER_CHUNK_SIZE`

smelter-core/src/pipeline/encoder/libopus.rs
```diff
@@ -48,34 +63,57 @@ impl AudioEncoder for OpusEncoder {
...
+                    .flat_map(|(l, r)| {
+                        [(*l * i16::MAX as f64) as i16, (*r * i16::MAX as f64) as i16]
+                    })
+                    .collect(),
+            };
+
+            let data = match self.encoder.encode(&raw_samples, &mut self.output_buffer) {
+                Ok(len) => Bytes::copy_from_slice(&self.output_buffer[..len]),
+                Err(err) => {
+                    error!("Opus encoding error: {}", err);
```
JBRS307:
```suggestion
                    error!("Opus encoding error: {err}");
```

smelter-core/src/pipeline/utils/audio_buffer.rs
```diff
@@ -0,0 +1,137 @@
...
+                    samples.extend_from_slice(&batch[range])
+                }
+                _ => {
+                    error!("Wrong channel layout");
+                }
+            }
+        }
+
+        // fill with zero if channel layouts would mismatch
+        let range = 0..sample_count - samples.len();
```
JBRS307:
```suggestion
        let range = 0..(sample_count - samples.len());
```

It took me a while to read that, I think this is more readable, but no strong opinion

smelter-core/src/codecs.rs
```diff
@@ -70,6 +70,7 @@ pub enum OutputPixelFormat {
 
 pub(crate) trait AudioEncoderOptionsExt {
     fn sample_rate(&self) -> u32;
+    fn channel(&self) -> AudioChannels;
```
JBRS307:
```suggestion
    fn channels(&self) -> AudioChannels;
```

smelter-core/src/audio_mixer/mixer.rs
```diff
@@ -14,6 +14,27 @@ use crate::{
 
 use crate::prelude::*;
 
+/// Audio mixer responsible for generating output samples from input samples.
+///
+/// It meets following constraints:
+/// - Input samples can change theirs sample rate and channel layout dynamically.
+/// - Input samples set PTS, should not have any numerical error. end pts of one set is the same
```
wkazmierczak:
```suggestion
/// - Input samples set PTS, should not have any numerical error. end_pts of one set is the same
```

smelter-core/src/pipeline/encoder/libopus.rs
```diff
@@ -48,34 +65,59 @@ impl AudioEncoder for OpusEncoder {
...
                 dts: None,
                 is_keyframe: false,
                 kind: MediaKind::Audio(AudioCodec::Opus),
-            }],
-            Err(err) => {
-                error!("Opus encoding error: {}", err);
-                vec![]
-            }
+            });
+            self.encoded_samples += 960;
```
JBRS307: `SAMPLES_PER_BATCH` should be here too.

PR #1698, opened 2026-02-09
[ts-sdk] Update RTMP input api

wkazmierczak:
Update RTMP input api (field `url` changed to pair of fields `app`, `streamKey`)

### DIFF
--- a/ts/smelter/src/types/input.ts
+++ b/ts/smelter/src/types/input.ts
@@ -150,9 +150,13 @@ export type RegisterWhepClientInput = {
 export type RegisterRtmpServerInput = {
   type: 'rtmp_server';
   /**
-   * URL where the RTMP server will listen for incoming stream. Format: `rtmp://<ip_address>:<port>`
+   * The RTMP application name. This is the first path segment of the RTMP stream URL that Smelter listens on for incoming streams. Format: `rtmp://<ip_address>:<port>/<app>/<stream_key`
    */
-  url: string;
+  app: string;
+  /**
+   * The RTMP stream key. This is the second path segment of the RTMP stream URL that Smelter listens on for incoming streams. Format: `rtmp://<ip_address>:<port>/<app>/<stream_key`
+   */
+  streamKey: string;
   /**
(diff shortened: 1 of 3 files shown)


ts/smelter/src/types/input.ts
```diff
@@ -150,9 +150,13 @@ export type RegisterWhepClientInput = {
 export type RegisterRtmpServerInput = {
   type: 'rtmp_server';
   /**
-   * URL where the RTMP server will listen for incoming stream. Format: `rtmp://<ip_address>:<port>`
+   * The RTMP application name. This is the first path segment of the RTMP stream URL that Smelter listens on for incoming streams. Format: `rtmp://<ip_address>:<port>/<app>/<stream_key`
    */
```
JBRS307, line 154:
```suggestion
   * The RTMP application name. This is the first segment of the URL path on which the Smelter RTMP server listens for incoming streams. Format: `rtmp://<ip_address>:<port>/<app>/<stream_key>`
  */
```

ts/smelter/src/types/input.ts
```diff
@@ -150,9 +150,13 @@ export type RegisterWhepClientInput = {
...
-  url: string;
+  app: string;
+  /**
+   * The RTMP stream key. This is the second path segment of the RTMP stream URL that Smelter listens on for incoming streams. Format: `rtmp://<ip_address>:<port>/<app>/<stream_key`
+   */
```
JBRS307, line 158:
```suggestion
   * The RTMP stream key. This is the second segment of the URL path on which the Smelter RTMP server listens for incoming streams. Format: `rtmp://<ip_address>:<port>/<app>/<stream_key>`
   */
```

ts/smelter/src/types/input.ts
JBRS307: No strong opinion, but i think this sounds better.

wkazmierczak: I prefer my version, so I'll leave as it is

PR #1701, opened 2026-02-09
Prevent spam while waiting for audio/video config on RTMP input

wkazmierczak:


smelter-core/src/pipeline/rtmp/rtmp_input/connection.rs
```diff
@@ -42,8 +48,8 @@ struct RtmpConnectionState {
     video_decoders: RtmpServerInputVideoDecoders,
     buffer: InputBuffer,
 
-    video_handle: Option<DecoderThreadHandle>,
-    audio_handle: Option<DecoderThreadHandle>,
+    video_state: ConfigState,
```
wkazmierczak: I do not like this naming, but for now I do not have a better idea. `struct` and `field` name

smelter-core/src/pipeline/rtmp/rtmp_input/connection.rs
```diff
@@ -34,6 +34,12 @@ pub(crate) struct RtmpConnectionOptions {
     pub buffer: InputBuffer,
 }
 
+enum ConfigState {
+    WaitingForConfig,
+    WaitingLogged,
+    Received(DecoderThreadHandle),
+}
```
wkozyra95:
```suggestion
enum TrackState {
    BeforeFirstEvent,
    ConfigMissing,
    Ready(DecoderThreadHandle),
}
```

smelter-core/src/pipeline/rtmp/rtmp_input/connection.rs
```diff
@@ -132,7 +138,7 @@ impl RtmpConnectionState {
 
         match handle {
             Ok(handle) => {
-                self.video_handle = Some(handle);
+                self.video_state = ConfigState::Received(handle);
```
wkozyra95: init should return handle

PR #1704, opened 2026-02-09
Fixed inconsistent types in api.

### LINKED ISSUE #1697: Make types of api fields consistent.
Sometimes paths is as `Arc<Path>` sometimes as `String`. It should be consistent.

JBRS307:
Closes #1697 

- Merge after #1692 

---

Changes types used in api to `Arc<str>` or `Arc<Path>` wherever `String` or `PathBuf` are used. Main goal is to keep api types consistent, as `Arc<>` types are used in some places, and structs in other.

### DIFF
--- a/integration-tests/src/bin/benchmark/scenes.rs
+++ b/integration-tests/src/bin/benchmark/scenes.rs
@@ -22,10 +22,10 @@ pub type SceneBuilderFn = fn(ctx: &SceneContext, output_id: &OutputId) -> Compon
 
 pub fn example_image() -> (RendererId, RendererSpec) {
     (
-        RendererId(Arc::from("example_image")),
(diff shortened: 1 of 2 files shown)


integration-tests/src/bin/benchmark/scenes.rs
```diff
@@ -25,7 +25,7 @@ pub fn example_image() -> (RendererId, RendererSpec) {
         RendererId(Arc::from("example_image")),
         RendererSpec::Image(ImageSpec {
             src: ImageSource::LocalPath {
-                path: example_image_path().to_string_lossy().to_string(),
+                path: Arc::from(example_image_path().as_path()),
```
wkazmierczak: Maybe `into()` will be better in those cases

PR #1711, opened 2026-02-11
[rtmp] Integrate `amf3` decoder with the `RTMP` server

### LINKED ISSUE #1690: [rtmp] Integrate `amf3` decoder with the server.

JBRS307:
Closes #1690 

- Merge after #1720 

---

After this is merged the server will still not respond that it handles `amf3`. That will be the case only after encoder is implemented.

---

Additionally, decoding of the sign extended `U29` was done properly. Last time I misunderstood the spec.


rtmp/src/amf0/encoding.rs
```diff
@@ -31,6 +31,7 @@ fn encode_value(buf: &mut BytesMut, value: &Amf0Value) -> Result<(), AmfEncoding
             class_name: _class_name,
             properties: _properties,
         } => unimplemented!(),
+        Amf0Value::AvmPlus(_amf3_value) => unimplemented!(),
```
JBRS307, line 34: Will be done in the next PR

rtmp/src/amf3/decoding.rs
JBRS307, line 1: The decoder now holds any type that implements `Buf` trait from the `bytes` crate. It was done this way so it can either own the `Bytes` object or take reference `&mut Bytes`

rtmp/src/flv/tag/scriptdata/extended.rs
JBRS307, line 1: This one clearly corresponds to `amf3`, however I really do not want to make any `amf` related code public

rtmp/src/amf0/decoding.rs
```diff
@@ -2,31 +2,31 @@ use bytes::{Buf, Bytes};
...
+struct Amf0DecoderState<T> {
+    buf: T,
     // According to spec (https://rtmp.veriskope.com/pdf/amf0-file-format-specification.pdf),
     // complex types are Object, ECMA Array, Strict Array and Typed Objext.
     complexes: Vec<Amf0Value>,
 }
 
-impl AmfDecoderState {
-    fn new(amf_bytes: Bytes) -> Self {
+impl<T> Amf0DecoderState<T>
```
wkozyra95: why use generic if you know specific type?

JBRS307: ~Here it is done just for consistency with how AMF3 decoder works.~

EDIT: Reverted these changes

rtmp/src/flv/tag/scriptdata/extended.rs
wkozyra95: to much nesting move that level up

Tag directory also needs to be flattern, but this is something I'll do in my PR

PR #1723, opened 2026-02-17
Use RTMP client from `rtmp` crate

wkozyra95:


smelter-api/src/output/mp4_into.rs
```diff
@@ -97,6 +97,7 @@ impl Mp4VideoEncoderOptions {
                     .unwrap_or_default()
                     .into_iter()
                     .collect(),
+                output_format: core::H264BitstreamFormat::AnnexB,
```
JBRS307: Why does mp4 output have annexb format?

smelter-api/src/output/hls_into.rs
```diff
@@ -97,6 +97,7 @@ impl HlsVideoEncoderOptions {
                     .unwrap_or_default()
                     .into_iter()
                     .collect(),
+                output_format: core::H264BitstreamFormat::AnnexB,
```
JBRS307: Same as with mp4, why annexb here?

smelter-api/src/output/mp4_into.rs
wkazmierczak: yeah I think it should be avcc

PR #1726, opened 2026-02-17
[rtmp] Parsing sound rate and channel configuration from `AAC` audio specific config.

JBRS307:


rtmp/src/flv/audio.rs
```diff
@@ -44,6 +44,8 @@ pub enum AudioCodec {
     DeviceSpecific,
 }
 
+pub struct AudioSpecificConfig {}
+
```
JBRS307: Delete

rtmp/src/message/event.rs
```diff
@@ -139,8 +135,10 @@ pub(super) fn event_into_raw(
                 codec: AudioCodec::Aac,
                 sample_rate: AudioTagSoundRate::Rate44000,
                 sample_size: AudioTagSampleSize::Sample16Bit,
-                channels: config.channels,
-                data: config.data,
+                channels: config
+                    .channels()
+                    .map_err(|_| SerializationError::AscParsingError)?,
+                data: config.data(),
```
JBRS307: Handle it with stereo as default.

rtmp/src/events.rs
```diff
@@ -30,9 +32,121 @@ pub struct AacAudioData {
 
 #[derive(Clone)]
 pub struct AacAudioConfig {
-    pub channels: AudioChannels,
-    pub sample_rate: u32,
-    pub data: Bytes, // TODO: Audio specific config
+    data: Bytes, // TODO: Audio specific config
+}
+
+impl AacAudioConfig {
```
JBRS307: Add checks if there is enough data for cases that require more than 2 bytes.

rtmp/src/events.rs
```diff
@@ -30,9 +32,121 @@ pub struct AacAudioData {
...
+
+        let object_type = (self.data[0] >> 3) & 0x1F;
+        let frequency_index = match object_type {
+            31 => (self.data[1] >> 1) & 0xF,
+            _ => {
+                let high = self.data[0] & 0x7;
+                let low = (self.data[1] >> 7) & 0x1;
+                (high << 1) | low
+            }
+        };
```
JBRS307: Add comments with bit length of fields

rtmp/src/events.rs
```diff
@@ -172,3 +288,90 @@ fn bytes_debug(data: &[u8]) -> String {
...
+    use bytes::Bytes;
+
+    use crate::{AacAudioConfig, AudioChannels};
+
+    #[test]
+    fn test_sound_frequency() {
+        // Encoded with sample rate 44100 Hz.
+        let asc_bytes = Bytes::from_iter([0b00010_010, 0b0_0000000]);
+        let asc = AacAudioConfig::new(asc_bytes);
+        assert_eq!(asc.sample_rate().unwrap(), 44_100);
```
JBRS307: This test case does not add any value.

rtmp/src/events.rs
JBRS307: For `pub` functions too.

smelter-core/src/pipeline/rtmp/rtmp_output.rs
```diff
@@ -39,7 +39,7 @@ pub struct RtmpClientOutput {
 struct AudioConfig {
     extradata: Bytes,
     channels: AudioChannels,
-    sample_rate: u32,
+    _sample_rate: u32,
```
JBRS307: This is now unread, I left it because I don't know if it is going to be needed or not

wkazmierczak (approved):

rtmp/src/events.rs
```diff
@@ -172,3 +311,85 @@ fn bytes_debug(data: &[u8]) -> String {
...
+        assert_eq!(asc.sample_rate().unwrap(), 48_000);
+
+        // Encoded with sample rate 48000 Hz. object_type == 31
+        let asc_bytes = Bytes::from_iter([0b11111_000, 0b000_0011_0]);
+        let asc = AacAudioConfig::new(asc_bytes);
+        assert_eq!(asc.sample_rate().unwrap(), 48_000);
+
+        // Encoded with custom sample rate (2137 Hz).
+        let asc_bytes = Bytes::from_iter([
+            0b00010_111,
```
wkozyra95: If you want to make it more readable, I could see adding `_` every 4 bytes, but not be a separator between fields.

If you want to make it clear you can do something like this if the fields do not go between bits.

```suggestion
            // 5 bits - fileda, 4 bits - field b
            0b0001_0111,
```

rtmp/src/events.rs
```diff
@@ -30,9 +32,144 @@ pub struct AacAudioData {
...
+        let (object_type, frequency_index) = self.object_type_frequency_index()?;
+
+        // 4 bit channel_configuration field
+        let channel_configuration = match (object_type, frequency_index) {
+            (31, 15) => {
+                if self.data.remaining() < 6 {
+                    return Err(ParseError::NotEnoughData);
+                }
+                let high = self.data[4] & 0x1;
+                let low = (self.data[5] >> 5) & 0x7;
```
wkozyra95: I think it's clearer if you do
```suggestion
                let low = (self.data[5] & 0b11100000 ) >> 5 ;
```
from the syntax you see that you are taking top 3 bits and shifting them to get the number

With your version you have to:
- first imagine what 0x7 looks in bits
- you need to make toperations in your head to identify what it does

smelter-core/src/pipeline/rtmp/rtmp_output.rs
wkozyra95: remove it

rtmp/src/events.rs
```diff
@@ -30,9 +32,144 @@ pub struct AacAudioData {
 
 #[derive(Clone)]
 pub struct AacAudioConfig {
```
wkozyra95: move to a separate file e.g. aac.rs

rtmp/src/message/event.rs
```diff
@@ -139,8 +137,11 @@ pub(super) fn event_into_raw(
                 codec: AudioCodec::Aac,
                 sample_rate: AudioTagSoundRate::Rate44000,
                 sample_size: AudioTagSampleSize::Sample16Bit,
-                channels: config.channels,
-                data: config.data,
+                channels: config.channels().map_err(|error| {
+                    error!(%error, "Audio specific config parse error.");
```
wkozyra95: Don't log it, return it

wkozyra95 (approved):
I'm fine with merging this but the API needs to be different.

If you have some idea, you can try to improve it, if not, I'll address it in follow up. I don't have specific structure in mind, but I don't like the fact  that those methods on config return errors.

PR #1727, opened 2026-02-17
Add options to expose WebRTC endpoints from container

wkozyra95:
- limit ports to a specific range
- provide the host ip that should be used in SDP

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -361,7 +361,7 @@ fn framerate_from_str(s: &str) -> Result<Framerate, &'static str> {
 
 fn port_range_from_str(s: &str) -> Result<(u16, u16), &'static str> {
     const ERROR_MESSAGE: &str =
-        "Port range needs to be in the \"START:END\" format, where START < END are both < 65535.";
+        "Port range needs to be in the \"START:END\" format, where START < END.";
     let Some((num_str, den_str)) = s.split_once(':') else {
         return Err(ERROR_MESSAGE);
     };


.github/workflows/rust_check.yml
```diff
@@ -151,4 +151,4 @@ jobs:
 
             - name: 📄 Generate JSON schema
               run: |
-                  cargo run -p tools --bin generate_json_schema_openapi -- --check
+                  cargo run -p tools --bin generate_from_types -- --check
```
wkozyra95, line 154: unrelated to the PR, recent rename was inconsistent, so some scripts were broken

smelter-core/src/pipeline.rs
```diff
@@ -68,7 +68,9 @@ pub struct PipelineOptions {
     pub chromium_context: Option<Arc<ChromiumContext>>,
 
     pub whip_whep_server: PipelineWhipWhepServerOptions,
```
wkazmierczak, line 70: Wondering if rename this as well for consistency, but no strong opinion since there is some distinction

src/config.rs
```diff
@@ -339,6 +359,17 @@ fn framerate_from_str(s: &str) -> Result<Framerate, &'static str> {
     }
 }
 
+fn port_range_from_str(s: &str) -> Result<(u16, u16), &'static str> {
+    const ERROR_MESSAGE: &str =
+        "Port range needs to be in the \"START:END\" format, where START < END are both < 65535.";
```
wkazmierczak:
```suggestion
        "Port range needs to be in the \"START:END\" format, where START < END are both <= 65535.";
```
Nitpick

smelter-core/src/pipeline.rs
wkozyra95: I was considering that, but
- I want to avoid renaming envs
- The other options are for webrtc, this is specific for WHIP/WHEP
- I did not want to spend time on that right know. I'm open to getting back to it in the future

src/config.rs
wkozyra95:
```suggestion
        "Port range needs to be in the \"START:END\" format, where START < END.";
```

PR #1732, opened 2026-02-18
[rtmp] Different api for `AacAudioConfig`

JBRS307:
Suggestion of api that doesn't return `Result` on getters. If approved I will merge it before merging the base to master.

### DIFF
--- a/rtmp/src/events/aac.rs
+++ b/rtmp/src/events/aac.rs
@@ -65,28 +65,12 @@ impl AacAudioConfig {
                 }
                 match object_type {
                     31 => {
-                        let first_chunk = data[1] & 0b0000_0001;
-                        let second_chunk = data[2];
-                        let third_chunk = data[3];
-                        let fourth_chunk = (data[4] & 0b1111_1110) >> 1;
-
-                        let first_byte = (first_chunk << 7) | (second_chunk >> 1);
-                        let second_byte = (second_chunk << 7) | (third_chunk >> 1);
-                        let third_byte = (third_chunk << 7) | fourth_chunk;
-
-                        u32::from_be_bytes([0, first_byte, second_byte, third_byte])
+                        let raw = u32::from_be_bytes([data[1], data[2], data[3], data[4]]);
+                        (raw & 0x01_FF_FF_FE) >> 1
                     }
                     _ => {
-                        let first_chunk = data[1] & 0b0111_1111;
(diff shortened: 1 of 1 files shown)


smelter-core/src/pipeline/rtmp/rtmp_output.rs
```diff
@@ -126,7 +126,8 @@ impl RtmpClientOutput {
         }
 
         if let Some(config) = audio_config {
-            let config = AacAudioConfig::new(config.extradata.clone());
+            let config =
+                AacAudioConfig::try_from(config.extradata.clone()).map_err(RtmpError::from)?;
```
wkozyra95, line 130:
```suggestion
            let config = AacAudioConfig::try_from(config.extradata.clone())?;
```
? is automatic `map_err(Into::into)`

rtmp/src/events/aac.rs
```diff
@@ -36,15 +60,15 @@ impl AacAudioConfig {
...
+                if data.remaining() < 5 {
                     return Err(ParseError::NotEnoughData);
                 }
                 match object_type {
                     31 => {
-                        let first_chunk = self.data[1] & 0b0000_0001;
-                        let second_chunk = self.data[2];
-                        let third_chunk = self.data[3];
-                        let fourth_chunk = (self.data[4] & 0b1111_1110) >> 1;
+                        let first_chunk = data[1] & 0b0000_0001;
```
wkozyra95:
```
                        let value = u32::from_be_bytes([data[1], data[2], data[3], data[4]])
                        (value & 0x00FFFFFF) >> 1
```
It goes against the approach i suggested for single byte operations, but I think it's a lot cleaner and conveys operation. It wouldn't work if it was not whole 3 bytes

wkozyra95 (approved):

smelter-core/src/pipeline/rtmp/rtmp_output.rs
JBRS307: This does not compile. I because there are 2 layers of nested errors it just can't convert it.

JBRS307: There is enum variant in `RtmpError` called `ParsingError`. It is currently unused, I will refactor every function that throws `ParseError` to throw `RtmpError` with this variant, but in followup PR.

PR #1734, opened 2026-02-19
Add option to mux WebRTC connections on single port

wkozyra95:
I'm waiting with merging for now. If it will work well for FishJam I'll merge it. 
Another option to consider is upgrading to sansio based implementation on webrtcrs master might have fix that issue.

### DIFF
--- a/smelter-core/src/pipeline/webrtc/mod.rs
+++ b/smelter-core/src/pipeline/webrtc/mod.rs
@@ -1,23 +1,7 @@
-use std::{
-    io,
-    net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4},
-    sync::{Arc, Weak},
-    time::Duration,
-};
+use std::sync::Arc;
 
-use async_trait::async_trait;
-use tokio::{net::UdpSocket, runtime::Runtime, sync::oneshot};
-use tracing::{error, info, warn};
-use webrtc::{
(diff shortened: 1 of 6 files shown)


src/state.rs
```diff
@@ -131,7 +131,12 @@ pub fn pipeline_options_from_config(
             },
             false => PipelineWhipWhepServerOptions::Disable,
         },
-        webrtc_port_range: opt.webrtc_udp_port_range,
+        webrtc_port_strategy: opt.webrtc_udp_port_strategy.clone().map(|s| match s {
```
wkazmierczak: maybe consistent name

smelter-core/src/pipeline/webrtc/mod.rs
```diff
@@ -89,19 +102,184 @@ impl<T> Iterator for AsyncReceiverIter<T> {
     }
 }
 
-fn default_setting_engine(ctx: &Arc<PipelineCtx>) -> SettingEngine {
-    let mut setting_engine = SettingEngine::default();
-    if !ctx.webrtc_nat_1to1_ips.is_empty() {
+#[derive(Clone)]
+pub(crate) enum WebrtcSettingEngineCtx {
```
wkazmierczak: I think moving `WebrtcSettingEngineCtx` to separate file will be better

PR #1736, opened 2026-02-19
[vk-video] Add conversion helpers

### LINKED ISSUE #1731: [vk-video] Add helpers for `NV12` <-> `RGBA` conversion

noituri:
Closes #1731

### DIFF
--- a/vk-video/src/lib.rs
+++ b/vk-video/src/lib.rs
@@ -15,9 +15,9 @@ mod vulkan_transcoder;
 #[cfg(vulkan)]
 pub(crate) mod wrappers;
 
-#[cfg(vulkan)]
+#[cfg(all(vulkan, feature = "wgpu"))]
 mod wgpu_helpers;
-#[cfg(vulkan)]
+#[cfg(all(vulkan, feature = "wgpu"))]
 pub use wgpu_helpers::*;
 
 #[cfg(vulkan)]


vk-video/examples/encode_wgpu.rs
noituri, line 1: This is `texture_as_input` example from before `NV12` encoding was added + rgba to nv12 converter helper

vk-video/src/shaders/nv12_to_rgba.wgsl
```diff
@@ -0,0 +1,47 @@
...
+@group(1) @binding(0) var sampler_: sampler;
+
+@fragment
+fn fs_main(input: VertexOutput) -> @location(0) vec4<f32> {
+    let conversion_bias = vec3<f32>(0.0, 0.5, 0.5);
+    let conversion_weights = mat3x3<f32>(
+        1.0, 0.0, 1.5748,
+        1.0, -0.1873, -0.4681,
+        1.0, 1.8556, 0.0
+    );
```
jerzywilczek, line 39: Maybe we should at least add an issue to support different kinds of YUV in the converters? I mean YUVJ etc.

vk-video/src/wgpu_helpers.rs
```diff
@@ -0,0 +1,112 @@
...
+pub use rgba_to_nv12::*;
+
+/// Represents mapping between `InputTexture` and `OutputTexture` used by the converters.
+/// `InputTexture` is converted into `OutputTexture`.
+///
+/// Can be created with [`WgpuNv12ToRgbaConverter`] or [`WgpuRgbaToNv12Converter`].
+#[derive(Clone)]
+pub struct WgpuTextureMapping<InputTexture, OutputTexture> {
+    pub(crate) input_bg: wgpu::BindGroup,
+    _input_texture: PhantomData<InputTexture>,
```
jerzywilczek: why is this nencessary? just to have the first generic parameter?

vk-video/src/wgpu_helpers/nv12_to_rgba.rs
```diff
@@ -0,0 +1,195 @@
...
+        });
+        let sampler_bgl = device.create_bind_group_layout(&wgpu::BindGroupLayoutDescriptor {
+            label: None,
+            entries: &[wgpu::BindGroupLayoutEntry {
+                binding: 0,
+                visibility: wgpu::ShaderStages::FRAGMENT,
+                ty: wgpu::BindingType::Sampler(wgpu::SamplerBindingType::Filtering),
+                count: None,
+            }],
+        });
```
jerzywilczek: if our `super` is called `wgpu_helpers`, maybe parts of code like this should be moved to some common place next to this module and it's sibling.

vk-video/src/wgpu_helpers.rs
```diff
@@ -0,0 +1,112 @@
...
+
+pub use nv12_to_rgba::*;
+pub use rgba_to_nv12::*;
+
+/// Represents mapping between `InputTexture` and `OutputTexture` used by the converters.
+/// `InputTexture` is converted into `OutputTexture`.
+///
+/// Can be created with [`WgpuNv12ToRgbaConverter`] or [`WgpuRgbaToNv12Converter`].
+#[derive(Clone)]
+pub struct WgpuTextureMapping<InputTexture, OutputTexture> {
```
jerzywilczek: I don't really get why this is necessary.

I understand that this is for making sure that the texture we get as output works for us, but since there is no vulkan in this code and everything is based on wgpu, shouldn't we just do things the wgpu way? i.e. allow the user to give us any texture, and then wgpu will crash/report the mistake through an error scope? I think that would make the API more fitting for a wgpu middleware. Also, we have to do stuff like this when we interact with vulkan, because the safety of the program depends on it. Wgpu should validate everything for us and not allow unsafety.

noituri: I did it this way because I prefer when things are caught by the compiler during compilation rather than runtime. But I do agree that it's not very wgpu-like and we don't  even have this type of api for the encoder, so I'll change it

vk-video/src/wgpu_helpers.rs
noituri: changed the implementation, this thing is gone

PR #1751, opened 2026-02-25
Interleave writes to RTMP output

wkozyra95:


smelter-core/src/pipeline/rtmp/rtmp_output.rs
```diff
@@ -234,62 +226,124 @@ impl Output for RtmpClientOutput {
...
+                        channels,
+                    ))?;
                 }
-            },
-        };
-
-        if received_video_eos.unwrap_or(true) && received_audio_eos.unwrap_or(true) {
-            break;
+                (None, None) => break,
+            }
```
JBRS307: Suggestion: remove `if { ... } else { ... }` and move the body of `else` to the `(false, false)` section. Same logic, less nesting.

JBRS307 (approved):

PR #1754, opened 2026-02-26
[rtmp] Rework errors

wkozyra95:


rtmp/src/error.rs
```diff
@@ -109,15 +122,21 @@ pub enum AudioTagParseError {
...
-pub enum AudioSpecificConfigParseError {
+pub enum AacConfigParseError {
     #[error("Invalid frequency index: {0}")]
     InvalidFrequencyIndex(u8),
 
     #[error("Invalid audio channel value in AAC audio specific config: {0}")]
     InvalidAudioChannel(u8),
+
+    #[error("Not enough data, config too short")]
+    ToShort,
```
JBRS307:
```suggestion
    TooShort,
```

rtmp/src/error.rs
```diff
@@ -1,88 +1,98 @@
...
 #[derive(Error, Debug, Clone, PartialEq)]
-pub enum SerializationError {
+pub enum RtmpMessageSerializeError {
     #[error("Error encoding amf0: {0}")]
     Amf0Encoding(#[from] AmfEncodingError),
 
-    #[error("Unsupported video codec: {0:?}")]
-    UnsupportedVideoCodec(VideoCodec),
+    #[error("Failed to serialize message: {0}")]
+    InternalError(String),
```
JBRS307, line 64: In the code it looks much worse than it sounded in discussions. Currently it is used only 4 times and can be easily typed either in there or by creating separate type for `InternalError`. And we really should not be making our life harder by having to parse error strings.

wkozyra95: This is an error that "should not happen" it means there is a bug in the implementation on our or the other side, so this is not something code can handle.

PR #1756, opened 2026-02-26
[vk-video] Make `wgpu` optional

noituri:
I had to change the adapter/device api a little bit

### DIFF
--- a/vk-video/CHANGELOG.md
+++ b/vk-video/CHANGELOG.md
@@ -6,9 +6,9 @@
 - `EncoderParameters` has an extra field, which determines whether stream parameters are inlined in the output stream.
 - Changed adapter and device creation API
 
-
 ### ✨ New features
 - One-to-many transcoders via `VulkanDevice::create_transcoder`
+- Made `wgpu` dependency optional via `wgpu` feature (enabled by default)
 
 ### 🐛 Bug fixes
 
--- a/vk-video/Cargo.toml
+++ b/vk-video/Cargo.toml
@@ -34,7 +34,7 @@ memchr = "2.7.4"
 rustc-hash = "2.1.1"
 thiserror = "1.0.59"
 tracing = "0.1.40"
-naga = "28.0.0"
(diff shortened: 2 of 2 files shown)


vk-video/src/adapter.rs
```diff
@@ -270,32 +252,74 @@ impl<'a> VulkanAdapter<'a> {
...
     }
 }
 
+// TODO: maybe there should be a way of specifying power preference / device preference (like wgpu)
+/// Describes a [`VulkanAdapter`].
+/// Used by [`VulkanInstance::create_adapter`]
+#[cfg(feature = "wgpu")]
+pub struct VulkanAdapterDescriptor<'a> {
+    pub supports_decoding: bool,
+    pub supports_encoding: bool,
```
noituri, line 287: This will be changed to something that's more specific

vk-video/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### 💥 Breaking changes
 - `EncoderParameters` has an extra field, which determines whether stream parameters are inlined in the output stream.
+- Changed adapter and device creation API
+
```
jerzywilczek, line 8: I think you should also state that wgpu is optional now, maybe in new features?

.github/workflows/rust_check.yml
```diff
@@ -149,6 +149,9 @@ jobs:
             - name: 📎 Run clippy (no-default-features)
               run: cargo clippy --workspace --all-targets --no-default-features -- -D warnings
 
+            - name: 📎 Run clippy (vk-video, no-default-features)
+              run: cargo clippy --package vk-video --all-targets --no-default-features -- -D clippy::todo -D clippy::uninlined_format_args -D warnings
```
wkozyra95, line 153:
```suggestion
              run: cargo clippy --package vk-video --all-targets --no-default-features -- -D warnings
```
It should work automatically because there is 
```
[lints]
workspace = true
```
in Cargo.toml

PR #1758, opened 2026-02-27
[rtmp] Add support for UserControlMessages

wkozyra95:


rtmp/src/message/user_control.rs
```diff
@@ -0,0 +1,156 @@
...
+            }
+        };
+        result.freeze()
+    }
+}
+
+// https://rtmp.veriskope.com/docs/spec/#717user-control-message-events
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[allow(unused)]
+enum UserControlMessageKind {
```
wkazmierczak, line 121:
```suggestion
enum UserControlEventType {
```

rtmp/src/message/user_control.rs
```diff
@@ -0,0 +1,156 @@
+use std::time::Duration;
+
+use bytes::{BufMut, Bytes, BytesMut};
+
+use crate::RtmpMessageParseError;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[allow(unused)]
+pub(crate) enum UserControlMessage {
```
wkazmierczak, line 9:
```suggestion
pub(crate) enum UserControlEvent {
```

rtmp/src/message/user_control.rs
wkazmierczak: And not sure if this enum is necessary. Maybe just method on UserControlEvent?

wkozyra95: I just moved it, normally I would write 2 methods to parse, but if we have separate type and it's fully private I think it's fine

PR #1762, opened 2026-02-27
[rtmp] Improve negotiation lifecycle + extract command messages

wkozyra95:


rtmp/src/transport/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use std::{
+    io::{self, Read, Write},
+    net::TcpStream,
+    time::Duration,
+};
+
+use crate::RtmpConnectionError;
+use tls::TlsClientStream;
+
```
wkazmierczak:
```suggestion
use std::{
    io::{self, Read, Write},
    net::TcpStream,
    time::Duration,
};

use crate::RtmpConnectionError;
use tls::TlsClientStream;

mod tls;
```

rtmp/src/message/mod.rs
```diff
@@ -1,43 +1,66 @@
...
+
     UserControl(UserControlMessage),
+    CommandMessage {
+        msg: CommandMessage,
+        stream_id: u32,
+    },
 
     // Explanation why it is a sequence of amf0 values and not amf3 values:
     // https://zenomt.github.io/rtmp-errata-addenda/rtmp-errata-addenda.html#name-object-encoding-3-2
     CommandMessageAmf3 {
```
wkazmierczak, line 60: So this will also be deleted in the current approach?

rtmp/src/server/connection.rs
```diff
@@ -1,68 +1,245 @@
...
+            ]
+            .into_iter()
+            .map(|(k, v)| (k.into(), v)),
+        );
+        let info = HashMap::from_iter(
+            [
+                ("level", "status".into()),
+                ("code", "NetConnection.Connect.Success".into()),
+                ("description", "Connection succeeded".into()),
+                ("objectEncoding", Amf0Value::Number(0 as f64)), // AMF0 encoding
```
JBRS307, line 181:
```suggestion
                ("objectEncoding", Amf0Value::Number(0.0)), // AMF0 encoding
```

rtmp/src/message/command.rs
```diff
@@ -0,0 +1,842 @@
...
+/// a command object (Object or Null), followed by optional arguments.
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) enum CommandMessage {
+    // ---------------------------------------------------------------
+    // 7.2.1 NetConnection Commands
+    // ---------------------------------------------------------------
+    /// 7.2.1.1 Client -> Server
+    Connect {
+        transaction_id: u32,
+        /// key-value pairs e.g. app, audioCodec, videoCodecs, objectEncoding
```
wkazmierczak:
```suggestion
        /// key-value pairs e.g. app, audioCodecs, videoCodecs, objectEncoding
```

rtmp/src/message/command.rs
```diff
@@ -0,0 +1,842 @@
...
+    ///
+    /// transaction_id = 0
+    /// command_object = null
+    OnStatus(Amf0Value),
+
+    /// _result or _error response
+    Result(Result<CommandMessageOk, CommandMessageError>),
+
+    /// Close command (NetConnection)
+    Close {
```
wkazmierczak, line 101: `close` is also one of the NetConnection Commands, so i think it would be better to move it under 7.2.1

rtmp/src/message/mod.rs
wkozyra95: maybe, not sure how amf3 support will look

rtmp/src/server/connection.rs
wkozyra95: It's an integer that needs to be sent as a float so I prefer cast int

PR #1766, opened 2026-03-03
Fixed docker hanging when run with `--cpus 1`

### LINKED ISSUE #1763: Investigate hangup when starting in docker container with less than 2 cpu cores

JBRS307:
Closes #1763

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -2,7 +2,7 @@ use crossbeam_channel::Receiver;
 use signal_hook::{consts, iterator::Signals};
 use smelter_render::error::ErrorStack;
 use tokio::runtime::Builder;
-use tracing::{debug, error, info, trace};
+use tracing::{debug, error, info};
 
 use std::{env, net::SocketAddr, process, sync::Arc, thread};
 use tokio::runtime::Runtime;
@@ -72,24 +72,22 @@ pub fn run_api(
 fn init_runtime() -> Runtime {
     const MINIMUM_WORKER_THREADS: usize = 3;
 
-    let available_threads = thread::available_parallelism();
-    trace!(?available_threads, "Available threads detected.");
-    let available_threads = available_threads.map_or(MINIMUM_WORKER_THREADS, |v| v.get());
-
-    let thread_count = env::var("TOKIO_WORKER_THREADS").map_or(available_threads, |v| {
-        let threads = v.trim().parse();
-        match threads {
+    let worker_threads = if let Ok(thread_count) = env::var("TOKIO_WORKER_THREADS") {
+        match thread_count.trim().parse() {
             Ok(t) if t > 0 => t,
-            Ok(_) | Err(_) => panic!("TOKIO_WORKER_THREADS must be a number greater than 0."),
+            _ => panic!("TOKIO_WORKER_THREADS must be a number greater than 0."),
         }
-    });
-    debug!(
-        worker_threads = thread_count,
-        "Number of runtime worker threads."
-    );
+    } else if let Ok(thread_count) = thread::available_parallelism() {
+        thread_count.get()
+    } else {
+        MINIMUM_WORKER_THREADS
+    };
+    let worker_threads = usize::max(worker_threads, MINIMUM_WORKER_THREADS);
+
+    debug!(worker_threads, "Initializing Tokio runtime");
     Builder::new_multi_thread()
         .enable_all()
-        .worker_threads(thread_count)
+        .worker_threads(worker_threads)
         .build()
         .unwrap()
 }


src/server.rs
```diff
@@ -15,7 +15,33 @@ pub fn run() {
     init_logger(config.logger.clone());
 
     info!("Starting Smelter with config:\n{:#?}", config);
-    let runtime = Arc::new(Runtime::new().unwrap());
+    let runtime = Arc::new({
```
wkozyra95: logic is fine, but refactor it, at least move that to function e.g. init_runtime

src/server.rs
```diff
@@ -15,7 +15,33 @@ pub fn run() {
...
+        const MINIMUM_WORKER_THREADS: usize = 3;
+        let available_threads = thread::available_parallelism()
+            .map(|val| val.get())
+            .unwrap_or(MINIMUM_WORKER_THREADS);
+        if available_threads >= MINIMUM_WORKER_THREADS {
+            debug!(
+                api_worker_threads = available_threads,
+                "Number of worker threads used for API."
```
wkozyra95: This is not only for API, Webrtc is using that too

src/server.rs
```diff
@@ -15,7 +15,33 @@ pub fn run() {
...
+            );
+            Runtime::new().unwrap()
+        } else {
+            let threads = env::var("TOKIO_WORKER_THREADS")
```
wkozyra95:
```suggestion
            let thread_count = env::var("TOKIO_WORKER_THREADS")
```

src/server.rs
```diff
@@ -15,7 +15,33 @@ pub fn run() {
...
+                .ok()
+                .and_then(|v| v.trim().parse().ok())
+                .unwrap_or(MINIMUM_WORKER_THREADS);
+            debug!(
+                api_worker_threads = threads,
+                "Number of worker threads used for API."
```
wkozyra95: the same as above

src/server.rs
```diff
@@ -15,7 +15,33 @@ pub fn run() {
...
+        const MINIMUM_WORKER_THREADS: usize = 3;
+        let available_threads = thread::available_parallelism()
+            .map(|val| val.get())
+            .unwrap_or(MINIMUM_WORKER_THREADS);
+        if available_threads >= MINIMUM_WORKER_THREADS {
+            debug!(
+                api_worker_threads = available_threads,
+                "Number of worker threads used for API."
+            );
+            Runtime::new().unwrap()
```
wkozyra95: up to you, but in general, I would avoid different initialization modes, less risk for unexpected changes. Check what `Builder::new_multi_thread()` does by default and replicate the behavior explictlly

wkozyra95: I know that i said it's up to you, but the new approach does not make sense, you are creating the builder, run some methods twice, it's still might be a different behavior dependint on which if/else branch is selected

JBRS307: I think I misunderstood the first time. I'll do it with one `Builder`.

src/server.rs
```diff
@@ -69,6 +69,37 @@ pub fn run_api(
     })
 }
 
+fn init_runtime() -> Runtime {
+    const MINIMUM_WORKER_THREADS: usize = 3;
+    let mut runtime_builder = Builder::new_multi_thread();
+    runtime_builder.enable_all();
```
noituri:
```suggestion
    let mut runtime_builder = Builder::new_multi_thread().enable_all();
```

src/server.rs
```diff
@@ -69,6 +69,37 @@ pub fn run_api(
...
+            .and_then(|v| v.trim().parse().ok())
+            .unwrap_or(MINIMUM_WORKER_THREADS);
+        debug!(
+            worker_threads = thread_count,
+            "Number of runtime worker threads used."
+        );
+        runtime_builder
+            .enable_all()
+            .worker_threads(thread_count)
+            .build()
```
noituri:
```suggestion
        runtime_builder
            .worker_threads(thread_count)
            .build()
```

noituri (commented):
NIT:

I think tokio will set worker threads to `std::thread::available_parallelism()` unless they are explicitly provided. So you could calculate the correct `thread_count` and then build the runtime, instead of branching on `available_threads >= MINIMUM_WORKER_THREADS`

e.g.
```rust
let available_threads = thread::available_parallelism()
   .map(|val| val.get())
   .unwrap_or(MINIMUM_WORKER_THREADS)
   .max(MINIMUM_WORKER_THREADS);
let thread_count = env::var("TOKIO_WORKER_THREADS")
   .ok()
   .and_then(|v| v.trim().parse().ok())
   .unwrap_or(available_threads);

// build runtime ....
```

src/server.rs
```diff
@@ -69,6 +69,40 @@ pub fn run_api(
...
+        .ok()
+        .map(|v| {
+            let val: usize = v
+                .parse()
+                .expect("Failed to parse TOKIO_WORKER_THREADS. Must be a number greater than 0.");
+            if val == 0 {
+                panic!("TOKIO_WORKER_THREADS must be greater than 0.");
+            } else {
+                val
+            }
```
noituri: i think this `else` is not necessary

JBRS307: I think that function used inside of `Option::map()` needs to return the final value, [spec](https://doc.rust-lang.org/std/option/enum.Option.html#method.map)

src/server.rs
```diff
@@ -69,6 +69,40 @@ pub fn run_api(
     })
 }
 
+fn init_runtime() -> Runtime {
+    const MINIMUM_WORKER_THREADS: usize = 3;
+
+    let available_threads = thread::available_parallelism().ok().map(|v| v.get());
+    trace!(available_parallelism=?available_threads, "Available cpus detected.");
```
wkozyra95:
```suggestion
    trace!(?available_threads, "Available CPUs detected.");
```
or change name of the variable

src/server.rs
wkozyra95:
```suggestion
            match val {
                0 => panic!("TOKIO_WORKER_THREADS must be greater than 0."),
                val => val,
            }
```

src/server.rs
```diff
@@ -69,6 +69,40 @@ pub fn run_api(
...
+    let available_threads = available_threads
+        .unwrap_or(MINIMUM_WORKER_THREADS)
+        .max(MINIMUM_WORKER_THREADS);
+
+    let thread_count = env::var("TOKIO_WORKER_THREADS")
+        .ok()
```
wkozyra95: what is this ok() here?

src/server.rs
```diff
@@ -69,6 +69,40 @@ pub fn run_api(
...
+            } else {
+                val
+            }
+        })
+        .unwrap_or(available_threads);
+
+    debug!(
+        worker_threads = thread_count,
+        "Number of runtime worker threads."
+    );
```
wkozyra95:
```suggestion
    debug!(thread_count, "Number of runtime worker threads.");
```

PR #1775, opened 2026-03-05
[vk-video] Add a possibility to not embed parameters in the stream

jerzywilczek:
This patch adds an option to the encoder to not embed parameters in the output stream, and a way to request the encoder to provide just the parameters to facilitate transferring them out-of-bound

### DIFF
--- a/vk-video/src/vulkan_video.rs
+++ b/vk-video/src/vulkan_video.rs
@@ -286,6 +286,22 @@ impl BytesEncoder {
     ) -> Result<EncodedOutputChunk<Vec<u8>>, VulkanEncoderError> {
         self.vulkan_encoder.encode_bytes(frame, force_keyframe)
     }
+
+    /// Retrieve encoded SPS NAL units from the video session parameters, in Annex B.
+    ///
+    /// Useful when `inline_stream_params` is `false` and the parameters need to be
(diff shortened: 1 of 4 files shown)


vk-video/src/vulkan_video.rs
```diff
@@ -286,6 +286,18 @@ impl BytesEncoder {
     ) -> Result<EncodedOutputChunk<Vec<u8>>, VulkanEncoderError> {
         self.vulkan_encoder.encode_bytes(frame, force_keyframe)
     }
+
+    /// Retrieve encoded SPS/PPS NAL units from the video session parameters.
+    ///
+    /// Useful when `inline_stream_params` is `false` and the parameters need to be
```
wkozyra95, line 292: Are they in AnnexB format? If so at that in the comment

vk-video/src/vulkan_video.rs
```diff
@@ -286,6 +286,18 @@ impl BytesEncoder {
...
+    /// sent out-of-band (e.g. in RTMP or MP4 headers).
+    pub fn stream_parameters(
+        &self,
+        write_sps: bool,
```
wkozyra95: this api does not feel natural, maybe separate method for sps and pps? Or if it's problematic we could skipp the options in public API and just produce both

PR #1777, opened 2026-03-05
[rtmp] Add RTMPS support for server

wkazmierczak:


smelter-core/src/pipeline.rs
```diff
@@ -97,7 +97,11 @@ pub enum PipelineWhipWhepServerOptions {
 
 #[derive(Debug)]
 pub enum PipelineRtmpServerOptions {
-    Enable { port: u16 }, // server config
+    Enable {
+        port: u16,
+        tls_cert_file: Option<Arc<str>>,
+        tls_key_file: Option<Arc<str>>,
```
wkozyra95, line 103: If those values are corelated then they should never be split into separate enums

rtmp/src/server/listen_thread.rs
```diff
@@ -25,15 +25,21 @@ pub(super) fn start_listener_thread(
     listener.set_nonblocking(true).unwrap();
     let on_connection = Arc::new(Mutex::new(on_connection));
 
+    let protocol = if config.tls.is_some() {
+        "RTMPS"
+    } else {
+        "RTMP"
+    };
+    info!("{protocol} server running on port {port}");
+
```
wkozyra95:
```suggestion
```

rtmp/src/server/listen_thread.rs
```diff
@@ -25,15 +25,21 @@ pub(super) fn start_listener_thread(
...
     let shutdown = Arc::new(AtomicBool::new(false));
     let server = Arc::new(Mutex::new(RtmpServer { config, shutdown }));
 
-    info!("RTMP server running on port {port}");
-
     let server_weak: Weak<Mutex<RtmpServer>> = Arc::downgrade(&server);
 
     thread::Builder::new()
-        .name("RTMP server".to_string())
+        .name(format!("{protocol} server"))
```
wkozyra95:
```suggestion
        .name("RTMP server".to_string())
```

rtmp/src/server/listen_thread.rs
```diff
@@ -25,15 +25,21 @@ pub(super) fn start_listener_thread(
...
     let on_connection = Arc::new(Mutex::new(on_connection));
 
+    let protocol = if config.tls.is_some() {
+        "RTMPS"
+    } else {
+        "RTMP"
+    };
+    info!("{protocol} server running on port {port}");
+
+    let tls = config.tls.clone();
```
wkozyra95, line 30:
```suggestion
    let tls = config.tls;
```
This clone does not do anything

rtmp/src/server/listen_thread.rs
```diff
@@ -50,8 +56,9 @@ pub(super) fn start_listener_thread(
                         info!("New connection from: {peer_addr:?}");
 
                         let on_connection_clone = on_connection.clone();
+                        let tls = tls.clone();
                         thread::spawn(move || {
```
wkozyra95, line 62: why are you passing that to handle_connection? Actual handler should know as little as possible about the underlying transport

rtmp/src/transport/tls.rs
```diff
@@ -55,3 +58,45 @@ impl Write for TlsClientStream {
...
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.0.write(buf)
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        self.0.flush()
+    }
+}
+
+fn build_server_tls_config(tls: &TlsConfig) -> Result<Arc<ServerConfig>, RtmpConnectionError> {
```
wkozyra95: don't see the point of separating this function:
- TlsServerStream:new does not do anything else
- entire struct TlsServerStream does not have any other logic. If it has a bunch of methods I would see the point of extracting a larger chunk, so the rest of the logic is not hidden.

smelter-core/src/pipeline.rs
wkazmierczak: I'll do it in following PR

PR #1778, opened 2026-03-06
Rust `1.94`

JBRS307:


wkozyra95 (approved):
fine, but maybe for the future let's wait with updating rust (one version behind) unless there is something specific in a new version that we need

PR #1779, opened 2026-03-06
Bitrate stats for outputs.

JBRS307:
Bitrate stats for the following outputs:
- WHIP
- WHEP
- HLS
- RTMP
- MP4
- ~RTP~, Will be done later

### DIFF
--- a/smelter-core/src/stats/output_state.rs
+++ b/smelter-core/src/stats/output_state.rs
@@ -75,7 +75,7 @@ impl OutputStatsState {
                 state.handle_event(event)
             }
             (state, event) => {
-                error!(?state, ?event, "Wrong event type for input")
+                error!(?state, ?event, "Wrong event type for output.")
             }
         }
     }


smelter-core/src/pipeline/webrtc/handle_keyframe_requests.rs
```diff
@@ -8,6 +8,7 @@ use webrtc::{
 
 use crate::PipelineCtx;
 
+// TODO: (@jbrs) Think of how to do nack handling here for stats
```
JBRS307:
```suggestion
```

smelter-core/src/pipeline/encoder/encoder_thread_audio.rs
```diff
@@ -76,6 +87,10 @@ where
 
     fn run(self) {
         for event in self.stream {
+            if let Some(make_event) = self.chunk_size_event {
+                self.stats_sender
+                    .send(vec![make_event(event.data_size(), &self.output_ref)]);
```
wkozyra95: You are putting this `vec!` everywhere. Do no do that if you send one event

smelter-core/src/pipeline/encoder/encoder_thread_video.rs
```diff
@@ -21,11 +21,16 @@ pub(crate) struct VideoEncoderThreadOptions<Encoder: VideoEncoder> {
...
+    pub output_ref: Ref<OutputId>,
+    pub chunk_size_event: Option<fn(u64, &Ref<OutputId>) -> StatsEvent>,
 }
 
 pub(crate) struct VideoEncoderThread<Encoder: VideoEncoder> {
     stream: Box<dyn Iterator<Item = EncodedOutputEvent>>,
     chunks_sender: Sender<EncodedOutputEvent>,
+    output_ref: Ref<OutputId>,
+    stats_sender: StatsSender,
+    chunk_size_event: Option<fn(u64, &Ref<OutputId>) -> StatsEvent>,
```
wkozyra95: This is feels very bolted on. I would explore other options before doing something like this.

Also you are doing protocol specific types either way, so I don't really see the point of trying to push that logic to the decoder/encoder. It would be a lot cleaner in the protocol code.

smelter-core/src/pipeline/rtmp/rtmp_output.rs
```diff
@@ -329,18 +345,24 @@ fn run_synced_av(
             (false, false) => match (&pending_video, &pending_audio) {
                 (Some(video), Some(audio)) => {
                     if video.pts <= audio.pts {
+                        rtmp_stats_sender
+                            .bytes_sent_event(video.data_size(), StatsTrackKind::Video);
                         client.send(video_chunk_to_event(pending_video.take().unwrap()))?;
```
JBRS307, line 350: This thing should be performed after actual data is sent, not before, check everywhere

smelter-core/src/protocols/channel/encoded_output.rs
```diff
@@ -33,6 +33,21 @@ pub struct EncodedOutputChunk {
...
+impl EncodedOutputChunk {
+    pub fn data_size(&self) -> u64 {
+        self.data.len() as u64
+    }
+}
+
+impl EncodedOutputEvent {
+    pub fn data_size(&self) -> u64 {
+        match self {
+            Self::Data(data) => data.data.len() as u64,
```
JBRS307: Make use of the internal function that is already implemented

smelter-core/src/stats/output_state/rtmp.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    pub fn report(&mut self) -> RtmpOutputTrackStatsReport {
+        // RtmpOutputTrackStatsReport {
+        //     last_10_seconds: RtmpOutputTrackSlidingWindowStatsReport {
+        //         bitrate_avg: self.bitrate_10_secs.sum()
+        //             / self.bitrate_10_secs.window_size().as_secs(),
+        //     },
+        // }
```
JBRS307:
```suggestion
```

smelter-core/src/pipeline/hls/hls_output.rs
```diff
@@ -332,6 +344,7 @@ fn write_chunk(
     output_ctx: &mut ffmpeg::format::context::Output,
     frame_duration: Duration,
     timestamp_offset: Duration,
+    hls_stats_sender: &HlsOutputStatsSender,
```
wkozyra95: You are in hls file, `hls_` prefix is not helpful. I think you are thinking about changes in this PR, but not how they fit the specific inputs.

The changes fit exactly the same pattern regardless of how it works for specific input. e.g. sometimes passing one more argument is fine and sometimes is not.

smelter-core/src/pipeline/hls/hls_output.rs
```diff
@@ -383,7 +396,23 @@ fn write_chunk(
         packet.set_flags(ffmpeg::packet::Flags::KEY)
     }
 
-    if let Err(err) = packet.write(output_ctx) {
-        error!("Failed to write packet to HLS file: {}.", err);
+    match packet.write(output_ctx) {
+        Ok(_) => {
+            hls_stats_sender.bytes_sent_event(chunk.data_size(), chunk.kind.into());
```
wkozyra95: I think this is the correct approach, but I'm not sure if it's worth the extra logic, up to you

smelter-core/src/pipeline/rtmp/rtmp_output.rs
```diff
@@ -329,22 +345,30 @@ fn run_synced_av(
             (false, false) => match (&pending_video, &pending_audio) {
                 (Some(video), Some(audio)) => {
                     if video.pts <= audio.pts {
+                        let data_size = video.data_size();
                         client.send(video_chunk_to_event(pending_video.take().unwrap()))?;
```
wkozyra95, line 350: The same as above, I would prefer to reverse the order than the complexity it introduces. Of course, it would be great to have both, but I don't see it without more in depth look into that.

smelter-core/src/pipeline/rtmp/rtmp_output.rs
```diff
@@ -353,3 +377,16 @@ fn run_synced_av(
 
     Ok(())
 }
+
+struct RtmpOutputStatsSender {
```
wkozyra95, line 381: Are you sure this pattern is necessary for each protocol?

smelter-core/src/pipeline/rtp/mod.rs
```diff
@@ -25,6 +25,12 @@ pub enum RtpInputEvent {
     LostPacket,
 }
 
+impl RtpPacket {
+    pub fn data_size(&self) -> u64 {
```
wkozyra95: If it's needed I would do
```suggestion
    pub fn len(&self) -> usize {
```

smelter-core/src/pipeline/hls/hls_output.rs
JBRS307: I reversed the order so no additional logic is introduced. In case of error the sending loop is broken anyway, so info about one chunk that was not sent will not cause any harm.

smelter-core/src/pipeline/rtmp/rtmp_output.rs
JBRS307: It definitely looks nicer in the code, as sending this events manually looks uglier. Besides, I think it will be easier to add more stats later by having a simple wrapper for each protocol.

smelter-core/src/pipeline/hls/hls_output.rs
JBRS307: Another insight. The order where stats event is sent only if actual chunk is sent successfully will be necessary if we would like to add stats that count total packets sent. But for now it can stay like this.

PR #1780, opened 2026-03-06
Simplify connection handler for RTMP input

wkozyra95:


rtmp/examples/server.rs
```diff
@@ -15,13 +15,12 @@ fn main() {
     };
 
     let on_connection = Box::new(|conn: RtmpConnection| {
```
wkazmierczak, line 14: I think in current API `RtmpConnection` is not the best name. Maybe `RtmpStream` or something similar

wkozyra95: Why? I planned to rename it to RtmpServerConnection so it's not confused with client, but don't see the problem with the word Connection.

PR #1781, opened 2026-03-06
[vk-video] Change the deocde input buffer to not overwrite it in the transcoder

jerzywilczek:


vk-video/src/vulkan_transcoder.rs
```diff
@@ -240,6 +242,9 @@ impl Transcoder {
         self.decoder
             .tracker
             .mark_waited(resized_images.data.decoder_wait_value);
+        self.decoder
+            .free_input_buffer(resized_images.data.input_buffer);
+
         self.resizing_pipeline
             .mark_command_buffers_completed(resized_images.data.decoder_wait_value);
         self.resizing_pipeline
```
jerzywilczek, line 251: I dont like this whole section very much, any ideas on how to make it better?

noituri: you pass DecodeSubmission (which has reference to the decoder) to the resizing_pipeline and it gives you back ResizeSubmission so maybe you could create a `mark_as_handled` method on ResizeSubmission which would do the cleanup

noituri: managing the lifetimes could be difficult with this tbh

PR #1782, opened 2026-03-06
[rtmp] Implement shutdown logic

wkozyra95:
- Implement shutdown logic (previously shouting down server would not shutdown connections)
- Extract logic for storing shared state
- Hide channel implementation details

### DIFF
--- a/rtmp/src/server/instance.rs
+++ b/rtmp/src/server/instance.rs
@@ -64,9 +64,9 @@ struct ServerInstance {
     shutdown_condition: ShutdownCondition,
 }
 
-impl Drop for RtmpServer {
+impl Drop for ServerInstance {
     fn drop(&mut self) {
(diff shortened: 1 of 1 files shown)


rtmp/src/server/listener_thread.rs
```diff
@@ -0,0 +1,98 @@
...
+    let thread_handle = thread::Builder::new()
+        .name("RTMP connection thread".to_string())
+        .spawn(move || {
+            if let Err(err) = run_connection_thread(&ctx_clone, transport) {
+                error!(?err, "Connection terminated with an error");
+            }
+        })
+        .unwrap();
+
+    ctx.lock().unwrap().thread_handle = Some(thread_handle);
```
wkazmierczak, line 96: I think no method is called on `thread_handle`. It is just stored in state.

PR #1860, opened 2026-03-26
[rtmp] Handle E-RTMP video messages

wkazmierczak:


rtmp/src/lib.rs
```diff
@@ -15,3 +15,17 @@ pub use error::*;
 pub use events::*;
 pub use flv::*;
 pub use server::*;
+
+pub(crate) const VIDEO_FOURCC_LIST: [&str; 6] = ["av01", "vp09", "vp08", "hvc1", "vvc1", "avc1"];
```
wkazmierczak, line 19: Not sure if those consts can be here

rtmp/src/client.rs
```diff
@@ -156,7 +171,8 @@ impl RtmpClientState {
                 Err(err) => return Err(err.into()),
             };
 
-            if let Some(_response) = state.try_match_connect_response(&msg)? {
+            if let Some(response) = state.try_match_connect_response(&msg)? {
+                self.peer_supports_enhanced = connect_response_supports_enhanced(&response);
```
wkozyra95: I would do that inside try_match_connect_response and return 2 values

rtmp/src/client/negotiation.rs
```diff
@@ -0,0 +1,266 @@
...
+            stream_key: stream_key.to_string(),
+            publishing_type: "live".to_string(),
+        },
+        stream_id,
+    })?;
+    Ok(())
+}
+
+pub(super) fn connect_response_supports_enhanced(response: &CommandMessageConnectSuccess) -> bool {
+    map_supports_enhanced(&response.properties) || map_supports_enhanced(&response.information)
```
wkozyra95: too small, I would inline that

rtmp/src/client/negotiation.rs
```diff
@@ -0,0 +1,266 @@
...
+        AmfValue::Object(map) => map,
+        AmfValue::EcmaArray(map) => map,
+        _ => return false,
+    };
+
+    map.iter().any(|(k, v)| {
+        let AmfValue::Number(mask) = v else {
+            return false;
+        };
+        *mask > 0.0 && (k == "*" || VIDEO_FOURCC_LIST.contains(&k.as_str()))
```
wkozyra95: hard to read, I would consider match with 3 entries (None, Some(mask) if mask > 0, Some(mask))

wkozyra95 (approved):

PR #1885, opened 2026-04-08
Lanczos3 mipmap cache

jerzywilczek:
A very simple cache for textures with mipmaps, to avoid allocating them each frame.

### DIFF
--- a/smelter-render/src/transformations/layout.rs
+++ b/smelter-render/src/transformations/layout.rs
@@ -25,6 +25,7 @@ pub(crate) trait LayoutProvider: Send {
 pub(crate) struct LayoutNode {
     layout_provider: Box<dyn LayoutProvider>,
     shader: Arc<LayoutShader>,
+    mip_cache: HashMap<usize, crate::wgpu::utils::MippedTexture>,
 }
 
 /// When rendering we cut this fragment from texture and stretch it on
@@ -155,6 +156,7 @@ impl LayoutNode {
         Self {
             layout_provider,
             shader,
+            mip_cache: HashMap::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


smelter-render/src/transformations/layout.rs
```diff
@@ -288,6 +289,8 @@ impl LayoutNode {
         );
 
         ctx.wgpu_ctx.queue.submit(Some(encoder.finish()));
+
+        self.mip_cache = mipped_textures;
```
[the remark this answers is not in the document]
jerzywilczek: I will not consider that

smelter-render/src/transformations/layout.rs
```diff
@@ -25,6 +25,7 @@ pub(crate) trait LayoutProvider: Send {
 pub(crate) struct LayoutNode {
     layout_provider: Box<dyn LayoutProvider>,
     shader: Arc<LayoutShader>,
+    mip_cache: HashMap<usize, crate::wgpu::utils::MippedTexture>,
```
wkozyra95, line 28:
```suggestion
    mip_cache: HashMap<usize, MippedTexture>,
```
any reason for not importing that type?

wkozyra95 (commented):
Not related to this PR, but also make sure to test both a case where:
- layout child is an input texture
- layout child is another Node texture (e.g. Text or Image)

I don't expect  any issues, but accessing them is a bit different and there were bugs in the past that happens only in one scenario, but not in the other

wkozyra95 (approved):

smelter-render/src/transformations/layout.rs
jerzywilczek: forgetfullness

PR #1893, opened 2026-04-13
`MP4` better `elst` offset handling.

JBRS307:
- New snapshots: https://github.com/smelter-labs/smelter-snapshot-tests/pull/79

---

Offset calculation and move is now performed using the Seek function. This way frames that will not be displayed are not pushed to decoder.


wkozyra95 (approved):
PR name is very unclear. Also a bit  misleading, you did not change anything related to seek functionality, You improved a case where there is no seeking on mp4 input

PR #1894, opened 2026-04-13
[rtmp] Add `tcUrl` to `connect` message - RTMP client

wkazmierczak:


rtmp/src/client.rs
```diff
@@ -64,7 +64,14 @@ impl RtmpClient {
             last_ack: 0,
         };
 
-        let stream_id = state.negotiate_connection(&config.app, &config.stream_key)?;
+        let tc_url = format!(
```
wkozyra95: pass options to negotiation, and create url just before, or extract it to a function or method on config

wkozyra95 (approved):

PR #1897, opened 2026-04-14
`MP4` Handling of `elst` empty edit.

JBRS307:


smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -108,37 +111,51 @@ impl<Reader: Read + Seek + Send + 'static> Mp4FileReader<Reader> {
...
-    /// This implementation synchronizes using first elst box. If box is missing or
-    /// first edit is empty then zero offset is used.
-    fn calculate_offset(track: &Mp4Track) -> Duration {
-        let first_elst = track
+    /// Calculates the media-time offset and presentation delay from the edit list.
+    ///
+    /// Returns `(offset, delay)` where:
+    /// - `offset` is derived from `media_time` of the first non-empty edit (used for seeking)
+    /// - `delay` is the sum of `segment_duration` of all leading empty edits (presentation delay)
+    fn calculate_offset(track: &Mp4Track, movie_timescale: u32) -> (Duration, Duration) {
```
JBRS307, line 133: Rename that to something like `calculate_elst_edits`

smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -108,37 +111,51 @@ impl<Reader: Read + Seek + Send + 'static> Mp4FileReader<Reader> {
...
 
-    /// This implementation synchronizes using first elst box. If box is missing or
-    /// first edit is empty then zero offset is used.
-    fn calculate_offset(track: &Mp4Track) -> Duration {
-        let first_elst = track
+    /// Calculates the media-time offset and presentation delay from the edit list.
+    ///
+    /// Returns `(offset, delay)` where:
+    /// - `offset` is derived from `media_time` of the first non-empty edit (used for seeking)
+    /// - `delay` is the sum of `segment_duration` of all leading empty edits (presentation delay)
```
JBRS307, line 132: Makes no sense, rewrite

smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -108,37 +111,51 @@ impl<Reader: Read + Seek + Send + 'static> Mp4FileReader<Reader> {
...
+            .map(|elst| elst.entries.as_slice())
+            .unwrap_or(&[]);
+
+        let mut delay_ticks: u64 = 0;
+        let mut offset = Duration::ZERO;
+
+        for entry in entries {
+            // u32::MAX value is the result of overflowing -1
+            if entry.media_time == u32::MAX as u64 || entry.media_time == u64::MAX {
+                // Empty edit (media_time == -1): accumulate delay
```
JBRS307:
```suggestion
```

smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -283,13 +300,16 @@ impl<Reader: Read + Seek + Send + 'static> TrackChunks<'_, Reader> {
...
 
         let dts = Duration::from_secs_f64(start_time as f64 / self.track.timescale as f64)
-            .saturating_sub(self.seek);
+            .saturating_sub(self.seek)
+            + delay;
         let pts = Duration::from_secs_f64(
             (start_time as f64 + rendering_offset as f64) / self.track.timescale as f64,
         )
-        .saturating_sub(self.seek);
+        .saturating_sub(self.seek)
```
wkozyra95, line 311: always add first before subtracting, split that code for readability

smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -283,13 +300,16 @@ impl<Reader: Read + Seek + Send + 'static> TrackChunks<'_, Reader> {
         let start_time = sample.start_time;
         let sample_duration =
             Duration::from_secs_f64(sample.duration as f64 / self.track.timescale as f64);
+        let delay = self.track.delay;
 
         let dts = Duration::from_secs_f64(start_time as f64 / self.track.timescale as f64)
-            .saturating_sub(self.seek);
+            .saturating_sub(self.seek)
+            + delay;
```
wkozyra95, line 307:
```suggestion
```
dts absolute value does not matter so you can remove it for simplicity

smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -108,37 +111,50 @@ impl<Reader: Read + Seek + Send + 'static> Mp4FileReader<Reader> {
...
+
         Some(Track {
             sample_count: track.sample_count(),
             timescale: track.timescale(),
             track_id,
             duration: track.duration(),
             decoder_options: DecoderOptions::H264(h264_config),
-            offset: Self::calculate_offset(track),
+            offset,
+            delay,
```
wkozyra95, line 123: Better names are needed, plus add comments describing what they are in the Track definition. Technically, what you call a delay is offset

Also normally I don't recommend, but this is complex enough that it is worth documenting there where and how those values should be used

smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -149,28 +168,60 @@ pub(crate) struct Track<Reader: Read + Seek + Send + 'static> {
...
+                        track_seek: track_start_offset,
+                        next_sample_index: start_index,
+                        present_from_index: present_index,
+                    }
+                }
+                None => TrackChunks {
+                    track: self,
+                    track_seek: Duration::ZERO,
+                    next_sample_index: 1,
+                    present_from_index: 1,
```
JBRS307: For this to happen someone would have to provide forged mp4 with `media_time` larger than duration.

JBRS307: I am not sure if we sure keep it or just ignore that case

JBRS307:
There is still one little visual issue I'd like to address, but it is not crucial at the moment. With delay instead of black screen we see the first frame for the duration of the delay.

smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -149,28 +169,60 @@ pub(crate) struct Track<Reader: Read + Seek + Send + 'static> {
...
+        // Position on the raw media-sample timeline (the STTS-described timeline where
+        // each stored sample has a fixed time) from which sample reading should start.
+        // Computed so that any portion of `user_seek` that falls inside the leading
+        // `presentation_delay` (black screen) is absorbed by that black screen instead
+        // of being skipped over real samples. When `user_seek <= presentation_delay`
+        // this collapses to `track_start_offset`, meaning "start at the very first
+        // sample the `elst` box tells us to actually play"; excess seek beyond the
+        // black screen advances further into media.
+        let media_seek =
+            self.track_start_offset + user_seek.saturating_sub(self.presentation_delay);
```
wkozyra95:
```suggestion
            (self.track_start_offset + user_seek).saturating_sub(self.presentation_delay);
```

smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -149,28 +169,60 @@ pub(crate) struct Track<Reader: Read + Seek + Send + 'static> {
...
-                track: self,
-                seek: Duration::ZERO,
-                next_sample_index: 1,
-                present_from_index: 1,
+            None => match self.find_seek_start_sample(self.track_start_offset) {
+                Some((start_index, present_index)) => {
+                    let track_start_offset = self.track_start_offset;
+                    TrackChunks {
+                        track: self,
+                        track_seek: track_start_offset,
```
wkozyra95:
```suggestion
                    TrackChunks {
                        track: self,
                        track_seek:,
```

smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -149,28 +169,60 @@ pub(crate) struct Track<Reader: Read + Seek + Send + 'static> {
...
+        let media_seek =
+            self.track_start_offset + user_seek.saturating_sub(self.presentation_delay);
+
+        // Used in `sample_into_chunk` to shift each sample's pts so that the user's
+        // seek point becomes pts 0. If the user seeks into the leading black screen
+        // (user_seek < presentation_delay), the first real sample gets a positive pts
+        // equal to the remaining black screen, which the pipeline fills with black.
+        let track_seek = self.track_start_offset + user_seek;
+
+        match self.find_seek_start_sample(media_seek) {
```
wkozyra95, line 203: let's change behavior for seeking duraiton longer than video, I would prefer to not play from the begining.

Solution that would work is returning last keyframe and packets after keyframe with present: false. However try to find cleaner approach, maybe just return index > last sample would work too

smelter-core/src/pipeline/mp4/reader.rs
JBRS307: This breaks the logic. The value of seek provided by user has to be reduced by the delay. The amount that is left is the actual amount that we want to seek into the actual track. If seek is smaller than the delay, we set it to we don't want to seek into the track more than offset, in that case we only use that value to move the `pts` in the `smaple_into_chunk()`

JBRS307: I made the change to make it more readable without logic change.

wkozyra95 (approved):

smelter-core/src/pipeline/mp4/reader.rs
```diff
@@ -190,14 +231,18 @@ impl<Reader: Read + Seek + Send + 'static> Track<Reader> {
     /// Returns `(start_index, present_from_index)` for the given seek position.
     /// `start_index` is the last sync sample before seek (for decoder warmup).
     /// `present_from_index` is the first sample at or after seek.
-    /// Returns `None` if seek is past the end.
-    fn find_seek_start_sample(&self, seek: Duration) -> Option<(u32, u32)> {
+    /// If seek is past the end, returns `Err` with index one past the last sample
+    fn find_seek_start_sample(&self, seek: Duration) -> Result<(u32, u32), u32> {
```
wkozyra95: I wouldn't return that as an error

PR #1901, opened 2026-04-15
Duration based channel (demuxer -> decoder)

wkozyra95:


smelter-core/src/queue/video_input.rs
```diff
@@ -253,7 +253,7 @@ impl VideoInputReceiver {
     pub fn new(delay: Duration) -> (Self, Sender<Frame>) {
         let (sender, receiver) = bounded(1);
         let track = Self {
-            max_size: Duration::from_secs(1),
+            max_size: Duration::from_millis(100),
```
wkozyra95, line 256: it will always one elelement more than max, so if framerate is 1fps it will still work (there will be 2 frames in the buffer)

smelter-core/src/queue/queue_input.rs
```diff
@@ -16,6 +15,29 @@ use crate::{
 
 use crate::prelude::*;
 
+pub(crate) struct QueueSender<T>(crossbeam_channel::Sender<T>);
+
+impl<T> QueueSender<T> {
```
wkozyra95, line 20: I'm mostly using it to clean up imports QueueSender is used in the same files as the duration_bounded channel which is anoying to deal with

smelter-core/src/pipeline/utils/input_buffer.rs
```diff
@@ -36,10 +22,12 @@ impl<T: TimedValue> InputDelayBuffer<T> {
 
     pub fn write(&mut self, item: T) {
         self.buffer.push_back(item);
-        if !self.ready
-            && let (Some(first), Some(last)) = (self.buffer.front(), self.buffer.back())
-        {
-            self.ready = last.timestamp_range().1.abs_diff(first.timestamp_range().0) > self.size
+        if !self.ready {
+            let first_ts = self.buffer.iter().find_map(|i| i.timestamp_range());
```
wkozyra95, line 26: TimedValue can return None, so I can't just check first element

smelter-core/src/pipeline/utils/channel.rs
jerzywilczek, line 1: There is a rather unsettling bug in this code. There are two scenarios I came up with that break it, one more likely, and one that maybe can't happen given our current usage of the channel (but might in the future)

scenario 1:

There are two producers blocked on pushing into a full channel. A consumer eats some values and makes the channel not full. He wakes up one producer (`notify_one`). One producer wakes up, pushes their value and quits. The pushed value returns `None` form its `timestamp_range()`, which means the channel is still not full, but the other producer was not woken up and doesn't push their value.

scenario 2 is more or less the same thing, but with no "untimed" values. Let's say the channel has capacity of 1 second and contains a single, 1.5 second long value. Two producers with values timed for 0.5s each are waiting. When the value is consumed, only one of them wakes up and the other never pushes their value.

There are two possible fixes for this, one is to simply change `notify_one` to `notify_all` (which is not ideal, since many threads can be woken up without the need). The other way, which I think is better, is for the producer to after pushing a new value check whether the channel is full, and `notify_one` on the `not_full` condvar if it isn't. This way, all blocked producers that have a chance of pushing can be woken up.

I think the second fix is better. While it is a bit more complex, the increase in complexity is very small, and we get the optimal behavior out of it.

jerzywilczek (changes requested):
I only looked at the channel in detail and just gave the rest a brief look, but everything other than the channel seems fine.

jerzywilczek (approved):

wkazmierczak (approved):

PR #1903, opened 2026-04-15
Better WHIP WHEP server errors

wkazmierczak:


smelter-core/src/error.rs
```diff
@@ -285,6 +286,7 @@ pub enum ErrorType {
     UserError,
     EntityNotFound,
     Conflict,
+    UpstreamError(StatusCode),
```
wkozyra95:
```suggestion
    Other(StatusCode),
```
or

```suggestion
    Custom(StatusCode),
```

I also don't like the fact that you are using StatusCode from reqwest to return error from axum, it works just because they both are based on the same internals

smelter-core/src/error.rs
```diff
@@ -335,6 +338,14 @@ impl From<&RegisterInputError> for PipelineErrorInfo {
             {
                 PipelineErrorInfo::new(WHEP_REQUEST_FAILED, ErrorType::UserError)
             }
+            RegisterInputError::InputError(_, InputInitError::Whep(err))
+                if matches!(err.as_ref(), WebrtcClientError::BadStatus(_, _)) =>
+            {
+                let WebrtcClientError::BadStatus(status, _) = err.as_ref() else {
+                    unreachable!()
```
wkozyra95: definitely not unreachable

smelter-core/src/error.rs
```diff
@@ -407,6 +419,14 @@ impl From<&RegisterOutputError> for PipelineErrorInfo {
             {
                 PipelineErrorInfo::new(WHIP_REQUEST_FAILED, ErrorType::UserError)
             }
+            RegisterOutputError::OutputError(_, OutputInitError::WhipInitError(err))
+                if matches!(err.as_ref(), WebrtcClientError::BadStatus(_, _)) =>
+            {
+                let WebrtcClientError::BadStatus(status, _) = err.as_ref() else {
+                    unreachable!()
```
wkozyra95: the same

wkozyra95 (approved):

PR #1907, opened 2026-04-17
Expose video encoder info on registered outputs

michael-dm:
Adds `Pipeline::output_video_encoder_info(&output_id) -> Option<VideoEncoderInfo>` so downstream consumers can read the encoder's `resolution` and H.264 `avcC` extradata without re-parsing NAL units from the encoded chunk stream.

Useful for anyone building a custom muxer or transport on top of `register_encoded_data_output`: populate `codec_private` for Matroska/MP4, feed `description` into a browser WebCodecs `VideoDecoder`, or hand `extradata` to an ffmpeg decoder in another process.

Purely additive: one new public method on `Pipeline`, one new public struct `VideoEncoderInfo`, and a default-`None` method on the crate-internal `Output` trait. No existing code needs to change.

Happy to reshape if preferred — alt (B): return a handle struct `{ receiver, video_encoder_info, audio_encoder_info }` from `register_encoded_data_output` (atomic with registration, breaking).
Alt (C): add a `VideoEncoderReady` variant to `EncodedOutputEvent` (async delivery via the existing stream, mildly breaking for exhaustive matches).

### DIFF
--- a/smelter-core/src/pipeline/instance.rs
+++ b/smelter-core/src/pipeline/instance.rs
@@ -160,16 +160,16 @@ impl Pipeline {
         pipeline: &Arc<Mutex<Self>>,
         output_id: OutputId,
         register_options: RegisterEncodedDataOutputOptions,
-    ) -> Result<Receiver<EncodedOutputEvent>, RegisterOutputError> {
+    ) -> Result<EncodedDataOutputHandle, RegisterOutputError> {
         register_pipeline_output(
             pipeline,
             output_id,
             register_options.video,
             register_options.audio,
             |ctx, output_ref| {
-                let (output, result) =
+                let (output, handle) =
                     EncodedDataOutput::new(ctx, output_ref, register_options.output_options)?;
-                Ok((Box::new(output), result))
+                Ok((Box::new(output), handle))
             },
         )
     }
--- a/integration-tests/examples/encoded_channel_output.rs
+++ b/integration-tests/examples/encoded_channel_output.rs
@@ -92,7 +92,7 @@ fn main() {
 
     Pipeline::register_input(&state.pipeline().unwrap(), input_id.clone(), input_options).unwrap();
 
-    let output_receiver = Pipeline::register_encoded_data_output(
+    let output = Pipeline::register_encoded_data_output(
         &state.pipeline().unwrap(),
         output_id.clone(),
         output_options,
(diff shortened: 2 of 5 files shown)


smelter-core/src/pipeline/instance.rs
```diff
@@ -239,6 +239,20 @@ impl Pipeline {
...
+    /// output.
+    ///
+    /// Returns `None` if the output is not registered, has no video track, or
+    /// uses an output kind that does not expose encoder info (currently only
+    /// outputs registered via [`Pipeline::register_encoded_data_output`] do).
+    ///
+    /// The returned [`VideoEncoderInfo::extradata`] is populated only for
+    /// encoders that produce a decoder configuration record (e.g. H.264 in
+    /// AVCC bitstream mode).
+    pub fn output_video_encoder_info(&self, output_id: &OutputId) -> Option<VideoEncoderInfo> {
```
wkozyra95: Will you need support for other output? I would prefer to return that config on registration instead.

Instead of returning `Receiver<EncodedOutputEvent>` from `EncodedDataOutput::new`, you can return some struct that has both the receiver and that data. (and change signature of Pipeline::register_encoded_data_output)

smelter-core/src/pipeline/output.rs
```diff
@@ -40,6 +40,13 @@ pub(crate) trait Output: Send {
     fn audio(&self) -> Option<OutputAudio<'_>>;
     fn video(&self) -> Option<OutputVideo<'_>>;
     fn kind(&self) -> OutputProtocolKind;
+
+    /// Snapshot of the video encoder configuration (resolution and optional
+    /// decoder configuration record). `None` when the output has no video
+    /// track or the encoder does not expose this information.
+    fn video_encoder_info(&self) -> Option<VideoEncoderInfo> {
```
wkozyra95: If you need that for only this output, then refer to my previous comment)

If you will need that config on other outputs too then, I would prefer to move those 2 fields to OutputVideo and only extract those 2 fields inside `output_video_encoder_info` function into `VideoEncoderInfo`

michael-dm:
Took your suggestion — reshape pushed as `b74429a3`. `register_encoded_data_output` now returns an `EncodedDataOutputHandle { receiver, video: Option<VideoEncoderInfo>, audio: Option<AudioEncoderInfo> }`; the `Pipeline` getter and trait method are gone.

wkozyra95 (approved):

PR #1908, opened 2026-04-17
Fix queue logic for higher input FPS

wkozyra95:


smelter-core/src/queue/queue_thread.rs
```diff
@@ -119,7 +119,7 @@ impl QueueThreadAfterStart {
     }
 
     fn run(mut self) {
-        let ticker = tick(Duration::from_millis(10));
+        let ticker = tick(Duration::from_millis(5));
```
wkozyra95, line 122: It defines how often the queue tries to do something

smelter-core/src/queue/video_queue.rs
```diff
@@ -65,29 +65,35 @@ impl VideoQueue {
         next_pts: Duration,
         queue_start_pts: Duration,
     ) -> bool {
-        if !self.ahead_of_time_processing && self.sync_point + next_pts > Instant::now() {
-            return false;
-        }
+        let input_status: HashMap<_, _> = self
```
wkozyra95: Before `is_ready_for_pts` was not called until queue already wanted to pop the queue, so effectively, the enqueue in the queue logic was only called once per frame.

Now this will be called at minimum every 5 milliseconds

smelter-core/src/queue/audio_queue.rs
```diff
@@ -61,24 +61,35 @@ impl AudioQueue {
...
+            .filter_map(|(input_id, weak)| {
+                weak.audio(|input| {
+                    let is_ready = input.is_ready_for_pts(pts_range, queue_start_pts);
+                    let is_required = input.required();
+                    (input_id, (is_ready, is_required))
+                })
+            })
+            .collect();
+
+        self.inputs
```
wkozyra95: not related to the bug, but this logic needs to be both in audio and video track

smelter-core/src/queue/audio_input.rs
```diff
@@ -243,7 +243,7 @@ impl AudioInputReceiver {
     pub fn new(delay: Duration) -> (Self, Sender<InputAudioSamples>) {
         let (sender, receiver) = bounded(1);
         let track = Self {
-            max_size: Duration::from_secs(1),
+            max_size: Duration::from_millis(100),
```
wkozyra95, line 246: Unrelated to the bug, I had it changed in one of the other follow-up PRs, but I needed to test if lowering it does not cause the same problem with higher fps

wkazmierczak (approved):

PR #1913, opened 2026-04-21
Preproces frames before sending them via side channel

wkozyra95:
Extracted part of the rendering pipeline to process frames. For now, it does not do much, code for rescaling and wgpu textures is not used.

In follow-up, I will expose options for configuring resolution and conversions to e.g bgr

### DIFF
--- a/smelter-render/src/state/frame_pre_processor.rs
+++ b/smelter-render/src/state/frame_pre_processor.rs
@@ -1,0 +1,249 @@
+use std::io::Write;
+use std::sync::Arc;
+
+use bytes::BufMut;
+use crossbeam_channel::bounded;
+use tracing::error;
+
+use crate::{
+    Frame, RenderingMode, Resolution,
+    wgpu::{
+        WgpuCtx,
+        texture::{RgbaLinearTexture, RgbaSrgbTexture, TextureExt, utils::pad_to_256},
+    },
+};
+
+use super::{input_texture::InputTexture, node_texture::NodeTexture};
+
+/// Stand-alone single-frame processor. Takes one `Frame` and runs it through
+/// a fixed pipeline:
+///
+///   1. upload: `Frame` (source format) -> `input_texture`
+///   2. convert: `input_texture` -> `node_texture` (RGBA at source resolution)
+///   3. rescale (optional, only when a target resolution is requested):
+///      `node_texture` -> `rescale_texture` (RGBA at target resolution)
+///   4. output - one of:
+///        - `process_to_texture`: clone the final RGBA texture and return it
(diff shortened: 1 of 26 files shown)


smelter-render/src/state/frame_pre_processor.rs
```diff
@@ -0,0 +1,249 @@
...
+            let rescale_texture = self.rescale_texture.as_ref().unwrap().texture();
+            self.download(rescale_texture, resolution)
+        } else {
+            let resolution = self.node_texture.resolution().unwrap();
+            self.ensure_download_buffer(resolution);
+
+            let node_texture = self.node_texture.state().unwrap().texture();
+            self.download(node_texture, resolution)
+        }
+    }
```
JBRS307, line 107:
```suggestion
        match resolution {
            Some(resolution) => {
                self.rescale_node_texture(resolution);
                self.ensure_download_buffer(resolution);

                let rescale_texture = self.rescale_texture.as_ref().unwrap().texture();
                self.download(rescale_texture, resolution)
            }
             None => {
                let resolution = self.node_texture.resolution().unwrap();
                self.ensure_download_buffer(resolution);

                let node_texture = self.node_texture.state().unwrap().texture();
                self.download(node_texture, resolution)
            }
        }
```

Nit pick, maybe `match` would be better

smelter-render/src/state/frame_pre_processor.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        let output = bytes::BytesMut::with_capacity(row_bytes * resolution.height);
+        let mut writer = output.writer();
+
+        {
+            let range = buffer.slice(..).get_mapped_range();
+            for chunk in range.chunks(padded_row_bytes) {
+                writer.write_all(&chunk[..row_bytes]).unwrap();
+            }
+        }
```
JBRS307, line 199:
```suggestion
        let range = buffer.slice(..).get_mapped_range();
        for chunk in range.chunks(padded_row_bytes) {
            writer.write_all(&chunk[..row_bytes]).unwrap();
        }
```

No name conflict so no need for separate block

JBRS307 (approved):

smelter-render/src/state/frame_pre_processor.rs
wkozyra95: it would panic on unmap

smelter-render/src/state/frame_pre_processor.rs
wkozyra95: maybe in follow-up, this could will be updated probably soon

PR #1921, opened 2026-04-22
[vk-video] Make sure all nalus contain their start codes

noituri:
Changed nalu splitter so that it won't include start code of the next nalu in the current nalu bytes. Also fixed some bugs

### DIFF
--- a/vk-video/src/parser/nalu_splitter.rs
+++ b/vk-video/src/parser/nalu_splitter.rs
@@ -25,10 +25,14 @@ fn find_start_of_next_nalu(buf: &[u8]) -> Option<usize> {
     // this could mean either that it's the longer 0, 0, 0, 1 code, or that there is one byte of
     // the previous nalu and the shorter code. This is what is checked here.
     if buf[0] != 0 && buf[1..4] == [0, 0, 1] {
-        return Some(5);
+        return Some(1);
     }
 
-    FINDER.find(&buf[2..]).map(|i| i + 5)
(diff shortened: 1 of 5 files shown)


vk-video/src/parser/nalu_splitter.rs
```diff
@@ -55,7 +55,7 @@ impl NALUSplitter {
 
         // This will cause the whole start code to be reprocessed when the beginning of the next start code
         // is at the end of current buffer.
-        self.previous_search_end = self.buffer.len().saturating_sub(3);
+        self.previous_search_end = self.buffer.len().saturating_sub(4);
```
noituri, line 62: I think `3` was incorrect here, it would not work if someone would push 2 nalus like that:
1. `0 0 0 1 X X 0 0 0`:  `find_start_of_next_nalu()` can't find start code which ends the nalu which is correct. So it sets `previous_search_end` to `6`
2. Then we push `1 Y Y Y 0 0 0 1` so we end up with buffer that is `0 0 0 1 X X 0 0 0 1 Y Y Y 0 0 0 1` but we start from the index `6` so we look for the start of the next nalu in this slice `0 0 0 1 Y Y Y 0 0 0 1`, we ignore the start code at the beginning and find the next start code at the end so we have one nalu `0 0 0 1 X X 0 0 0 1 Y Y Y` instead of 2 nalus

jerzywilczek (approved):

PR #1923, opened 2026-04-22
[rtmp] `RtmpEvent` refactor

wkazmierczak:


rtmp/src/flv/video.rs
```diff
@@ -1,24 +1,24 @@
 use bytes::{BufMut, Bytes, BytesMut};
 
-use crate::{RtmpMessageSerializeError, error::FlvVideoTagParseError};
+use crate::{RtmpMessageSerializeError, RtmpVideoCodec, error::FlvVideoTagParseError};
 
 /// Struct representing legacy flv VIDEODATA.
 /// Check <https://veovera.org/docs/legacy/video-file-format-v10-1-spec.pdf#page=74> for more info.
 #[derive(Debug, Clone)]
 pub struct VideoTag {
```
wkozyra95, line 11: In general I try to follow something like this:
- fields are always fully private or fully pub
- if you want to limit visibility, limit visibility of the struct

I don't see the case where you want to keep struct pub, but want to hide internals.

rtmp/src/client.rs
```diff
@@ -84,44 +84,38 @@ impl RtmpClient {
...
-
                 RtmpMessage::Video {
-                    video: VideoMessage::Enhanced(video),
+                    video: VideoMessage::Data(data),
                     stream_id: self.stream_id,
                 }
             }
+            RtmpEvent::VideoConfig(config) => {
+                if config.codec != RtmpVideoCodec::H264 && !self.state.peer_supports_enhanced {
+                    return Err(RtmpMessageSerializeError::InternalError(
```
wkozyra95: this is not internal error, this something user did wrong, on server this might be considered internal error, but not on client.

rtmp/src/error.rs
```diff
@@ -76,6 +74,12 @@ pub enum RtmpMessageSerializeError {
     #[error("Error encoding amf0: {0}")]
     Amf0Encoding(#[from] AmfEncodingError),
 
+    #[error(transparent)]
+    AudioCodecConversion(#[from] crate::AudioCodecConversionError),
```
wkozyra95: import, drop `crate::`

rtmp/src/events.rs
```diff
@@ -3,118 +3,193 @@ use std::{collections::HashMap, time::Duration};
...
+
+    fn try_from(codec: RtmpVideoCodec) -> Result<Self, Self::Error> {
+        match codec {
+            RtmpVideoCodec::H264 => Ok(LegacyFlvVideoCodec::H264),
+            _ => Err(VideoCodecConversionError::UnsupportedLegacy(codec)),
+        }
+    }
+}
+
+impl TryFrom<RtmpVideoCodec> for ExVideoFourCc {
```
wkozyra95: if conversion is between public and private type I would keep it next to private one

wkozyra95 (approved):

rtmp/src/flv/video.rs
wkazmierczak: yeah, thats a leftover I missed

PR #1924, opened 2026-04-22
Always pick older frame on video input (instead of closest)

wkozyra95:
This will stop freeze frame at the start when the side channel delay is enabled

### DIFF
--- a/smelter-core/src/queue/video_input.rs
+++ b/smelter-core/src/queue/video_input.rs
@@ -98,10 +98,6 @@ impl VideoQueueInput {
             // Partially duplicate get_frame logic, we can't call it directly
             // because we don't want to tiger eos event.
             let offset = self.resolve_offset(pts, queue_start_pts)?;
-            // if pts is before offset we don't want to return it yet
-            if pts < offset {
-                return None;
-            }
             self.receiver.get_for_pts(pts.saturating_sub(offset))
         });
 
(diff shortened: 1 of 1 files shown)


smelter-core/src/queue/video_input.rs
```diff
@@ -98,10 +98,6 @@ impl VideoQueueInput {
             // Partially duplicate get_frame logic, we can't call it directly
             // because we don't want to tiger eos event.
             let offset = self.resolve_offset(pts, queue_start_pts)?;
-            // if pts is before offset we don't want to return it yet
```
wkozyra95, line 101: I think this is not necessary now, because get_frame will return None before offset, because pts of input start from zero in the input frame of reference, which means it start from offset in queue frame of reference

smelter-core/src/queue/video_input.rs
```diff
@@ -284,8 +284,14 @@ impl VideoInputReceiver {
...
 
+    /// Receiver is ready for pts if:
+    /// - it already finished
+    /// - it has any frame in buffer that is newer than pts
+    ///
+    /// If first pts is newer it is still considered ready, but get_for_pts
+    /// will not return that frame for that pts
     fn is_ready_for_pts(&mut self, pts: Duration) -> bool {
-        if self.state() == ReceiverState::Done || self.disconnected {
+        if self.disconnected {
```
wkozyra95, line 294: Done implies disconnected

smelter-core/src/queue/video_input.rs
```diff
@@ -297,27 +303,23 @@ impl VideoInputReceiver {
...
+    /// After this call, only the front frame of `self.buffer` might be older than `pts`;
+    /// all remaining frames are newer.
     fn prepare_for_pts(&mut self, pts: Duration) {
         loop {
             self.try_enqueue();
-            if self.buffer.is_empty() {
-                return;
+            let mut dropped = false;
+
+            // if second element is older than pts then we can drop firs one
```
wkozyra95, line 313:
```suggestion
            // if second element is older than pts then we can drop the first one
```

PR #1926, opened 2026-04-22
Working POC of `srt` input with `mpegts` crate

JBRS307:
So we don't have to use FFmpeg demuxer


wkozyra95:
## `mpegts` crate audit

Audited all 9 source files against ISO/IEC 13818-1:2023 / ITU-T H.222.0 v9.

### Overall

Sound POC. Bit-level parsing of TS header, PSI sections, PES header, and PTS/DTS is correct. A handful of real defects (one panic risk) and a few design gaps worth noting before this grows beyond POC status.

### Verified correct

- `packet.rs`: TS header decode (sync, TEI/PUSI/priority, 13-bit PID, scrambling, AFC, CC) matches §2.4.3.2 / Table 2-2. AFC bit decode correct; AFC=`00` and AFC=`10` both end up with `has_payload=false` which produces the right skip behavior.
- `pes.rs`: 33-bit PTS/DTS decode is correct — verified bit ranges `[32..30]/[29..22]/[21..15]/[14..7]/[6..0]` against Table 2-21. The list of stream_ids that skip the extended header (`BC/BE/BF/F0/F1/F2/F8/FF`) matches the "normal stream_id" carveout in §2.4.3.7.
- `psi/mod.rs`: `pointer_field` handling (§2.4.4.2) is right; `3 + section_length` total per Table 2-30/2-33.
- `psi/pat.rs` / `psi/pmt.rs`: section offsets and length gates match Table 2-30 and Table 2-33; `PID & 0x1FFF`, `program_info_length & 0x0FFF`, `ES_info_length & 0x0FFF` masks applied.
- `stream_type.rs`: Table 2-34 values correct; 0x81/0x87 for AC-3/E-AC-3 are user-private but conventional (ATSC A/52).

### Defects

**1. `pes.rs:61-65` — potential panic on malformed PES header**
`read_timestamp(&buf[9..14])` and `&buf[14..19]` slice unconditionally once `pts_dts_flags` is set. The only length guard is `buf.len() < payload_offset`, where `payload_offset = 9 + header_data_length`. If a sender declares `pts_dts_flags = 10` but `header_data_length = 3`, `payload_offset = 12`, the length check passes, and slicing `buf[9..14]` panics. Same failure mode for PTS+DTS when `header_data_length < 10`. Fix: explicitly require `header_data_length >= 5` (PTS) / `>= 10` (PTS+DTS) before slicing, or use `.get(9..14)` + `ok_or(Error::PesTooShort)`.

**2. `demuxer.rs:147-161` — PMT version not tracked**
`streams.clear()` + `ProgramDiscovered` is emitted on every PMT section. PMTs are periodic (hundreds of ms cadence), so this produces a flood of duplicate events. `version_number` / `current_next_indicator` exist precisely to avoid this — track the last parsed version per `program_number` and short-circuit unchanged sections.

**3. `demuxer.rs:147` — stale assemblers linger across PMT changes**
When a PMT update drops a PID, `streams` is cleared and rebuilt, but `pes_assemblers` keeps entries for removed PIDs. They'll be finalized on `flush()` but never emitted since `streams.get(&pid)` returns `None` → silent drop. Either drop assemblers whose PID is no longer in `streams`, or surface a warning when `finalize_pes` can't resolve the stream type.

**4. `demuxer.rs:99-115` — weak resync**
`process_buffer` advances by 1 byte when the current byte isn't `0x47`, but does not verify alignment (that `buffer[cursor + 188]` is also `0x47`) before parsing. On a stream with a genuine 0x47 in payload, the parser will parse ~188 bytes of garbage, `TsPacket::parse` usually rejects it, but only after one packet is consumed. Industry practice requires ≥3 consecutive `0x47`s at 188-byte stride before accepting the lock.

### Minor / observational

- `packet.rs:58`: when `af_len == 0`, `adaptation` is returned as an empty slice. Per §2.4.3.4 this means "one stuffing byte (the length byte itself)"; flags don't exist at length=0. Current behavior is fine, worth a note.
- `psi/*.rs`: CRC_32 (Annex A) is not validated. Acceptable for a lenient demuxer, but call it out — corrupt PSI will silently produce wrong stream lists.
- `pes.rs`: marker bits (`1`) between timestamp groups are not verified. Acceptable.
- `demuxer.rs:268`: ADTS `ticks_per_frame = (1024 * blocks * 90000) / sample_rate` uses integer division. For 44.1/22.05/11.025 kHz this drifts (~0.01 tick per frame). Fine for short PES but accumulates over long captures; accumulate sample counts instead of multiplying `frame_index` if drift matters.
- `demuxer.rs:134`: PAT picks "first program with `program_number != 0`". Multi-program TS have other programs ignored silently — document this or surface all PMT PIDs.
- `stream_type.rs`: missing common entries (0x06 private/AC-3-in-DVB, 0x15 metadata, 0x20 H.264 SVC, 0x1C/0x1D/0x1E MPEG-4 audio variants). Add as needed.
- No continuity-counter tracking → packet loss is invisible. Add CC mismatch detection as a diagnostic counter.

### Not implemented (flag for follow-up)

- PCR parsing (§2.4.2.2). Demuxer emits PTS/DTS in 90 kHz but has no PCR to anchor it — downstream can't compute arrival time or schedule jitter buffers per the spec timing model.
- CAT (0x0001), TSDT (0x0002), and descriptor parsing (registration, AC-3, HEVC). Non-blocking for generic H.264/AAC SRT ingest.

PR #1929, opened 2026-04-23
Fix failing socket binding for long file paths

wkozyra95:
move dir for macos, handle a gracefully a case where path is still to long

### DIFF
--- a/smelter-core/src/pipeline/instance.rs
+++ b/smelter-core/src/pipeline/instance.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::HashMap,
+    path::Path,
     sync::{Arc, Mutex, Weak},
     thread,
     time::Duration,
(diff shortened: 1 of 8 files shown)


smelter-core/src/pipeline/instance.rs
```diff
@@ -564,6 +565,10 @@ fn create_pipeline(opts: PipelineOptions) -> Result<Pipeline, InitPipelineError>
         .into();
     std::fs::create_dir_all(&download_dir).map_err(InitPipelineError::CreateDownloadDir)?;
 
+    if let Some(dir) = opts.side_channel_socket_dir.as_deref() {
```
wkozyra95, line 568: I'm creating it here because config is sometimes called multiple times, so it creates the dir multiple times

smelter-core/src/queue/side_channel/mod.rs
```diff
@@ -23,14 +23,19 @@ pub struct VideoSideChannel {
 }
 
 impl VideoSideChannel {
-    pub fn new(ctx: &Arc<PipelineCtx>, input_ref: &Ref<InputId>, socket_dir: &Path) -> Self {
+    pub fn new(
```
wkozyra95, line 26: I considered returning an error, but this is such a weird edge case that i think it is enough to just make sure it does not panic or crush input.

JBRS307 (approved):

PR #1934, opened 2026-04-23
[vk-video] Use Box in H264 & H265 parameters

krakow10:
I noticed this is manually converting to and from NonNull with unsafe, why not just directly use Box?

### DIFF
--- a/vk-video/src/codec/h265/parameters.rs
+++ b/vk-video/src/codec/h265/parameters.rs
@@ -4,8 +4,8 @@ use crate::{VulkanDecoderError, codec::h265::H265Codec, vulkan_encoder::FullEnco
 
 pub(crate) struct VkH265VideoParameterSet {
     pub(crate) vps: vk::native::StdVideoH265VideoParameterSet,
-    _profile_tier_level: Option<Box<vk::native::StdVideoH265ProfileTierLevel>>,
-    _dec_pic_buf_mgr: Option<Box<vk::native::StdVideoH265DecPicBufMgr>>,
+    _profile_tier_level: Box<vk::native::StdVideoH265ProfileTierLevel>,
+    _dec_pic_buf_mgr: Box<vk::native::StdVideoH265DecPicBufMgr>,
 }
 
 fn profile_tier_level(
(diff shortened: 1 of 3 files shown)


noituri (approved):
Thanks for noticing this!

vk-video/src/codec/h265/parameters.rs
```diff
@@ -1,14 +1,12 @@
-use std::ptr::NonNull;
-
 use ash::vk;
 
 use crate::{VulkanDecoderError, codec::h265::H265Codec, vulkan_encoder::FullEncoderParameters};
 
+#[expect(unused)]
```
jerzywilczek: Can you rename the fields so that they begin with an underscore? I mean `profile_tier_level` -> `_profile_tier_level`? I would prefer that instead of `#[expect(unused)]`. And I mean in the whole PR, not just this instance.

jerzywilczek (changes requested):
Wow, thanks for noticing.

the `NonNull` gymnastics are necessary in some places, and I think we just assumed linking any two `ash` structs together required this. A nice catch, makes the code cleaner.

Could you rebase to the current `master`, look at `vk-video/CHANGELOG.md` and add your change, as well as the PR link and your username to the `Bug Fixes` section?

vk-video/src/codec/h265/parameters.rs
krakow10: Done

krakow10:
Rebased and added changelog

jerzywilczek:
@krakow10 its a bit late for me right now, I'll take a look on monday

vk-video/src/codec/h265/parameters.rs
```diff
@@ -1,14 +1,11 @@
...
 
 use crate::{VulkanDecoderError, codec::h265::H265Codec, vulkan_encoder::FullEncoderParameters};
 
 pub(crate) struct VkH265VideoParameterSet {
     pub(crate) vps: vk::native::StdVideoH265VideoParameterSet,
-
-    profile_tier_level: Option<NonNull<vk::native::StdVideoH265ProfileTierLevel>>,
-    dec_pic_buf_mgr: Option<NonNull<vk::native::StdVideoH265DecPicBufMgr>>,
+    _profile_tier_level: Option<Box<vk::native::StdVideoH265ProfileTierLevel>>,
+    _dec_pic_buf_mgr: Option<Box<vk::native::StdVideoH265DecPicBufMgr>>,
```
jerzywilczek: Sorry for not noticing this before, I think these aren't optional according to the spec, and were only optional because I at some point thought they were. Since we're touching them anyway, could we make them mandatory?

jerzywilczek (changes requested):

vk-video/src/codec/h265/parameters.rs
krakow10: Done

jerzywilczek (approved):

PR #1937, opened 2026-04-24
[rtmp] E-RTMP audio parsing

wkazmierczak:


rtmp/src/message/audio.rs
```diff
@@ -58,8 +56,37 @@ impl AudioMessage {
...
+                track_id: TrackId::PRIMARY,
+                codec,
+                data,
+            }),
+            ExAudioPacket::CodedFrames(data) => Self::Data(AudioData {
+                track_id: TrackId::PRIMARY,
+                codec,
+                pts,
+                // ExAudio coded frames do not carry channel flags in the packet header.
+                channels: AudioChannels::Stereo,
```
wkazmierczak, line 85: In E-RTMP `channels` need to be statefull, I'll address it in following PR.

wkozyra95 (approved):

PR #1941, opened 2026-04-27
[integration-test] RTP compare tool

wkozyra95:
audio
20260428_11h12m22s_grim

video
20260428_12h35m21s_grim

### DIFF
--- a/integration-tests/src/common.rs
+++ b/integration-tests/src/common.rs
@@ -94,6 +94,24 @@ pub fn save_failed_test_dumps<P: AsRef<Path>>(
     fs::write(path.join(format!("actual_dump_{file_name}")), actual_dump).unwrap();
 }
 
+/// Write only the `actual` dump to the failed-snapshots dir. Used
+/// when the expected snapshot is missing entirely (so there's nothing
+/// to pair it with) or in any other path that produced an `actual`
+/// without a corresponding `expected` to diff against.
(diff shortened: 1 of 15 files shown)


wkozyra95:
I'm merging, but if you have any feedback I'll address it separatlly

integration-tests/src/common.rs
```diff
@@ -94,6 +94,24 @@ pub fn save_failed_test_dumps<P: AsRef<Path>>(
...
+/// to pair it with) or in any other path that produced an `actual`
+/// without a corresponding `expected` to diff against.
+pub fn save_failed_actual_dump<P: AsRef<Path>>(actual_dump: &Bytes, snapshot_filename: P) {
+    let path = failed_snapshots_dir_path();
+    let _ = fs::create_dir_all(&path);
+    let file_name = snapshot_filename
+        .as_ref()
+        .file_name()
+        .unwrap()
+        .to_string_lossy();
```
JBRS307, line 108:
```suggestion
    let file_name = snapshot_filename
        .as_ref()
        .file_name()
        .unwrap()
        .to_str()
        .unwrap();
```

It should never happen, but I think this way is better, because if it SOMEHOW fails then we will know instead of just getting weird name

JBRS307 (commented):
LGTM

PR #1942, opened 2026-04-28
[vk-video][gpu-video] Rename vk-video to gpu-video

jerzywilczek:


noituri (approved):
You changed `vk-video` feature name to `gpu-video` in smelter. I don't know if we consider it a breaking change in smelter but maybe it should be included in the changelog? Also maybe we should mention the rename in our changelog as well?

wkozyra95:
> I don't know if we consider it a breaking change in smelter but maybe it should be included in the changelog? 

CHANGELOG for smelter is maintained from the server perspective, not as a library, so I wouldn't consider it a breaking change, no strong opinion though

wkozyra95 (approved):

PR #1947, opened 2026-04-29
[ts-sdk] Export missing types from smelter-core

wkozyra95:


ts/smelter-core/src/index.ts
```diff
@@ -1,8 +1,15 @@
 import * as Output from './api/output';
 import * as Input from './api/input';
+import * as Image from './api/image';
```
AHGIJMKLKKZNPJKQR, line 4: **suggestion:** Consider re-exporting `Api` from `@swmansion/smelter` as well, because it's used by `ApiClient`

AHGIJMKLKKZNPJKQR (approved):

ts/smelter-core/src/index.ts
wkozyra95: I intentionally skipped it. I feel like it should be imported from the smelter, but I also don't have anything agains adding it, so I added it.

PR #1950, opened 2026-04-29
`SRT` poc

JBRS307:
POC of `srt` input and output, with encryption included.

---

Input:
```ts
type SrtInput = {
  type: "srt";
  video?: InputSrtVideoOptions;
  audio?: boolean;
  required?: boolean;
  offset_ms?: number;
  side_channel?: SideChannel;
  encryption?: {
    passphrase: string;
    encryption: "aes128" | "aes192" | "aes256";
  };
};
```
Output:
```ts
type SrtOutput = {
  type: "srt";
  video?: OutputSrtVideoOptions;
  audio?: OutputSrtAudioOptions;
  encryption?: {
    passphrase: string;
    encryption: "aes128" | "aes192" | "aes256";
  };
};
```

Use the interactive demo for simple showcase, it will print `ffmpeg` and `ffplay` commands to send and receive the `srt` stream. These commands will not work on basic *Homebrew* build as it does not have `srt` enabled.


JBRS307:
@wkozyra95 I think it is completed for now

PR #1953, opened 2026-04-29
Proper handling of single track streams for `WHIP` input.

JBRS307:
Handled by checking what codecs are present in the `OfferCodecs` struct.


wkozyra95 (commented):
What if the answer did not include those codecs? What if even the answer had that track, but the sender was not able to send it e.g. if we send h264 with level 3.1 and sender needs to send a higher resolution.

Even if we know that all SDP are exactly correct, there are probably other reasons why on_track wouldn't be called. This approach minimizes the number of cases where this would happen, but it does not fix it.

The same problem exists for WHEP input

PR #1954, opened 2026-04-29
Fix H264 WHIP stream `smelter-editor` -> `smelter`

wkazmierczak:
I am not entirely sure if this is the correct approach, but it fixes a bug that causes an invisible H264 stream when streaming from `smelter-editor` to `smelter`.

I guess that to do it properly it is necessary to dive into our implementation of AUSplitter.

Fixed in #1957


wkozyra95 (changes requested):
I don't think this is acceptable fix, it needs to be debuged. If the issue is here:
- try before last change Mikołaj did, he mentioned that something related to this code landed recentlly
- if issue was existing before, try modifying smelter do dump h264 to file, so Jerzy or Mikołaj can investigate it
- one other option is to dump rtp (tcp) format, so it can be reproduced with integration test for easier debuging

PR #1961, opened 2026-05-04
Use FEC only when there is a gap in sequence numbers

wkozyra95:


smelter-core/src/pipeline/decoder/libopus.rs
```diff
@@ -82,42 +95,6 @@ impl OpusDecoder {
...
-
-    fn calculate_fec_buf_size(&self, stream_gap: Duration) -> usize {
-        // 120 samples is 2.5 ms with 48kHz sample rate. For FEC it is mandatory that buffer size
-        // is a multiple of 2.5 ms and of the same size (or at least as close as possible) to the
-        // size of lost chunks.
-        let lost_samples = stream_gap.as_secs_f64() * self.decoded_sample_rate as f64;
-        let fec_buf_size = 120 * (lost_samples / 120.0f64).round() as usize;
-
-        2 * fec_buf_size // Multiplication by the number of channels
-    }
```
JBRS307, line 119: The output buf size for FEC must be a multiple of 2.5ms, as close as possible to the actual number of lost samples. This ensured that, now it is not ensured, and fec decode will not work if it is not equal.

PR #1963, opened 2026-05-05
[rtmp] Make audio channels stateful

wkazmierczak:


rtmp/src/protocol/message_stream.rs
```diff
@@ -60,20 +56,46 @@ impl RtmpMessageStream {
     pub fn write_msg(&mut self, msg: RtmpMessage) -> Result<(), RtmpStreamError> {
         self.writer.write_msg(&mut self.stream, msg)
     }
+
+    /// Updates the peer-advertised ack window after parsing a `WindowAckSize`.
```
wkozyra95: I understand saying that we set that based on message x, but saying that we set it after parsing is weird. It's like saying, "To fix that bug we need to open a code editor."

rtmp/src/message/audio.rs
```diff
@@ -6,10 +6,41 @@ use crate::{
     AacAudioConfig, AudioChannels, AudioConfig, AudioData, AudioTag, AudioTagAacPacketType,
     AudioTagSampleSize, AudioTagSoundRate, ExAudioPacket, ExAudioTag, FlvAudioData,
     LegacyFlvAudioCodec, RtmpAudioCodec, RtmpMessageParseError, RtmpMessageSerializeError, TrackId,
-    message::AUDIO_CHUNK_STREAM_ID,
+    message::{AUDIO_CHUNK_STREAM_ID, AudioTrackState, RtmpStreamState, TrackKey},
     protocol::{MessageType, RawMessage},
 };
 
+#[derive(Clone, Copy)]
+struct AudioPacketContext {
```
wkozyra95: This seems like state, I wouldn't keep it here, unless it literally stored inside parser state, but I don't think this is a case

smelter-core/src/pipeline/rtmp/rtmp_output.rs
```diff
@@ -128,6 +128,10 @@ impl RtmpClientOutput {
                 track_id: TrackId::PRIMARY,
                 codec: RtmpAudioCodec::Aac,
                 data: config.extradata.clone(),
+                channels: match config.channels {
```
wkozyra95, line 131: it might be needed for opus, but if not conside removing it and parsing asc yourself

wkozyra95 (approved):

smelter-core/src/pipeline/rtmp/rtmp_output.rs
wkazmierczak: It'll be needed for opus

PR #1964, opened 2026-05-06
Resize jitter-buffer at a (per sec rate instead per packet)

wkozyra95:
Jitter buffer algo:
- starts with 600ms
- on each package incoming from the network (the inner struct shared between tracks)
  - decide trend based on effective buffer (queue_sync_point.elapsed() - packet_pts)
  - Below 80ms - increase buffer immediately by one second (thrtled to trigger at most every 3 seconds)
  - Below 160ms - select trend GrowFast
  - Below 240ms - select trend Grow
  - Above 320ms - and all packets in the past 10 seconds were above then select Shrink
  - Above 2000ms -  and all packets in the past 10 seconds were above, then select ShrinkFast
  - If packet pts is larger than max pts, then modify buffer based on trend. It can only be modified as a percentage of the distance from last, e.g. if GrowFast then increase by diff * 0.03
- Per-track wrapper
  - There might be cases where audio does not increase, but video does, or the other way around, so it could cause a situation where the video increased buffer and then we got an audio packet that would jump by a large value. That is why there is an additional per-track layer that is smoothing it out, so the single step is not larger than 4%.

When popping a packet from the buffer
- shared struct stores target size
- per-track struct stores actual size (and tries to converge on target)
- actual pts from per-track struct are used when modifying popped packet pts
- When calculating next_pts to decide whether the packet needs to be popped, we use `min(target_size, real_size)`. I think it is a good trade-off, that at worst it is causing pop a bit earlier

### DIFF
--- a/smelter-core/src/pipeline/rtp/rtp_input/jitter_buffer.rs
+++ b/smelter-core/src/pipeline/rtp/rtp_input/jitter_buffer.rs
@@ -24,11 +24,11 @@ struct JitterBufferPacket {
 
 #[derive(Debug, Clone, Copy)]
 pub enum RtpJitterBufferMode {
-    /// If I receive packet for sample `x`, then don't wait
-    /// for missing packets older than `(x-size)`
+    /// Release packets once the PTS span of buffered packets exceeds the window.
+    /// Wall clock is not consulted.
     FixedWindow(Duration),
-    /// Packet needs to be returned from buffer before instant.elapsed() gets bigger
-    /// than pts. (with some extra fixed buffer e.g. 80ms)
+    /// Release packets when their output PTS (PTS + dynamic buffer size) is about
+    /// to be reached on the wall clock, with `MIN_DECODE_TIME` of slack.
     RealTime,
 }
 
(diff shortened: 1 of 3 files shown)


smelter-core/src/pipeline/rtp/rtp_input/jitter_buffer.rs
```diff
@@ -93,30 +93,39 @@ impl RtpJitterBufferSharedContext {
...
+    /// Next expected sequence number (last returned from `read_packet` + 1).
     next_seq_num: Option<u64>,
     on_stats_event: Box<dyn FnMut(RtpJitterBufferStatsEvent) + 'static + Send>,
 }
 
 /// We are assuming here that it is enough time to decode. Might be
 /// problematic in case of B-frames, because it would require processing multiple
 /// frames before
-const MIN_DECODE_TIME: Duration = Duration::from_millis(30);
+const MIN_DECODE_TIME: Duration = Duration::from_millis(80);
```
wkozyra95, line 110: It needs to be at least 40ms, otherwise fec even if it triggers it would reach the queue too late

JBRS307 (approved):

PR #1967, opened 2026-05-07
`WHIP` input required handled.

JBRS307:
This solution works by calling the `.queue_new_track()` after the `.on_track()`. It utilizes shared state to check which tracks (audio or video) actually send data and add only them into the queue.
Queue is started if **one** of the following happens:
- 2 seconds pass
- `.on_track()` handle is called for both audio and video.


JBRS307:
The other approach goes in

PR #1977, opened 2026-05-11
[integration-tests] Move tests to GPU VM

noituri:
Tests are run as a separate job which runs on a larger github machine that has access to nvidia gpu.
The job will only run after `build_linux` and `lint` jobs.

There was also a problem with segfaults and deadlocks in nvidia's driver, so I added sleep at the end so that the threads have a chance to finish.

Snapshots: https://github.com/smelter-labs/smelter-snapshot-tests/pull/84

Closes: https://github.com/software-mansion/smelter/issues/1864

### DIFF
--- a/.github/workflows/rust_check.yml
+++ b/.github/workflows/rust_check.yml
@@ -11,7 +11,7 @@ concurrency:
     cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
 
 jobs:
-    build_and_test_linux:
+    build_linux:
         runs-on: ubuntu-24.04
         steps:
             - name: 🛠 Install system dependencies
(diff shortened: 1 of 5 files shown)


.github/workflows/rust_check.yml
```diff
@@ -155,3 +126,61 @@ jobs:
             - name: 📄 Generate JSON schema
               run: |
                   cargo run -p tools --bin generate_from_types -- --check
+
+    test_linux:
+        needs: [build_linux, lint]
```
wkozyra95: if we don't run tests then maybe lint could be part of the build now, at this point it just builds, e.g. build_maco already runs clippy so it would be even more consistent

wkozyra95 (approved):

wkazmierczak (approved):

PR #1983, opened 2026-05-12
[gpu-video] Expose the H265 encoder in the API

jerzywilczek:
Also refactor docs and the readme.


noituri (approved):
I also tested it on amd and everything works

PR #1988, opened 2026-05-13
Add E-RTMP support to Smelter

wkazmierczak:
Currently on H264, AAC output we allow only legacy RTMP.

What worked(in every below case legacy H264+AAC works fine):
- RTMP input
  - ffmpeg (VP9, Opus, forced enhanced AAC(if ffmpeg command worked as it should))
  - OBS (VP9, forced enhanced AAC(this is based on ffmpeg so same case as above))
 
- RTMP output
  - Twitch (forced enhanced H264) ? forcing enhanced AAC inaudible, Opus inaudible
  - YouTube (Opus) ? forcing enhanced H264 and AAC doesn't work
  - MediaMTX (VP9, Opus, forced enhanced H264, forced enhanced AAC)

### DIFF
--- a/rtmp/src/client/negotiation.rs
+++ b/rtmp/src/client/negotiation.rs
@@ -16,8 +16,8 @@ const CONNECT_TRANSACTION_ID: u32 = 1;
 const CREATE_STREAM_TRANSACTION_ID: u32 = 2;
 
 use crate::{
-    CAPS_EX_MODEX, CAPS_EX_RECONNECT, CAPS_EX_TIMESTAMP_NANO, FOURCC_INFO_CAN_DECODE,
-    FOURCC_INFO_CAN_ENCODE, FOURCC_INFO_CAN_FORWARD,
+    CAPS_EX_MODEX, CAPS_EX_RECONNECT, CAPS_EX_TIMESTAMP_NANO, FOURCC_INFO_CAN_ENCODE,
+    FOURCC_INFO_CAN_FORWARD,
 };
 
 use crate::{AUDIO_FOURCC_LIST, VIDEO_FOURCC_LIST};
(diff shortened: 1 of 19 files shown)


rtmp/src/client/negotiation.rs
```diff
@@ -241,29 +234,54 @@ pub(super) fn send_publish(
...
-    // TODO: include audio capability indicators once enhanced audio is implemented.
-    let has_fourcc_list = map
-        .get("fourCcList")
-        .is_some_and(fourcc_list_supports_video);
-    let has_video_info_map = map
-        .get("videoFourCcInfoMap")
-        .is_some_and(video_fourcc_info_map_supports_video);
+fn supports_timestamp_nano_mod_ex(caps_ex_bits: u8) -> bool {
+    caps_ex_bits & (CAPS_EX_MODEX | CAPS_EX_TIMESTAMP_NANO)
+        == CAPS_EX_MODEX | CAPS_EX_TIMESTAMP_NANO
```
wkozyra95:
```suggestion
    let has_modex = caps_ex_bits & CAPS_EX_MODEX != 0;
    let has_ts_nano = caps_ex_bits & CAPS_EX_TIMESTAMP_NANO) != 0;
    has_modex && has_ts_nano
```

rtmp/src/client/negotiation.rs
```diff
@@ -241,29 +234,54 @@ pub(super) fn send_publish(
...
-    has_fourcc_list || has_video_info_map
+fn parse_caps_ex_bits(map: &HashMap<String, AmfValue>) -> u8 {
+    match map.get("capsEx") {
+        Some(AmfValue::Number(bits)) if bits.is_finite() && *bits > 0.0 => {
+            let bits = bits.floor();
+            if bits > u8::MAX as f64 {
+                u8::MAX
+            } else {
+                bits as u8
+            }
```
wkozyra95:
```suggestion
            bits.floor().clamp(0.0, u8::MAX as f64);
```
and you can dro that if guard too

wkozyra95: go through this PR and check for other easy simplification like this, to be clear I don't wan't shorter code, but in this case it's a lot more readable too

smelter-api/src/output/rtmp.rs
```diff
@@ -50,6 +50,31 @@ pub enum RtmpClientVideoEncoderOptions {
         /// Raw FFmpeg encoder options. See [docs](https://ffmpeg.org/ffmpeg-codecs.html) for more.
         ffmpeg_options: Option<HashMap<Arc<str>, Arc<str>>>,
     },
+    #[serde(rename = "ffmpeg_vp8")]
+    FfmpegVp8 {
```
wkozyra95, line 55: Add some comment on the variant that requires ertmp support on the other side

smelter-core/src/pipeline/rtmp/rtmp_input/connection.rs
```diff
@@ -212,22 +226,42 @@ impl RtmpConnectionState {
...
+            }
+            RtmpAudioCodec::Opus => {
+                let options = AudioDecoderThreadOptions {
+                    ctx: self.ctx.clone(),
+                    decoder_options: (),
+                    samples_sender,
+                    input_buffer_size: RTMP_MAX_BUFFER,
+                };
+                AudioDecoderThread::<OpusDecoder>::spawn(input_ref, options)
+                    .map_err(RtmpConnectionError::InitAudioDecoder)?
```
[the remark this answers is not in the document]
wkozyra95: I don't think this is true, but wroth veryfing. I think opus can be decoded to whathever you want, so we can always force stereo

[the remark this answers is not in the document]
wkazmierczak: I've verified it. Even if mono stream arrives, when `Channels::Stereo` is set, the mono input is automatically upmixed to stereo and everything works fine

rtmp/src/message/serialize.rs
```diff
@@ -11,7 +11,10 @@ use crate::{
 };
 
 impl RtmpMessageOutgoing {
-    pub fn into_raw(self) -> Result<RawMessage, RtmpMessageSerializeError> {
+    pub fn into_raw(
+        self,
+        supports_timestamp_nano_mod_ex: bool,
```
wkozyra95: create structs

wkozyra95 (approved):

PR #1993, opened 2026-05-14
`smelter-api` add deserialization tests

wkozyra95:
In the past it was not worth it to maintain seprate test suite for parsing and serializing, and e2e tests handled testing parsing. Now it is a lot easier to maintain those tests, so I plan to remove using json values directlly from most places and use Rust types

### DIFF
--- a/smelter-api/src/common/id.rs
+++ b/smelter-api/src/common/id.rs
@@ -4,17 +4,41 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
-#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema)]
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema, PartialEq)]
 pub struct ComponentId(Arc<str>);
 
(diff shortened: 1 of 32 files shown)


smelter-api/src/common/id.rs
```diff
@@ -4,18 +4,42 @@ use schemars::JsonSchema;
...
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema, PartialEq)]
 pub struct RendererId(Arc<str>);
 
-#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema)]
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema, PartialEq)]
 pub struct OutputId(Arc<str>);
 
-#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema)]
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema, PartialEq)]
 pub struct InputId(Arc<str>);
```
noituri, line 17: tbh this and the other structs could also derive `Eq`.

noituri (approved):

PR #1998, opened 2026-05-18
Add MoQ server input.

JBRS307:
- #2009

Based on the `moq-lite` crate, however adds a few more dependencies for QUIC server and CMAF handling.

Currently supports only `H264` video and `AAC` audio.

---

Yet to be done:
- [x] Properly close all sessions on `/reset`
- [x] Properly close session in input removal
- [x] Sequential catalog handling
- [x] Refactor for new `moq-mux` version

---
What will NOT be done in the experimental:
- Handling of catalog changes (catalog currently read only once per session)
- Handling other containers than CMAF


Cargo.toml
```diff
@@ -78,6 +78,13 @@ tower-http = { version = "0.6.1", features = ["cors"] }
 tracing-subscriber = { version = "0.3.20", features = ["json", "env-filter"] }
 pollster = "0.3.0"
 regex = "1.11.2"
+hang = "0.16"
+moq-lite = "0.16"
+moq-mux = "0.4"
+web-transport-quinn = "0.11"
```
JBRS307: Adding `hang` is required as the `moq-mux` has public functions that return types from it, but does not re-export them.

JBRS307: Additionally, it needs a check which of these deps are needed in workspace and which only in the `smelter-core`

smelter-core/src/pipeline.rs
```diff
@@ -78,6 +80,19 @@ pub struct PipelineOptions {
...
+}
+
+#[derive(Debug)]
+pub enum PipelineMoqServerOptions {
+    Enable {
+        port: u16,
+
+        // XXX: I am not sure if this is the best solution from semantic POV, using `rtmp::TlsConfig` in
+        // MoQ
+        tls_config: Option<TlsConfig>,
```
JBRS307: That is a good question

smelter-core/src/stats/input/mod.rs
JBRS307, line 1: All `stats` related code should be put to another PR, remove it from here.

smelter-api/src/input/moq.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+use super::SideChannel;
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema)]
+#[serde(deny_unknown_fields)]
+pub struct MoqInput {
```
wkozyra95: I think we should use "server" in the input name

smelter-core/src/pipeline.rs
wkozyra95: agree, create separate type

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,210 @@
...
+            port,
+            origin: Origin::random().produce(),
+            inputs: MoqInputsState::default(),
+            tls_config,
+        })
+    }
+}
+
+pub struct MoqServerHandle {
+    tasks: Vec<tokio::task::JoinHandle<()>>,
```
wkozyra95: any reason that this needs to be a dynamic structure?

src/config.rs
```diff
@@ -357,6 +361,36 @@ fn try_read_config() -> Result<Config, String> {
...
+            .map_err(|_| "SMELTER_MOQ_SERVER_PORT has to be valid port number")?,
+        Err(_) => 4443,
+    };
+
+    let moq_enable = match env::var("SMELTER_START_MOQ_SERVER") {
+        Ok(enable) => bool_env_from_str(&enable).unwrap_or(true),
+        Err(_) => true,
+    };
+
+    let moq_tls_cert_file = match env::var("SMELTER_MOQ_TLS_CERT_FILE") {
```
wkozyra95: does the moq server even make sense without certs? If no then you can skip enable option, by default do not enable it unless certs are provided

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+fn start_wt_server(
+    addr: std::net::SocketAddr,
+    chain: &[rustls_pki_types::CertificateDer<'static>],
+    key: &rustls_pki_types::PrivateKeyDer<'static>,
+    rt: Arc<Runtime>,
+) -> Result<web_transport_quinn::Server, InitPipelineError> {
+    // HACK: This is needed because even though `.with_certificate()` is sync it runs async code
+    // underneath and requires a runtime
```
wkozyra95: It's worth preserving the comment that with_certificate requires runtime, but instead of hack, I would just make this function async

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,210 @@
...
+            .with_certificate(chain.to_vec(), key.clone_key())
+        {
+            Ok(server) => return Ok(server),
+            Err(error) => {
+                warn!("Failed to start MoQ server. Retrying ...");
+                debug!(%error, "MoQ server failed to start.");
+                last_error = Some(error);
+            }
+        }
+        thread::sleep(Duration::from_millis(1000));
```
wkozyra95: this could be async sleep, if you make the function async

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+}
+
+pub fn spawn_moq_server(
+    ctx: Arc<PipelineCtx>,
+    state: &Arc<MoqPipelineState>,
+) -> Result<MoqServerHandle, InitPipelineError> {
+    let port = state.port;
+    let addr = std::net::SocketAddr::from(([0, 0, 0, 0, 0, 0, 0, 0], port));
```
wkozyra95: I assume this listens on both ipv4 and ipv6? Maybe some add some comment

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,210 @@
...
+                }
+            }
+            None => {
+                info!(path = %path_str, "MoQ broadcast unannounced");
+            }
+        }
+    }
+}
+
+fn start_wt_server(
```
wkozyra95: I don't know what wt is

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,210 @@
...
+    })
+}
+
+async fn run_accept_loop(mut wt_server: web_transport_quinn::Server, moq_server: moq_lite::Server) {
+    while let Some(request) = wt_server.accept().await {
+        let moq = moq_server.clone();
+        tokio::spawn(async move {
+            let session = match request.ok().await {
+                Ok(session) => {
+                    info!("WebTransport session created.");
```
wkozyra95: Is it correct to refer to web transport? My understanding is that WebTransport is a browser API for QUICK, but if you run outside of browser it does not have anything to do with it, you just use quick then.

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,210 @@
...
+                        warn!(
+                            "MoQ broadcast path not matched: {}",
+                            ErrorStack::new(&err).into_string()
+                        );
+                        continue;
+                    }
+                };
+
+                let ctx_inner = ctx.clone();
+                let inputs_inner = moq_inputs.clone();
```
wkozyra95:
```suggestion
                let moq_inputs = moq_inputs.clone();
```

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,210 @@
...
+                    Err(err) => {
+                        warn!(
+                            "MoQ broadcast path not matched: {}",
+                            ErrorStack::new(&err).into_string()
+                        );
+                        continue;
+                    }
+                };
+
+                let ctx_inner = ctx.clone();
```
wkozyra95:
```suggestion
                let ctx = ctx.clone();
```

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,462 @@
...
+    container: Hang,
+    description: Option<Bytes>,
+}
+
+struct DiscoveredTracks {
+    video: Option<DiscoveredVideo>,
+    audio: Option<DiscoveredAudio>,
+}
+
+pub(crate) fn spawn_broadcast_handler(
```
wkozyra95: refactor

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,462 @@
...
+
+    let audio = match catalog.audio.renditions.first_key_value() {
+        Some((name, config)) if let HangAudioCodec::AAC(_) = config.codec => {
+            match Hang::try_from(&config.container) {
+                Ok(container) => Some(DiscoveredAudio {
+                    name: name.clone(),
+                    container,
+                    description: config.description.clone(),
+                }),
+                Err(error) => {
```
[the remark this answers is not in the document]
JBRS307: Aware of that, this code is not meant to be production ready for now

smelter-core/src/stats/input/mod.rs
```diff
@@ -65,6 +65,7 @@ impl InputStatsState {
             InputProtocolKind::Rtmp => InputStatsState::Rtmp(RtmpInputState::new()),
             InputProtocolKind::Mp4 => InputStatsState::Mp4(Mp4InputState::new()),
             InputProtocolKind::Hls => InputStatsState::Hls(HlsInputState::new()),
+            InputProtocolKind::Moq => unimplemented!(),
             InputProtocolKind::V4l2 => unimplemented!(),
             InputProtocolKind::DeckLink => unimplemented!(),
             InputProtocolKind::RawDataChannel => unimplemented!(),
```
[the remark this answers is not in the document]
JBRS307: It does not emit anything

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -0,0 +1,218 @@
...
+                    .is_some_and(|c| c.starts_with("avc1") || c.starts_with("avc3"))
+        })
+        .and_then(|t| match msf_track_to_hang(t) {
+            Ok((container, description)) => Some(DiscoveredVideo {
+                name: t.name.clone(),
+                container,
+                description,
+            }),
+            Err(reason) => {
+                warn!(track=%t.name, "Skipping MSF video track: {reason}");
```
JBRS307:
```suggestion
            Err(error) => {
                warn!(track=%t.name, error, "Skipping MSF video track.");
```

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -0,0 +1,218 @@
...
+                && t.codec.as_deref().is_some_and(|c| c.starts_with("mp4a"))
+        })
+        .and_then(|t| match msf_track_to_hang(t) {
+            Ok((container, description)) => Some(DiscoveredAudio {
+                name: t.name.clone(),
+                container,
+                description,
+            }),
+            Err(reason) => {
+                warn!(track=%t.name, "Skipping MSF audio track: {reason}");
```
JBRS307:
```suggestion
            Err(error) => {
                warn!(track=%t.name,  error, "Skipping MSF audio track.");
```

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,167 @@
...
+    sessions: MoqSessions,
+}
+
+impl Drop for MoqServer {
+    fn drop(&mut self) {
+        self.accept_task.abort();
+        self.announce_task.abort();
+        let mut sessions = self.sessions.lock().unwrap();
+        for (_, mut session) in sessions.drain() {
+            session.close(moq_lite::Error::Cancel);
```
JBRS307: This is closed on drop, just `.drain()` should work.

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,165 @@
...
+    let mut server_result: Option<Result<moq_native::Server, anyhow::Error>> = None;
+    for _ in 0..5 {
+        match config.clone().init() {
+            Ok(server) => {
+                server_result = Some(Ok(server));
+                break;
+            }
+            Err(error) => {
+                warn!("Failed to start MoQ server. Retrying ...");
+                server_result = Some(Err(error));
```
JBRS307: Add debug log with exact cause

smelter-api/src/input/moq_server.rs
```diff
@@ -0,0 +1,39 @@
...
+use utoipa::ToSchema;
+
+use super::SideChannel;
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema)]
+#[serde(deny_unknown_fields)]
+pub struct MoqInputServer {
+    /// The MoQ broadcast path that Smelter listens on for incoming streams.
+    /// A publishing client must announce this broadcast path for the input to receive media.
+    pub broadcast_path: Arc<str>,
```
JBRS307: This field is needed as it may be set to `some/broadcast/path`. Input with that name would breaks API

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+pub async fn spawn_moq_server(
+    ctx: Arc<PipelineCtx>,
+    state: &Arc<MoqPipelineState>,
+) -> Result<MoqServer, InitPipelineError> {
+    let port = state.port;
+
+    let mut config = ServerConfig::default();
+    config.bind = Some(format!("[::]:{port}"));
```
JBRS307, line 86: This is the only accepted syntax, but I was not able to connect with IPv6 using `moq-cli`, however the error I got was "failed DNS lookup".

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,166 @@
...
+    sessions: MoqSessions,
+}
+
+impl Drop for MoqServer {
+    fn drop(&mut self) {
+        self.accept_task.abort();
+        self.announce_task.abort();
+        let mut sessions = self.sessions.lock().unwrap();
+        sessions.clear();
+    }
```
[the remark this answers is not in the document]
JBRS307: Drop closes the session, the per session task is then stopped afterwards.

integration-tests/examples/demo/inputs/moq.rs
```diff
@@ -0,0 +1,104 @@
...
+            "decoder_map": {
+                "h264": "ffmpeg_h264",
+            }
+        })
+    }
+
+    pub fn on_after_registration(&self) -> Result<()> {
+        println!("Publish to this input using moq-cli:");
+        println!(
+            "ffmpeg -re -readrate_initial_burst 7 -i <SOURCE_MP4> -c copy -movflags cmaf -f mp4 - | moq-cli publish --tls-disable-verify --url https://localhost:4443 --broadcast {} fmp4",
```
wkozyra95: is this increasing or decreasing input burst? Either way option like this only hides whether the initial buffer option is sane. Unless you can expect that are clients will send burst like this?

src/config.rs
```diff
@@ -342,6 +346,45 @@ fn try_read_config() -> Result<Config, String> {
...
     };
 
+    let moq_server_port = match env::var("SMELTER_MOQ_SERVER_PORT") {
+        Ok(moq_port) => moq_port
+            .parse::<u16>()
+            .map_err(|_| "SMELTER_MOQ_SERVER_PORT has to be valid port number")?,
+        Err(_) => 4443,
+    };
+
+    let moq_enable = match env::var("SMELTER_START_MOQ_SERVER") {
```
wkozyra95: Bumping You did not respond to https://github.com/software-mansion/smelter/pull/1998#discussion_r3259837312


I get that self signed cert might make sense for debuging, but if it is the only case then defaults that made sense for webrt and rtmp do not make sense here

wkozyra95: also it's a bit to much code, extract moq stuff to function

smelter-core/Cargo.toml
```diff
@@ -48,6 +48,12 @@ pollster = { workspace = true }
 audioadapter-buffers = "2.0.0"
 audioadapter = "2.0.0"
 async-trait = "0.1.89"
+hang = "0.18"
+moq-msf = "0.2"
+moq-mux = "0.5"
+mp4-atom = "0.11"
```
wkozyra95, line 55: might be worth consider to follow up to unify crates that we are using for mp4 and moq inputs

smelter-core/src/pipeline.rs
```diff
@@ -77,6 +79,16 @@ pub struct PipelineOptions {
     pub webrtc_nat_1to1_ips: Arc<Vec<String>>,
 
     pub rtmp_server: PipelineRtmpServerOptions,
+    pub moq_server: PipelineMoqServerOptions,
+}
+
+#[derive(Debug)]
+pub enum PipelineMoqServerOptions {
```
wkozyra95: nit: not important, but you could preserve the order of fields when defining types

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,166 @@
...
+type MoqSessions = Arc<Mutex<HashMap<u64, Session>>>;
+pub struct MoqServer {
+    accept_task: tokio::task::JoinHandle<()>,
+    announce_task: tokio::task::JoinHandle<()>,
+    sessions: MoqSessions,
+}
+
+impl Drop for MoqServer {
+    fn drop(&mut self) {
+        self.accept_task.abort();
```
wkozyra95, line 75: Are you sure this will work if drop happens in code outside tokio runtime?

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,166 @@
...
+        self.announce_task.abort();
+        let mut sessions = self.sessions.lock().unwrap();
+        sessions.clear();
```
wkozyra95:
```suggestion
        self.sessions.lock().unwrap().clear();
```

I assume this is in case if someone closens sessions object we still want to drop the content?
If we really need that add some comment explainign why?

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,166 @@
...
+    Ok(MoqServer {
+        accept_task,
+        announce_task,
+        sessions: moq_sessions,
+    })
+}
+
+async fn run_accept_loop(mut server: moq_native::Server, moq_sessions: MoqSessions) {
+    while let Some(request) = server.accept().await {
+        let moq_sessions = moq_sessions.clone();
```
wkozyra95: the cleaner pattern would be always create Weak, and try to upgrade it every time you need it. You would not need that hack with cleanup then

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,166 @@
...
+async fn run_announce_loop(
+    mut origin_consumer: OriginConsumer,
+    moq_inputs: MoqInputsState,
+    ctx: Arc<PipelineCtx>,
+) {
+    while let Some((path, broadcast)) = origin_consumer.announced().await {
+        let path_str = path.to_string();
+        match broadcast {
+            Some(broadcast) => {
+                info!(path = %path_str, "MoQ broadcast announced");
```
wkozyra95:
```suggestion
                info!(%path, "MoQ broadcast announced");
```
I think this would already do the same

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,396 @@
...
+    video: &DiscoveredVideo,
+    frame_sender: QueueSender<Frame>,
+) -> Result<DecoderThreadHandle, MoqConnectionError> {
+    // Only CMAF H264 is allowed right now, other codecs are rejected before this function
+
+    let avcc_bytes = video
+        .description
+        .clone()
+        .ok_or(MoqConnectionError::InvalidAvcc)?;
+
```
wkozyra95:
```suggestion
    let avcc_bytes = video
        .description
        .clone()
        .ok_or(MoqConnectionError::InvalidAvcc)?;
```

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,396 @@
...
+    broadcast: BroadcastConsumer,
+) -> Option<tokio::task::JoinHandle<()>> {
+    let queue_input = input.queue_input.upgrade()?;
+
+    let input_ref = input_ref.clone();
+    let decoders = input.decoders.clone();
+    let rt = ctx.tokio_rt.clone();
+
+    let handle = rt.spawn(async move {
+        let input_id_str = input_ref.to_string();
```
wkozyra95: Don't do that, I would expect variables like this only in parsing code. Plus from what I see in all the places where you using you could just use input_id

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,396 @@
...
+    samples_sender: QueueSender<InputAudioSamples>,
+) -> Result<DecoderThreadHandle, MoqConnectionError> {
+    // Only AAC is allowed right now, different codecs are rejected before this function is called
+
+    let asc = audio
+        .description
+        .clone()
+        .ok_or(MoqConnectionError::MissingAsc)?;
+    let aac_decoder_options = AudioDecoderOptions::FdkAac(FdkAacDecoderOptions { asc: Some(asc) });
+
```
wkozyra95:
```suggestion
    let asc = audio
        .description
        .clone()
        .ok_or(MoqConnectionError::MissingAsc)?;
    let aac_decoder_options = AudioDecoderOptions::FdkAac(FdkAacDecoderOptions { asc: Some(asc) });
```

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,396 @@
...
+    decoder_handle: Option<DecoderThreadHandle>,
+    broadcast: &BroadcastConsumer,
+) {
+    if let Some(audio) = audio
+        && let Some(decoder_handle) = decoder_handle
+    {
+        match broadcast.subscribe_track(&Track::new(&audio.name)) {
+            Ok(track) => {
+                let consumer =
+                    ContainerConsumer::new(track, audio.container).with_latency(MOQ_BUFFER);
```
wkozyra95: what does with_latency here mean?

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,396 @@
...
+
+    #[error("Container read error")]
+    ContainerError(#[source] moq_mux::Error),
+}
+
+async fn read_video_track(
+    mut consumer: ContainerConsumer<Container>,
+    decoder_handle: DecoderThreadHandle,
+) -> Result<(), MoqConnectionError> {
+    let mut first_pts: Option<Duration> = None;
```
wkozyra95: What about a/v synchronization, even if moq media are synchronized in protocol, by reseting it this way you are breaking this synchronization

smelter-core/src/pipeline/moq/server_input.rs
wkozyra95, line 1: skip moq in name

smelter-core/src/protocols/moq.rs
```diff
@@ -0,0 +1,43 @@
...
+pub struct MoqServerInputOptions {
+    pub broadcast_path: Arc<str>,
+    pub decoders: MoqServerInputDecoders,
+    pub queue_options: QueueInputOptions,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct MoqServerInputDecoders {
+    pub h264: Option<VideoDecoderOptions>,
+    pub aac: Option<AudioDecoderOptions>,
```
wkozyra95: what is this option suppose to do? I see you are setting it to None, but even if it was supported, are you expecint user to privde asc config?

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,166 @@
...
+    state: &Arc<MoqPipelineState>,
+) -> Result<MoqServer, InitPipelineError> {
+    let port = state.port;
+
+    let mut config = ServerConfig::default();
+    config.bind = Some(format!("[::]:{port}"));
+    config.tls = state.tls_config.clone();
+
+    let mut server_result: Option<Result<moq_native::Server, anyhow::Error>> = None;
+    for _ in 0..5 {
```
wkozyra95: I would extract that, especially that you are currently doing extra sleep after last error and with separate function you could early exit

smelter-api/src/input/moq_server.rs
wkozyra95: Not sure I understand.
- Are you saying that we can't enforce on client what the path is? If so ok.
- Are you saying that `/`(or different special characters) are the problem? If so then it's not the issue we already handle that by url encoding input ids in whip

Argument in favor of current setup is that client might have path test/h264, test/av1 and user provides prefix and we pick optimal format for us. Not relevant for now, but worth to consider to avoid breaking changes latter.

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,396 @@
...
+                return;
+            }
+        };
+
+        let has_video = discovered.video.is_some();
+        let has_audio = discovered.audio.is_some();
+
+        // TODO: This has to be handled in a more reliable way that does not introduce high latency,
+        // probably jitter buffer.
+        let (video_sender, audio_sender) = queue_input.queue_new_track(QueueTrackOptions {
```
wkozyra95: By calling this function here, you are saying that zero pts represents ctx.queue_ctx.effective_last_pts() + buffer
Note that you are not taking effective_last_pts from first packet, if the time between first packet and this call is large enough this might waste buffer space for somethign that is not related to regular jitter

wkozyra95 (changes requested):

src/config.rs
JBRS307: > if it is the only case then defaults that made sense for webrt and rtmp do not make sense here

I don't understand it

smelter-api/src/input/moq_server.rs
JBRS307: https://www.ietf.org/archive/id/draft-ietf-moq-transport-18.html#name-representing-namespace-and-

In that case, I think removing the broadcast path will be just fine as recommended is something else

smelter-core/src/pipeline/moq/server.rs
JBRS307: Neither docs nor other sources do not mention that it is not allowed. Besides, the `JoinHandle` holds low level `RawTask` and operates on that, I think it is going to work properly

smelter-core/Cargo.toml
JBRS307: It is `moq-mux` dependency. I need a type from this crate that is in a field I read, but is not re-exported.

smelter-core/src/pipeline/moq/connection.rs
JBRS307: I added a comment briefly describing what it does.

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,424 @@
...
+    #[error("Decoder channel closed")]
+    ChannelClosed,
+
+    #[error("Container read error")]
+    ContainerError(#[source] moq_mux::Error),
+}
+
+/// Normalizes a raw track timestamp against the first PTS observed across all
+/// tracks of the broadcast, so audio and video share the same zero point.
+fn normalize_pts(first_pts: &Mutex<Option<Duration>>, raw_pts: Duration) -> Duration {
```
JBRS307:
```suggestion
fn normalize_pts(first_pts: &Arc<Mutex<Option<Duration>>>, raw_pts: Duration) -> Duration {
```

smelter-core/src/pipeline/moq/connection.rs
JBRS307: The fixed buffer has been set to 1 second. It will have to be handled differently in the near future

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -0,0 +1,229 @@
...
+                    description,
+                })
+            }
+            _ => {
+                warn!("Only CMAF container is supported.");
+                None
+            }
+        },
+        Some((name, config)) => {
+            warn!(track=%name, codec=%config.codec, "Unsupported audio codec, skipping track");
```
wkozyra95:
```suggestion
            warn!(track, ?codec, "Unsupported audio codec, skipping track");
```
or something like that

Try to keep logs short, do not do stuff like key=%value unless it really provides some benefit. You tend to do stuff like this often I remember commenting on stuff like this in previous PRs. Please keep that in mind for the future

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -0,0 +1,229 @@
...
+        Some((name, config)) => {
+            warn!(track=%name, codec=%config.codec, "Unsupported video codec, skipping track");
+            None
+        }
+        None => None,
+    };
+
+    let audio = match catalog.audio.renditions.first_key_value() {
+        Some((name, config)) if let AudioCodec::AAC(_) = config.codec => match &config.container {
+            CatalogContainer::Cmaf { init, .. } => {
```
wkozyra95: readability asside this very unintuive/inconsitent approach, first you match with match then if let then match again, why?

If anything it should be:

```
        Some((name, config)) => match (&config.container, config.codec) {
```

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -0,0 +1,229 @@
...
+    broadcast: &BroadcastConsumer,
+) -> Result<DiscoveredTracks, MoqCatalogError> {
+    let catalog_track = broadcast
+        .subscribe_track(&moq_net::Track::new(moq_msf::DEFAULT_NAME))
+        .map_err(MoqCatalogError::CatalogSubscribeError)?;
+
+    let mut consumer = MsfConsumer::new(catalog_track);
+
+    let catalog = consumer
+        .next()
```
wkozyra95: In few places you take iterator or correction and only take first element from it. It's hard to judge for me whether it's appropriate at a specific place, but go through all the cases, and unless it literally does not make sense to read second element then add comment why did you decided to only look at the first one

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -0,0 +1,229 @@
...
+    let codec = cmaf.trak().mdia.minf.stbl.stsd.codecs.first().ok_or(
+        MoqCatalogError::CodecConfigExtractionError("CMAF init segment contains no codec entries"),
+    )?;
+
+    let mut buf = Vec::new();
+    match codec {
+        mp4_atom::Codec::Avc1(avc1) => {
+            avc1.avcc.encode_body(&mut buf).map_err(|_| {
+                MoqCatalogError::CodecConfigExtractionError("failed to encode AVCDecoderConfig")
+            })?;
```
wkozyra95, line 186:
```suggestion
            let mut buf = BytesMut::new();
            avc1.avcc.encode_body(&mut buf).map_err(|_| {
                MoqCatalogError::CodecConfigExtractionError("failed to encode AVCDecoderConfig")
            })?;
            return buf.freeze()
```
(not tested, but something like this should work, BytesMut implements BufMut)

Modying by arguments is sometimes necessary, but unless you e.g. reusing the same memory in the loop or something like that, keep it always as contained as possible

smelter-core/src/pipeline/moq/certificate.rs
```diff
@@ -0,0 +1,183 @@
...
+    } else {
+        generate(&cert_path, &key_path)?
+    };
+
+    let fingerprint = fingerprint(&cert_der);
+    warn!(
+        "Using auto-generated, self-signed MoQ TLS certificate. Generated/loaded from {}. \
+         Make sure to configure proper certs in production. Cert SHA-256 (for client serverCertificateHashes): {fingerprint}",
+        dir.display()
+    );
```
wkozyra95:
```suggestion
    let fingerprint = fingerprint(&cert_der);
    warn!(
        "Using auto-generated, self-signed MoQ TLS certificate (sha-256: {fingerprint}, location: {}. \
         Make sure to configure proper certs in production.",
        dir.display()
    );
```
or
```suggestion
    let sha256_fingerprint = fingerprint(&cert_der);
    warn!(
       sha256_fingerprint,
       cert_path,
        "Using auto-generated, self-signed MoQ TLS certificate. Make sure to configure proper certs in production."
    );
```

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,424 @@
...
+};
+
+use bytes::Bytes;
+use moq_mux::catalog::hang::Container;
+use moq_mux::container::Consumer as ContainerConsumer;
+use moq_native::moq_net::{BroadcastConsumer, Error as MoqError, Track};
+use smelter_render::error::ErrorStack;
+use tracing::{info, trace, warn};
+
+use crate::pipeline::moq::connection::catalog::{MoqCatalogError, read_catalog};
```
wkozyra95: cleanup imports
- you are importing pipeline here and in block below
- A lot of types explicitly imported is already in prelude

smelter-core/src/pipeline/moq/connection.rs
wkozyra95: 2 things:
- If you say "we check" I would understood that our implementation is doing something, but this is library option so I assume this is not the case here
- Example is does not clarify most of the cases. I'm not saying it has to cover all, but this calculation does not say much. Latency is smaller than a group, so it is the same as if you set latency to 0
- Comment  provides information, but reaching any conclusion from it takes long time, ti might be quicker to jump to implementation. you don't need to describe exact logic, you should focus more on outcome/intent


I would explain it like that:
```
// .with_latency() defines how long we wait for group. Group dalay is a difference between group start timestamp and highest received timestamp. 
```

and if you really want to have example with numbers
```
// e.g. If latency is 300ms a group with timestamps between 200-400ms is not complete yet, we will wait for that group untill timestamp 500ms (or higher) is received
```

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,424 @@
...
+        });
+
+        if let Some(v) = &discovered.video {
+            info!(input_id = %input_ref, track = %v.name, "Discovered MoQ video track");
+        }
+        if let Some(a) = &discovered.audio {
+            info!(input_id = %input_ref, track = %a.name, "Discovered MoQ audio track");
+        }
+
+        let (video_decoder_handle, audio_decoder_handle) = spawn_decoders(
```
wkozyra95: This seem like very artificial division, you are effectively handling None bunch of times for the same case. It seems to me that better approach would be to have one handle video call/handle audio call

Structure like this could make sense if e.g. you need to spawn decoders, do something common and then do something pre track again.

Also side point, in most cases it is cleaner to handle "null-ability" in top level code
```
if let Some(video) = video {
  handle_video()
}
```
than just handling option in `handle_video`

smelter-core/src/pipeline/moq/connection/catalog.rs
JBRS307: This is being changed in the followup, I want to add Opus as the MoQ streamer tool audio does not work on linux.

wkozyra95: The comment was for syntax. If you add opus, my comment still stands, I don't see why this change would have to be delayed for followup pr

JBRS307: I changed it after second thought in this PR, so it is done

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,190 @@
...
+    while let Some(request) = server.accept().await {
+        let moq_sessions = match weak_sessions.clone().upgrade() {
+            Some(s) => s,
+            None => break,
+        };
+        tokio::spawn(async move {
+            match request.ok().await {
+                Ok(session) => {
+                    info!("MoQ session established");
+                    debug!(moq_version=?session.version());
```
wkozyra95: I don't see the point, just add version to info log

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,190 @@
...
+
+async fn try_start_server(config: ServerConfig) -> Result<moq_native::Server, anyhow::Error> {
+    for _ in 0..4 {
+        match config.clone().init() {
+            Ok(server) => {
+                return Ok(server);
+            }
+            Err(error) => {
+                warn!("Failed to start MoQ server. Retrying ...");
+                debug!(%error, "Failed to start MoQ server.");
```
wkozyra95:
```suggestion
```
Maybe print error with warn! at the last try, but I suspect it's get prined downstream, so just removing fine

smelter-core/src/pipeline/moq/server.rs
wkozyra95: That applies to all things in drop, what about dropping sessions outside of context? The easiest way is to just test that

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,395 @@
...
+        // Shared across audio and video so both tracks are normalized against
+        // the same first PTS, preserving A/V synchronization. Whichever track
+        // produces the first frame sets the common zero point for both.
+        let first_pts = Arc::new(Mutex::new(None));
+
+        let first_pts_inner = first_pts.clone();
+        let video_fut = async {
+            if let (Some(video), Some(frame_sender)) = (video, video_sender) {
+                let decoder_handle =
+                    spawn_video_decoder(&ctx, &input_ref, &decoders, &video, frame_sender);
```
wkozyra95: move spawn_video_decoder to inside run_video_track, also consider creating hard divide between async and sync code, you only have one async method that you call in sync code and all the logic is handled inside async

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,395 @@
...
+                );
+            }
+        }
+        Err(err) => {
+            warn!("Failed to subscribe to MoQ video track: {err}");
+        }
+    }
+}
+
+async fn run_audio_track(
```
wkozyra95, line 301: I would keep this implementation close to read function, they are very coupled, so you need to look at both or split code in a different way e.g.

- run_audio_track returns result
- read_audio_track inline inside run_audio_track (but do not just copy paste it highlly indented)
- handle printing error in calling code

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,361 @@
...
+        };
+
+        tokio::join!(video_fut, audio_fut);
+        info!(input_id = %input_ref, "MoQ broadcast connection closed");
+    });
+
+    Some(handle)
+}
+
+fn process_video_config(
```
wkozyra95: method that spwans decoder should not be called like this

smelter-core/src/pipeline/moq/state.rs
```diff
@@ -0,0 +1,114 @@
...
+use tracing::error;
+
+use crate::queue::WeakQueueInput;
+
+use crate::prelude::*;
+
+#[derive(Clone, Default)]
+pub(crate) struct MoqInputsState(Arc<Mutex<HashMap<Ref<InputId>, MoqInputState>>>);
+
+pub(crate) struct MoqInputState {
```
JBRS307, line 16: Rename that to `MoqServerInputState` so there is no mess later when adding the client input

smelter-core/src/pipeline/moq/server.rs
JBRS307: It occupies the IPv6 socket as well

smelter-core/src/pipeline/moq/state.rs
```diff
@@ -0,0 +1,110 @@
...
+
+impl MoqInputState {
+    fn new(options: MoqInputStateOptions) -> Self {
+        Self {
+            queue_input: options.queue_input,
+            decoders: options.decoders,
+            should_close: Arc::new(false.into()),
+            connection_handle: None,
+            session: None,
+        }
```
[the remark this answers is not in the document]
JBRS307: You are wrong

smelter-core/src/pipeline/moq/certificate.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    let mut file = fs::OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .mode(0o600)
+        .open(key_path)?;
+    file.write_all(pem.as_bytes())?;
+    Ok(())
+}
```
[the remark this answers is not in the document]
JBRS307: Only linux and macos supported

smelter-core/src/pipeline/moq/server.rs
JBRS307: This is the only `.abort()` that is left now. It is called only on shutdown or reset, when we destroy the current state anyway.

src/config.rs
wkozyra95: Maybe enabling server by default does not make sense for moq, maybe it should be it should be disabled by default if certs are missing?

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -0,0 +1,218 @@
...
+async fn read_hang_catalog(
+    broadcast: &BroadcastConsumer,
+) -> Result<DiscoveredTracks, MoqCatalogError> {
+    let catalog_track = broadcast
+        .subscribe_track(&hang::Catalog::default_track())
+        .map_err(MoqCatalogError::CatalogSubscribeError)?;
+
+    let mut consumer = HangConsumer::new(catalog_track);
+
+    // The `.next()` method of the catalog consumer yields the next available catalog update.
```
wkozyra95: ok now I see what it's doing, but if that is the case then this is bad naming, if you know that your have iterator over catalog updates, then calling something consumer is like calling variable `value`. This section can be written a lot cleaner without need for extra comments

smelter-core/src/pipeline/moq/certificate.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    Ok(cert_der.as_ref().to_vec())
+}
+
+/// Generate a fresh self-signed certificate and persist it to disk.
+///
+/// Mirrors moq-native's own generator: SANs for `localhost`/`127.0.0.1`/`::1` and a
+/// 14-day validity window starting yesterday (WebTransport caps self-signed certs at
+/// two weeks, and the back-dated start tolerates clock drift).
+fn generate(cert_path: &Path, key_path: &Path) -> Result<Vec<u8>, SelfSignedTlsError> {
```
wkozyra95: misleading name, function like that should only return certs not write to disc, also It would be good to split it the logic, generate and validation don't need to be coupled with reading and writing to disc.

This is very localized and does not affect the rest of the code base, so I don't care about it much, but when you thinking about ways to split code you need to consider more carefully what is a better division point

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,396 @@
...
+    queue_input: QueueInput,
+    broadcast: BroadcastConsumer,
+    should_close: Arc<AtomicBool>,
+) {
+    info!(input_id = %input_ref, "MoQ broadcast connection established");
+
+    let mut discovered = match read_catalog(&broadcast).await {
+        Ok(d) => d,
+        Err(err) => {
+            warn!(
```
wkozyra95: return error

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,396 @@
...
+    let first_pts = Arc::new(Mutex::new(None));
+    let rt = ctx.tokio_rt.clone();
+
+    let ctx_inner = ctx.clone();
+    let input_ref_inner = input_ref.clone();
+    let decoders_inner = decoders.clone();
+    let broadcast_inner = broadcast.clone();
+    let first_pts_inner = first_pts.clone();
+    let should_close_inner = should_close.clone();
+    let video_task = rt.spawn(async move {
```
wkozyra95: this is not great, needs refactor

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,205 @@
...
+    session: Session,
+    rt: Arc<tokio::runtime::Runtime>,
+}
+
+impl MoqSession {
+    fn new(session: Session, rt: Arc<tokio::runtime::Runtime>) -> Self {
+        Self { session, rt }
+    }
+
+    fn session(&self) -> &Session {
```
wkozyra95: I think Deref trait is for that (or something similiar)

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,205 @@
...
+    moq_inputs: MoqServerState,
+    ctx: Arc<PipelineCtx>,
+) {
+    while let Some(request) = server.accept().await {
+        let Some(url) = request.url() else {
+            if let Err(error) = request.close(400).await {
+                warn!(%error, "Error while rejecting MoQ connection.");
+            }
+            warn!("Rejected MoQ connection, unable to extract URL from the request.");
+            continue;
```
wkozyra95: If you need patterns like this (logging on every "early exist", bunch of continue) it usually suggest to me that wrong thing was extracted to function, you could leave 
```
 let accept_task = tokio::spawn(async {
    while let Some(request) = server.accept().await {
        if let Err(err) = handel_incoming_connection(...).await  {
             warn!(...)
        }
    }
    server.close().await
});
```

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,205 @@
...
+async fn handle_session(
+    session: MoqSession,
+    mut origin_consumer: OriginConsumer,
+    moq_inputs: MoqServerState,
+    ctx: Arc<PipelineCtx>,
+    input_ref: Ref<InputId>,
+) {
+    info!(moq_version=?session.session().version(), "MoQ session established");
+
+    let Some((path, Some(broadcast))) = origin_consumer.announced().await else {
```
wkozyra95, line 203:
```suggestion
    // waiting for the first announced path from session
    let Some((path, Some(broadcast))) = origin_consumer.announced().await else {
```
it's not clear from code that it is iterator

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,205 @@
...
+                input.connection_handle = Some(handle);
+                input.session = Some(session);
+            }
+            None => {
+                warn!("Failed to handle MoQ broadcast, input queue was dropped.");
+            }
+        }
+        Ok(())
+    }) {
+        warn!(
```
wkozyra95: it's one place so it's fine, but as soon it grows it needs to be refactored

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,205 @@
...
+    fn drop(&mut self) {
+        let _guard = self.rt.enter();
+        self.session.close(hang::moq_net::Error::Cancel);
+        tracing::info!("MoQ session closed!");
+    }
+}
+
+pub struct MoqPipelineState {
+    pub port: u16,
+    pub inputs: MoqServerState,
```
wkozyra95: you renamed types, but field is still named inputs

wkozyra95 (commented):
audit variable names, there is bunch of inconsistent naming after some type name changes

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,398 @@
...
+        .await
+        .inspect_err(|error| {
+            warn!(
+                input_id = %input_ref,
+                "MoQ broadcast error: {}",
+                ErrorStack::new(error).into_string()
+            );
+        })
+        .ok();
+    });
```
JBRS307: I am not a fan of how that looks now, as the error is now handled inside spawn. Neither I nor claude have better ideas.

smelter-core/src/pipeline/moq/connection.rs
JBRS307: https://github.com/software-mansion/smelter/pull/1998/changes#r3434428576

smelter-core/src/pipeline/moq/server.rs
JBRS307: ~I don't think it will be used in more than one place anytime soon, so for now I'll keep it~

Did a minor change so it looks nicer and there is no log redundancy

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -0,0 +1,222 @@
...
+
+async fn run_accept_loop(
+    mut server: moq_native::Server,
+    moq_inputs: MoqServerState,
+    ctx: Arc<PipelineCtx>,
+) {
+    while let Some(request) = server.accept().await {
+        let (session, consumer, input_ref) =
+            match handle_incoming_connection(request, &moq_inputs, &ctx).await {
+                Ok(session) => session,
```
JBRS307: Correct from the logic point of view, semantically terrible

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -0,0 +1,214 @@
...
+    };
+
+    let audio = match catalog.audio.renditions.first_key_value() {
+        Some((name, config)) => match (&config.container, &config.codec) {
+            (CatalogContainer::Cmaf { init, .. }, AudioCodec::AAC(_)) => {
+                let wire = fmp4::Wire::from_init(init)?;
+                let description = match extract_codec_description(&wire) {
+                    Ok(config) => Some(config),
+                    Err(error) => {
+                        warn!(%error, "Failed to extract audio config from CMAF container.");
```
wkozyra95, line 148: what are the consequences? config failed, but you still return DiscoveredAudio

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,461 @@
...
+struct TrackCtx {
+    ctx: Arc<PipelineCtx>,
+    input_ref: Ref<InputId>,
+    broadcast: BroadcastConsumer,
+    decoders: MoqServerInputDecoders,
+    first_pts: Arc<Mutex<Option<Duration>>>,
+    should_close: Arc<AtomicBool>,
+}
+
+impl TrackCtx {}
```
wkozyra95:
```suggestion
```

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,461 @@
...
+            pts,
+            dts: None,
+            kind: MediaKind::Video(VideoCodec::H264),
+            present: true,
+        };
+
+        decoder_handle
+            .chunk_sender
+            .send(PipelineEvent::Data(chunk))
+            .map_err(|_| MoqConnectionError::ChannelClosed)?;
```
wkozyra95:
```suggestion
        if decoder_handle.chunk_sender.send(PipelineEvent::Data(chunk)).is_err() {
           debug!("Failed to send chunk, Channel closed.")
           break;
        }
```
and add eos after the loop

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,461 @@
...
+
+        let chunk = EncodedInputChunk {
+            data: payload,
+            pts,
+            dts: None,
+            kind: MediaKind::Audio(AudioCodec::Aac),
+            present: true,
+        };
+
+        decoder_handle
```
wkozyra95: the same as video

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -0,0 +1,461 @@
...
+#[derive(thiserror::Error, Debug)]
+enum MoqConnectionError {
+    #[error("MoQ track error")]
+    TrackError(#[from] MoqError),
+
+    #[error("MoQ catalog error: {0}")]
+    CatalogError(#[from] MoqCatalogError),
+
+    #[error("Failed to initialize H264 decoder")]
+    InitVideoDecoder(#[source] DecoderInitError),
```
wkozyra95: Up to you, but I would just create one InitDecoder variant with [from], which would simplify `map_err` above to just `?`

wkozyra95 (approved):

smelter-core/src/pipeline/moq/connection/catalog.rs
JBRS307: I will take that into account when adding other containers.

PR #1999, opened 2026-05-18
vp9 depayloader failes after few frames

wkozyra95:
Not sure if this was always happening, I think there was cases where vp9 worked normally, so maybe I'm hitting this case on some browser. (explanation in the comment)

Apparently, this should happen only in flexible mode which is rare. The same issue exists in pion and webrtc-rs

### DIFF
--- a/smelter-core/src/pipeline/rtp/depayloader.rs
+++ b/smelter-core/src/pipeline/rtp/depayloader.rs
@@ -70,7 +70,7 @@ pub enum DepayloadingError {
 struct BufferedDepayloader<T: Depacketizer + Default + 'static> {
     kind: MediaKind,
     buffer: Vec<Bytes>,
-    depayloader: T,
+    _marker: std::marker::PhantomData<T>,
 }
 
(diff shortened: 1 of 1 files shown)


smelter-core/src/pipeline/rtp/depayloader.rs
```diff
@@ -89,7 +89,17 @@ impl<T: Depacketizer + Default + 'static> Depayloader for BufferedDepayloader<T>
         packet: RtpPacket,
     ) -> Result<Vec<EncodedInputEvent>, DepayloadingError> {
         trace!(?packet, "RTP depayloader received new packet");
-        let chunk = self.depayloader.depacketize(&packet.packet.payload)?;
+        // Workaround for a webrtc-rs `Vp9Packet::depacketize` bug — construct a fresh `T` per packet
```
wkozyra95, line 92: I'm assuming this is a bug, because h264 depayloder is statefull, so it does not make sense for this one to be recreated for each chunk

wkazmierczak (approved):

JBRS307 (approved):

PR #2016, opened 2026-05-22
Fixed the "unused import" warning when building WASM.

JBRS307:


smelter-render/src/wgpu.rs
```diff
@@ -61,7 +60,7 @@ impl Drop for WgpuErrorScope {
         };
         for scope in scopes.into_iter().rev() {
             if let Some(err) = pollster::block_on(scope.pop()) {
-                error!("Wgpu error in dropped scope: {err}");
+                tracing::error!("Wgpu error in dropped scope: {err}");
```
JBRS307, line 63: The only place where that macro is used

wkozyra95 (approved):
Fix pr title, it does not make sense currentlly e.g. Fix unused import when building for WASM

PR #2017, opened 2026-05-22
[ts-sdk] Experimental MoQ server input API

JBRS307:
- #2009

---

- Please review this after #1998

PR #2018, opened 2026-05-25
Experimental MoQ client input

JBRS307:
- #2019

- This should be reviewed and merged after #2017 

---

It has different use case as MoQ server input. We need a way to both take streams for endpoint and put smelter in between the streamer and relay and to take the stream from the relay. For that we need both client and server input.


smelter-core/src/protocols/moq.rs
```diff
@@ -8,16 +8,26 @@ use crate::queue::QueueInputOptions;
...
 }
 
 #[derive(Debug, Clone, PartialEq)]
-pub struct MoqServerInputDecoders {
+pub struct MoqInputDecoders {
     pub h264: Option<VideoDecoderOptions>,
     pub aac: Option<AudioDecoderOptions>,
 }
 
+pub type MoqClientInputDecoders = MoqInputDecoders;
```
JBRS307: Remove

smelter-core/src/pipeline/moq/moq_client_input.rs
```diff
@@ -0,0 +1,111 @@
...
+    decoders: MoqInputDecoders,
+    queue_input: QueueInput,
+    url: Url,
+    broadcast_path: Arc<str>,
+) -> anyhow::Result<()> {
+    let mut config = moq_native::ClientConfig::default();
+
+    // TODO: (@jbrs) This is fine for the experimental, however will need to be addressed in the
+    // "complete" version.
+    config.tls.disable_verify = Some(true);
```
JBRS307: Add env for that. In release mode default it to false, in debug to true

JBRS307:
Will be back to that later

PR #2022, opened 2026-05-27
Configure side channel delay per input

wkozyra95:
Originally I wanted to have one value to make sure that all the streams are synced correctly. However this is causing a lot of issues, e.g.
- If we need speech to text, the delay has to be long, effectively in our demo adding new mp4 would have delay of 10 seconds
- Mp4 when looping, have a gap between start/end equal the value of the delay. I'll try to address that in follow up

Also supporting per input value is easy, so I'm leaning towards that approach.

I also considered different value for audio and video, but that would cause desync.


wkazmierczak (commented):
LGTM

PR #2024, opened 2026-05-27
Remove `endpoint_override` option from WHIP input

wkozyra95:
we don't need override anymore because ts sdk wil now use the same value for input and route(no `global:` prefix). As a result we don't need concept of endpoint_id, because segment in webrtc path will always be input_id

### DIFF
--- a/ts/smelter-core/src/offline/compositor.ts
+++ b/ts/smelter-core/src/offline/compositor.ts
@@ -91,10 +91,7 @@ export class OfflineSmelter {
     _smelterInternals.assertGlobalInputId(inputId);
 
     const inputRef = { type: 'global', id: inputId } as const;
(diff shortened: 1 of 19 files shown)


ts/smelter-core/src/offline/compositor.ts
```diff
@@ -91,10 +91,7 @@ export class OfflineSmelter {
     _smelterInternals.assertGlobalInputId(inputId);
 
     const inputRef = { type: 'global', id: inputId } as const;
-    const result = await this.api.registerInput(inputRef, intoRegisterInput(inputId, request));
-    if (request.type === 'whip_server') {
```
wkozyra95, line 95: server already sends that

wkozyra95:
@JBRS307 This might be relevant to your work on moq input

wkazmierczak (approved):

JBRS307 (approved):

PR #2025, opened 2026-05-27
Use `wgpu` from `master` branch

noituri:


integration-tests/src/render_tests/harness/utils.rs
```diff
@@ -165,7 +165,7 @@ fn download_buffer(
     r.recv().unwrap().unwrap();
     let mut buffer = buffer.writer();
     {
-        let range = source.slice(..).get_mapped_range();
+        let range = source.slice(..).get_mapped_range().unwrap();
```
noituri, line 168: in `wgpu 29` this would just panic but in `wgpu master` this now returns error

integration-tests/examples/silly.wgsl
```diff
@@ -1,3 +1,5 @@
+enable wgpu_binding_array;
```
wkozyra95, line 1: This is a breaking change, add to CHANGELOG and update docs

wkozyra95 (approved):

integration-tests/examples/silly.wgsl
noituri: already did

PR #2032, opened 2026-05-28
[rtmp] E-RTMP client reconnect handling

wkazmierczak:


rtmp/src/reconnect.rs
```diff
@@ -0,0 +1,102 @@
...
+    if get_str(map, "level") != Some("status") {
+        return None;
+    }
+    Some(ReconnectRequest {
+        tc_url: get_str(map, "tcUrl").map(str::to_string),
+        description: get_str(map, "description").map(str::to_string),
+    })
+}
+
+fn get_str<'a>(map: &'a HashMap<String, AmfValue>, key: &str) -> Option<&'a str> {
```
wkozyra95: I think this is bad, but it's bad hard to explain
- This function does 3 things (field access, null check, and type cast), while having extremely generic name
- It would be fine to add something similar on AmfValue itself, but here it looks extremely out of place
- It looks  like you wanted helper to do specific thing and just implemented something that handle the case 100%, instead of considering what general purpose stuff you could add that would solve it at 80%

This approach would be a kind of acceptable if you put this entire function in the func that is calling it, but even then I think ther are better approaches.

What I would do:
- do not hide early exists in if conditions, very hard to read
- extract all fields you need with pattern match at one step

rtmp/src/client.rs
```diff
@@ -47,12 +54,39 @@ struct RtmpClientState {
     window_size: Option<u64>,
     /// last ack sent to client
     last_ack: u64,
+    ex_capabilities: ExCapabilities,
+    /// Background connection being established for E-RTMP reconnect.
+    pending_reconnect: Option<PendingReconnection>,
+    /// Stored for replay on reconnect.
+    stored_video_config: Option<VideoConfig>,
```
wkozyra95: definitionally reconsider this, both naming and purpose, if we need to keep configs, maybe they should not be sent as event, but provided on start instead?
Or maybe the good way is just disconnect, return appropriate error to user and let them handle second connection (unless reconnects needs some specific knowledge), I don't know how the reconnect is suppose to work, I'm just throwing it out.

rtmp/src/client.rs
```diff
@@ -232,6 +368,54 @@ impl RtmpClientState {
...
+        let resolved_url = match resolve_reconnect_url(&current_tc_url, request.tc_url.as_deref()) {
+            Ok(url) => url,
+            Err(err) => {
+                warn!(?err, "Ignoring E-RTMP ReconnectRequest with invalid tcUrl");
+                return;
+            }
+        };
+
+        let mut new_config = config.clone();
+        if let Err(err) = apply_tc_url(&mut new_config, &resolved_url) {
```
wkozyra95: unclear what it does, what is for, probablly wrong place for that logic

rtmp/src/client.rs
wkazmierczak: If I understand spec correctly, E-RTMP require overlap to avoid frame gap, so disconnect with error is not an option.

wkazmierczak: And when it comes to configs I think they should remain as events. RTMP allows updating config multiple times during one stream.

On the other hand, from what I read, it is rare to have more than one video/audio config in one RTMP stream, but it is theoretically possible.

rtmp/src/client.rs
```diff
@@ -120,19 +200,123 @@ impl RtmpClient {
...
         };
         self.state.stream.write_msg(event)?;
 
-        // try read any pending messages
         while let Some(msg) = self.state.stream.try_read_msg()? {
+            if let Some(request) = try_match_reconnect_request(&msg) {
+                debug!(?request, "Received NetConnection.Connect.ReconnectRequest");
+                self.start_reconnect(request);
+                continue;
+            }
```
wkozyra95: not a place for that, this should be tight loop that just handles any pending incoming messages,

rtmp/src/client.rs
```diff
@@ -120,19 +200,123 @@ impl RtmpClient {
...
+        let current_tc_url = self.config.tc_url();
+        let resolved_url = match resolve_reconnect_url(&current_tc_url, request.tc_url.as_deref()) {
+            Ok(url) => url,
+            Err(err) => {
+                warn!(?err, "Ignoring E-RTMP ReconnectRequest with invalid tcUrl");
+                return;
+            }
+        };
+
+        let new_config = match self.config.update_with_tc_url(&resolved_url) {
```
wkozyra95: name implies mut, it is misleading method like should be `RtmpClientConfig::from_tc_url()`.

However after looking at resolve_reconnect_url this separation seem backward, why would you split this logic into 2 parts. You first trying reconstruct tc_url and then the object that would have for free method to return tc url?

Also by spliting it into 2 parts, 2 function calls take up 20 lines poluting actual content of the function

rtmp/src/client.rs
```diff
@@ -68,35 +128,54 @@ impl RtmpClient {
...
+    /// new connection is established in a background thread while data
+    /// continues flowing on the old connection. At the next media boundary
+    /// (video keyframe, or any audio chunk for audio-only streams), the
+    /// client switches to the new connection and replays codec configs
+    /// and metadata.
     pub fn send<T>(&mut self, event: T) -> Result<(), RtmpStreamError>
     where
         RtmpEvent: From<T>,
     {
+        let reconnect_ready = self
```
wkozyra95: The general approach feels very forced,  (how to fit reconnect with minimal effort/minimal chages).
Stuff like:
- connecting in extra thread, but in the meantime we are still sending data and just checking if it finished
- random if statement ploted in event handling 

So question:
- Does protocol expects you to keep sending data or you should/can abort after starting reconnect? 
- What is the purpose of reconnect, is it suppose to be glitch/gap free switch, or visible stop is expected?
- Reconnect requests are suggestions, or will it actually drop?

If it is really suppose to be glitch free I get the thread approach, but I think it can be done cleaner without heavy coupling

wkazmierczak: Based on spec:

Message Flow When Handling NetConnection.Connect.ReconnectRequest
1. Prior to the shutdown of the live streaming server or when the server intends to remap the client to another server instance, it dispatches an onStatus command to the client with a code of NetConnection.Connect.ReconnectRequest. If the server aims to remap the client, it MUST set the tcUrl property in the Info Object. In order to avoid a disruption, the server managing the original connection
(commonly referred to as the "old server") SHOULD continue processing messages from the client until the client disconnects.
2. When the client receives the NetConnection.Connect.ReconnectRequest event, it persists in streaming to/from the current server up to the next appropriate media boundary, such as a keyframe. Subsequently, it establishes a connection with a new server and disconnects from the old server. If the Info Object includes the tcUrl property, the client uses this URL for the reconnection process. Absent this property, the client defaults to the tcUrl for the current connection.
3. While the client can establish a new connection before severing the original one, it SHOULD exercise caution to ensure the Quality of Service (QoS) is not compromised.


Q1: Protocol expects to keep sending
Q2: It suppose to be glitch/gep free switch
Q3: Reconnect itself is suggestion. Does not mean actual drop

rtmp/src/client.rs
wkazmierczak: Previously it was used in a different place but currently it does not make sense

rtmp/src/client.rs
wkazmierczak: I was wondering where to put it. It cannot be in `default_msg_handler()` because it would be strange that something with name `default` returns something specific for reconnect, so I put right before it. I wasn't really convinced, because it also feels a bit strange but I had no better idea

rtmp/src/client.rs
```diff
@@ -150,75 +178,49 @@ impl RtmpClient {
...
-                RtmpMessageOutgoing::DataMessage {
-                    data: DataMessage::OnMetaData(metadata),
-                    stream_id: self.stream_id,
-                }
-            }
-        };
-        self.state.stream.write_msg(event)?;
+    /// Spec: switch only at a media boundary — a video keyframe, or (for
+    /// audio-only streams) any audio chunk. (E-RTMPv2, Reconnect message flow.)
+    fn is_switch_boundary(&self, event: &RtmpEvent) -> bool {
```
wkozyra95: I would inline that (but keep comment)

rtmp/src/client.rs
```diff
@@ -150,75 +178,49 @@ impl RtmpClient {
...
     {
-        let reconnect_ready = self
+        let event = RtmpEvent::from(event);
+
+        // Honor a finished background reconnect, but only at a media boundary.
+        self.maybe_switch_connection(&event);
+        // Remember codec configs / metadata so they can be replayed after a switch.
+        self.media_config.record(&event);
+        // Build + write on whatever connection is now current.
+        let msg = self.state.build_outgoing(event, self.stream_id);
```
wkozyra95: maybe into_message

rtmp/src/client.rs
```diff
@@ -389,6 +379,45 @@ impl RtmpClientState {
...
+                video: VideoMessage::Data(data),
+                stream_id,
+            },
+            RtmpEvent::VideoConfig(config) => RtmpMessageOutgoing::Video {
+                video: VideoMessage::Config(config),
+                stream_id,
+            },
+            RtmpEvent::AudioData(data) => {
+                let channels = self
+                    .audio_channels
```
wkozyra95: you can use recorded values instead

rtmp/src/client.rs
```diff
@@ -68,70 +156,156 @@ impl RtmpClient {
...
+    fn handle_incoming(&mut self, msg: RtmpMessageIncoming) -> Result<(), RtmpStreamError> {
+        if let Some(request) = try_match_reconnect_request(&msg) {
+            debug!(?request, "Received NetConnection.Connect.ReconnectRequest");
+            self.start_reconnect(request);
+            return Ok(());
+        }
+        self.state.default_msg_handler(msg)
+    }
+
+    fn do_reconnect(&mut self) {
```
wkozyra95: It's used in one place, so inline it, just write the condition differnt way so it's not nested

rtmp/src/client.rs
```diff
@@ -68,70 +156,156 @@ impl RtmpClient {
...
-                    .insert(TrackKey::new(self.stream_id, config.track_id), channels);
-                RtmpMessageOutgoing::Audio {
-                    audio: AudioMessage::Config(config),
+            Err(_) => {
+                warn!("E-RTMP reconnect thread panicked, continuing on current connection");
+                return;
+            }
+        };
+
+        let _ = self
```
wkozyra95: always at least log with debug! the error

rtmp/src/client.rs
```diff
@@ -68,70 +156,156 @@ impl RtmpClient {
...
+                msg: CommandMessage::DeleteStream {
+                    transaction_id: 0,
                     stream_id: self.stream_id,
-                    channels,
-                }
+                },
+                stream_id: CONTROL_MESSAGE_STREAM_ID,
+            });
+
+        self.config = pending.new_config;
```
wkozyra95: consider renaming connection_opts (and type to RtmpConnectionOptions)

rtmp/src/client.rs
```diff
@@ -68,70 +156,156 @@ impl RtmpClient {
...
+        }
+        Ok(())
+    }
+
+    fn maybe_switch_connection(&mut self, event: &RtmpEvent) {
+        let ready = self
+            .pending_reconnect
+            .as_ref()
+            .is_some_and(|p| p.handle.is_finished());
+        if ready && self.is_switch_boundary(event) {
```
wkozyra95: might be worth checking if thread finished

wkazmierczak: but I am already checking if finished:
```
let ready = self
    .pending_reconnect
    .as_ref()
    .is_some_and(|p| p.handle.is_finished());
```

rtmp/src/amf0/mod.rs
```diff
@@ -40,6 +40,15 @@ pub enum AmfValue {
     },
 }
 
+impl AmfValue {
+    pub fn as_str(&self) -> Option<&str> {
```
wkozyra95, line 44:
```suggestion
    pub fn try_as_str(&self) -> Option<&str> {
```
or `checked_`

rtmp/src/transport/mod.rs
```diff
@@ -18,16 +18,22 @@ pub(crate) enum RtmpTransport {
     TlsServer(Box<TlsServerStream>),
 }
 
+/// Bounded TCP connect timeout for all client connect attempts. Caps how long
+/// the calling thread can block in [`TcpStream::connect_timeout`] when the
+/// target is unreachable or blackholed, so background reconnect workers honor
```
wkozyra95, line 23: blackholed?

rtmp/src/transport/mod.rs
```diff
@@ -79,6 +85,30 @@ impl RtmpTransport {
     }
 }
 
+fn connect_tcp(host: &str, port: u16, timeout: Duration) -> Result<TcpStream, RtmpConnectionError> {
```
wkozyra95, line 88: why is this needed?

rtmp/src/transport/mod.rs
```diff
@@ -79,6 +85,30 @@ impl RtmpTransport {
...
+    let mut last_err: Option<io::Error> = None;
+    let mut any_addr = false;
+    for addr in (host, port).to_socket_addrs()? {
+        any_addr = true;
+        match TcpStream::connect_timeout(&addr, timeout) {
+            Ok(socket) => return Ok(socket),
+            Err(err) => last_err = Some(err),
+        }
+    }
+    Err(last_err
```
wkozyra95, line 98: I would rethink the structure/formating

rtmp/src/client.rs
```diff
@@ -21,148 +25,360 @@ mod negotiation;
...
     }
+
+    /// Resolve an E-RTMP `ReconnectRequest` target into a new config.
+    ///
+    /// `request_tc_url` is resolved against the current config's tcUrl per the
+    /// E-RTMP v2 spec:
+    /// - `None` → reconnect to current.
+    /// - absolute → used as-is.
+    /// - relative / protocol-relative / path-only → resolved against current.
+    pub(crate) fn resolve_reconnect(
```
wkozyra95, line 50:
```suggestion
    pub(crate) fn resolve_reconnect_options(
```

smelter-core/src/pipeline/rtmp/rtmp_output.rs
```diff
@@ -114,7 +114,7 @@ impl RtmpClientOutput {
         video_config: &Option<VideoConfig>,
         audio_config: &Option<AudioConfig>,
     ) -> Result<RtmpClient, RtmpClientError> {
-        let mut client = RtmpClient::connect(RtmpClientConfig {
+        let mut client = RtmpClient::connect(rtmp::RtmpConnectionOptions {
```
wkozyra95, line 117:
```suggestion
        let mut client = RtmpClient::connect(RtmpConnectionOptions {
```

Is there some naming colision here?

PR #2037, opened 2026-05-29
[rtmp] Add codec selection to RTMP client/server config

wkazmierczak:


smelter-core/src/pipeline/rtmp/rtmp_output.rs
```diff
@@ -114,13 +114,16 @@ impl RtmpClientOutput {
...
-            app: connection_opts.app,
-            stream_key: connection_opts.stream_key,
-            use_tls: connection_opts.use_tls,
-        })?;
+        let config = RtmpClientConfig::new(
+            connection_opts.host,
+            connection_opts.port,
+            connection_opts.app,
+            connection_opts.stream_key,
+            connection_opts.use_tls,
```
wkozyra95: If we are going with a builder-like pattern then use_tls and port should have a default value and be set via method.

I'm not sure about host

wkozyra95 (approved):

PR #2038, opened 2026-05-29
[rtmp] Remove unused code, extract `OUTGOING_CHUNK_SIZE` const

wkazmierczak:


wkozyra95 (approved):
do not name it pre release clanup, express what it's actually doing,

PR #2039, opened 2026-05-29
[gpu-video] New device and adapter API

noituri:
The general idea behind this PR is to make the API nicer for the user. Currently to use the library a user has to use our wrapper types. These types wrap:
- vulkan instance + wgpu instance
- vulkan adapter + wgpu adapter
- vulkan device/queue + wgpu device/queue

Users could retrieve the wgpu types from those wrapper types.

In this new approach, the initialization process is more similar to the typical wgpu init process:
- create wgpu instance
- get adapter:
  -  There's `VideoAdapterExt` trait implemented on `wgpu::Adapter` which provides methods to query video capabilities of the adapter and to create `wgpu::Device` capable of video operations
- User gets  `wgpu::Device` (which has video caps) and a normal `wgpu::Queue` 

Decoders and encoders can be created with just a `wgpu::Device` via `VideoDeviceExt` trait.

In the library there's a global registry which maps `wgpu::Device` to `VideoDevice`.
Users who don't need `wgpu` can just use the current initialization process and get `VideoDevice` instead of `wgpu::Device`

I also started renaming some things and there's even more things that should be renamed but I don't want to add bunch of unrelated things in this PR


gpu-video/src/adapter/wgpu_api.rs
```diff
@@ -0,0 +1,50 @@
...
+        desc: &VideoDeviceDescriptor,
+    ) -> Result<(wgpu::Device, wgpu::Queue), VulkanInitError> {
+        let hal_adapter = unsafe { self.as_hal::<VkApi>().unwrap() };
+        let physical_device = hal_adapter.raw_physical_device();
+
+        let instance = hal_adapter.shared_instance();
+        let instance = VideoInstance::new_unowned(
+            instance.raw_instance().clone(),
+            // TODO: does entry even need arc?
+            instance.entry().clone().into(),
```
noituri: `Entry` already has `Arc` inside so it should be fine to keep just `Entry` without `Arc` in VulkanInstance

gpu-video/src/wgpu_helpers/global_registry.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// TODO: Maybe it would be better to just panic? It would be kinda inline with what wgpu would probably do?
+#[derive(Debug, thiserror::Error)]
+pub enum RegistryError {
+    #[error(
+        "Could not find the device in registry. Make sure the device was created with video capabilities"
+    )]
+    DeviceNotFound,
+}
```
noituri: currently all error enums used in the `VideoDeviceExt` trait also include `RegistryError`. I don't like it. I thought about creating a wrapper error enum which would wrap `RegistryError` and some generic error (`DecoderError`, `VulkanInitError`....). But that looked messy as well

gpu-video/src/wrappers.rs
```diff
@@ -28,11 +28,32 @@ pub(crate) struct Instance {
     pub(crate) video_queue_instance_ext: ash::khr::video_queue::Instance,
     pub(crate) video_encode_queue_instance_ext: ash::khr::video_encode_queue::Instance,
     pub(crate) debug_utils_instance_ext: ash::ext::debug_utils::Instance,
+    pub(crate) destroy_instance_on_drop: bool,
```
noituri, line 31: i didn't want to play with `std::mem::forget`. We also had a memory leak because of it some time ago...

gpu-video/src/adapter/wgpu_api.rs
```diff
@@ -0,0 +1,68 @@
...
+        let physical_device = hal_adapter.raw_physical_device();
+        let instance = hal_adapter.shared_instance();
+        let instance = VideoInstance::new_unowned(
+            instance.raw_instance().clone(),
+            instance.entry().clone().into(),
+            &VideoInstanceDescriptor {
+                enable_validations: instance.extensions().contains(&vk::EXT_DEBUG_UTILS_NAME),
+                ..Default::default()
+            },
+        );
```
jerzywilczek: wait, each time we query we make a new `Instance`?? it seems quite strange. mind explaining the motivation behind this? why are we not putting instances in a registry too?

gpu-video/src/adapter/wgpu_api.rs
jerzywilczek: i think only one copy of a .so will exist in a system anyway. so as long as the Entry is Clone, we don't have to care

jerzywilczek (changes requested):
as we talked, Im not a huge fan of non-wgpu things not going through the registry. i am still uncertain whether its the correct course or not to change that.

gpu-video/src/adapter/wgpu_api.rs
noituri: ye, I also don't like it but I couldn't put this in a registry because we won't know when to remove it

gpu-video/examples/decode.rs
```diff
@@ -21,15 +24,19 @@ fn main() {
 
     let h264_bytestream = std::fs::read(&args[1]).unwrap_or_else(|_| panic!("read {}", args[1]));
 
-    let vulkan_instance = VulkanInstance::new().unwrap();
+    let vulkan_instance = VideoInstance::new(&VideoInstanceDescriptor {
```
jerzywilczek: rename the bindings to vulkan_*

gpu-video/src/adapter/wgpu_api.rs
```diff
@@ -0,0 +1,59 @@
...
+
+/// [`wgpu::Adapter`] extension that exposes video capabilities of an adapter.
+pub trait VideoAdapterExt {
+    /// Retrieves information about adapter and its video capabilities.
+    /// Returns `None` if it doesn't support any video operations.
+    fn video_adapter_info(&self) -> Option<VideoAdapterInfo>;
+
+    /// Creates a device capable of creating video decoders and encoders
+    /// via [`VideoDeviceExt`](crate::VideoDeviceExt)
+    fn request_device_with_video_support(
```
jerzywilczek, line 15: shouldnt this be called `create_device_with_video_support`, to be more in line with the wgpu naming?

gpu-video/src/device.rs
```diff
@@ -265,60 +253,166 @@ pub struct EncoderParametersH265 {
     pub output_parameters: EncoderOutputParameters<H265Profile>,
 }
 
-/// Open connection to a coding-capable device. Also contains a [`wgpu::Device`], a [`wgpu::Queue`] and
-/// a [`wgpu::Adapter`].
-pub struct VulkanDevice {
-    #[cfg(feature = "wgpu")]
-    pub(crate) wgpu_ctx: WgpuContext,
+/// Extension that exposes video capabilities of a device.
+/// The device must be created with [`VideoAdapterExt::request_device_with_video_support`](`crate::VideoAdapterExt::request_device_with_video_support`).
```
jerzywilczek:
```suggestion
/// Extension that exposes the video capabilities of a device.
/// The device must have been created using [`VideoAdapterExt::request_device_with_video_support`](`crate::VideoAdapterExt::request_device_with_video_support`).
```

also, say what will happen if it hasn't been created using our method.

gpu-video/src/lib.rs
```diff
@@ -7,6 +7,9 @@ pub(crate) mod codec;
 #[cfg(vulkan)]
 mod device;
 #[cfg(vulkan)]
+#[cfg_attr(not(feature = "wgpu"), allow(dead_code))]
+mod global_registry;
```
jerzywilczek: what is this used for when wgpu is off?

smelter-core/src/graphics_context/vulkan_context.rs
```diff
@@ -20,80 +25,113 @@ pub fn create_vulkan_graphics_ctx(
...
     Ok(GraphicsContext {
-        device: device.wgpu_device().into(),
-        queue: device.wgpu_queue().into(),
-        adapter: device.wgpu_adapter().into(),
-        instance: instance.wgpu_instance().into(),
-        vulkan_ctx: Some(VulkanCtx { instance, device }),
+        device: device.into(),
+        queue: queue.into(),
+        adapter: adapter.into(),
+        instance: instance.into(),
```
jerzywilczek, line 121: these still in arcs? maybe remove them?

smelter-core/src/pipeline/decoder/vulkan_h264.rs
```diff
@@ -25,22 +26,23 @@ impl VideoDecoder for VulkanH264Decoder {
...
-                })
-            }
-            None => Err(DecoderInitError::VulkanContextRequiredForVulkanDecoder),
+        if ctx.graphics_context.vulkan_ctx.is_none() {
+            return Err(DecoderInitError::VulkanContextRequiredForVulkanDecoder);
         }
+
+        info!("Initializing Vulkan H264 decoder");
+        let device = ctx.wgpu_ctx.device.video().map_err(|_| {
+            DecoderInitError::VulkanDecoderError(VideoDecoderError::VideoDeviceWithoutWgpu)
```
jerzywilczek: `VideoDeviceWithoutWgpu`? not the other way around lol?

Cargo.toml
```diff
@@ -62,7 +62,12 @@ glyphon = { git = "https://github.com/smelter-labs/glyphon", rev = "33b3dfe" }
 futures-util = "0.3.30"
 tokio = { version = "1", features = ["full"] }
 schemars = { version = "0.8.22", features = ["preserve_order"] }
-utoipa = { version = "5.4.0", features = ["rc_schema", "preserve_order", "preserve_path_order", "axum_extras"] }
+utoipa = { version = "5.4.0", features = [
+    "rc_schema",
+    "preserve_order",
+    "preserve_path_order",
+    "axum_extras",
+] }
```
jerzywilczek, line 70: what'd you do that formatted all tomls

jerzywilczek (approved):
i think its good, but we need to talk a bit irl

gpu-video/src/adapter/wgpu_api.rs
noituri: wgpu uses `request_device` so it's inline with their naming

smelter-core/src/graphics_context/vulkan_context.rs
noituri: yup, I will refactor it soon-ish, I just didn't want to do it in this PR

smelter-core/src/pipeline/decoder/vulkan_h264.rs
noituri: true xd

Cargo.toml
noituri: helix uses taplo underneath

gpu-video/examples/decode.rs
noituri: I'm assuming you meant video_*

gpu-video/src/lib.rs
noituri: it's not. I'll hide it behind a wgpu feature

gpu-video/examples/decode.rs
jerzywilczek: i did

gpu-video/src/adapter/wgpu_api.rs
jerzywilczek: no it doesnt. you request an adapter, but create a device

noituri: bruh

PR #2044, opened 2026-06-02
Added endpoint with MoQ cert

JBRS307:
This is needed so that the browser can skip TLS verification if needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1016,15 +1016,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "conducer"
-version = "0.3.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8d2cb64e61144d6960a830d3e6f2ba3a61d5c0ca689e87e11dc9effb96dcfff5"
-dependencies = [
- "smallvec",
-]
-
-[[package]]
 name = "console_error_panic_hook"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


smelter-core/src/pipeline/moq/server.rs
```diff
@@ -67,6 +69,21 @@ pub async fn spawn_moq_server(
...
+    let certificate_sha256 = server
+        .tls_info()
+        .read()
+        .unwrap()
+        .fingerprints
+        .first()
+        .cloned();
+    if certificate_sha256.is_none() {
+        warn!("MoQ server has no certificate fingerprint, /moq/certificate.sha256 will return 404");
+    }
```
JBRS307: This needs to be thought on. Should it work like that or should fingerprint failure stop the MoQ server?

src/routes/status.rs
```diff
@@ -158,3 +158,14 @@ pub async fn stats_handler(
...
+pub async fn moq_certificate_handler(
+    State(state): State<Arc<ApiState>>,
+) -> Result<impl IntoResponse, ApiError> {
+    let pipeline = state.pipeline()?;
+    let fingerprint = pipeline.lock().unwrap().moq_certificate_sha256();
+    match fingerprint {
+        Some(fp) => Ok(([(axum::http::header::CONTENT_TYPE, "text/plain")], fp).into_response()),
+        None => Ok(StatusCode::NOT_FOUND.into_response()),
+    }
+}
```
JBRS307: Add this to openapi (`utoipa` macro)

smelter-core/src/pipeline/instance.rs
```diff
@@ -79,6 +79,12 @@ impl Pipeline {
         self.stats_monitor.report()
     }
 
+    pub fn moq_certificate_sha256(&self) -> Option<String> {
+        self.moq_server
+            .as_ref()
+            .and_then(|s| s.certificate_sha256.clone())
+    }
+
```
JBRS307:
```suggestion
```

PR #2050, opened 2026-06-03
Fix the RGBA-NV12 shader to convert to the limited range NV12.

jerzywilczek:


smelter-render/src/wgpu/format/rgba_to_nv12.wgsl
```diff
@@ -39,7 +42,9 @@ fn fs_main_uv(input: VertexOutput) -> @location(0) vec2<f32> {
         -0.3854, -0.4542,
          0.5,    -0.0458,
     );
+    let conversion_scale = 224.0 / 255.0;
     let conversion_bias = vec2<f32>(0.5, 0.5);
 
-    return clamp(conversion_weights * color + conversion_bias, vec2(0.0, 0.0), vec2(1.0, 1.0));
+    let uv = conversion_weights * color * conversion_scale + conversion_bias;
```
wkozyra95: Are you sure about that? in yuv implementation we multiplay by scale after adding that 0.5 bias and still add that 16/255

wkozyra95: ok acording to Claude it's more or less the same, it only claims that you should use 128/255 instead of 0.5

Up to you, but I would prefer consistent implementation with rgba_to _yuv.wgsl

wkozyra95 (approved):

Issue #2009, opened 2026-05-21
MoQ server input


JBRS307:
- [x] basic server code: #1998 
  - [x] stats: #2060 
  - [x] opus audio (this is important because currently it is not possible to stream any audio from browser on linux): #2075
  - [x] vp8, vp9: #2094 
  - [x] legacy container: #2075 
  - [x] LOC container: #2075 
  - [x] auth: #2078
- [x] ts-sdk: #2095 
- [x] docs: https://github.com/smelter-labs/smelter-website/pull/280

PR #2060, opened 2026-06-09
Added stats to `moq_server` input.

JBRS307:
- #2009


ts/smelter/src/api.generated.ts
JBRS307, line 1: Remove that, ts-sdk will go later

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -114,13 +115,15 @@ async fn handle_broadcast(
 
     let discovered = read_catalog(&broadcast).await?;
 
+    let stats_sender = MoqStatsSender::new(input_ref.clone(), ctx.stats_sender.clone());
```
wkozyra95: You can create it inside BroadcastHandler::new, it's not relvant code for the core function it is best to hide it

wkozyra95 (approved):

PR #2067, opened 2026-06-11
Queue tests

wkozyra95:


smelter-core/src/queue/queue_input.rs
```diff
@@ -44,7 +46,8 @@ impl<T> std::fmt::Debug for QueueSender<T> {
 }
 
 pub(super) struct InnerQueueInput {
-    ctx: Arc<PipelineCtx>,
+    queue_ctx: QueueContext,
```
wkozyra95, line 49: Mocking the entire PipelineCtx would be too much, so I decided to just extract what is needed

smelter-core/src/queue/tests/video.rs
```diff
@@ -0,0 +1,1413 @@
...
+
+        // desync regular clock from queue clock
+        sleep(OFFSET);
+
+        queue.start();
+        (queue, input)
+    }
+
+    #[test]
+    fn offset_from_start_delivered_early() {
```
JBRS307, line 59: This needs a brief comment, because from just reading the name I do not know what do "offset from start" and "delivered early" mean

smelter-core/src/queue/tests/harness.rs
```diff
@@ -0,0 +1,561 @@
...
+//! - send frames/samples via `TestInput` (or `stream_video_then_eos` for sends
+//!   that must not block the test thread, e.g. before queue start)
+//! - `start()` the queue
+//! - read output with `next_video_batch`/`next_audio_batch` and compare against
+//!   expected `VideoBatch`/`AudioBatch` values; all PTS in summaries are relative
+//!   to the queue start, so expectations are deterministic for `Pts`/`FromStart`
+//!   offsets. For `QueueTrackOffset::None` the offset depends on wall clock —
+//!   assert with tolerance instead of equality.
+
+#![allow(dead_code)]
```
JBRS307, line 19: Should it stay?

smelter-core/src/queue/tests/harness.rs
```diff
@@ -0,0 +1,561 @@
...
+        DEFAULT_AUDIO_CHUNK_DURATION, Queue, QueueAudioOutput, QueueContext, QueueInput,
+        QueueInputOptions, QueueOptions, QueueSender, QueueTrackOptions, QueueVideoOutput,
+    },
+    types::Ref,
+};
+
+/// Distance between queue creation and start to desync clocks
+pub const OFFSET: Duration = Duration::from_micros(123_456);
+
+pub const OUTPUT_FRAMERATE: Framerate = Framerate { num: 50, den: 1 };
```
JBRS307, line 40: Nit pick, but maybe we should pick something more standard, 30 or 60?

JBRS307 (commented):
Tests need short descriptions like in the `integration-tests`. It is difficult to quickly understand what does the given test do just by the name.

smelter-core/src/queue/tests/harness.rs
wkozyra95: nope it's 20ms far easier to calculate

smelter-core/src/queue/tests/video.rs
wkozyra95: it means that input has offset `FromStart` and input packets are delivered earlier than they are needed, naming pattern is consistent in the file, and any more precise definition would be overkill

smelter-core/src/queue/tests/harness.rs
wkozyra95: yes, I did not write audio test, so types there will be dead code

PR #2075, opened 2026-06-17
Added Opus codec and LOC, Legacy containers to MoQ server input.

JBRS307:
- #2009

---

This one is important. Streaming from browser with our MoQ streamer tool does not work with AAC on Linux, as there are no supported encoders.

### DIFF
--- a/smelter-core/src/pipeline/moq/connection/catalog.rs
+++ b/smelter-core/src/pipeline/moq/connection/catalog.rs
@@ -120,7 +120,6 @@ fn discover_video(
     };
 
     let description = match (&config.description, &container) {
-        (Some(desc), _) => Some(desc.clone()),
         (None, Container::Cmaf(wire)) => match extract_codec_description(wire) {
             Ok(desc) => Some(desc),
             Err(error) => {
@@ -128,7 +127,7 @@ fn discover_video(
                 return Ok(None);
             }
         },
-        (None, _) => config.description.clone(),
+        _ => config.description.clone(),
     };
 
     Ok(Some(DiscoveredVideo {
@@ -164,16 +163,15 @@ fn discover_audio(
 
     // Decoder config extraction is necessary only for AAC. Opus is self-contained and does not need
     // description
-    let description = match (&config.description, &container) {
-        (Some(desc), _) => Some(desc.clone()),
-        (None, Container::Cmaf(wire)) => match extract_codec_description(wire) {
+    let description = match (&config.description, &codec, &container) {
+        (None, AudioCodec::Aac, Container::Cmaf(wire)) => match extract_codec_description(wire) {
             Ok(desc) => Some(desc),
             Err(error) => {
-                warn!(%error, "Failed to extract video decoder config from container; skipping video track.");
+                warn!(%error, "Failed to extract audio decoder config from container; skipping audio track.");
                 return Ok(None);
             }
         },
-        (None, _) => config.description.clone(),
+        _ => config.description.clone(),
     };
 
     Ok(Some(DiscoveredAudio {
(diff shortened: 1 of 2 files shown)


smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -205,3 +242,18 @@ fn extract_codec_description(cmaf: &fmp4::Wire) -> Result<Bytes, MoqCatalogError
         )),
     }
 }
+
+fn audio_codec(codec: &MoqAudioCodec) -> Result<AudioCodec, &'static str> {
```
wkozyra95: why?

JBRS307: ~I didn't want to modify this string in 2 places after adding more codecs, but will change that to return `Option`~

JBRS307: Reworked

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -171,6 +97,87 @@ async fn read_msf_catalog(
...
+        CatalogContainer::Cmaf { init, .. } => Container::Cmaf(fmp4::Wire::from_init(init)?),
+        _ => {
+            warn!("Unsupported video container. Only CMAF is supported.");
+            return Ok(None);
+        }
+    };
+
+    let description = match (catalog_type, &container) {
+        (CatalogType::Msf, Container::Cmaf(wire)) => Some(extract_codec_description(wire)?),
+        (_, _) => config.description.clone(),
```
JBRS307: Add some explanation

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -171,6 +97,87 @@ async fn read_msf_catalog(
...
+            return Ok(None);
+        }
+    };
+
+    // TODO: (@jbrs): It needs to be reconsidered how decoder config should be handled,
+    // where should it be extracted from the container, here or in the decoder.
+    // Return to that when adding additional containers.
+    let description = match (catalog_type, &container) {
+        (CatalogType::Msf, Container::Cmaf(wire)) => Some(extract_codec_description(wire)?),
+        (_, _) => config.description.clone(),
```
JBRS307: Add some explanation

JBRS307:
I merged the #2084 because after refactor it turned out to be just +4 lines.

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -109,66 +87,95 @@ async fn read_msf_catalog(
...
+    let audio = discover_audio(&catalog.audio)?;
+
+    if video.is_none() && audio.is_none() {
+        return Err(MoqCatalogError::CatalogNoTracks);
+    }
+
+    Ok(DiscoveredTracks { video, audio })
+}
+
+fn discover_video(
```
wkozyra95:
```suggestion
fn find_first_video(
```
discover suggest activity, you are searching through data structure (catalog), also given the internal logic adding first in the name would be helpfull for to understand which video track would be chosen

maybe event find_first_video_x
where x would something like track/renditon/.. not sure what moq terminology is correct here

wkozyra95: I don't like the DiscoveredVideo type too, but I don't think I understand well enough what it represents to suggest alternative

wkozyra95 (approved):

smelter-core/src/pipeline/moq/connection/catalog.rs
JBRS307: These types contain track info:
- name
- container
- codec
- Optional description

Renamed them to `VideoTrack` and `AudioTrack`

JBRS307: Plus `DiscoveredTracks` was removed

wkozyra95 (approved):

PR #2077, opened 2026-06-17
[gpu-video] Add decoder tests

noituri:
Snapshots: https://github.com/smelter-labs/smelter-snapshot-tests/pull/91


integration-tests/Cargo.toml
```diff
@@ -29,7 +29,7 @@ serde_json = { workspace = true }
 ffmpeg-next = { workspace = true }
 socket2 = { workspace = true }
 bytes = { workspace = true }
-webrtc = {workspace = true }
```
jerzywilczek, line 32: maybe we should just add a taplo check to the lint CI

integration-tests/src/gpu_video_tests/h264_decode_tests.rs
jerzywilczek, line 1: would be nice to hava a basic big buck bunny test, to have something regular instead of just problematic streams

integration-tests/src/gpu_video_tests/harness/decode_test_runner.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fn verify_decoders(
+        &self,
+        mut reference_decoder: BufferedDecoder,
+        mut gv_decoders: Vec<BufferedDecoder>,
+    ) {
+        let mut source = File::open(gpu_video_dumps_dir_path().join(&self.dump_file_path)).unwrap();
+
+        let mut buffer = [0; 256];
```
jerzywilczek:
```suggestion
        let mut buffer = [0; 4096];
```
256 is very small, lets keep it an even page

jerzywilczek (approved):

integration-tests/Cargo.toml
noituri: agreed

PR #2078, opened 2026-06-19
Added authentication via `auth_token` to MoQ server input

JBRS307:
- #2009

---

The user MUST provide the `auth_token` during input registration.

When connecting to the MoQ server via token must be provided as query param e.g. `https://example.com/input_name?token=<THE_PROVIDED_TOKEN>`

### DIFF
--- a/integration-tests/examples/demo/inputs/moq.rs
+++ b/integration-tests/examples/demo/inputs/moq.rs
@@ -11,6 +11,7 @@ use serde_json::json;
 #[derive(Debug, Serialize, Deserialize)]
 pub struct MoqInput {
     pub name: String,
+    pub auth_token: String,
 }
 
 impl MoqInput {
@@ -19,7 +20,8 @@ impl MoqInput {
             "type": "moq_server",
             "decoder_map": {
                 "h264": "ffmpeg_h264",
-            }
+            },
+            "auth_token": self.auth_token,
         })
     }
 
@@ -32,12 +34,14 @@ impl MoqInput {
 
 pub struct MoqInputBuilder {
     name: String,
+    auth_token: String,
 }
 
 impl MoqInputBuilder {
     pub fn new() -> Self {
         let name = Self::generate_name();
-        Self { name }
+        let auth_token = "example".to_string();
+        Self { name, auth_token }
     }
 
     fn generate_name() -> String {
@@ -48,7 +52,7 @@ impl MoqInputBuilder {
     }
 
     pub fn prompt(self) -> Result<Self> {
-        self.prompt_name()
+        self.prompt_name()?.prompt_token()
     }
 
     fn prompt_name(self) -> Result<Self> {
@@ -60,12 +64,29 @@ impl MoqInputBuilder {
         }
     }
 
+    fn prompt_token(self) -> Result<Self> {
+        let token_input = Text::new("Auth token (ESC for \"example\"):").prompt_skippable()?;
+
+        match token_input {
+            Some(token) if !token.trim().is_empty() => Ok(self.with_token(token)),
+            None | Some(_) => Ok(self),
+        }
+    }
+
     pub fn with_name(mut self, name: String) -> Self {
         self.name = name;
         self
     }
 
+    pub fn with_token(mut self, auth_token: String) -> Self {
+        self.auth_token = auth_token;
+        self
+    }
+
     pub fn build(self) -> MoqInput {
-        MoqInput { name: self.name }
+        MoqInput {
+            name: self.name,
+            auth_token: self.auth_token,
+        }
     }
 }
--- a/smelter-api/src/input/moq_server.rs
+++ b/smelter-api/src/input/moq_server.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -9,6 +9,9 @@ use super::SideChannel;
 #[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema)]
 #[serde(deny_unknown_fields)]
 pub struct MoqInputServer {
+    /// Token used for authentication in MoQ server input. The broadcaster must provide
+    /// it as a `token` query parameter when connecting
+    pub auth_token: Arc<str>,
     /// (**default=`false`**) If input is required and the stream is not delivered
     /// on time, then Smelter will delay producing output frames.
     pub required: Option<bool>,
--- a/smelter-core/src/pipeline/moq/server.rs
+++ b/smelter-core/src/pipeline/moq/server.rs
@@ -151,7 +151,8 @@ async fn handle_incoming_connection(
     ctx: &Arc<PipelineCtx>,
 ) -> Result<(MoqSession, OriginConsumer, Ref<InputId>), MoqServerError> {
     let Some(url) = request.url() else {
-        reject_request(request, 400).await;
+        // This error is safe to ignore as it comes only if the connection is already dead.
+        _ = request.close(400).await;
         return Err(MoqServerError::UrlNotFound);
     };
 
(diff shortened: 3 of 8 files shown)


smelter-core/src/pipeline/moq/server.rs
```diff
@@ -151,6 +151,7 @@ async fn handle_incoming_connection(
         }
         return Err(MoqServerError::UrlNotFound);
     };
+    tracing::info!("{url:#?}");
```
JBRS307:
```suggestion
```

integration-tests/examples/demo/inputs/moq.rs
```diff
@@ -19,7 +19,9 @@ impl MoqInput {
             "type": "moq_server",
             "decoder_map": {
                 "h264": "ffmpeg_h264",
-            }
+            },
+            // TODO: (@jbrs) Before merge parametrize it.
```
JBRS307: resolve

smelter-api/src/input/moq_server.rs
```diff
@@ -9,6 +9,8 @@ use super::SideChannel;
 #[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema)]
 #[serde(deny_unknown_fields)]
 pub struct MoqInputServer {
+    /// Token used for authentication in MoQ protocol.
```
JBRS307:
```suggestion
    /// Token used for authentication in MoQ server input.
```

JBRS307 (commented):
~There is a bug, that causes messages about invalid or missing token to display twice~
FIXED

smelter-core/src/protocols/moq.rs
```diff
@@ -28,6 +29,12 @@ pub enum MoqServerError {
     #[error("URL path \"{0}\" not found among registered inputs.")]
     PathNotFound(Arc<str>),
 
+    #[error("Invalid bearer token for input \"{0}\"")]
+    InvalidToken(InputId),
+
+    #[error("Missing bearer token for input \"{0}\"")]
+    MissingToken(InputId),
```
JBRS307:
```suggestion
    #[error("Invalid authentication token for input \"{0}\"")]
    InvalidToken(InputId),

    #[error("Missing authentication token for input \"{0}\"")]
    MissingToken(InputId),
```

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -178,6 +178,22 @@ async fn handle_incoming_connection(
         }
     };
 
+    let Some(auth_token) = url.query_pairs().find_map(|(k, v)| match k == "token" {
```
wkozyra95: hard to decipher this logic

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -178,6 +178,22 @@ async fn handle_incoming_connection(
...
+        true => Some(v),
+        false => None,
+    }) else {
+        if let Err(error) = request.close(401).await {
+            warn!(%error, "Error while rejecting MoQ connection.");
+        }
+        return Err(MoqServerError::MissingToken(input_ref.id().clone()));
+    };
```
noituri: maybe split this into 2 parts? There's a lot happening here

smelter-core/src/pipeline/moq/state.rs
```diff
@@ -88,6 +93,26 @@ impl MoqServerState {
...
+            .get(input_ref)
+            .ok_or(MoqServerError::InputNotFound(input_ref.id().clone()))?;
+
+        match provided_token == input.auth_token.as_ref() {
+            true => Ok(()),
+            false => {
+                let nanos = rand::rng().random_range(0..1000);
+                std::thread::sleep(Duration::from_nanos(nanos));
+                Err(MoqServerError::InvalidToken(input_ref.id().clone()))
+            }
```
noituri: why?

JBRS307: It is work in progress, I am converting to draft.

But to answer the question, it was meant to prevent non-constant time string comparison attack (but it won't be done this way)

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -151,33 +151,48 @@ async fn handle_incoming_connection(
     ctx: &Arc<PipelineCtx>,
 ) -> Result<(MoqSession, OriginConsumer, Ref<InputId>), MoqServerError> {
     let Some(url) = request.url() else {
-        if let Err(error) = request.close(400).await {
-            warn!(%error, "Error while rejecting MoQ connection.");
-        }
+        reject_request(request, 400).await;
```
wkozyra95: If you see patterns like that it should give you a pause, you are handling an error in a bunch of places, but still return it. You attempted to extract the repeated part by extracting it to a function, but I don't think this is helping here.

There is no point in forcing cleanup code here, return the error, and let caller handle closing

JBRS307: The `reject_request` here handles the error caused by the call to `request.close()`. The error returned is the cause of request rejection which I think is more important to return even if there is a problem with.

HOWEVER, after investigating the matter I think it is safe to completely ignore this error, as it may arise in 2 cases here:
1. Invalid status code, which cannot happen as that is hardcoded to valid 4xx values
2. Connection is already dead which is something we want anyway.

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -146,48 +161,40 @@ async fn run_accept_loop(
...
         Ok(decoded) => decoded.into_owned(),
         Err(error) => {
-            if let Err(error) = request.close(400).await {
-                warn!(%error, "Error while rejecting MoQ connection.");
-            }
             return Err(MoqServerError::UrlDecodeFailed(error));
         }
     };
 
     let input_ref = match moq_inputs.find_by_url(&input_name) {
```
wkozyra95:
```
 let input_ref = match moq_inputs.find_by_url(&input_name)?;
```

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -146,48 +161,40 @@ async fn run_accept_loop(
...
-        if let Err(error) = request.close(400).await {
-            warn!(%error, "Error while rejecting MoQ connection.");
-        }
         return Err(MoqServerError::UrlNotFound);
     };
 
     let input_name_encoded = url.path().trim_start_matches('/');
     let input_name = match urlencoding::decode(input_name_encoded) {
         Ok(decoded) => decoded.into_owned(),
         Err(error) => {
```
wkozyra95:
```
        Err(error) =>  return Err(MoqServerError::UrlDecodeFailed(error));
```
if formatter allows

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -119,17 +119,32 @@ async fn run_accept_loop(
...
-                    warn!(
-                        "MoQ connection rejected: {}",
-                        ErrorStack::new(&error).into_string()
-                    );
-                    continue;
-                }
-            };
+        let (origin, input_ref) = match handle_incoming_connection(&request, &moq_inputs).await {
+            Ok(sci) => sci,
+            Err(error) => {
```
wkozyra95:
```suggestion
            Err(err) => {
```
nit, but let's keep it consistent (in other places too)

smelter-core/src/pipeline/moq/server.rs
```diff
@@ -119,17 +119,32 @@ async fn run_accept_loop(
...
+                warn!(
+                    "MoQ connection rejected: {}",
+                    ErrorStack::new(&error).into_string()
+                );
+                let rejection_code = match error {
```
wkozyra95: it's fine here, but it's borderline, if the amount of logic grew, it should be extracted.

wkozyra95 (approved):

smelter-core/src/pipeline/moq/server.rs
JBRS307: I'll keep `error` wherever it is used directly in the `warn!`, like `warn!(%error, "something...")`, as it looks nicer.  Will shorten that in other places.

PR #2080, opened 2026-06-19
[gpu-video] Fix decoding video with a dynamic resolution

noituri:
This was only an issue on my integrated AMD GPU, discrete AMD GPU had no problems with decoding this.

If new SPS changes resolution to a smaller one, we don't recreate DPB which causes the video resource info to contain larger `coded_extent`. So now when an IDR frame is decoded, the correct `coded_extent` is set for the current SPS.

### DIFF
--- a/gpu-video/src/vulkan_decoder.rs
+++ b/gpu-video/src/vulkan_decoder.rs
@@ -350,7 +350,6 @@ impl<'a> VulkanDecoder<'a> {
             )))?;
 
         let cropped_extent = sps.size()?;
-        let coded_extent = sps.coded_size()?;
         let color_space = ColorSpace::from(sps);
         let color_range = ColorRange::from(sps);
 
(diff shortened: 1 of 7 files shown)


gpu-video/src/vulkan_decoder.rs
```diff
@@ -359,6 +360,10 @@ impl<'a> VulkanDecoder<'a> {
                 self.tracker.command_buffer_pools.decode.begin_buffer()?,
                 &mut self.tracker,
             )?;
+            // TODO: Will this be safe if we start decoding without CPU wait?
+            video_session_resources
+                .decoding_images
+                .update_coded_extent(coded_extent)?;
         }
```
jerzywilczek: why not as part of ensuring session?

gpu-video/src/vulkan_decoder.rs
```diff
@@ -359,6 +360,10 @@ impl<'a> VulkanDecoder<'a> {
                 self.tracker.command_buffer_pools.decode.begin_buffer()?,
                 &mut self.tracker,
             )?;
+            // TODO: Will this be safe if we start decoding without CPU wait?
```
jerzywilczek: this just changes metadata of the dpb. the commands that were already submitted should be fine, as they copy the metadata afair

gpu-video/src/vulkan_decoder.rs
noituri: I completely missed the above `ensure_sessions`

jerzywilczek (approved):

PR #2081, opened 2026-06-22
Compare `WebRTC` and `RTMP` server auth keys using `SHA3-512` hashes instead of plain text passwords.

JBRS307:
The goal of that is to mitigate timing attacks based on non-constant time string comparison.


smelter-core/src/pipeline/webrtc/whep_output/state.rs
```diff
@@ -127,32 +127,34 @@ impl WhepOutputsState {
...
+    /// SHA3-512 hash (lowercase hex) of the bearer token, not the plaintext token.
+    pub bearer_token_hash: Option<Arc<str>>,
     pub video_options: Option<WhepVideoConnectionOptions>,
     pub audio_options: Option<WhepAudioConnectionOptions>,
 }
 
 #[derive(Debug)]
 pub(crate) struct WhepOutputConnectionState {
-    pub bearer_token: Option<Arc<str>>,
+    /// SHA3-512 hash (lowercase hex) of the bearer token, not the plaintext token.
```
wkozyra95: don't store hash, store token and calculate hash just for comparison

noituri (approved):
lgtm

smelter-core/src/pipeline/utils/authentication.rs
```diff
@@ -0,0 +1,9 @@
+use sha3::{Digest, Sha3_512};
+
+/// Validates a provided token against the expected one by comparing their
+/// SHA3-512 hashes. Returns `true` when they match.
+pub(crate) fn validate_token(expected_token: &str, provided_token: &str) -> bool {
```
wkozyra95: Inline that function (duplicate)
- authentication is bad name, this function is used by authentication code, but itself does not do it
- I would expect function named this way to return Result no bool
- 3 lines of code is not worth an extra util

smelter-core/src/pipeline/webrtc/bearer_token.rs
```diff
@@ -19,7 +18,7 @@ pub(super) fn generate_token() -> Arc<str> {
     Arc::from(token)
 }
 
-pub(super) async fn validate_token(
+pub(super) fn validate_bearer_token(
```
wkozyra95: it's not a big problem, but you can revert name change after inlining validate_token

PR #2083, opened 2026-06-23
Fix panic on malformed NALU

wkozyra95:


JBRS307 (approved):

noituri (approved):
Did we recently encounter this?

wkozyra95:
> Did we recently encounter this?

I hit that when trying to launch mp4 generated on my phone(it was hdr with 10bit pixel format). it expected that it didn't work, but it should not cause panic. I did not investigate why that file would cause that issue

PR #2090, opened 2026-06-25
[gpu-video] Make `VideoInstance` and `VideoAdapter` backend agnostic

noituri:
The plan is to have each backend specific object wrapped in `Video*Backend` trait.
In this PR I only did instance and adapter.

Errors are split into backend specific and backend agnostic (user-facing) errors. If some error can't be converted to backend agnostic error, it's converted to string and put into `VideoBackendError`

### DIFF
--- a/gpu-video/src/backends/vulkan/vulkan_adapter.rs
+++ b/gpu-video/src/backends/vulkan/vulkan_adapter.rs
@@ -1,0 +1,509 @@
+use ash::vk;
+use std::ffi::CStr;
+use tracing::{debug, debug_span, warn};
+
+use crate::{
+    VideoDeviceInitError,
+    adapter::{DeviceType, VideoAdapterBackend, VideoAdapterInfo},
+    backends::vulkan::vulkan_instance::VulkanInstance,
+    capabilities::{DecodeCapabilities, EncodeCapabilities},
+    device::{
+        DECODE_CODEC_EXTENSIONS, DECODE_EXTENSIONS, ENCODE_CODEC_EXTENSIONS, ENCODE_EXTENSIONS,
+        REQUIRED_EXTENSIONS, VideoDevice, VideoDeviceDescriptor,
+        caps::{NativeDecodeCapabilities, NativeEncodeCapabilities},
+        queues::{QueueIndex, QueueIndices},
+    },
+};
+
+#[cfg(feature = "wgpu")]
+mod wgpu_api;
+#[cfg(feature = "wgpu")]
+pub(crate) use wgpu_api::*;
+
+/// Represents a handle to a physical device.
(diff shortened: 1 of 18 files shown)


gpu-video/src/backends/vulkan/vulkan_adapter.rs
noituri, line 1: all of it comes from `adapter.rs` + renames + new error enum

gpu-video/src/backends/vulkan/vulkan_instance.rs
noituri, line 1: similar changes to adapter

gpu-video/src/device.rs
```diff
@@ -745,6 +745,25 @@ impl Deref for EncodingDevice {
...
+
+impl From<VulkanDeviceInitError> for VideoDeviceInitError {
+    fn from(err: VulkanDeviceInitError) -> Self {
+        Self::BackendError(crate::VideoBackendError {
+            message: err.to_string(),
+            source: Box::new(err),
+        })
+    }
+}
+
```
noituri, line 772: I'm afraid that in the future we may forget to correctly handle this. Not every vulkan error has to be a backend error

gpu-video/src/backends/vulkan/vulkan_adapter.rs
jerzywilczek: I very much appreciate this comment

gpu-video/src/device.rs
jerzywilczek: then turn it into a match statement
```rs
match err {
    VulkanDeviceInitError::VkError(_) | VulkanDeviceInitError::WgpuError(_) => err.to_string()
}
```

gpu-video/src/instance.rs
```diff
@@ -18,135 +16,51 @@ pub struct VideoInstanceDescriptor {
     pub enable_api_dump: bool,
 }
 
+pub(crate) trait VideoInstanceBackend {
+    fn iter_adapters<'a>(
+        &'a self,
+    ) -> Result<Box<dyn Iterator<Item = VideoAdapter<'a>> + 'a>, VideoInstanceInitError>;
```
jerzywilczek, line 22: i know you probably tried, but why can't we do `-> impl Iterator<...>`?

jerzywilczek: Is it because we need `VideoInstanceBackend` to be object-safe and rpitit makes a trait non-object-safe? sorry, a bit rusty on the new features with very long shorthand names

jerzywilczek (approved):
I think this is nice

gpu-video/src/instance.rs
noituri: yes that's the reason behind it. I tried to get around this but in the end this was the simplest solution

PR #2091, opened 2026-06-26
[gpu-video] Make `VideoDevice` backend agnostic

noituri:
This one will be probably by far the largest PR in this task. Many things had to be moved to `backends/vulkan/`.

Some methods in `VulkanDevice` are private under `expose-backends` feature. I'll expose them in a separate PR

Depends on: https://github.com/software-mansion/smelter/pull/2090


gpu-video/src/capabilities.rs
noituri, line 1: moved user-facing caps from device/caps

gpu-video/src/device/caps.rs
```diff
@@ -59,50 +66,6 @@ pub(crate) fn query_video_format_properties<'a>(
...
-    pub max_height: u32,
-    /// The supported rate control modes in bitflag form
-    pub supported_rate_control: vk::VideoEncodeRateControlModeFlagsKHR,
-    /// Maximum number of back references a P-frame can have
-    pub max_references: u32,
-    /// The count of [Vulkan Video encode quality levels](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#encode-quality-level)
-    pub quality_levels: u32,
-}
-
 #[derive(Debug, Clone)]
```
noituri, line 106: moved to capabilities.rs

gpu-video/src/backends/vulkan/vulkan_device.rs
noituri, line 1: copied from device.rs, added `VideoDeviceBackend` impl + errors.

`VulkanDevice` is exposed via `expose-backends` feature but currently it can't be used for anything useful (I'll work on it later)

gpu-video/src/capabilities.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+/// Rate control capabilities supported by the encode profile
+#[derive(Debug, Clone, Copy)]
+pub struct RateControlCapabilities {
+    pub vbr_supported: bool,
+    pub cbr_supported: bool,
+}
+
```
noituri, line 60: previously a flag from vulkan was used here

gpu-video/src/device.rs
```diff
@@ -204,6 +174,14 @@ impl From<ColorSpace> for CodecColorDescription {
...
 }
 
+#[derive(Debug, Clone, Copy)]
+pub enum EncoderPreset {
+    /// A hint for the encoder to prioritize speed over quality
+    Speed,
+    /// A hint for the encoder to prioritize quality over speed
+    Quality,
+}
+
```
noituri: replacement for vulkan's quality level

gpu-video/src/backends/vulkan/vulkan_device.rs
```diff
@@ -0,0 +1,734 @@
...
+    }
+
+    #[cfg(feature = "transcoder")]
+    fn create_transcoder(
+        self: Arc<Self>,
+        parameters: crate::parameters::TranscoderParameters,
+    ) -> Result<crate::vulkan_transcoder::Transcoder, crate::vulkan_transcoder::VideoTranscoderError>
+    {
+        crate::vulkan_transcoder::Transcoder::new(self, parameters)
+    }
```
jerzywilczek: wouldnt we want to extract these to separate traits? not a fan of putting these `cfg`s on the methods when implementing this trait

gpu-video/src/device.rs
```diff
@@ -257,532 +234,52 @@ pub struct EncoderParametersH265 {
...
-
-#[derive(thiserror::Error, Debug)]
-pub enum VulkanDeviceInitError {
-    #[error("Vulkan error: {0}")]
-    VkError(#[from] vk::Result),
+    fn create_wgpu_textures_decoder_h264(
+        self: Arc<Self>,
+        wgpu_device: wgpu::Device,
+        parameters: DecoderParameters,
+    ) -> Result<crate::WgpuTexturesDecoder, VideoDecoderError>;
```
jerzywilczek: same here

gpu-video/src/vulkan_video.rs
```diff
@@ -27,8 +18,10 @@ pub mod parameters {
     #[cfg(feature = "wgpu")]
     pub use crate::wgpu_helpers::WgpuConverterParameters;
 
+    // TODO: What about these
     pub use ash::vk::VideoDecodeUsageFlagsKHR as DecoderUsageFlags;
 
+    // TODO: And what about these
```
jerzywilczek: What about these indeed

gpu-video/src/lib.rs
```diff
@@ -5,41 +5,39 @@ pub mod backends;
...
 mod vulkan_encoder;
 #[cfg(all(vulkan, feature = "transcoder"))]
 mod vulkan_transcoder;
-#[cfg(all(vulkan, feature = "wgpu"))]
-pub(crate) mod wgpu_helpers;
-#[cfg(vulkan)]
-pub(crate) mod wrappers;
 
-// TODO: This could be inlined with lib.rs when metal support is added
+// TODO: Rename to prelude? Or exports?
```
jerzywilczek: `exports` imo

jerzywilczek (commented):
Generally looks good, however I am a bit worried that implementing a new backend is not super simple. I think we should have a `Backend` trait which contains all traits and things that need to be implementedto add a backend. I also think we should try to eradicate almost all `#[cfg(...)]`s (like I noted in two places here). Theyre getting too spaghettified, I can no longer tell what implications will adding a `#[cfg]` on a method have in other places in the codebase.

gpu-video/src/vulkan_video.rs
jerzywilczek: I think we need to add our enums for these, and map them to no-ops on VT (I think maybe one of them maps cleanly to a VT functionality, the rest have no equivalents)

noituri: i forgor about these. I put them in the enum as you suggested.

gpu-video/src/backends/vulkan/codec/h264.rs
```diff
@@ -142,3 +144,19 @@ impl<'a> CodecSpecificEncoderQualityLevelProperties
...
+            }
+            H264Profile::Main => {
+                ash::vk::native::StdVideoH264ProfileIdc_STD_VIDEO_H264_PROFILE_IDC_MAIN
+            }
+            H264Profile::High => {
+                ash::vk::native::StdVideoH264ProfileIdc_STD_VIDEO_H264_PROFILE_IDC_HIGH
+            }
+        }
+    }
+}
```
noituri, line 162: this was moved from `vulkan_video`. Same for `H265Profile`

gpu-video/src/backends/vulkan/vulkan_device.rs
```diff
@@ -0,0 +1,530 @@
...
+        }
+        if framerate.numerator == 0 {
+            return Err(VideoEncoderError::ParametersError {
+                field: "framerate",
+                problem: format!("Framerate is {framerate:?}. The numerator should be != 0.",),
+            });
+        }
+        let usage_flags = encoder_parameters.usage_flags.unwrap_or_default().into();
+        let tuning_mode = vk::VideoEncodeTuningModeKHR::DEFAULT;
+        let content_flags = encoder_parameters.content_flags.unwrap_or_default().into();
```
noituri: what should be the `tuning_mode` here? We already have `EncoderPreset` which sets the quality level, but maybe it should also set the `tuning_mode`? I avoided adding `EncoderTuningMode` because it seems similar to `EncoderPreset`. Any thoughts? Tuning mode has more variants than encoder preset but idk what would it map to on metal

gpu-video/src/backends.rs
noituri, line 1: lmk what you think about this. The heatwave already fried my brain so I probably missed something.
All vulkan `cfg`s are limited to this file (except places I didn't refactor yet). There's still plenty of `cfg(feature = "wgpu")`, but I don't think we can easily get rid of them unfortunately.

I also thought about making something like this:
```rs
trait Backend {
  type Adapter: VideoAdapterBackend;
  type Instance: VideoInstanceBackend;
  type Device: VideoDeviceBackend;
}
```

so it would be clear what has to be implemented by a new backend, but this is not dyn-compatible so it would require a bit of trait magic (probably the same thing wgpu does). Idk if it's worth the hassle.

Currently if you'd try to create a new backend, you would implement `CoreBackend` and `WgpuBackend`, this would force you to implement `VideoInstanceBackend`, which would then force you to implement `VideoAdapterBackend` and then `VideoDeviceBackend` etc etc.

In the end the compiler would lead you to the complete implementation of the backend but it's an **ugly** way of doing it.

gpu-video/src/backends.rs
```diff
@@ -1,2 +1,49 @@
...
+        &self,
+        wgpu_adapter: &wgpu::Adapter,
+        desc: &crate::parameters::VideoDeviceDescriptor,
+    ) -> Result<(wgpu::Device, wgpu::Queue), crate::VideoDeviceInitError>;
+}
+
+pub(crate) fn default_backend() -> impl CoreBackend {
+    #[cfg(vulkan)]
+    vulkan::VulkanBackend
+}
```
noituri, line 40: If we'd have multiple backends supported on one platform, it would probably make sense  to have some logic that chooses the correct backend and then `VideoInstance` would store it.

gpu-video/src/exports.rs
```diff
@@ -27,11 +18,34 @@ pub mod parameters {
...
+    }
+
+    #[derive(Debug, Clone, Copy, Default)]
+    pub enum EncoderContent {
+        #[default]
+        Default,
+        Camera,
+        Desktop,
+        Rendered,
+    }
```
noituri, line 49: Didn't add `EncoderTuningMode` for the reasons explain in `vulkan_device.rs`

gpu-video/src/backends.rs
jerzywilczek: you can leave it as-is, i think its ok

gpu-video/src/backends/vulkan/vulkan_device.rs
```diff
@@ -550,12 +353,19 @@ impl VulkanDevice {
...
+        let (quality_level, tuning_mode) = match encoder_parameters.preset {
+            EncoderPreset::HighQuality => (
+                native_profile_caps
+                    .quality_level_properties
+                    .len()
+                    .saturating_sub(1) as u32,
+                vk::VideoEncodeTuningModeKHR::HIGH_QUALITY,
+            ),
+            EncoderPreset::Balanced => (
+                native_profile_caps.quality_level_properties.len() as u32 / 2,
```
jerzywilczek, line 365: <3

jerzywilczek: life is good man

jerzywilczek (approved):

gpu-video/src/backends/vulkan/vulkan_device.rs
noituri: life is good <3

PR #2093, opened 2026-06-29
Added `ffmpeg_options` to HLS output.

### LINKED ISSUE #2092: Add fMP4 (CMAF) segment support to HLS output
The HLS output currently emits MPEG-TS segments only. Add an option (e.g. segment_type: "mpeg_ts" | "fmp4", default mpeg_ts) on HlsOutput so segments can be fMP4/CMAF.

Why it touches the encoder
fMP4/ISOBMFF requires AVCC H.264 (length-prefixed NALUs + avcC), whereas MPEG-TS uses Annex-B. We write encoded packets straight to the muxer with no bitstream filter, so the encoder output must match the container. Today bitstream_format is hardcoded to AnnexB (hls_into.rs:100,116). For fMP4 it must be Avcc, derived from the new option (fmp4 ⟹ Avcc, mpeg_ts ⟹ AnnexB).

Note: the hardcoded segment_format/segment_list_* options in `hls_output.rs:79` are segment-muxer options and are ignored by the hls muxer.

JBRS307:
- Closes #2092

### DIFF
--- a/smelter-core/src/pipeline/hls/hls_output.rs
+++ b/smelter-core/src/pipeline/hls/hls_output.rs
@@ -75,14 +75,16 @@ impl HlsOutput {
             None => None,
         };
 
-        // 0 means no list size limit
-        let list_size = options.max_playlist_size.unwrap_or(0).to_string();
         let mut ffmpeg_options = FfmpegOptions::from(&[
             ("segment_format", "mpegts"),
             ("segment_list_type", "m3u8"),
             ("segment_list_flags", "cache+live"),
             ("hls_flags", "delete_segments"),
-            ("hls_list_size", list_size.as_str()),
+            (
+                "hls_list_size",
+                // 0 means no list size limit
+                &options.max_playlist_size.unwrap_or(0).to_string(),
+            ),
         ]);
         ffmpeg_options.append(&options.raw_options);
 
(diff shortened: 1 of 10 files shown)


ts/smelter/src/api.generated.ts
JBRS307, line 1: MoQ changes here where added with auto generation. I will add MoQ fully to the `ts-sdk` only after the server input is completed.

smelter-core/src/pipeline/hls/hls_output.rs
```diff
@@ -75,20 +75,19 @@ impl HlsOutput {
...
             ("segment_format", "mpegts"),
             ("segment_list_type", "m3u8"),
             ("segment_list_flags", "cache+live"),
             ("hls_flags", "delete_segments"),
-            (
-                "hls_list_size",
-                // 0 means no list size limit
-                &options.max_playlist_size.unwrap_or(0).to_string(),
-            ),
+            ("hls_list_size", list_size.as_str()),
```
wkozyra95: Not a strong opinion, but I prefer the original structure

ts/smelter/src/api.generated.ts
```diff
@@ -76,6 +76,27 @@ export type RegisterInput =
        */
       side_channel?: SideChannel | null;
     }
+  | {
```
wkozyra95: revert moq changes

wkozyra95 (approved):

PR #2094, opened 2026-06-29
Added `vp8` and `vp9` to MoQ server input.

JBRS307:
- #2009 

---

One word about API for `decoder_map`. It mirrors the RTMP server input, i. e. Only one option for `h264` which chooses software vs hardware decoding. If this option is set and detected codec is vp8 or vp9 it will be ignored and take no effect.

### DIFF
--- a/smelter-core/src/pipeline/moq/connection.rs
+++ b/smelter-core/src/pipeline/moq/connection.rs
@@ -380,17 +380,31 @@ fn spawn_video_decoder(
     video: &VideoTrack,
     frame_sender: QueueSender<Frame>,
 ) -> Result<DecoderThreadHandle, MoqConnectionError> {
-    let transformer = match (&video.description, &video.codec, &video.container) {
-        (None, VideoCodec::H264, Container::Cmaf(_)) => {
-            return Err(MoqConnectionError::MissingAvcc);
-        }
-        (Some(desc), VideoCodec::H264, _) => {
-            let h264_config = H264AvcDecoderConfig::parse(desc.clone())
-                .map_err(|_| MoqConnectionError::InvalidAvcc)?;
+    let (decoder_opt, transformer) = match &video.codec {
+        VideoCodec::H264 => {
+            let transformer = match (&video.description, &video.container) {
+                (None, Container::Cmaf(_)) => {
+                    return Err(MoqConnectionError::MissingAvcc);
+                }
+                (Some(desc), _) => {
+                    let h264_config = H264AvcDecoderConfig::parse(desc.clone())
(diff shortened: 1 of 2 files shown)


smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -115,8 +117,8 @@ fn find_first_video(video: &hang::catalog::Video) -> Result<Option<VideoTrack>,
         CatalogContainer::Loc => Container::Loc,
     };
 
-    let description = match (&config.description, &container) {
-        (None, Container::Cmaf(wire)) => match extract_codec_description(wire) {
+    let description = match (&config.description, &codec, &container) {
+        (None, VideoCodec::H264, Container::Cmaf(wire)) => match extract_codec_description(wire) {
```
wkozyra95: This looks weird, for me it signals that you are doing it in the wrong place. It seems to me that optimal logic would be to run extract codec always and just put return value inside. And handle None/Some when spawning decoder

There is no clear signal why is only called for h264

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -404,15 +400,18 @@ fn spawn_video_decoder(
         input_buffer_size: MOQ_MAX_BUFFER,
     };
 
-    let h264_decoder = match decoders.h264 {
-        Some(decoder) => decoder,
-        None => match ctx.graphics_context.has_vulkan_decoder_support() {
-            true => VideoDecoderOptions::VulkanH264,
-            false => VideoDecoderOptions::FfmpegH264,
-        },
+    let decoder_opt = match &video.codec {
```
wkozyra95: look at entire function, you have here 3 steps where each of them is doing custom logic per codec

smelter-core/src/pipeline/moq/connection/catalog.rs
```diff
@@ -115,15 +117,15 @@ fn find_first_video(video: &hang::catalog::Video) -> Result<Option<VideoTrack>,
...
     };
 
-    let description = match (&config.description, &container) {
-        (None, Container::Cmaf(wire)) => match extract_codec_description(wire) {
-            Ok(desc) => Some(desc),
+    let description = match &config.description {
+        Some(desc) => Some(desc.clone()),
+        None => match extract_codec_description(&container) {
+            Ok(desc) => desc,
             Err(error) => {
```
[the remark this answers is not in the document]
JBRS307: May sound scary but is wrong. It is handled in the `handle_video_track` or `handle_audio_track`. If this happens smelter will not subscribe to failing track, if the other track is configured properly it will play.

wkozyra95 (approved):

PR #2097, opened 2026-06-30
Do not block V4L2 input registration on the first frame

ackanir:
The V4L2 registration path dequeued and discarded the first frame inline before returning.

1. This tied two concerns: registering the device vs reading frames from it.
2. The DQBUF runs on the caller thread and blocks until the device delivers a frame. A device that opens and STREAMONs but never produces one (a marginal USB capture, a capture card with no input signal) wedges the caller indefinitely.

This PR moves the skip onto the reader thread, which already owns the capture loop: `new_input` now returns as soon as STREAMON succeeds and doesn't block, and the first-frame discard happens on the first loop iteration.


wkozyra95 (approved):

jerzywilczek (approved):
sure, looks alright to me

Issue #2019, opened 2026-05-25
MoQ client input


JBRS307:
- [x] Client code: #2099 
- [x] ts-sdk: #2121
- [x] docs: https://github.com/smelter-labs/smelter-website/pull/280

PR #2099, opened 2026-07-01
Added MoQ client input.

JBRS307:
- #2019 

---

For receiving the broadcast it uses exactly the same code as MoQ server input. The `smelter-core/src/pipeline/moq/connection.rs`


smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -56,41 +53,40 @@ struct TrackCtx {
...
-                queue_input,
-                broadcast,
-                should_close,
-            )
-            .await;
+            let broadcast_result =
+                handle_broadcast(ctx, input_ref.clone(), queue_input, broadcast_ctx).await;
             if let Err(error) = broadcast_result {
                 warn!(
                     "broadcast failed: {}",
```
JBRS307: Check this comment, optionally add `receiver` info if relevant

smelter-core/src/protocols/moq.rs
```diff
@@ -51,6 +60,21 @@ pub enum MoqServerError {
...
+#[derive(Debug, thiserror::Error)]
+pub enum MoqClientError {
+    #[error("Invalid MoQ relay URL \"{0}\": {1}")]
+    InvalidUrl(Arc<str>, #[source] url::ParseError),
+
+    #[error("Unsupported MoQ relay URL scheme \"{0}\", only \"https\" is supported.")]
+    InvalidScheme(String),
+
+    #[error("Failed to initialize MoQ client: {0}")]
+    ClientInitFailed(#[source] anyhow::Error),
```
wkozyra95: do not use anyhow in public api

smelter-core/src/pipeline/moq/input/server_input.rs
```diff
@@ -1,12 +1,7 @@
...
 
-use crate::{
-    pipeline::{
-        input::Input,
-        moq::state::{MoqInputStateOptions, MoqServerState},
-    },
-    queue::QueueInput,
-};
+use crate::pipeline::moq::server::state::{MoqServerInputStateOptions, MoqServerState};
+use crate::{pipeline::input::Input, queue::QueueInput};
```
wkozyra95: fix imports

smelter-api/src/input/moq_client.rs
```diff
@@ -0,0 +1,44 @@
...
+#[serde(deny_unknown_fields)]
+pub struct MoqClientInput {
+    /// URL of the MoQ relay to connect to. Must use the `https://` scheme.
+    pub endpoint_url: Arc<str>,
+    /// Path of the broadcast to subscribe to on the relay.
+    pub broadcast_path: Arc<str>,
+    /// (**default=`false`**) Skips validation of the relay's TLS certificate.
+    /// Only enable this on trusted networks — it leaves the connection vulnerable
+    /// to man-in-the-middle attacks.
+    pub disable_tls_verification: Option<bool>,
```
wkozyra95: don't see use case for that, if it is needed for development then global env that disables it on all inputs might make more sense,

smelter-core/src/pipeline/moq/input/client_input.rs
```diff
@@ -0,0 +1,143 @@
...
+    should_close: Arc<AtomicBool>,
+    _session: MoqSession,
+}
+
+impl MoqClientInput {
+    pub fn new_input(
+        ctx: Arc<PipelineCtx>,
+        input_ref: Ref<InputId>,
+        options: MoqClientInputOptions,
+    ) -> Result<(Input, InputInitInfo, QueueInput), InputInitError> {
```
wkozyra95, line 31: stup span with input id and input type so the logs are tagged aproriatlly, make sure to pass them to spawned code too

smelter-core/src/pipeline/moq/input/client_input.rs
```diff
@@ -0,0 +1,143 @@
...
+                }
+            };
+
+            let broadcast_ctx = BroadcastCtx {
+                broadcast,
+                decoders,
+                should_close,
+                endpoint_kind: MoqEndpointKind::Client,
+            };
+            if start_broadcast_handler_task(
```
wkozyra95: refactor this

smelter-core/src/pipeline/moq/input/client_input.rs
```diff
@@ -0,0 +1,143 @@
...
+        )?;
+
+        let input = Self {
+            queue_input: queue_input.downgrade(),
+            decoders: options.decoders,
+            should_close: Arc::new(false.into()),
+            _session: session,
+        };
+
+        input.spawn_broadcast_handler(ctx, input_ref, consumer, options.broadcast_path);
```
wkozyra95: the logic flow here is wierd and I'm not sure if it is needed or just consequence of adding stuf on top of each other without considering entire flow.

- First you run async code with block_on `connect`
- Then you spawn task (without waiting for result)
  - Inside that task you are spawning other task without waitiing for results

smelter-core/src/pipeline/moq/input/client_input.rs
```diff
@@ -0,0 +1,143 @@
...
+use moq_native::ClientConfig;
+use smelter_render::InputId;
+use tracing::{info, warn};
+use url::Url;
+
+use crate::prelude::*;
+
+pub struct MoqClientInput {
+    queue_input: WeakQueueInput,
+    decoders: MoqInputDecoders,
```
wkozyra95:
```suggestion
```
This struct will be store in the pipeline, you added those 2 fields just so you don't have to pass that as arguments, this is not a good approach,

If you want to have struct to simplify the flow then create new private one and implement those methods on it

smelter-core/src/pipeline/moq/input/client_input.rs
```diff
@@ -0,0 +1,148 @@
...
+        let rt = ctx.tokio_rt.clone();
+
+        let span = span!(
+            Level::INFO,
+            "MoQ client input",
+            input_id = input_ref.to_string()
+        );
+        rt.spawn(
+            async move {
+                let broadcast = loop {
```
wkozyra95: let's not do that

smelter-core/src/pipeline/moq/input/connection.rs
```diff
@@ -188,8 +138,15 @@ impl BroadcastHandler {
         let epoch = EpochShared::new();
 
         let single_track_stream = video.is_none() || audio.is_none();
+        let BroadcastCtx {
```
wkozyra95: If you need to unpack it and use in different places in code I would move that to the start, but in general I would prefer accessing fields instead of unpacking

smelter-core/src/pipeline/moq/mod.rs
```diff
@@ -1,9 +1,36 @@
...
-pub(crate) use certificate::SelfSignedTlsError;
+use hang::moq_net::Session;
+use std::{ops::Deref, sync::Arc};
+
+pub use input::{MoqClientInput, MoqServerInput};
+pub(crate) use server::SelfSignedTlsError;
 pub(super) use server::{MoqPipelineState, MoqServer, spawn_moq_server};
-pub use server_input::MoqServerInput;
+
+pub(crate) struct MoqSession {
```
wkozyra95:
```suggestion
pub(super) struct MoqSession {
```
Is Moq session used any higher?

smelter-core/src/pipeline/moq/input/client_input.rs
```diff
@@ -0,0 +1,148 @@
...
+        broadcast_path: Arc<str>,
+        should_close: Arc<AtomicBool>,
+        decoders: MoqInputDecoders,
+        queue_input: WeakQueueInput,
+    ) {
+        let rt = ctx.tokio_rt.clone();
+
+        let span = span!(
+            Level::INFO,
+            "MoQ client input",
```
wkozyra95: You are attaching it only here, but not for connect. You can instead create a span and enter it at the start of the input register, and instead of paasing it via funciton args, get current span and pass it to instrument

smelter-core/src/pipeline/moq/input/connection.rs
JBRS307: I moved that to the start of the function. However, I left the unpacking here as the field names of this sctruct are the same as in the `TrackCtx` struct created in this method from these fields and accessing by fields in that context is ugly.

smelter-core/src/pipeline/moq/input/client_input.rs
JBRS307: solved by extracting to function

smelter-core/src/pipeline/moq/input/client_input.rs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+impl Drop for MoqClientInput {
+    fn drop(&mut self) {
+        self.should_close
+            .store(true, std::sync::atomic::Ordering::Relaxed);
+    }
+}
+
+async fn await_broadcast(
```
wkozyra95:
```suggestion
async fn wait_for_broadcast(
```

smelter-core/src/pipeline/moq/input/client_input.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        let Some((path, Some(broadcast))) = consumer.announced().await else {
+            warn!(%broadcast_path, "MoQ session closed before announcing required broadcast.");
+            return None;
+        };
+
+        if path.as_str().trim_start_matches("/") == broadcast_path.as_ref().trim_start_matches("/")
+        {
+            return Some(broadcast);
+        }
```
wkozyra95:
```suggestion
        let incoming_path = path.as_str().trim_start_matches("/");
        let expected_path = broadcast_path.as_ref().trim_start_matches("/");
        if incoming_path == expected_path {
            return Some(broadcast);
        }
```

wkozyra95 (approved):

PR #2105, opened 2026-07-02
Configurable layout limit

wkozyra95:


smelter-render/src/transformations/layout/shader.rs
```diff
@@ -23,13 +26,27 @@ pub struct LayoutShader {
...
         let scope = WgpuErrorScope::push(&wgpu_ctx.device);
 
+        assert!(
+            SHADER_SOURCE.contains(MAX_LAYOUTS_WGSL_DECLARATION),
+            "MAX_LAYOUTS_COUNT declaration not found in apply_layouts.wgsl"
+        );
+        let shader_source = SHADER_SOURCE.replace(
+            MAX_LAYOUTS_WGSL_DECLARATION,
+            &format!("const MAX_LAYOUTS_COUNT: u32 = {max_layouts_count};"),
+        );
```
jerzywilczek, line 42: there actually is a dedicated way to do this:
https://docs.rs/wgpu/latest/wgpu/struct.PipelineCompilationOptions.html

I think it would be better to use it.

wkozyra95: Apparentlly it has to be a const expression not override expressions

> A variable in the [uniform](https://www.w3.org/TR/WGSL/#address-spaces-uniform) address space is a uniform buffer variable. Its [store type](https://www.w3.org/TR/WGSL/#store-type) must be a [host-shareable](https://www.w3.org/TR/WGSL/#host-shareable) [constructible](https://www.w3.org/TR/WGSL/#constructible) type, and must satisfy the [address space layout constraints](https://www.w3.org/TR/WGSL/#address-space-layout-constraints).

and to be host sharable it has to be [creation-fixed footprint](https://www.w3.org/TR/WGSL/#creation-fixed-footprint)

PR #2107, opened 2026-07-03
[gpu-video] Make `Transcoder` backend agnostic

noituri:


gpu-video/build.rs
```diff
@@ -15,13 +15,25 @@ fn main() {
...
+#[cfg(all(
+    any(
+        windows,
+        all(
+            unix,
+            not(any(target_os = "macos", target_os = "ios", target_os = "emscripten"))
+        )
+    ),
+    feature = "transcoder"
+))]
```
noituri, line 28: :/

gpu-video/src/lib.rs
noituri, line 1: all of those vulkan cfgs can be removed once the metal backend is added. Currently it doesn't compile on macos because `default_backend()` needs to return a value

jerzywilczek (approved):

PR #2109, opened 2026-07-06
[gpu-video] Fall back to a compute-capable queue when no dedicated one exists

fschutt:
## What

`VulkanAdapter` selection requires a queue family that exposes `COMPUTE` without `GRAPHICS` (a dedicated async-compute queue). Devices that only advertise `COMPUTE` on the combined graphics family are rejected in `VulkanAdapter::new` — it returns `None`, so the whole adapter is dropped with "Cannot find a suitable physical device".

NVIDIA Maxwell cards (GTX 950/960/970/980) are the common case: they expose `COMPUTE` only on the graphics family, so they fail this check even though they fully support `VK_KHR_video_decode_h264`.

## Testing

With this patch, I can get GPU Video decoding working on my NVIDIA GTX 960:

image


gpu-video/src/backends/vulkan/vulkan_adapter.rs
```diff
@@ -159,6 +159,19 @@ impl<'a> VulkanAdapter<'a> {
...
+            // combined graphics family, and insisting on a graphics-free
+            // compute queue rejects them even though they otherwise support
+            // video decode.
+            .or_else(|| {
+                queues.iter().enumerate().find(|(_, q)| {
+                    q.queue_family_properties
+                        .queue_flags
+                        .contains(vk::QueueFlags::COMPUTE)
+                })
+            })
```
jerzywilczek: There is a problem with this.
In vulkan, queue access needs to be synchronized by the API user. This means that two threads cannot be submitting to the same queue at the same time. wgpu needs to have a queue, and as far as I remember, it uses family 0, index 0 now (which is the one that supports graphics, compute and transfer). If this fallback happens, wgpu will use the same queue as this one. This would mean that if a user will be submitting on this queue and on wgpu from two separate threads simultaneously, things will break. I think it should be doable to, in case of this fallback, request two queues from this family, and hand one of them to wgpu and the other to gpu-video.

jerzywilczek (changes requested):
Seems doable if we solve this problem

jerzywilczek:
since I'm leaving for holiday tomorrow, @noituri would you take over managing this PR?

noituri:
> since I'm leaving for holiday tomorrow, @noituri would you take over managing this PR?

sure!

fschutt:
Ok its still in progress, haven't forgotten about it

gpu-video/src/backends/vulkan/vulkan_adapter.rs
fschutt: Ok, addressed in https://github.com/software-mansion/smelter/pull/2109/changes/d4930f57d616fd16818811c34f72e212171d1191 

I would try to rebuild azul again to verify it still works after that, but otherwise it should be ready.

gpu-video/src/backends/vulkan/vulkan_device/queues.rs
```diff
@@ -86,27 +86,34 @@ pub(crate) struct QueueIndices<'a> {
...
-            (self.compute.family_index != self.transfer.family_index)
-                .then_some((self.compute.family_index, self.compute.queue_count)),
+            Some((self.compute.family_index, self.compute.queue_count)),
             Some((
                 self.graphics_transfer_compute.family_index,
                 self.graphics_transfer_compute.queue_count,
             )),
         ]
         .into_iter()
         .flatten()
```
noituri, line 107: could you extract the queues to a separate variable and then do `queues.into_iter().flatten()` here?
I think this would be more readable

gpu-video/src/backends/vulkan/vulkan_adapter.rs
```diff
@@ -247,7 +260,13 @@ impl<'a> VulkanAdapter<'a> {
...
                 compute: QueueIndex {
                     family_index: compute_queue_idx,
-                    queue_count: queue_counts[compute_queue_idx] as usize,
+                    // On fallback to wgpu's family, take two queues: index 0 for wgpu,
+                    // index 1 for gpu-video (the fallback guarantees two exist).
+                    queue_count: if compute_queue_idx == graphics_transfer_compute_queue_idx {
+                        2
+                    } else {
+                        queue_counts[compute_queue_idx] as usize
+                    },
```
noituri, line 269: I would just use `queue_counts[compute_queue_idx] as usize` here and change the `graphics_transfer_compute`'s queue count from `1` to `queue_counts[graphics_transfer_compute_queue_idx]`.

This is more consistent with how the current fallback to the `transfer` queue works
(even though requesting all queues is probably suboptimal, but this will get refactored at some point in the future)

gpu-video/src/backends/vulkan/vulkan_device.rs
```diff
@@ -205,6 +205,12 @@ impl VulkanDevice {
                 } else {
                     transfer_queue.clone()
                 }
+            } else if queue_indices.compute.family_index
+                == queue_indices.graphics_transfer_compute.family_index
+                && queue_indices.compute.queue_count > 1
```
noituri, line 210: if you apply the other suggestions, this would also need to change

```suggestion
                && queue_indices.graphics_transfer_compute.queue_count > 1
```

noituri (commented):
Looks great, just couple of nitpicks + make sure the CI checks pass

gpu-video/src/backends/vulkan/vulkan_device/queues.rs
```diff
@@ -86,27 +86,34 @@ pub(crate) struct QueueIndices<'a> {
 
 impl QueueIndices<'_> {
     pub(crate) fn queue_create_infos(&self) -> Vec<QueueCreateInfo> {
-        [
+        // One create info per family, with the max queue_count any role wants;
```
noituri, line 89:
```suggestion
        // One create info per family, with the max queue_count requested for it;
```

a bit more readable i think

PR #2110, opened 2026-07-06
[smelter-render] Replace mip-based Lanczos rescaling with a separable resampler

michael-dm:
Builds on the Lanczos3 support from #1882. The mip-based path has a few issues, all from the 6x6 kernel sampling one mip level without scaling to the leftover ratio:

1. At power-of-two ratios the samples land on mip texel centers where every non-center weight is zero, so a 2:1 downscale is plain bilinear-box output at 36 taps/pixel.
2. At fractional ratios the leftover decimation isn't covered by the kernel, so it aliases.
3. For ratios in (1, 2) the source still enters `mip_levels_needed`, so a full-res mip-0 copy runs every frame and is never sampled (33 MB/frame for 4K).

This replaces the mip pyramid + 2D kernel with a separable resampler (horizontal then vertical pass) whose kernel scales with the ratio, so one path covers up- and downscaling. Ratios past 4x get a single box reduction first (averaging a 2^k box equals k cascaded 2:1 halvings) to keep the tap count bounded. The passes run through a linear `Rgba16Float` intermediate; a unorm one would clamp the kernel's negative lobes and lose ~17 dB on upscales. The mipmap generator and the `scaling_filter`/`mip_level` plumbing are removed, and `apply_layouts.wgsl` goes back to a plain `textureSample`.

Quality vs a float64 Lanczos3 reference, computed from the renderer's own 1:1 dump of a high-frequency grid (a worst case; real content sits higher):

| ratio | master | this PR |
|---|---|---|
| 1.5x | 23.0 dB | 52.7 dB |
| 2.0x | 17.4 dB | 51.3 dB |
| 3.0x | 16.1 dB | 50.4 dB |
| 5.0x | 24.3 dB | 43.3 dB |
| 8.0x | 28.8 dB | 53.4 dB |

Scaling-stage GPU time, M5 Pro, medians over 300 frames (us):

| case | master | this PR |
|---|---|---|
| 2.0x up (960->1920) | 1341 | 341 |
| 2.0x (3840->1920) | 1454 | 1064 |
| 3.7x (4K into tile) | 571 | 914 |
| 5.0x (4K into tile) | 422 | 431 |
| 8.0x (4K into tile) | 289 | 405 |

Faster than master at <=2x. At 3-4x it's slower, which is the tap count of actually filtering the signal the mip path throws away. Past 4x the box reduction keeps it bounded: a 16-up grid of 4K sources at 8x fits a 60 fps frame, where the unbounded kernel doesn't. The 4x cutoff is a single constant.

sheet-sweep


smelter-render/src/transformations/layout.rs
```diff
@@ -181,39 +185,43 @@ impl LayoutNode {
...
+                    source_index: *index,
+                    crop: crop.clone(),
+                    dst,
+                });
+                *crop = Crop {
+                    top: 0.0,
+                    left: 0.0,
+                    width: dst.width as f32,
+                    height: dst.height as f32,
+                };
```
jerzywilczek: I think this breaks.

If one axis has scale different than one, and the other axis a 1:1 scale, but with an integer offset, then the resampler will skip the second axis entirely (teh offset is only checked for being an integer there), and this code will set the crop offset to 0. This will cause the offset of the second axis to be skipped by both sides.

smelter-render/src/transformations/layout/resampler.rs
```diff
@@ -0,0 +1,335 @@
...
+        };
+
+        // Stronger shrink first, to minimize the intermediate size and tap count.
+        let mut passes: Vec<AxisMapping> = mappings
+            .iter()
+            .zip(levels)
+            .map(|(mapping, levels)| mapping.on_reduced_source(levels))
+            .filter(|mapping| !mapping.is_identity())
+            .collect();
+        passes.sort_by(|a, b| b.scale().total_cmp(&a.scale()));
```
jerzywilczek: maybe we can avoid alloc + sort for an array that is at most 2 elements long?

smelter-render/src/transformations/layout/resampler.rs
```diff
@@ -0,0 +1,335 @@
...
+            &make_layout(8),
+            &downsample_shader,
+            wgpu::TextureFormat::Rgba16Float,
+        );
+
+        scope.pop()?;
+        Ok(Self {
+            pipeline: build(wgpu_ctx.default_view_format()),
+            pipeline_f16: build(wgpu::TextureFormat::Rgba16Float),
+            downsample,
```
jerzywilczek, line 238: both `build(...)` calls should happen before the `scope` is popped.

smelter-render/src/transformations/layout/resampler.rs
```diff
@@ -0,0 +1,335 @@
...
+const LABEL: Option<&str> = Some("Lanczos3 resampler");
+
+/// Ratios up to this use the scaled kernel alone; beyond it a single box pass
+/// reduces the source first, keeping the tap count bounded and the box away from
+/// ratios where its softening would show.
+const KERNEL_BUDGET: f32 = 4.0;
+
+/// Separable Lanczos3 resampler: the kernel scales with the ratio, so one pair
+/// of passes covers any up/downscale without mip pyramids. Everything before the
+/// final pass runs in linear Rgba16Float, since a unorm intermediate would clamp
```
jerzywilczek, line 27: Have you verified whether `Rgba16Float` works in wasm?

smelter-render/src/transformations/layout/resampler.rs
```diff
@@ -0,0 +1,335 @@
...
+        &mut self,
+        wgpu_ctx: &Arc<WgpuCtx>,
+        shader: &ResamplerShader,
+        source: &NodeTextureState,
+        crop: &Crop,
+        dst: Resolution,
+        encoder: &mut wgpu::CommandEncoder,
+    ) {
+        let mappings = axis_mappings(crop, dst);
+        let levels = mappings.map(|m| m.predecimate_levels());
```
jerzywilczek, line 315: I think this should be clamped to <= 32, either inside `predecimate_levels` or here. It may be very hard to trigger (or even impossible), but the clamp cant hurt us, and we'll all feel safer.

smelter-render/src/transformations/layout/resampler.rs
```diff
@@ -0,0 +1,335 @@
...
+        let factor = levels.map(|l| 1u32 << l);
+
+        // Box-reduce the ratio above the budget (a 2^k box == k cascaded 2:1
+        // halvings, so one pass suffices), then the kernel takes the residual.
+        let (src_view, src_res) = if factor == [1, 1] {
+            (source.view(), source.resolution())
```
jerzywilczek, line 322: shouldn't we `self.reduced = None` here? similarily to :274?

smelter-render/src/transformations/layout/layout_renderer.rs
```diff
@@ -2,13 +2,17 @@ use std::sync::Arc;
...
+pub struct LayoutRenderer {
+    pub(super) shader: Arc<LayoutShader>,
+    pub(super) resampler: Arc<ResamplerShader>,
+}
 
 impl LayoutRenderer {
     pub fn new(wgpu_ctx: &Arc<WgpuCtx>) -> Result<Self, CreateShaderError> {
         let shader = Arc::new(LayoutShader::new(wgpu_ctx)?);
-        Ok(Self(shader))
+        let resampler = Arc::new(ResamplerShader::new(wgpu_ctx)?);
```
jerzywilczek: Seems to me the resampler does not need to be initialized in cpu-optimized rendering mode

jerzywilczek (changes requested):
Looks pretty good so far, I'd say that maybe a change of structure in `LayoutNode::render` would be quite welcome, so that bugs like the one where the crop offset is not applied are easier to detect.

michael-dm:
> Looks pretty good so far, I'd say that maybe a change of structure in `LayoutNode::render` would be quite welcome, so that bugs like the one where the crop offset is not applied are easier to detect.

Thanks for the review, let me know if the fixes seem correct.

smelter-render/src/transformations/layout/resampler.rs
```diff
@@ -0,0 +1,468 @@
...
+        // halvings, so one pass suffices), then the kernel takes the residual.
+        let (src_view, src_res) = if factor == [1, 1] {
+            self.reduced = None;
+            (source.view(), source.resolution())
+        } else {
+            let full = source.resolution();
+            let reduced = Resolution {
+                width: full.width / factor[0] as usize,
+                height: full.height / factor[1] as usize,
+            };
```
jerzywilczek, line 328:
```suggestion
            let reduced = Resolution {
                width: full.width.div_ceil(factor[0] as usize),
                height: full.height.div_ceil(factor[1] as usize),
            };
```

I think the kernel later uses full / factor as a float, so we need that extra pixel

jerzywilczek (approved):
just this one thing and seems good

wkozyra95:
You can switch submodule to this PR https://github.com/smelter-labs/smelter-snapshot-tests/pull/93 (I just created it from artifacts from CI, so It is possible that it will still fail on pipeline tests)

PR #2112, opened 2026-07-08
Correct A/V sync for all MoQ inputs

JBRS307:
Current implementation works ONLY if audio and video use the same epoch. This was wrong assumption.

---

What happens here:
- Basing on the timestamps of the first received frames from a/v tracks we determin if we want to normalize them to some shared anchor or use live edge estimation. Decision is based on the diff of the first frames from both tracks.
- If there is only one track always normalize to the first frame's pts (we know how many tracks to expect from the `catalog`, before we start receiving the stream)
- If there is weird change in pts compared to previous frame (time went backwards or pts is significantly bigger than the amount of wall-clock time passed suggests) AND we are on the keyframe -> this is an epoch discontinuity. In that case we ALWAYS switch to live-edge estimation for that track and change the epoch.

**WARNING**: This solution breaks if the stream is not actually live, but rather delivered in bursts (like HLS)
- If timestamps are normalized to one shared anchor by the publisher it is mitigated and if bursts are not larger than 2 seconds everything somewhat works.
- If timestamps are NOT normalized then the stream breaks. Solving this requires implementing proper jitter buffer which is out of scope of this PR.


smelter-core/src/pipeline/moq/connection.rs
JBRS307, line 1: The epoch implementation is VERY stretched. And tests DEFINITELY require to be put in a separate file (probably with the rest of the new code, but I think it should be refactored by human hand)

smelter-core/src/pipeline/moq/connection/timestamp_aligner.rs
```diff
@@ -350,30 +371,45 @@ impl TimestampAligner {
...
 
         let Some(other_first) = self.shared.first_offset(self.kind.other()) else {
-            return SkewDecision::Pending;
+            // Counterpart's first frame not seen yet, so the skew is not
+            // measurable. Hold out for it until the warmup deadline, then give up
+            // on verifying a single-epoch pair and put the stream on live-edge.
+            return match self.warmup_deadline_passed(elapsed, started) {
+                true => Some(self.shared.decide_mode(SyncMode::LiveEdge)),
+                false => None,
+            };
```
JBRS307: Extract to variable and return it, this looks uncanny

smelter-core/src/pipeline/moq/connection/timestamp_aligner.rs
```diff
@@ -0,0 +1,452 @@
...
+    /// Resolve the stream's sync mode: follow the shared decision if there is
+    /// one, otherwise measure. `None` means still undecided (counterpart's first
+    /// frame not seen and the deadline has not fired).
+    ///
+    /// The two measuring branches always agree between the tracks — they read the
+    /// same set-once first offsets — so whichever track measures first commits the
+    /// verdict the other would have reached anyway. Only the deadline branch here
+    /// and the discontinuity setter in [`Self::on_chunk`] add information a track
+    /// cannot derive on its own, and those are exactly the cases where a track
+    /// follows a mode it did not measure.
```
JBRS307:
```suggestion
```

smelter-core/src/pipeline/moq/connection/timestamp_aligner.rs
```diff
@@ -0,0 +1,452 @@
...
+    fn resolve_mode(&self, elapsed: Duration, started: Duration) -> Option<SyncMode> {
+        if let Some(mode) = self.shared.mode() {
+            return Some(mode);
+        }
+
+        // Single track: no counterpart => trivially single-epoch.
```
JBRS307:
```suggestion
```

smelter-core/src/pipeline/moq/connection.rs
```diff
@@ -33,8 +36,13 @@ use crate::prelude::*;
 use self::catalog::{MoqCatalogError, read_catalog};
 
 mod catalog;
+mod timestamp_aligner;
 
-const MOQ_BUFFER: Duration = Duration::from_secs(1);
+/// Buffer added to a new track's PTS offset during queue initialization. This gives frames
+/// time to arrive and be decoded before their scheduled presentation, absorbing network jitter and
+/// decode latency at the cost of a fixed delay. The 2200ms value is set based on Twitch
+/// broadcasting guidelines.
```
wkozyra95, line 44: I wouldn't mention Twitch at all

wkozyra95 (approved):

PR #2118, opened 2026-07-15
[gpu-video] merge vt decoder

jerzywilczek:


gpu-video/src/backends/video_toolbox/decoder/wgpu_api.rs
```diff
@@ -0,0 +1,209 @@
...
+            .as_hal::<MtlApi>()
+            .ok_or(VTInitError::NotMetalBackend)?
+            .raw_device()
+            .clone()
+    };
+
+    let texture_attributes = unsafe {
+        cf::CFDictionary::<cf::CFString, cf::CFNumber>::from_slices(
+            &[cv::kCVMetalTextureUsage],
+            &[cf::CFNumber::new_i64(mtl::MTLTextureUsage::ShaderRead.0 as i64).as_ref()],
```
noituri, line 183: shouldn't this have the same usages as the output `wgpu::Texture`?

gpu-video/src/backends/video_toolbox/decoder/wgpu_api.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    fn to_wgpu_texture(
+        &self,
+        device: &wgpu::Device,
+        buffer: &cv::CVBuffer,
+    ) -> Result<wgpu::Texture, VTDecoderError> {
+        let Some(cache) = &self.texture_cache else {
+            panic!("decoder asked to output to wgpu texture, but not configured for that initially")
+        };
```
noituri, line 78: I'm not saying this needs to be changed, but vulkan decoder would return an error here instead of panicing

gpu-video/src/backends/video_toolbox/decoder/wgpu_api.rs
```diff
@@ -0,0 +1,209 @@
...
+
+        let width = cv::CVPixelBufferGetWidth(buffer);
+        let height = cv::CVPixelBufferGetHeight(buffer);
+        let y_width = cv::CVPixelBufferGetWidthOfPlane(buffer, 0);
+        let y_height = cv::CVPixelBufferGetHeightOfPlane(buffer, 0);
```
noituri, line 83: Just curious, is there any difference between the buffer dimensions and the Y plane dimensions?

gpu-video/src/backends/video_toolbox/caps.rs
```diff
@@ -0,0 +1,185 @@
...
+        min_height: ASSUMED_MIN_DIMENSION,
+        max_height: ASSUMED_MAX_DIMENSION,
+        rate_control: RateControlCapabilities {
+            vbr_supported: true,
+            cbr_supported: supports_property(supported_properties, unsafe {
+                kVTCompressionPropertyKey_ConstantBitRate
+            }),
+        },
+        max_references: ASSUMED_ENCODE_MAX_REFERENCES,
+        quality_levels: ASSUMED_ENCODE_QUALITY_LEVELS,
```
noituri, line 118: we should probably remove the `quality_levels` on master

gpu-video/Cargo.toml
```diff
@@ -42,17 +48,76 @@ wgpu = { version = "30.0.0", optional = true }
...
 [dev-dependencies]
 clap = { version = "4.5.20", features = ["derive"] }
 tracing-subscriber = "0.3.20"
-wgpu = { version = "30.0.0", features = ["noop"] }
+wgpu = { version = "30.0.0", features = [
+    "noop",
+] }
 winit = "0.30"
 bytemuck = { version = "1.19.0", features = ["derive"] }
 pollster = { workspace = true }
```
noituri, line 113: `pollster` can be removed from `dev-dependencies`

gpu-video/src/lib.rs
```diff
@@ -6,37 +6,37 @@ pub mod parser;
...
-#[cfg(vulkan)]
+#[cfg(supported)]
 mod exports;
-#[cfg(vulkan)]
+#[cfg(supported)]
 pub use exports::*;
 
-// If vulkan is unsupported and parsers are not exposed
-#[cfg(not(any(vulkan, feature = "expose-parsers")))]
+// If supported is unsupported and parsers are not exposed
```
noituri: this sounds weird xd

gpu-video/src/frame_sorter.rs
```diff
@@ -61,13 +62,15 @@ pub(crate) struct FrameSorter<T> {
     frames: BinaryHeap<DecodeResult<T>>,
 }
 
-impl<T> FrameSorter<T> {
-    pub(crate) fn new() -> Self {
+impl<T> Default for FrameSorter<T> {
```
noituri, line 65: maybe just derive `Default`

gpu-video/src/backends/video_toolbox/decoder.rs
```diff
@@ -0,0 +1,538 @@
...
+                    pps: pps.clone(),
+                })?,
+                DecoderInstruction::Decode { decode_info, .. } => {
+                    results.extend(self.do_decode(decode_info.clone(), false)?)
+                }
+                DecoderInstruction::Idr { decode_info, .. } => {
+                    results.extend(self.do_decode(decode_info.clone(), true)?)
+                }
+                DecoderInstruction::Drop { .. } => {}
+            }
```
noituri: tbh, we could accept owned `instructions` here and avoid the clones

gpu-video/src/backends/video_toolbox/decoder.rs
```diff
@@ -0,0 +1,538 @@
...
+            .sps
+            .get(&sps_id)
+            .map(|sps| ColorRange::from(&sps.sps))
+            .unwrap_or(ColorRange::Limited);
+
+        let range_changed = self.session_color_range != Some(color_range);
+
+        if !self.needs_session_update && !range_changed && self.session.is_some() {
+            return Ok(());
+        }
```
noituri, line 210: Color range can only change when sps changes, no? This `range_changed` check seems unnecessary

noituri (approved):
Very nice, just couple of questions

gpu-video/src/backends/video_toolbox/decoder.rs
jerzywilczek: yes, I wanted to text you about that but I forgot before

gpu-video/src/frame_sorter.rs
jerzywilczek: impossible unfortunately. `derive(Default)` checks whether all generic parameters implement `Default`, even though it is not necessary in this case

gpu-video/src/backends/video_toolbox/decoder/wgpu_api.rs
jerzywilczek: copy source and copy dst are always enabled on all textures on metal

gpu-video/src/backends/video_toolbox/decoder/wgpu_api.rs
jerzywilczek: probably not, but cant hurt, right?

gpu-video/src/backends/video_toolbox/caps.rs
jerzywilczek: #2124

gpu-video/src/backends/video_toolbox/decoder.rs
jerzywilczek: do you think we should do it on master or here?

gpu-video/src/backends/video_toolbox/decoder.rs
jerzywilczek: actually, it can

if we get sps with id 0 and color range full, and sps with id 1 and color range limited at the start of the stream, and then AU0 uses sps 0 and AU1 uses sps 1, this is where `needs_session_update == false`, but `range_changed = true`

gpu-video/Cargo.toml
jerzywilczek: only on apple, it needs to be here for vulkan platforms

gpu-video/src/backends/video_toolbox/decoder.rs
noituri: I think both

gpu-video/Cargo.toml
noituri: I didn't notice that `pollster` was defined under apple deps

gpu-video/src/backends/video_toolbox/decoder.rs
jerzywilczek: so, you do it on master and then I rebase here?

noituri: https://github.com/software-mansion/smelter/pull/2126

PR #2123, opened 2026-07-16
Limit queue tracks per-input - fix looping short mp4

wkozyra95:


smelter-core/src/queue/queue_input.rs
```diff
@@ -105,16 +110,14 @@ impl InnerQueueInput {
         }
     }
 
-    fn queue_new_track(
-        &mut self,
+    fn new_pending_track(
```
wkozyra95, line 113: We need to create it immediately, but can't send it because it would hold a lock

JBRS307 (commented):
LGTM

Issue #2108, opened 2026-07-03
MoQ client output.


JBRS307:
- [x] Output client code: #2125 
- [x] ts-sdk: #2148 
- [x] docs: https://github.com/smelter-labs/smelter-website/pull/280

PR #2125, opened 2026-07-16
Added MoQ client output.

JBRS307:
- #2108


smelter-core/src/protocols/moq.rs
```diff
@@ -72,6 +104,26 @@ pub enum MoqClientError {
 
     #[error("Failed to connect to MoQ relay: {0}")]
     ConnectFailed(String),
+
+    #[error("MoQ relay refused to publish a broadcast under path \"{0}\".")]
```
JBRS307: This error message points that relay refused to publish, reword it as it is now misleading.

smelter-core/src/protocols/moq.rs
```diff
@@ -72,6 +104,26 @@ pub enum MoqClientError {
...
+    PublishFailed(Arc<str>),
+
+    #[error("Failed to create a MoQ broadcast: {0}")]
+    BroadcastInitFailed(String),
+
+    #[error("H264 encoder did not produce a decoder configuration record.")]
+    MissingH264DecoderConfig,
```
JBRS307:
```suggestion
    MissingH264EncoderConfig,
```

smelter-core/src/protocols/moq.rs
```diff
@@ -72,6 +104,26 @@ pub enum MoqClientError {
...
+    #[error("Failed to build a CMAF init segment: {0}")]
+    InitSegmentError(String),
+
+    #[error(
+        "Codec {codec:?} is not supported with the \"{container}\" container. Use \"legacy\" or \"loc\" instead."
+    )]
+    UnsupportedCodecContainer {
+        codec: VideoCodec,
+        container: MoqOutputContainer,
+    },
```
JBRS307: This is temporary, I want to support ALL combinations in this PR

smelter-core/src/error.rs
```diff
@@ -176,13 +176,29 @@ pub enum OutputInitError {
...
+}
+
+/// Error that can happen after registration
+#[derive(Debug, thiserror::Error, Clone)]
+pub enum OutputMoqClientRuntimeError {
+    #[error("Failed to write a frame to the MoQ broadcast: {0}")]
+    WriteError(String),
+
+    #[error("Frame timestamp is outside of the range supported by MoQ: {0}")]
+    InvalidTimestamp(String),
```
JBRS307: Check what is going on here and why is this type used only in the mp4

smelter-core/src/protocols/moq.rs
JBRS307: No

smelter-api/src/output/moq_client.rs
```diff
@@ -0,0 +1,140 @@
...
+use crate::*;
+
+#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, ToSchema)]
+#[serde(deny_unknown_fields)]
+pub struct MoqClientOutput {
+    /// URL of the MoQ relay to connect to. Must use the `https://` scheme.
+    pub endpoint_url: Arc<str>,
+    /// Path the broadcast will be published under on the relay.
+    pub broadcast_path: Arc<str>,
+    /// (**default=`"cmaf"`**) Container used to frame encoded media.
```
wkozyra95, line 16: I don't have better alternative, but I don't like that logic. The default that only works with h264 seems weird, but I also understand why you selected CMAF

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,563 @@
...
+        let (session, origin) = Self::connect(&ctx, &options.endpoint_url)?;
+        let state = Self::publish(&options, video_track, audio_track, session, origin)?;
+
+        let span = Span::current();
+        std::thread::Builder::new()
+            .name(format!("MoQ sender thread for output {output_ref}"))
+            .spawn(move || {
+                let _span = span.entered();
+                // moq-net spawns tasks (e.g. group cleanup) from the producer calls below.
+                let _rt_guard = ctx.tokio_rt.enter();
```
wkozyra95: Don't do that; it mixes async and sync in a very unpredictable way, hard to notice way. If something is async then it should be async.

Also I have no idea how this logic affects the tokio runtime scheduler

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,563 @@
...
+    // Each iteration either receives one chunk or writes the pending one with the
+    // lower PTS; it never does both.
+    loop {
+        let need_video = pending_video.is_none() && !video_eos;
+        let need_audio = pending_audio.is_none() && !audio_eos;
+
+        match (need_video, need_audio) {
+            //
+            // Receive phase. When `need_*` is true the matching channel is present.
+            //
```
wkozyra95:
```suggestion
            // Receive phase. When `need_*` is true the matching channel is present.
```

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,563 @@
...
+    }
+}
+
+/// Interleaves encoded chunks from the two encoders and writes them to the
+/// broadcast in PTS order. Mirrors the sync A/V muxing done by the RTMP output.
+fn run_moq_output_thread(
+    ctx: &Arc<PipelineCtx>,
+    output_ref: &Ref<OutputId>,
+    mut state: BroadcastState,
+    video_rx: Option<Receiver<EncodedOutputEvent>>,
```
wkozyra95: we don't use rx/tx in other places, keep it consistent

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,563 @@
...
+            //
+            // Write phase. Write the lower-PTS chunk (video first on ties).
+            //
+            (false, false) => {
+                let Some(chunk) = get_chunk_to_write(&mut pending_video, &mut pending_audio) else {
+                    // both tracks finished
+                    break;
+                };
+
+                if write_chunk(
```
wkozyra95: this call is longer than the function, I'm not saying to inline this, but reconsider the structure

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,563 @@
...
+    let mut pending_video: Option<EncodedOutputChunk> = None;
+    let mut pending_audio: Option<EncodedOutputChunk> = None;
+    // A missing channel means that media is not produced at all, i.e. already done.
+    let mut video_eos = video_rx.is_none();
+    let mut audio_eos = audio_rx.is_none();
+
+    // Each iteration either receives one chunk or writes the pending one with the
+    // lower PTS; it never does both.
+    loop {
+        let need_video = pending_video.is_none() && !video_eos;
```
wkozyra95: Good idea, it's definitely cleaner/more readable to reframe it around need video instead of the match in like in webrtc, but I don't like treating missing track as received EOS

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,563 @@
...
+                        // Disconnect without an explicit EOS is treated as EOS.
+                        _ => { video_eos = true; finish(state.video.as_mut(), "video"); }
+                    },
+                    recv(audio_rx.as_ref().unwrap()) -> msg => match msg {
+                        Ok(EncodedOutputEvent::Data(chunk)) => pending_audio = Some(chunk),
+                        _ => { audio_eos = true; finish(state.audio.as_mut(), "audio"); }
+                    },
+                }
+            }
+            (true, false) => match video_rx.as_ref().unwrap().recv() {
```
wkozyra95: Justifiable unwrap given how much it simplifies the logic, and it is close enough to the condition that enforces that it will not fail, but might be worth comenting

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,563 @@
...
+            let timestamp = Timestamp::from_micros(pts.as_micros() as u64)
+                .map_err(|err| OutputMoqClientRuntimeError::InvalidTimestamp(format!("{err}")))?;
+            producer
+                .write(Frame {
+                    timestamp,
+                    payload: chunk.data,
+                    keyframe: chunk.is_keyframe,
+                })
+                .map_err(|err| {
+                    OutputMoqClientRuntimeError::WriteError(ErrorStack::new(&err).into_string())
```
wkozyra95: This helper is needed when we actually ned to log it

smelter-core/src/pipeline/moq/output/init_segment.rs
```diff
@@ -0,0 +1,165 @@
...
+use crate::prelude::*;
+
+/// Timescale for video tracks. Encoder timestamps are microseconds, so any
+/// timescale works; 90 kHz is the MPEG convention.
+pub(super) const VIDEO_TIMESCALE: u32 = 90_000;
+
+/// Each MoQ track carries exactly one media track, so the id is always the same.
+pub(super) const TRACK_ID: u32 = 1;
+
+pub(super) fn h264(extradata: &[u8], resolution: Resolution) -> Result<Bytes, MoqClientError> {
```
wkozyra95: Unhelpful name; you can't assume that someone will use it as init_segment::h264. I'm not saying it has to be fully descriptive e.g. if there was modules h264 and opus and both had an init_segment function I would not have problem with that, because the name describes what it does, even though for full context you would still have to call  h264::init_segment.

btw I'm not saying to create h264/opus modules, it just hypothetical to illustrate the distinction between 2 cases

smelter-core/src/pipeline/moq/output/track.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+    let codec = match video {
+        Some(VideoEncoderOptions::FfmpegVp8(_)) => VideoCodec::Vp8,
+        Some(VideoEncoderOptions::FfmpegVp9(_)) => VideoCodec::Vp9,
+        _ => return Ok(()),
+    };
+    Err(MoqClientError::UnsupportedCodecContainer { codec, container })
+}
+
+pub(super) fn video(
```
wkozyra95: Same as above; the name says nothing about what it does. I don't like naming like that  event for private functions, although in that case that might be very subjective, so it's fine to leave it, but not for stuff that is exported.

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,563 @@
...
+        match (need_video, need_audio) {
+            //
+            // Receive phase. When `need_*` is true the matching channel is present.
+            //
+            (true, true) => {
+                crossbeam_channel::select! {
+                    recv(video_rx.as_ref().unwrap()) -> msg => match msg {
+                        Ok(EncodedOutputEvent::Data(chunk)) => pending_video = Some(chunk),
+                        // Disconnect without an explicit EOS is treated as EOS.
+                        _ => { video_eos = true; finish(state.video.as_mut(), "video"); }
```
wkozyra95: That "video" at a first glance looks like using string as enum. You need to check the implementation to see what it is doing. Also I think passing partial strings to be embeded in logs seems like very bad pattern.

 In this case I would pass enum and do not print it as part of string, but as a `warn!(?media, "The rest of the log")

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,563 @@
...
+}
+
+struct MoqClientOutputStatsSender {
+    stats_sender: StatsSender,
+    output_ref: Ref<OutputId>,
+}
+
+impl MoqClientOutputStatsSender {
+    fn bytes_sent_event(&self, size: usize, track_kind: StatsTrackKind) {
+        self.stats_sender.send(
```
wkozyra95: You call that in one place where you already pass both ctx and output_ref

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,563 @@
...
+    if let Err(err) = state.catalog.finish() {
+        warn!(
+            "Failed to close MoQ catalog: {}",
+            ErrorStack::new(&err).into_string()
+        );
+    }
+    // Dropping `state` closes the session, which unannounces the broadcast.
+}
+
+fn get_chunk_to_write(
```
wkozyra95: Reconsider how you divided stuff into functions and named them. I think you took too much from the webrtc structure while changing the logic

smelter-api/src/output/moq_client.rs
JBRS307: There is already a PR changing that, I think I will request a review on that today.

smelter-api/src/output/moq_client_into.rs
```diff
@@ -0,0 +1,241 @@
...
+
+impl MoqClientVideoEncoderOptions {
+    fn to_pipeline_options(
+        &self,
+        resolution: Resolution,
+        container: MoqOutputContainer,
+    ) -> Result<core::VideoEncoderOptions, TypeError> {
+        // CMAF describes H264 with an out-of-band `avcC` record in the init
+        // segment, so the encoder has to emit AVCC. Legacy and LOC carry the
+        // parameter sets inline, which means Annex B.
```
JBRS307:
```suggestion
```

smelter-api/src/output/moq_client_into.rs
```diff
@@ -0,0 +1,241 @@
...
+                        .collect(),
+                })
+            }
+        };
+        Ok(encoder_options)
+    }
+}
+
+/// The CMAF init segment we publish can only describe H264 video, so VP8/VP9
+/// have to opt into one of the containers that carries the payload as-is.
```
JBRS307:
```suggestion
```

smelter-core/src/codecs/aac.rs
```diff
@@ -167,6 +176,28 @@ fn freq_id_to_sample_rate(id: u8) -> Result<u32, AacAscParseError> {
     }
 }
 
+/// Inverse of [`freq_id_to_sample_rate`]. MPEG-4 part 3, 1.6.3.4.
+/// Returns `None` for sample rates that don't have a dedicated frequency index
+/// (those would require the escape value 15 with an explicit rate in the ASC).
```
JBRS307:
```suggestion
/// MPEG-4 part 3, 1.6.3.4.
/// Returns `None` for sample rates that don't have a dedicated frequency index
/// (those would require the escape value 15 with an explicit rate in the ASC).
```

smelter-core/src/codecs/aac.rs
```diff
@@ -212,6 +243,15 @@ mod tests {
...
 
+    #[test]
+    fn freq_index_round_trip() {
+        for id in 0x0..=0xc {
+            let sample_rate = freq_id_to_sample_rate(id).unwrap();
+            assert_eq!(sample_rate_to_freq_index(sample_rate), Some(id));
+        }
+        assert_eq!(sample_rate_to_freq_index(12345), None);
+    }
+
```
JBRS307:
```suggestion
```

integration-tests/examples/demo/outputs.rs
```diff
@@ -166,6 +175,12 @@ impl VideoEncoder {
                     "thread_type": "slice",
                 },
             }),
+            Self::FfmpegH264 => json!({
+                "type": "ffmpeg_h264",
+                "ffmpeg_options": {
+                    "bf": "0",
+                },
+            }),
```
JBRS307:
```suggestion
```

smelter-api/src/output/moq_client.rs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, JsonSchema, ToSchema, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum MoqOutputContainer {
+    /// Microsecond timestamp prefix followed by the raw codec payload.
+    Legacy,
+    /// Fragmented MP4. Each frame is a complete `moof`+`mdat` fragment.
+    Cmaf,
+    /// Low Overhead Container.
```
JBRS307:
```suggestion
    /// Low Overhead Container. (draft-ietf-moq-loc)
```

smelter-core/src/pipeline/moq/output/cmaf_init_segment.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+fn visual(resolution: Resolution) -> mp4_atom::Visual {
+    mp4_atom::Visual {
+        data_reference_index: 1,
+        width: resolution.width as u16,
+        height: resolution.height as u16,
+        ..Default::default()
+    }
+}
```
JBRS307: Put that below `video_cmaf_init`

smelter-core/src/error.rs
```diff
@@ -195,6 +203,15 @@ pub enum OutputWhipRuntimeError {
     PeerConnectionDisconnected,
 }
 
+#[derive(Debug, thiserror::Error, Clone)]
+pub enum OutputMoqClientRuntimeError {
```
wkozyra95: I missed that originally. Look at what errors we are reporting in other protocols. Either remove it, or provide some explanation why do you think it's aproriate/needed to send that erorrs over WebSocket. My main concern is this error my potentially be spam per packet

smelter-core/src/protocols/moq.rs
```diff
@@ -26,6 +26,38 @@ pub struct MoqInputDecoders {
...
+pub enum MoqOutputContainer {
+    /// Microsecond timestamp varint prefix followed by the raw codec payload.
+    Legacy,
+    /// Fragmented MP4. Each frame is a complete moof+mdat; the init segment is
+    /// published in the catalog.
+    #[default]
+    Cmaf,
+    /// Low Overhead Container.
+    Loc,
+}
```
wkozyra95:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum MoqOutputContainer {
    /// Microsecond timestamp varint prefix followed by the raw codec payload.
    Legacy,
    /// Fragmented MP4. Each frame is a complete moof+mdat; the init segment is
    /// published in the catalog.
    Cmaf,
    /// Low Overhead Container.
    Loc,
}
```
I don't think Default helps anywhere in the code, but it is unnecessarily confusing

smelter-core/src/pipeline/moq/output/catalog.rs
wkozyra95, line 1: catalog.rs or catalog_entry.rs

smelter-core/src/pipeline/moq/output/cmaf_init_segment.rs
wkozyra95, line 1: This file handles init segment for CMAF format, the most important part for someone looking for it is cmaf, ti needs to be in the name

smelter-core/src/pipeline/moq/output/init_segment.rs
```diff
@@ -0,0 +1,277 @@
...
+use crate::prelude::*;
+
+/// Timescale for video tracks. Encoder timestamps are microseconds, so any
+/// timescale works; 90 kHz is the MPEG convention.
+pub(super) const VIDEO_TIMESCALE: u32 = 90_000;
+
+/// Each MoQ track carries exactly one media track, so the id is always the same.
+pub(super) const TRACK_ID: u32 = 1;
+
+pub(super) fn h264_cmaf_init(
```
wkozyra95: If you want to make it shorter because init segement is too long then dropping word segment is definietlly wrong choice,

smelter-core/src/codecs/aac.rs
```diff
@@ -4,10 +4,19 @@ use bytes::Buf;
 
 use crate::prelude::*;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum AacBitstreamFormat {
+    /// Raw AAC access units; decoder config carried out-of-band (ASC extradata).
```
wkozyra95, line 9: where did you get "raw au" name from?

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,548 @@
...
+            match (need_video, need_audio) {
+                // Receive phase. When `need_*` is true the matching channel is present.
+                (true, true) => {
+                    tokio::select! {
+                        event_opt = self.video_receiver.as_mut().unwrap().recv() => self.handle_video_read(event_opt, state.video.as_mut()),
+                        event_opt = self.audio_receiver.as_mut().unwrap().recv() => self.handle_audio_read(event_opt, state.audio.as_mut()),
+                    }
+                }
+                (true, false) => {
+                    let event_opt = self.video_receiver.as_mut().unwrap().recv().await;
```
wkozyra95: what does _opt mean in this context?

smelter-api/src/output/moq_client_into.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+impl MoqClientVideoEncoderOptions {
+    fn to_pipeline_options(
+        &self,
+        resolution: Resolution,
+        container: MoqOutputContainer,
+    ) -> Result<core::VideoEncoderOptions, TypeError> {
+        let bitstream_format = match container {
```
wkozyra95:
```suggestion
        let h264_bitstream_format = match container {
```

smelter-core/src/codecs/aac.rs
JBRS307: this stands for raw access units

smelter-api/src/output/moq_client_into.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        let container = container.unwrap_or(MoqOutputContainer::Cmaf);
+
+        let (video_encoder_options, output_video_options) = match video {
+            Some(OutputMoqClientVideoOptions {
+                resolution,
+                send_eos_when,
+                encoder,
+                initial,
+            }) => {
```
wkozyra95, line 30:
```suggestion
            Some(opt) => {
```
nit, up to you, but I think syntax like this and e.g. opt.resolution is more redable

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,553 @@
...
+        let state = Self::publish(
+            &options,
+            video_track_config,
+            audio_track_config,
+            session,
+            origin,
+        )?;
+
+        let tokio_rt = ctx.tokio_rt.clone();
+        tokio_rt.spawn(
```
wkozyra95:
```suggestion
        ctx.tokio_rt.clone().spawn(
```
I think this would work too, but not tested

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,553 @@
...
+        })
+    }
+
+    fn connect(
+        ctx: &Arc<PipelineCtx>,
+        url: &str,
+    ) -> Result<(MoqSession, OriginProducer), MoqClientError> {
+        let url = Url::parse(url).map_err(|err| MoqClientError::InvalidUrl(Arc::from(url), err))?;
+
+        if !matches!(url.scheme(), "https" | "http") {
```
wkozyra95, line 156:
```suggestion
        if !["http", "https"].contains(url.scheme()) {
```
Another nit, up to you; personally, I prefer to avoid matches

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,553 @@
...
+        warn!(
+            "Failed to close MoQ catalog: {}",
+            ErrorStack::new(&err).into_string()
+        );
+    }
+
+    // Dropping `state` closes the session, which unannounces the broadcast.
+}
+
+struct InterleavedPacketSender {
```
wkozyra95: I named it sender because the logic of sending was part of the implementation. Maybe InterleavedPacketResolver

smelter-core/src/pipeline/moq/output/client_output.rs
```diff
@@ -0,0 +1,553 @@
...
+                }
+            }
+            (Some(_), None) => self.next_video.take(),
+            (None, Some(_)) => self.next_audio.take(),
+            (None, None) => None,
+        }
+    }
+}
+
+fn write(
```
wkozyra95: at least send_chunk

wkozyra95 (approved):

PR #2130, opened 2026-07-17
[gpu-video] Improve compile times a bit (on linux)

jerzywilczek:
For a long time, I wanted to check how much compile time can we save by managing dependencies more carefully

Turns out a bit, but not a ton:

| change | build with wgpu | build with no default features | example decode | example decode_wgpu |
|--------|----------|---------|-----------|----------------|
| before | 14.6s | 6.3s | 15.4s | 16.7s |
| after | 12.5s | 6.3s | 13.8s | 15.0s |
| change | -14.4% | 0% | -10.4% | -10.2% |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1487,17 +1487,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "derivative"
-version = "2.2.0"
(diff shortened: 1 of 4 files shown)


noituri (approved):
nice

PR #2141, opened 2026-07-21
Stop sender loop on disconnect in WHEP output

wkozyra95:
write_rtp does not throw errors, so after every client, the loop keeps running until output is unregistered


smelter-core/src/pipeline/webrtc/whep_output/stream_media_to_peer.rs
```diff
@@ -75,113 +137,79 @@ pub async fn stream_media_to_peer(
...
-            Err(TryGetEventError::Empty) => {}
+    fn resolve_from_state(&mut self) -> Result<EncodedOutputEvent, InterleaveResult> {
+        // Handle EOS for video
+        if let Some(EncodedOutputEvent::VideoEOS) = self.next_video {
+            return self.next_video.take().ok_or(InterleaveResult::Continue);
         }
-    }
 
-    ctx.event_emitter
-        .emit(Event::OutputDone(output_ref.id().clone()));
```
wkozyra95, line 108: This would be incorrect even if aborting was working correctlly. Output done should be sent on output end, not when one of the clients disconnects.

smelter-core/src/pipeline/webrtc/whep_output/stream_media_to_peer.rs
```diff
@@ -75,113 +137,79 @@ pub async fn stream_media_to_peer(
...
-                next_video_event.take().ok_or(TryGetEventError::Empty)
-            } else {
-                next_audio_event.take().ok_or(TryGetEventError::Empty)
+        match stream.payloader.payload(chunk) {
+            Ok(rtp_packets) => {
+                for rtp_packet in rtp_packets {
+                    if let Err(err) = stream.track.write_rtp(&rtp_packet.packet).await {
+                        return Err(WhepError::RtpWriteError(err));
+                    }
+                    trace!(?rtp_packet, "RTP packet written to track");
```
JBRS307: In this method we can easily check if it is audio or video chunk, I would add that info to the trace log.

JBRS307:
```suggestion
                    trace!(?rtp_packet, "RTP packet written to track.");
```

This way the log is more readable as the text part is nicely separated from the struct in debug formatting

JBRS307 (approved):

PR #2143, opened 2026-07-21
[gpu-video] Make decoders work asynchronously

### LINKED ISSUE #2036: [gpu-video] Use wait semaphores from the new wgpu release
I think it could be possibly used with normal core queue which would allow us to omit CPU wait? There's also a new way of adding singal semaphores

noituri:
This PR has been cooked for far too long and there are still things I'm not entirely sure about. lmk what you think.

TODO
- maybe check the performance

Closes: #2036


gpu-video/src/backends/vulkan/vulkan_decoder/decoders_h264.rs
```diff
@@ -0,0 +1,323 @@
...
+            }))
+        }
+    }
+
+    fn on_finish(&self, frames: Vec<OutputFrame<DownloadSubmission<wgpu::Texture>>>) {
+        for frame in frames {
+            if let Err(err) = frame.data.check_decode_results() {
+                self.on_error(err.into());
+            }
+        }
```
noituri: I kind of don't like it. I return the textures early but the errors from decoding will be passed to the user later. I think this can be a bit confusing?

smelter-core/src/pipeline/decoder/vulkan_h264.rs
```diff
@@ -64,40 +82,66 @@ impl VideoDecoderInstance for VulkanH264Decoder {
...
+
+        let mut frames = Vec::new();
+        for result in self.decode_result_receiver.try_iter() {
+            let frame = match result {
+                Ok(frame) => frame,
+                Err(err) => {
+                    handle_decode_err(err);
+                    continue;
+                }
+            };
```
noituri: it's kind of weird that now you'd have to handle decode errors 2 times (at schedule and at callback)

maybe errors should be split into 2 categories?:
1. Schedule Error
2. Actual Decode Error?

this would work for vulkan but would it be ok on metal? or other backends?

gpu-video/src/backends/vulkan/vulkan_decoder/decoders_h264.rs
noituri: talked this through offline

integration-tests/src/gpu_video_tests.rs
```diff
@@ -24,24 +24,20 @@ struct TestCase<Options> {
...
+        })
+        .collect::<Vec<_>>();
+    adapters.sort_by_key(|a| a.get_info().device_type == wgpu::DeviceType::DiscreteGpu);
+
+    let adapter = adapters.last().unwrap();
+
+    adapter
+        .request_device_with_video_support(&VideoDeviceDescriptor::default())
+        .unwrap()
 }
```
noituri, line 43: Waiter thread never exited because the device was stored in a static.
This caused segfaults on nvidia because the thread was waiting on the interrupt semaphore when the driver was executing `atexit`

gpu-video/examples/decode.rs
```diff
@@ -36,29 +36,36 @@ fn main() {
         .create_device(&VideoDeviceDescriptor::default())
         .unwrap();
 
+    let (frame_sender, frame_receiver) = std::sync::mpsc::channel::<OutputFrame<RawFrameData>>();
+
+    let waiter_thread_handle = std::thread::spawn(move || {
```
jerzywilczek:
```suggestion
    let writer_thread_handle = std::thread::spawn(move || {
```

gpu-video/examples/decode_wgpu.rs
```diff
@@ -35,7 +35,7 @@ fn main() {
     let mut decoder = device
         .video()
         .unwrap()
-        .create_wgpu_textures_decoder_h264(DecoderParameters::default())
+        .create_wgpu_textures_decoder_h264(&queue, DecoderParameters::default())
```
jerzywilczek, line 38: man, I hate both API solutions...

I hate having the wgpu API so different from the normal one
I would hate having to give a useless completion closure that fires immediately when using wgpu textures

gpu-video/src/backends/vulkan/waiter_thread.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    fn wait_value(&self) -> SemaphoreWaitValue {
+        *self.wait_for.lock().unwrap()
+    }
+
+    fn interrupt(&self) -> Result<(), VulkanCommonError> {
+        let mut wait_for = self.wait_for.lock().unwrap();
+        self.semaphore.signal(*wait_for)?;
+        *wait_for = SemaphoreWaitValue(wait_for.0 + 1);
+        Ok(())
```
jerzywilczek, line 172: what is the reasoning behind not using an `AtomicU64`? That we might fail to signal? because that's the only explanation I can see. This shouldn't be a worry, if a thread is waiting for value `n`, any `m >= n` wakes it up

gpu-video/src/backends/vulkan/waiter_thread.rs
```diff
@@ -0,0 +1,168 @@
...
+pub(crate) struct WaiterThread {
+    shared: Arc<SharedState>,
+    device: Arc<Device>,
+}
+
+impl WaiterThread {
+    pub(crate) fn spawn(device: Arc<Device>) -> Result<WaiterThreadHandle, VulkanCommonError> {
+        let shared_state = Arc::new(SharedState {
+            wait_requests: Mutex::new(Vec::new()),
+            quit_thread: AtomicBool::new(false),
```
jerzywilczek:
```suggestion
            should_quit: AtomicBool::new(false),
```

gpu-video/src/backends/vulkan/waiter_thread.rs
```diff
@@ -0,0 +1,168 @@
...
+
+            self.resolve_finished();
+        }
+    }
+
+    fn resolve_finished(&self) {
+        let finished = {
+            let mut requests = self.shared.wait_requests.lock().unwrap();
+
+            // semaphore values can increase as we iterate, so it's possible that the later submissions would qualify as finished while the early ones wouldn't
```
jerzywilczek:
```suggestion
            // we can only read each semaphore's value once, because if we check more times and it changes between two checks, we will run the finisher for a later frame before an earlier one
```

Damn, took me a good 15mins to understand this, maybe it's more readable this way?

gpu-video/src/backends/vulkan/waiter_thread.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+}
+
+pub(crate) struct SubmissionWaitRequest {
+    pub(crate) semaphore: Arc<TimelineSemaphore>,
+    pub(crate) wait_for: SemaphoreWaitValue,
+    pub(crate) on_finish: Box<dyn FnOnce() + Send>,
+}
+
+struct InterruptSemaphore {
```
jerzywilczek: maybe `Waker` instead of `Interrupt`?

gpu-video/src/backends/vulkan/vulkan_decoder.rs
```diff
@@ -328,8 +282,13 @@ impl<'a> VulkanDecoder<'a> {
                 )
         };
 
-        if let Some(pool) = video_session_resources.decode_query_pool.as_ref() {
-            pool.reset(cmd_buffer.buffer());
+        let decode_query_pool = video_session_resources
```
jerzywilczek: This isn't a query pool anymore, is it?

gpu-video/src/backends/vulkan/vulkan_decoder/decoders_h264.rs
```diff
@@ -0,0 +1,357 @@
...
+        frame.check_decode_results()?;
+
+        let metadata = frame.decode_metadata;
+        let width = metadata.cropped_width;
+        let height = metadata.cropped_height;
+
+        let data = unsafe {
+            frame
+                .frame
+                .download_data_from_buffer(width as usize * height as usize * 3 / 2)?
```
jerzywilczek: inda feel like calculating the frame size isn't really this guys responsibility, it feels out of place here. maybe the decoder should be handling this? thats the place where we get the knowledge to multiply by 3/2

gpu-video/src/backends/vulkan/vulkan_decoder/decoders_h264.rs
```diff
@@ -0,0 +1,357 @@
...
+            let (frame, semaphore_wait_value) =
+                submission.download_to_wgpu_texture(&self.wgpu_device, &self.wgpu_queue)?;
+
+            unordered_frames.push(DecodeResult {
+                frame: frame.frame.clone(),
+                metadata: frame.decode_metadata.clone(),
+            });
+
+            self.submission_tracker
+                .add_wait_request(semaphore_wait_value, move || frame.check_decode_results())?;
```
jerzywilczek: what does this do?

gpu-video/src/backends/vulkan/wrappers/sync.rs
```diff
@@ -105,7 +120,7 @@ impl<K: TrackerKind> Tracker<K> {
         })
     }
 
-    #[allow(dead_code)]
+    #[cfg_attr(not(feature = "wgpu"), allow(dead_code))]
```
jerzywilczek: let's switch allows to expects in cases like this where we change the attribute anyway

gpu-video/src/parser/reference_manager.rs
```diff
@@ -27,8 +27,10 @@ pub enum ReferenceManagementError {
     #[error("The H.264 bytestream is not spec compliant: {0}.")]
     IncorrectData(String),
 
-    #[error("Missing frame. Decoder is in a corrupted state. Waiting for IDR frame")]
-    MissingFrame,
+    /// Corrupted state can be caused by missing frames or invalid bitstream. This error can be only
+    /// cleared by decoding a valid IDR frame.
+    #[error("Decoder is in a corrupted state. Waiting for IDR frame")]
+    CorruptedState,
```
jerzywilczek, line 33: I was wondering about this. VT can just tell you to recreate the session, and in my PRs I forwarded that error to the user (expecting they should recreate the encoder/decoder). we should probablyt unify this into one of the options, just dont know which one. also dont know where to note this to not forget about it.

gpu-video/src/decoders.rs
```diff
@@ -15,76 +9,49 @@ mod wgpu_api;
...
-    pub(crate) parser: H264Parser,
-    pub(crate) reference_ctx: ReferenceContext,
-    pub(crate) frame_sorter: FrameSorter<RawFrameData>,
+pub struct BytesDecoderH264 {
+    pub(crate) backend: Box<dyn VideoDecoderBackend>,
 }
 
-impl BytesDecoder {
+impl BytesDecoderH264 {
     /// The result is a sequence of frames. The payload of each [`OutputFrame`] struct is a [`Vec<u8>`]. Each [`Vec<u8>`] contains a single
```
jerzywilczek: no longer true. all decoder docs should be revised for whether they still make sense

gpu-video/src/decoders.rs
```diff
@@ -15,76 +9,49 @@ mod wgpu_api;
...
     /// decoded frame in the [NV12 format](https://en.wikipedia.org/wiki/YCbCr#4:2:0).
-    pub fn decode(
-        &mut self,
-        frame: EncodedInputChunk<'_>,
-    ) -> Result<Vec<OutputFrame<RawFrameData>>, VideoDecoderError> {
+    ///
+    /// If [`DecoderParameters::max_in_flight_submissions`](crate::parameters::DecoderParameters::max_in_flight_submissions)
+    /// decode submissions are already in flight, this blocks until all submissions above the limit finish.
```
jerzywilczek, line 30: wondering if we should have a non-blocking method (try_decode) or one with a timeout

gpu-video/src/device.rs
```diff
@@ -56,15 +57,31 @@ pub enum MissedFrameHandling {
...
+    pub corrupted_state_handling: CorruptedStateHandling,
 
     /// A hint indicating what kind of content the decoder is going to be used for.
     pub usage_flags: crate::parameters::DecoderUsage,
+
+    /// Maximum number of decode submissions that can be in flight. When the limit is reached,
+    /// decoding blocks until the oldest submission finishes.
+    ///
+    /// **Defaults to 3**
+    pub max_in_flight_submissions: NonZeroU32,
```
jerzywilczek: hmm, zero kind of makes sense, like for a 'rendezvous' channel. passing a zero here could mean that the callback fires before `decode` returns.

gpu-video/src/exports.rs
```diff
@@ -182,25 +182,40 @@ impl std::fmt::Debug for VideoDevice {
 }
 
 impl VideoDevice {
-    pub fn create_bytes_decoder_h264(
+    /// Creates an H.264 decoder that sends each decoded frame via callback.
+    ///
+    /// Heavy work in the callback can delay the delivery of frames and block [`BytesDecoderH264::decode`].
+    /// On vulkan, the delivery is done on one thread that's shared between all decoders and
+    /// encoders, so a slow callback would affect them all.
```
jerzywilczek, line 192: maybe a tip to send results through a channel? not supporting a pool sounds kinda silly here, but it makes sense for us...

jerzywilczek (approved):

jerzywilczek:
also, would love some benchmarks, but they can happen once all async is done

gpu-video/src/backends/vulkan/waiter_thread.rs
noituri: assume we use atomics here. If 2 threads try to interrupt:
- Thread_1: 
```rs
  let v =  fetch_add(wait_for) // v = 1
```
- Thread_2:
 ```rs
  let v =  fetch_add(wait_for) // v = 2
````
- Thread_2 signals with `2`
- Thread_1 signals with `1` **oof**

File: gpu-video/examples/decode_wgpu.rs (hunk 5 above)
Comment (noituri) [reply gpu-video/examples/decode_wgpu.rs:38]: me too, but out of all options I considered this is the one I didn't dislike the most

File: gpu-video/src/parser/reference_manager.rs (hunk 14 above)
Comment (noituri) [reply gpu-video/src/parser/reference_manager.rs:33]: Yeah I was thinking about it when doing async encoder. We can discuss this offline later

File: gpu-video/src/backends/vulkan/waiter_thread.rs (hunk 6 above)
Comment (jerzywilczek) [reply gpu-video/src/backends/vulkan/waiter_thread.rs:172]: Very true

maybe put that in a comment so I can't forget it and rewrite it to use atomics in two years?

Comment (noituri) [reply gpu-video/src/backends/vulkan/waiter_thread.rs:172]: > maybe put that in a comment so I can't forget it and rewrite it to use atomics in two years?

rel

File: gpu-video/src/parser/reference_manager.rs (hunk 14 above)
Comment (jerzywilczek) [reply gpu-video/src/parser/reference_manager.rs:33]: well, if youre at the office, put it on the whiteboard

it is a bit worrying that the whiteboard only ever gets more text on it and never less

File: gpu-video/src/device.rs (hunk 17 above)
Comment (noituri) [reply gpu-video/src/device.rs]: > zero here could mean that the callback fires before decode returns.

I don't see when this would be useful

File: gpu-video/src/decoders.rs (hunk 16 above)
Comment (noituri) [reply gpu-video/src/decoders.rs:30]: yeah, I think having `try_decode` makes sense

Comment (noituri) [reply gpu-video/src/decoders.rs:30]: or maybe I could add `DecoderEvent::DecodeChunkTimeout(chunk, timeout)`. Then if someone needs it, they could use the "advanced" api via `BytesDecoderH264::process_event`

File: gpu-video/src/device.rs (hunk 17 above)
Comment (jerzywilczek) [reply gpu-video/src/device.rs]: well, you can achieve the same behavior if you just send through a rendezvous channel in the finisher closure and block on receiving, so yeah, it is maybe unnecessary to add zero here

File: gpu-video/src/decoders.rs (hunk 16 above)
Comment (jerzywilczek) [reply gpu-video/src/decoders.rs:30]: do you think it would make sense to only have this event for decoding, and use a zero-timeout for `try_decode` and a max timeout for `decode`?

Comment (noituri) [reply gpu-video/src/decoders.rs:30]: now that I look at it, this would make more sense if the timeout was provided in `process_events(event, Option<timeout>)` instead of adding another event timeout variant for every event that might need it.

Passing timeout for `SignalDataLoss` event would look kinda werid, but for all other events it would make sense so I guess that's ok

`decode` would just use `process_events` (as it does rn). I'm not sure if we need to add `try_decode`, to me it kind of falls into the case where someone should use `process_event`. On the other hand exposing this is not really difficult

Encoder would probably need similar api

File: gpu-video/src/backends/vulkan/vulkan_decoder/decoders_h264.rs [hunk 19]
```diff
@@ -99,74 +98,75 @@ impl VulkanDecoderH264 {
...
+
         Ok(())
     }
 
-    fn wait_if_full(&mut self) -> Result<(), VulkanDecoderError> {
+    fn wait_if_full(&mut self, timeout: Duration) -> Result<(), VulkanDecoderError> {
+        if self.max_in_flight == 0 {
+            return Ok(());
+        }
+
```
noituri, line 153: i tried to make it more elegant so that no extra `if`s are needed for this specific case, but it would change the decoder's behavior in a slightly weird way so I just ended up using `if`s

gpu-video/src/backends/vulkan/vulkan_decoder/decoders_h264.rs
noituri: `add_wait_request` on submission finish checks if the closure returns an error and signals the corrupted state. I have made it a little bit more readable when working on the async encoder. I backported these changes to this branch now so it should be more obvious

jerzywilczek (approved):

PR #2149, opened 2026-07-24
Refactored interleaved sender for `WHIP` output

JBRS307:


smelter-core/src/pipeline/webrtc/whip_output.rs
```diff
@@ -223,135 +224,27 @@ impl WhipClientTask {
...
-                (None, None) => break,
-                // we can't do anything here, but there are still receivers
-                // that can return something in the next loop.
-                //
-                // I don't think this can ever happen
-                (_, _) => (),
-            };
+                Err(error @ WhipError::UnexpectedVideoPacket) => error!(%error),
+                Err(error @ WhipError::UnexpectedAudioPacket) => error!(%error),
+            }
```
JBRS307: @wkozyra95 
I am not sure if this is the best approach, but did not have neither better nor cleaner ideas.

And one question, should we also break on this errors like on the write error?

wkozyra95: It's fine, I'm not sure if it's even worth returning that error up. In whep implementation we just log
```
            error!(?kind, "No stream of this kind");
```
This error can happen only in case of implementation mistake, so best to default to the least disruptive handling, so no breaking I think.

wkozyra95 (approved):
Btw, I wanted also refactor this input to be in line with other webrtc io in terms of structure. e.g. small mod.rs + output.rs + maybe stream to peer, like in whep code. While ago I started that refactor, but stopped in the middle. It can be in follow up, not necessarily in this PR

PR #2160, opened 2026-07-28
Preserve DeckLink hardware A/V timing

michael-dm:
Map DeckLink video and audio through one hardware stream clock. Video-only capture adds no fixed presentation lead; audio-enabled capture applies the same 40 ms lead to both tracks so audio reaches the queue without breaking A/V alignment.

### DIFF
--- a/smelter-core/src/pipeline/decklink/capture.rs
+++ b/smelter-core/src/pipeline/decklink/capture.rs
@@ -28,8 +28,7 @@ pub(super) struct ChannelCallbackAdapter {
     // dependency
     input: Weak<decklink::Input>,
     sync_point: Instant,
-    audio_offset: Mutex<Option<Duration>>,
-    video_offset: Mutex<Option<Duration>>,
+    stream_offset: Mutex<Option<Duration>>,
     last_format: Mutex<Format>,
 }
 
@@ -48,8 +47,7 @@ impl ChannelCallbackAdapter {
             span,
             input,
             sync_point: ctx.queue_ctx.sync_point,
-            audio_offset: Mutex::new(None),
-            video_offset: Mutex::new(None),
+            stream_offset: Mutex::new(None),
             last_format: Mutex::new(initial_format),
         }
     }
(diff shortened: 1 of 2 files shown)


smelter-core/src/pipeline/decklink/capture.rs
```diff
@@ -28,8 +28,12 @@ pub(super) struct ChannelCallbackAdapter {
...
     input: Weak<decklink::Input>,
     sync_point: Instant,
-    audio_offset: Mutex<Option<Duration>>,
-    video_offset: Mutex<Option<Duration>>,
+    /// One clock mapper for both media. Video `stream_time` and audio
+    /// `packet_time` are positions on the card's single stream clock, so a
+    /// shared offset carries the hardware's own A/V alignment into our
+    /// epoch. Per-medium offsets would overwrite that alignment with
+    /// whatever the callback arrival order happened to be.
+    stream_offset: Mutex<Option<Duration>>,
```
wkozyra95, line 31: I don't have DeckLink setup to verify that so I'm trusting that this offset should be shared. Orginally there was 2 offsets because I did not see any guarantee that both streams starts from the same epoch

smelter-core/src/pipeline/decklink/capture.rs
```diff
@@ -48,23 +52,32 @@ impl ChannelCallbackAdapter {
...
+    /// format change.
+    ///
+    /// Stamps are capture-domain: they say when the card saw the content,
+    /// not when it should be presented. A consumer that needs media to be
+    /// available before its presentation time owns that lead — it is the
+    /// only side that knows how far ahead it reads.
+    fn map_stream_time(&self, stream_time: Duration) -> Duration {
+        let mut guard = self.stream_offset.lock().unwrap();
+        let offset =
+            *guard.get_or_insert_with(|| self.sync_point.elapsed().saturating_sub(stream_time));
```
wkozyra95: That extra offset is necessary; it does not have to be 40ms, but there needs to be time for chunk to reach queue. For video it does not matter much, because we pick the last frame, so even with this implementation, we would `16ms` at 60fps input to reach the queue on time. However, for audio this would mean constant stuttering

wkozyra95 (approved):

PR #2161, opened 2026-07-28
Remove V4L2 presentation delay

michael-dm:
Timestamp V4L2 frames when they reach the Pipeline without adding a fixed presentation lead. This keeps the existing receipt-time behavior while removing 20 ms of latency from this video-only input.

### DIFF
--- a/smelter-core/src/pipeline/v4l2/v4l2_input.rs
+++ b/smelter-core/src/pipeline/v4l2/v4l2_input.rs
@@ -53,9 +53,7 @@ impl TryFrom<FourCC> for V4l2Format {
 ///
 /// - Register track with `QueueTrackOffset::Pts(Duration::ZERO)` which means
 ///   that PTS should be relative to queue `sync_point`.
-/// - PTS of each frame is `sync_point.elapsed() + 20ms` (real-time capture with a
-///   small fixed buffer to account for delivery latency). This effectively syncs
-///   with the queue on every frame.
+/// - PTS of each frame is relative to queue `sync_point` at receipt time.
 /// - Never block on sending.
 ///
 /// ### Unsupported scenarios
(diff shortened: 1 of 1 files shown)


smelter-core/src/pipeline/v4l2/v4l2_input.rs
```diff
@@ -273,6 +278,41 @@ impl V4l2DeviceConfig {
...
+/// Capture-domain PTS for one dequeued buffer.
+///
+/// The kernel's buffer timestamp is the honest capture edge, but it lives in
+/// `CLOCK_MONOTONIC` and we cannot read our own `Instant` epoch in that
+/// domain. So instead of anchoring the two clocks against each other, we
+/// measure the delivery delay *within* the kernel's domain and subtract it
+/// from our own receipt stamp: no cross-clock offset to establish, drift, or
+/// reset, and every frame corrects itself.
+fn capture_pts(ctx: &Arc<PipelineCtx>, meta: &Metadata) -> Duration {
+    let receipt = ctx.queue_ctx.sync_point.elapsed();
```
wkozyra95: I would prefer a similar approach to Decklink, storing offset from first pts and using it for calculation

smelter-core/src/pipeline/v4l2/v4l2_input.rs
```diff
@@ -344,7 +385,7 @@ impl InputState<'_> {
             };
 
             let frame = Frame {
-                pts: self.ctx.queue_ctx.sync_point.elapsed() + Duration::from_millis(20),
```
wkozyra95, line 347: The same problem as with Decklink PR, however, here we don't have audio, so I'm ok with dropping that extra 20ms offset

PR #2163, opened 2026-07-28
Remove FDK AAC decoder lookahead

michael-dm:
Disable the PCM limiter and interpolation concealment so decoded PCM matches its input PTS. Warn if FDK still reports nonzero output delay.


wkozyra95 (commented):
What is the goal here? Is it just fixing desync, or the teh extra 30ms latency is the problem too?

wkozyra95:
I addressed the sync issue in https://github.com/software-mansion/smelter/pull/2166, but if latency is also a problem I would prefer to address it as an input configuration option instead of always enabling those options

PR #2164, opened 2026-07-28
Compensate FDK AAC encoder delay

michael-dm:
Use FDK AAC's reported encoder delay to timestamp access units by the content they carry. Drop only frames containing no real content and retain the first mixed priming/content frame, avoiding a persistent AAC A/V shift without discarding its real samples.

### DIFF
--- a/smelter-core/src/pipeline/encoder/fdk_aac.rs
+++ b/smelter-core/src/pipeline/encoder/fdk_aac.rs
@@ -23,6 +23,7 @@ pub struct FdkAacEncoder {
     output_buffer: Vec<u8>,
     sample_rate: u32,
     samples_per_frame: u32,
+    codec_delay: u64,
 
     // This logic relies on the fact that input samples will always be continuous.
     first_input_pts: Option<Duration>,
@@ -114,6 +115,7 @@ impl AudioEncoder for FdkAacEncoder {
                 input_buffer: Vec::new(),
                 output_buffer: vec![0; info.maxOutBufBytes as usize],
                 sample_rate: options.sample_rate,
+                codec_delay: info.nDelay as u64,
                 first_input_pts: None,
                 encoded_samples: 0,
                 samples_per_frame: info.frameLength,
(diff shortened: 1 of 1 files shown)


smelter-core/src/pipeline/encoder/fdk_aac.rs
```diff
@@ -225,14 +232,26 @@ impl FdkAacEncoder {
...
+
+                // Swallow the codec delay: frames that still contain priming
+                // content are dropped entirely (their tail of real samples —
+                // under one frame, at the very head of the output — goes with
+                // them), and every emitted frame is stamped at the position
+                // of the *content* it carries. Without this, all AAC audio
+                // plays `codec_delay` (1600 samples for AAC-LC — 33 ms, one
+                // full frame at 29.97 fps) late relative to video, since
+                // neither raw AAC nor the muxers signal priming to players.
+                if frame_start < self.codec_delay {
```
wkozyra95: I would prefer to only drop frames that do not include real content. Would this work for you?

```suggestion
                if self.encoded_samples < self.codec_delay {
```

Technically, this earlly exit should not be necessary; just pts calculation should resolve all the issues, but it relies on all the downstream muxers handling it correctly

smelter-core/src/pipeline/encoder/fdk_aac.rs
```diff
@@ -223,16 +225,24 @@ impl FdkAacEncoder {
...
+                }
+                let first_input_pts = self.first_input_pts.unwrap_or_default();
+                let offset = Duration::from_secs_f64(
+                    frame_start.abs_diff(self.codec_delay) as f64 / self.sample_rate as f64,
+                );
+                let pts = if frame_start >= self.codec_delay {
+                    first_input_pts + offset
+                } else {
+                    first_input_pts.saturating_sub(offset)
+                };
```
wkozyra95:
```suggestion

                let first_pts = self.first_input_pts.unwrap_or_default();
                let offset = Duration::from_secs_f64(frame_start as f64 / self.sample_rate as f64);
                let codec_delay =
                    Duration::from_secs_f64(self.codec_delay as f64 / self.sample_rate as f64);
                let pts = (first_pts + offset).saturating_sub(codec_delay);
```

wkozyra95:
I had to update snapshots too, so opened a separate PR. Closing in favor of https://github.com/software-mansion/smelter/pull/2172

PR #2165, opened 2026-07-28
Preserve equal-rate audio batches

michael-dm:
Bypass rubato when input and output rates match. This preserves input PTS and forwards the final partial batch instead of dropping it at EOS.


wkozyra95 (approved):

wkozyra95:
To be honest, I'm not even sure why we're always resampling on output; I suspect it might be because the resampler also applies an anti-aliasing filter that can smooth out some audio artifacts, but at this point audio mixer should be responsible for that, so I'm fine with changing that now

PR #2168, opened 2026-07-31
Add native input side channels

michael-dm:
Adds an in-process destination to the existing input side channels. Native receivers get the same normalized timestamps and non-blocking, drop-on-full behavior as Unix socket clients, without socket serialization.

The existing API booleans continue to select Unix sockets. Core callers can select a bounded native channel explicitly.


wkozyra95:
Closing in favor of https://github.com/software-mansion/smelter/pull/2181. It's mostlly based on this PR with small refactor + added frame conversion in v4l2 and decklink inputs

PR #2169, opened 2026-07-31
Expose external output audio sender

michael-dm:
Allows an output audio sink to be fed directly when it is not owned by the Pipeline mixer. This supports already-mixed, Pipeline-clocked audio without an extra raw-input round trip or relay channel.

The caller owns PTS continuity and EOS. Mixer-owned outputs do not expose the sender.

### DIFF
--- a/smelter-core/src/pipeline/instance.rs
+++ b/smelter-core/src/pipeline/instance.rs
@@ -6,7 +6,7 @@ use std::{
     time::Duration,
 };
 
-use crossbeam_channel::{Receiver, bounded};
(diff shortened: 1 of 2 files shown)


wkozyra95 (commented):
If there is something more complex/harder to maintain that you specifically need we can discuss merging it, but this patch specifically seems very easy to maintain in your fork.

wkozyra95:
I think this should stay in fork. If you really need that, let me know; we can discuss that further, but for now I'm closing the PR

PR #2170, opened 2026-07-31
Schedule output start and stop by PTS

michael-dm:
Adds PTS-gated start and stop for independently prepared outputs, so callers can give several outputs the same boundaries without a persistent group abstraction. Stop reuses the existing EOS path.

Scheduled MP4 and HLS outputs use the requested start PTS as their common timestamp origin instead of whichever encoded packet arrives first. Immediate output behavior is unchanged.

I kept origin selection at the muxer boundary to limit the diff. Rebasing timestamps once at the generic output boundary may be a cleaner long-term model if you prefer that direction.

### DIFF
--- a/smelter-core/src/pipeline/output.rs
+++ b/smelter-core/src/pipeline/output.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{HashMap, HashSet},
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
 use crossbeam_channel::Sender;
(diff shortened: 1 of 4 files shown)


smelter-core/src/pipeline/output.rs
```diff
@@ -22,6 +23,26 @@ pub(crate) struct PipelineOutput {
     pub output: Box<dyn Output>,
     pub video_end_condition: Option<PipelineOutputEndConditionState>,
     pub audio_end_condition: Option<PipelineOutputEndConditionState>,
+    pub start_at: Option<Duration>,
+    pub stop_at: Option<Duration>,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum OutputTimestampOrigin {
```
wkozyra95, line 31: I see what it does, but not really what it is for, and how you intend to use it. Can you clarify that?

wkozyra95 (commented):
scheduled stopping is already possible by scheduling unregister,

scheduling output registration is not possible, but I'm open to adding it, I can implement it, but I want to first understand what is OutputTimestampOrigin for and how you use it to make sure the problem you are trying to solve with it can still be solved in new API.

michael-dm:
The issue we need to solve at start is this:
We record one MP4 containing video and audio alongside several video-only MP4s. All are scheduled for the same start time.
In a small test, both files contained the same 29 video frames, but the first video frame in the A/V file was timestamped 25-32 ms later. This happened because the A/V file chose an earlier audio packet as its local zero, while the video-only file chose its first video frame.

We persist the common recording start and later use it to align these files for editing. With the observed timestamps, stacking them at that start leaves the A/V video about one frame behind the video-only recordings. One possible behavior would be to use the first video packet as the local origin whenever an output contains video. That would align this case as well.

wkozyra95:
Closing in favor of https://github.com/software-mansion/smelter/pull/2183

PR #2171, opened 2026-08-03
Preserve exact MP4 loop timing

michael-dm:
MP4 loops currently follow reader completion rather than the container presentation boundary. AAC decoder priming and a final packet extending past that boundary can therefore change the duration of an iteration, producing an audio discontinuity or moving video off its declared cadence across loops.

Using current main unchanged does not leave a downstream repair point: by the time decoded media is observed, the queue has already selected the next track from reader EOS. Repairing it outside Smelter would duplicate MP4 timeline ownership.

This removes decoder priming from the initial decoded batches, reads the track presentation duration from `tkhd`, and chains finite queue tracks at that exact boundary. Samples beyond the boundary are clipped before entering the queue. Non-looping tracks keep their current behavior.

The changes stay in one PR because each part establishes the same invariant: decoding, queue transitions, and emitted media use the container presentation timeline.

### DIFF
--- a/smelter-core/src/pipeline/decoder/fdk_aac.rs
+++ b/smelter-core/src/pipeline/decoder/fdk_aac.rs
@@ -143,7 +143,7 @@ impl Decoder {
                 let info = unsafe { *fdk::aacDecoder_GetStreamInfo(self.instance) };
                 let raw_frame_size = (info.aacSamplesPerFrame * info.channelConfig) as usize;
 
-                let samples = match info.channelConfig {
+                let mut samples = match info.channelConfig {
                     1 => AudioSamples::Mono(
                         self.decoded_samples_buffer[..raw_frame_size]
                             .iter()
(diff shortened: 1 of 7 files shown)


wkozyra95 (commented):
queue is not a place to resolve logic like this, if there is something wrong with timestamps, it should be resolved in inputs directlly.

I agree that there is a problem with our implementation, we only support part of `elst` box functionality. This is intentional because there are some mechanisms that no one 
is using e.g. repeating specific segment.

I would need to look at the spec some more, but I think `tkhd` might not be correct value too, it's final duration (an outcome), but it does not say how much samples we should produce, that value depends on specifics of `elst` structure.

But before looking at I wanted to verify if this is the issue? Or the problem is with something else. For example if presentation time is 10 seconds but there is only 5 seonds of samples do you expect the loop to be 10 seconds or 5 seconds? If you have specific mp4s that cause problems that would be the most helpfull.

michael-dm:
Thanks @wkozyra95 for the feedback, here is a fixture with more context on our issue.

The file contains **13 video frames**, numbered 0–12, plus audio.

When smelter loops it:

- First playback: frames 0–12 are shown correctly.
- Later playbacks: the first frame is skipped, so they start at frame 1.

This happens because Smelter recalculates the timing origin for every loop. The audio packets do not align exactly with the 25 fps video frame boundaries, so the next loop can begin slightly too late, past frame 0.

Here, the desired behavior is simply: each new loop should start exactly one video duration after the previous one. That keeps every iteration on the same frame grid.

I verified that a simpler approach actually solved this skipped frame issue : https://github.com/software-mansion/smelter/pull/2173. Would that be a better approach for you ? I'm not yet sure if it creates audio clicks however.

https://github.com/user-attachments/assets/c9d0a876-17aa-4caa-9bb6-8be64edc78d2

PR #2173, opened 2026-08-04
Fix MP4 loop frame cadence

michael-dm:
Each queued iteration currently resolves a fresh shared A/V offset from whichever track first reaches the queue. AAC chunk cadence does not generally align with the video frame grid, so that offset can move the next iteration past its first video frame. A 25 fps fixture consequently produced frames 0–11 on its first iteration and 1–11 thereafter.

This change only lets an input describe the next track as a continuation of the preceding track. The queue still owns offset resolution, while the MP4 input supplies its known iteration duration. It does not add finite-track scheduling, trimming, or broader end-of-stream policy.
Supersedes #2171.

### DIFF
--- a/smelter-core/src/pipeline/mp4/mp4_input.rs
+++ b/smelter-core/src/pipeline/mp4/mp4_input.rs
@@ -117,6 +117,9 @@ impl Mp4Input {
         let video_duration = video_track.as_ref().and_then(|track| track.duration());
         let audio_track = Mp4FileReader::from_path(&source_file.path)?.try_new_aac_track();
         let audio_duration = audio_track.as_ref().and_then(|track| track.duration());
+        let loop_duration = video_duration
+            .or(audio_duration)
+            .filter(|_| options.should_loop);
 
         if video_track.is_none() && audio_track.is_none() {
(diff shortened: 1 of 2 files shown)


wkozyra95:
Closing in favor of https://github.com/software-mansion/smelter/pull/2177

Linked PR fixes the issue with the disappearing first frame, but it does not add any strict guarantees about loop duration and we still ignore duration from elst box and play all the samples. It prioritizes video instead of audio when establishing offset, so the offset will be snapped to the queue video grid instead of the audio. It will cause a symmetric issue for audio, but because audio is continuous, it's a lot less impactful.

There is one negative consequence if the previous track finished, and next point that queue should process is video e.g. pts 1066ms and next audio chunk would be 1080ms, we would lose that  14ms of audio (at most it will be `max(1/fps, audio_batch))`

But there is also an extra benefit of that approach. We always pick the closest pts (older one)  frame when selecting what should be shown for specific output PTS, but by definition it is causing desync between audio and video for max 1 frame. In this case, if input and queue fps are equal (really equal, not 30fps != 29.997 fps), then a/v sync will be fully maintained.

PR #2174, opened 2026-08-05
Release DeckLink capture configuration on drop

michael-dm:
## Summary

- shut down DeckLink streams, video, audio, and callbacks before releasing the input interface
- centralize teardown in the owning `decklink::Input` RAII wrapper
- remove the partial Smelter Core destructor

This prevents one capture session from leaving DeckLink input configuration active for the next session.

Validated with `cargo check -p decklink` and `cargo test -p decklink`.

### DIFF
--- a/decklink/cpp/api.cpp
+++ b/decklink/cpp/api.cpp
@@ -157,7 +157,13 @@ HResult input_flush_streams(IDeckLinkInput *input) {
   return static_cast<HResult>(input->FlushStreams());
 }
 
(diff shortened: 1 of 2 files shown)


michael-dm:
Closing: this teardown reorder was written while chasing an intermittent HDMI-audio silence that turned out to be unrelated (an ELD/pin-enable race on the playback side). We no longer carry this change and can't justify it with an observed failure.

PR #2175, opened 2026-08-05
gpu-video: introduce MemoryBarrier2Masks struct

krakow10:
The `Range` struct is abused here to reduce the argument count, but the values are absolutely not a range.  This was confusing when reading the code because of the semantic mismatch.  Introduce a dedicated struct to achieve the desired effect of reducing argument count.

### DIFF
--- a/gpu-video/src/backends/vulkan/wrappers/mem.rs
+++ b/gpu-video/src/backends/vulkan/wrappers/mem.rs
@@ -337,6 +337,14 @@ impl std::ops::Deref for Buffer {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct MemoryBarrier2Masks {
+    pub src_stage_mask: vk::PipelineStageFlags2,
+    pub dst_stage_mask: vk::PipelineStageFlags2,
+    pub src_access_mask: vk::AccessFlags2,
+    pub dst_access_mask: vk::AccessFlags2,
+}
+
 pub(crate) struct Image {
     pub(crate) image: vk::Image,
     allocation: vk_mem::Allocation,
@@ -414,16 +422,15 @@ impl Image {
         &self,
         command_buffer: vk::CommandBuffer,
         layout: &mut [vk::ImageLayout],
-        stages: std::ops::Range<vk::PipelineStageFlags2>,
-        accesses: std::ops::Range<vk::AccessFlags2>,
+        masks: MemoryBarrier2Masks,
         new_layout: vk::ImageLayout,
         subresource_range: vk::ImageSubresourceRange,
     ) -> Result<(), VulkanCommonError> {
         let barrier = vk::ImageMemoryBarrier2::default()
-            .src_stage_mask(stages.start)
-            .dst_stage_mask(stages.end)
-            .src_access_mask(accesses.start)
-            .dst_access_mask(accesses.end)
+            .src_stage_mask(masks.src_stage_mask)
+            .dst_stage_mask(masks.dst_stage_mask)
+            .src_access_mask(masks.src_access_mask)
+            .dst_access_mask(masks.dst_access_mask)
             .new_layout(new_layout)
             .src_queue_family_index(vk::QUEUE_FAMILY_IGNORED)
             .dst_queue_family_index(vk::QUEUE_FAMILY_IGNORED)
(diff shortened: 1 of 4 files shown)


jerzywilczek:
oh my, terribly sorry for not noticing this earlier
we must add a codeowners file, it will help me not miss PRs like this one in the future

reviewing now

gpu-video/src/backends/vulkan/wrappers/mem.rs
```diff
@@ -337,6 +337,14 @@ impl std::ops::Deref for Buffer {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct MemoryBarrier2Masks {
```
jerzywilczek, line 341: The name is not intuitive enough. I'd suggest something like `MemoryAccessRelation` and maybe renaming the fields from `src...`, `dst...` to `start...`, `end...`.

The struct name should not describe which slots the struct fills in the vulkan API, but rather what this struct describes itself: in this case, a happen-before relation between the start scope and the end scope. I'm not very certain about the suggested name and field names, you can come up with other, better ones, but the names should explain the struct in terms of the relation it establishes between two kinds of memory accesses.

This was my motivation behind the initial range-based design, as ranges carry the meaning that one thing is the start and the other is the end; clearly the idea wasn't perfect since it confused you, but thats what I wanted to achieve.

jerzywilczek: also, in a perfect world i would like the names to be shorter, to keep the struct ergonomic, as it's used quite a lot and in places that are rather noisy already. In the places where it is used, id like the code to read like this is just an expression of order in memory accesses, with no thinking of how it relates to specific vulkan mechanisms that are used to achieve the access ordering

jerzywilczek (changes requested):

gpu-video/src/backends/vulkan/wrappers/mem.rs
jerzywilczek: Oh, i also just remembered that the inspiration came from `wgpu` representing layout transitions this exact way (I mean by `start_layout..end_layout` ranges)

krakow10: > `start...`, `end...`

Start/End typically refers to space, like the start and end areas of a race, not temporal like before/after. My suggestion would be MemoryVisibilityChange From/To.

> The struct name should not describe which slots the struct fills in the vulkan API, but rather what this struct describes itself

I think that it's a bad idea to invent bespoke terminology here. The struct has one purpose: to pass mask arguments to vulkan MemoryBarrier2 or ImageMemoryBarrier2 structs. It should be named after that rather than having a name that only appears in this project. If you add other backends, such as supporting apple with VideoToolBox, this code would become a vulkan-specific implementation detail, meaning it does not need to describe its general purpose, because it is not general purpose.

> Oh, i also just remembered that the inspiration came from `wgpu` representing layout transitions this exact way (I mean by `start_layout..end_layout` ranges)

Maybe I should go bother them too :grin:

jerzywilczek: > Start/End refers to space

agreed, not a good name

> I think that it's a bad idea to invent bespoke terminology here

This is an issue of style, so a place where we may need to agree to disagree. When analyzing the more high-level parts of the code, we need to imagine the vulkan pipeline as a graph of ordered memory accesses. The language we use in the code should naturally reflect how we think of it. The purpose of an abstraction layer is to lead my brain through the higher-level aspects of the code so that I can analyze the high-level aspects of it and forget about the low-level details where needed. Readability is very important, but readability for maintainers is more important than readability for general public.

maybe we could have something more like:

```rust
struct MemoryAccess {
    stage: vk::PipelineStageFlags2,
    access: vk::AccessFlasg2,
}

struct MemoryDependency {
    from: MemoryAccess,
    to: MemoryAccess,
}
```

but using this is rather verbose.

the use of ranges in the current code, even though a little unorthodox, is very ergonomic and does express the (from, to) relationship.

jerzywilczek: @noituri opinion? we seem to be at a stalemate here

jerzywilczek: > but using this is rather verbose.

btw times like this are when you wish for zig's struct initialization syntax sugar

noituri: This mainly comes down to style preferences, so the opinions may vary from person to person, but
I really prefer using range here over a struct. 

I know it's a bit unconventional in rust, but as someone who joined the project later on, I didn't find it confusing. I'm afraid that using struct would make it more verbose. I'm assuming you already had to extract the masks to separate variables to avoid visual clutter, but now when I read `transition_layout()` I have to backtrack to where the masks are defined. It's a minor thing, but it kind of "breaks" the flow for me.

The range syntax is easier to inline because it avoids the clutter. Also, when I read range I read it as "from...to", which I think makes sense in the context it's used in.

> If you add other backends, such as supporting apple with VideoToolBox, this code would become a vulkan-specific implementation detail, meaning it does not need to describe its general purpose, because it is not general purpose.

This code is already vulkan specific so I don't think this is a problem.

I'm open to suggestion on how to make it more readable, but I'm not convinced that using struct like this is the solution.

krakow10:
Closing due to existing code being preferred by the actual maintainers.  Thanks for reviewing.

jerzywilczek:
No problem, take care! even though this didnt end up being merged, I hope that doesnt dissuade you from further contributions as theyre quite nice!

krakow10:
It definitely falls under the category of style change, so deferring style choices to you guys makes the most sense.  I'll probably keep reading the code as reference, so I may find things to contribute.

PR #2178, opened 2026-08-10
[gpu-video] Fix dropping image view before submission finishes

noituri:


jerzywilczek (approved):
strange we didnt catch that before

PR #2183, opened 2026-08-11
Register output with start time (HLS+MP4)

wkozyra95:
https://github.com/smelter-labs/smelter-snapshot-tests/pull/97

## What

Adds `start_at_ms` to MP4 and HLS output registration — the output is created when the register request is handled (file/playlist on disk, encoders up), but only starts receiving frames and samples once the queue reaches the given time. Not supported on live-connection protocols.

## Core change

Output registration is split into **create** and **start**. `register_pipeline_output` builds and inserts the output; `start_output` connects it to the renderer and audio mixer. Without `start_at_ms` the two happen back to back, otherwise the start becomes a scheduled queue event.

Outputs therefore have a pending state now: a scene update on a pending output is stored as its initial scene instead of going to the renderer, and its scene is validated at start time rather than during the register request.

MP4/HLS output timestamps are no longer anchored on the first chunk that arrives — the anchor is `start_at` when set, otherwise the lowest PTS across the first chunk of each track.

### DIFF
--- a/smelter-core/src/pipeline/encoder/fdk_aac.rs
+++ b/smelter-core/src/pipeline/encoder/fdk_aac.rs
@@ -230,7 +230,7 @@ impl FdkAacEncoder {
                 let frame_start = self.encoded_samples;
                 self.encoded_samples += self.samples_per_frame as u64;
 
-                if self.encoded_samples < self.codec_delay {
+                if self.encoded_samples <= self.codec_delay {
                     continue;
                 }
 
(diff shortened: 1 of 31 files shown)


smelter-core/src/pipeline/encoder/fdk_aac.rs
```diff
@@ -230,7 +230,7 @@ impl FdkAacEncoder {
                 let frame_start = self.encoded_samples;
                 self.encoded_samples += self.samples_per_frame as u64;
 
-                if self.encoded_samples < self.codec_delay {
+                if self.encoded_samples <= self.codec_delay {
```
wkozyra95, line 233: If the entire chunk is built from encoder padding, we are dropping it. This specific change fixes the case where padding is an exact multiple of a single frame

jerzywilczek (approved):

PR #2187, opened 2026-08-12
[gpu-video] VideoToolbox encoder

jerzywilczek:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/gpu-video/src/backends/video_toolbox/encoder.rs
+++ b/gpu-video/src/backends/video_toolbox/encoder.rs
@@ -1214,7 +1214,7 @@ fn apply_preset(session: &vt::VTCompressionSession, preset: EncoderPreset) {
     let Some(key) = cfstring_constant(symbol) else {
         // compression presets were added in macOS 26
         tracing::debug!(
-            "VideoToolbox has no compression presets on this macOS; skipping the {name} preset base configuration"
+            "VideoToolbox has no compression presets on this macOS version; skipping the {name} preset base configuration"
         );
         return;
     };
--- a/gpu-video/src/backends/video_toolbox/encoder.rs
+++ b/gpu-video/src/backends/video_toolbox/encoder.rs
@@ -1,0 +1,1357 @@
+use std::{
+    cell::OnceCell,
+    ffi::{c_int, c_void},
+    ops::Deref,
+    ptr::{NonNull, null, null_mut},
+    sync::mpsc,
+};
+
+use objc2_core_foundation as cf;
+use objc2_core_media as cm;
+use objc2_core_video as cv;
+use objc2_video_toolbox as vt;
+
+use crate::{
+    EncodedOutputChunk, InputFrame, RawFrameData, VideoEncoderError,
+    backends::video_toolbox::{
+        CVBufferExt, OSStatusError, allocate_retained, caps,
+        error::{OSStatusExt, VTEncoderError},
+    },
+    device::{ColorRange, ColorSpace, EncoderOutputParameters, VideoParameters},
+    encoders::{
+        VideoEncoderBackend, VideoEncoderParametersInfoH264, VideoEncoderParametersInfoH265,
+    },
+    parameters::{EncoderPreset, EncoderUsage, H264Profile, H265Profile, RateControl},
+};
+
+#[cfg(feature = "wgpu")]
+pub(crate) mod wgpu_api;
+
+const ANNEX_B_START_CODE: [u8; 4] = [0, 0, 0, 1];
+
+pub(crate) trait EncodeCodec: Send {
+    type Profile: Copy + std::fmt::Debug + Send;
+
+    /// parsed out of the session's current `CMFormatDescription`
+    type StreamParameters: Send;
+
+    const CODEC_TYPE: cm::CMVideoCodecType;
+
+    fn profile_level(profile: Self::Profile) -> &'static cf::CFString;
+
+    fn stream_parameters(
+        format_description: &cm::CMFormatDescription,
+    ) -> Result<Self::StreamParameters, OSStatusError>;
+
+    fn keyframe_prefix(parameters: &Self::StreamParameters) -> &[u8];
+
+    fn write_sample(
+        sample: &cm::CMBlockBuffer,
+        parameters: &Self::StreamParameters,
+        output: &mut Vec<u8>,
+    ) -> Result<(), VTEncoderError>;
+}
+
+pub(crate) struct ParameterSetInfo {
+    count: usize,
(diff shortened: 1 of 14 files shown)


gpu-video/src/backends/video_toolbox/encoder.rs
```diff
@@ -0,0 +1,1357 @@
...
+        if self.parameters_changed_mid_stream && !self.inline_stream_params {
+            return Err(VTEncoderError::ParametersDiverged);
+        }
+
+        let (presentation_time_stamp, duration) = self.next_frame_timing();
+
+        let frame_properties = Self::frame_properties(force_idr);
+        let frame_properties = frame_properties
+            .as_ref()
+            .map(|properties| properties.as_ref());
```
noituri, line 585: would `Self::frame_properties(force_idr).as_deref()` work here (and other places too)?

gpu-video/src/backends/video_toolbox/encoder.rs
```diff
@@ -0,0 +1,1357 @@
...
+    let (name, symbol) = match preset {
+        EncoderPreset::HighQuality => ("HighQuality", c"kVTCompressionPreset_HighQuality"),
+        EncoderPreset::Balanced => ("Balanced", c"kVTCompressionPreset_Balanced"),
+        EncoderPreset::LowLatency => ("HighSpeed", c"kVTCompressionPreset_HighSpeed"),
+    };
+
+    let Some(key) = cfstring_constant(symbol) else {
+        // compression presets were added in macOS 26
+        tracing::debug!(
+            "VideoToolbox has no compression presets on this macOS; skipping the {name} preset base configuration"
```
noituri:
```suggestion
            "VideoToolbox has no compression presets on this macOS version; skipping the {name} preset base configuration"
```

gpu-video/src/backends/video_toolbox/encoder/wgpu_api.rs
```diff
@@ -0,0 +1,310 @@
...
+        let value = state.next_fence_value;
+        state.next_fence_value += 1;
+
+        {
+            let hal_queue =
+                unsafe { wgpu_queue.as_hal::<MtlApi>() }.ok_or(VTInitError::NotMetalBackend)?;
+            unsafe { hal_queue.submit(&[], &[], (&state.fence, value)) }
+                .map_err(WgpuTextureEncoderError::from)?;
+        }
+
```
noituri, line 175: I think this submit can be avoided. There's a method on hal queue that lets you add signal event that's attached to the next submit. You can add the signal before submitting the texture copy

gpu-video/src/backends/video_toolbox/error.rs
```diff
@@ -301,6 +302,110 @@ impl From<VTDecoderError> for VideoDecoderError {
...
+    NoFormatDescription,
+
+    #[error("VideoToolbox invalidated the compression session and it could not be recovered")]
+    SessionInvalidated(#[source] Box<VTEncoderError>),
+
+    #[error(
+        "The stream parameters no longer match the ones this encoder returned earlier; recreate the encoder to continue"
+    )]
+    ParametersDiverged,
+
```
noituri, line 348: Is it something users should try to handle? It converges to `BackendError` so handling this is not so obvious

gpu-video/src/backends/video_toolbox/wgpu_api.rs
```diff
@@ -95,19 +103,185 @@ impl WgpuVideoDeviceBackend for VTDevice {
...
+    label: &str,
+) -> Result<wgpu::Texture, MetalTextureError> {
+    let width = cv::CVPixelBufferGetWidth(buffer);
+    let height = cv::CVPixelBufferGetHeight(buffer);
+    let y_width = cv::CVPixelBufferGetWidthOfPlane(buffer, 0);
+    let y_height = cv::CVPixelBufferGetHeightOfPlane(buffer, 0);
+    let uv_width = cv::CVPixelBufferGetWidthOfPlane(buffer, 1);
+    let uv_height = cv::CVPixelBufferGetHeightOfPlane(buffer, 1);
+
+    cache.0.flush(0);
```
noituri, line 207: why flush?

noituri (approved):
🚀

noituri:
sorry I force-pushed to this branch by accident xd

gpu-video/src/backends/video_toolbox/encoder/wgpu_api.rs
noituri: or not? According to the docs, submit from a different thread can signal the semaphore :/ eh

gpu-video/src/backends/video_toolbox/wgpu_api.rs
jerzywilczek: the cache only collects dropped textures on calling `flush`. I had the clanker check ffmpeg and chrome's webrtc (or similar large repos, cant recall) and they flush every frame, so I opted for the same thing

gpu-video/src/backends/video_toolbox/encoder/wgpu_api.rs
jerzywilczek: yeah, tragic

jerzywilczek: also, don't we use the new path on vulkan, shouldnt we fix this?

jerzywilczek:
> sorry I force-pushed to this branch by accident xd

dude....

gpu-video/src/backends/video_toolbox/encoder/wgpu_api.rs
noituri: in the new async decoder we use wait semaphores and my understanding is that wgpu chains submissions. 
So the next submission will wait for the decoding to finish, and the subsequent submission depends on the previous submission. So in this case everything works fine.

But with encoders we actually care which submission signals the semaphore, so this hacky code is needed :(

gpu-video/src/backends/video_toolbox/error.rs
jerzywilczek: made a new top-level error for forcing encoder recreation

gpu-video/src/backends/video_toolbox/encoder.rs
jerzywilczek: unfortunately, not really. `as_deref` will not infer the generic types the way we want

PR #2199, opened 2026-09-02
Filter decode_only frames by pts values

wkozyra95:
- Rename field (+ invert bool value)
- Extract logic in a decoder-independent way; up until  now, only h264 supported it
- Old approach: set a flag when receiving chunks that should be dropped, and drop frames that the decoder produced immediately after
- New approach: remember PTS ranges that should be dropped, and check if frame is in that range


wkozyra95:
does not work with mp4, drops initial frames (we lack negative PTS to support that properlly)

wkozyra95:
closing in favor of https://github.com/software-mansion/smelter/pull/2204

PR #2203, opened 2026-09-03
Replace Duration with Timestamp type (support negative)

wkozyra95:


jerzywilczek (approved):
i think it is fine. there are a few inconsistencies with rounding/saturating etc, but i dont think those are critical

PR #2218, opened 2026-09-16
Handle AAC encoder initial padding correctlly

wkozyra95:
The previous fix corrected the alignment, but cutting that padding will affect decoding of the real initial chunk.


smelter-core/src/pipeline/hls/hls_output.rs
```diff
@@ -120,6 +120,7 @@ impl HlsOutput {
             start_at,
             video_stream.is_some(),
             audio_stream.is_some(),
+            None,
```
wkozyra95, line 123: No way to do that for HLS, if playlist is mpeg-ts then there is no `elst` box equivalent and for fMP4 most players will ignore elst

